PR #207, opened 2014-11-27
Remove import * in character blueprint

pascalmolin:
this commit can be applied on sage 6.1 versions, it does not fix the issues with 6.4
(this will be the next pull request)

In d4118a88d0 I have disabled 3 tests which fail both on my machine and on beta.lmfdb.org -- from some time -- : I think those tests should be checked and commited again.
However I do not feel comfortable with this way of doing things so I can also revert d4118a88d0 : advice ?

### DIFF
--- a/lmfdb/HeckeCharacters.py
+++ b/lmfdb/HeckeCharacters.py
@@ -1,11 +1,12 @@
 # -*- coding: utf-8 -*-
 # HeckeCharacters.py
 
-from sage.all import *
+from sage.misc.cachefunc import cached_method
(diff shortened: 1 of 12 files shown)


JohnCremona:
Update: before merging this should work properly with Sage 6.5.

JohnCremona:
Should this request be closed?

haraldschilly:
@pascalmolin this looks like it has bitrotten. any chance you could try to rebase it on the current master?

davidfarmer:
Closing without merging because too much has happened since.

PR #208, opened 2014-11-27
Simplified setup

pascalmolin:
I think I have found a way to make the lmfdb setup much simpler. But some testing would be welcome.

With this patch, on my machine, with a new sage-4.6 freshly downloaded, following the (updated) instructions, that is typing only
```
sage -i gap_packages
sage -i database_gap
sage -i pip
sage -b
sage -pip install -r requirements.txt
```
sets up everything so that all tests (but one on tensor products) pass.

two comments:
- first, I suspect that the problems with the Dirichlet characters on sage 6.4 are due to the fact that everyone
  uses an outdated package dirichlet-conrey. Jonathan and I have made changes in the Arizona workshop.
- this issue is solved here since I use pip to install all packages at once from requirements.txt, including the dirichlet_conrey module directly from Jonathan's github

### DIFF
--- a/GettingStarted.md
+++ b/GettingStarted.md
@@ -6,7 +6,7 @@ Installation
      git clone git@github.com:LMFDB/lmfdb.git lmfdb
   ```
 
-* Make sure you have sage (>=6.4) installed and that
+* Make sure you have sage (>=6.5) installed and that
   `sage` is available from the commandline
 
 * Install dependencies (in the `lmfdb/` directory):
--- a/GettingStarted.md
+++ b/GettingStarted.md
@@ -1,83 +1,23 @@
 Installation
 ============
 
-  * To contribute, see below on sharing your work.  To simply run a copy of the site move into a new directory and type
+* To contribute, see below on sharing your work. To simply run a copy of the site move into a new directory and type
 ```
     git clone git@github.com:LMFDB/lmfdb.git lmfdb
 ```
 
-  * Install dependencies, i.e. you need Sage. Inside the Sage environment `sage -sh`:
+* Make sure you have sage (>=6.4) installed and that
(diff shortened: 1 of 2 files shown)


JohnCremona:
Pascal, are you still wanting this to be merged or is there more work to be done on it?

JohnCremona:
ping!

pascalmolin:
pip is included in sage6.5 and the libcrypto issues are gone, so it seems this new setup should work.
By the way, I noticed lmfdb is not compatible with pymongo 3.0, so the setup forces to use 2.8. 

I have tested it successfully on 3 machines (2 mac + 1 linux)

JohnCremona:
1. How can I tell which version of pymongo I am using?
2. What sort of incompatibility is there with lmfdb?  It would be nice to fix that, especially if making lmfdb work with pymongo 3.0 would not stop it from working with 2.8.

pascalmolin:
I get the list of packages installed in sage with

sage -pip freeze

As for pymongo 3.0, I did not look into all the interface changes, but at
least the module names are not the same.

-- 
Pascal

2015-04-08 18:06 GMT+02:00 John Cremona <notifications@github.com>:

JohnCremona:
We (John and Samuele) tested this on a fresh sage 6.6 + fresh lmfdb on a new laptop and all works as expected.
Once my pymongo pull request is also merged we can remove the forcing of pymongo version 2.8.

pdehaye:
What version of mongo is now running on the server? It might be helpful to
document if problems appear later during an upgrade.
Paul


Paul-Olivier Dehaye
SNF Assistant Professor of Mathematics
University of Zurich
http://user.math.uzh.ch/dehaye/contact_info.txt

PR #211, opened 2015-02-13
Minor elliptic curve cm handling

JohnCremona:
We now store CM discriminants in the database and so no longer have to compute them on the fly.
Also added a relevant test, and made a minor improvment to elliptic curve page layout.

### DIFF
--- a/CONTRIBUTORS.yaml
+++ b/CONTRIBUTORS.yaml
@@ -68,12 +68,16 @@ name: Tim Dokchitser
 url: http://www.maths.bris.ac.uk/~matyd/
 affil: University of Bristol
 ---
+name: Bas Edixhoven
+affil: Universiteit Leiden
+---
 name: Stephan Ehlen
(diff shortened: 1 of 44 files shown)


JohnCremona:
I inadvertently combined two things into one pull request.  Apart from the stuff mentioned in the title (which will not concern most people) the last commit & merge does:  making things work well with Sage 6.4 and 6.5.  I changed some .coeffs() calls with .coefficients(sparse=False) to get rid of deprecation warnings.  I fixed some Hecke Character code so that there are 4 tests which used to be skipped but which now work fine.  I made a small change to test.sh so that the defail works as before but you can prepend SAGE=sage-x.y.z on the command line to run tests with a particular version of Sage (which must be in your path) -- it just checks the SAGE environment variable.
There are still two failing tests but otherwise we should be able to switch the beta server to running 6.5 soon.  (The tests also fail on 6.4, 6.3, etc).

PR #217, opened 2015-03-27
Genus2

jvoight:
Output from LMFDB Dublin meeting--addition of genus 2 curves.


JohnCremona:
That's very impressive for a week's work!  I suggest that a couple of us at least do some testing before we merge, as there are a lot of additions.

jvoight:
Yes, please do check!

We had a fairly stable version as of Wednesday, and double-checked things this morning.  But it's all brand new, so many things need testing.

The L-function component is still partly in progress: Andy Booker had a version which would compute everything on the fly, but it was very slow.  Now he and Drew Sutherland are working on precomputing what is needed for the L-function page.  This may be in process for a bit.

JV

JohnCremona:
1.  I was pleased to see that you included a test_genus2_curves.py file, but much less impressed that all the tests in there are EllipticCurves/Q tests, in fact it is the same file renamed!!!  Please can you put in some genuine tests?  I think this is rather important, and have been trying to get into the habit of adding a test every time I add some functionality.  It also means that if others make changes which break your code, they will notice!

2. It also appears that there are still functions in the file genus2_curve.py which have been copied from the elliptic curves file with no sensible changes.  Some of the route-ings seem nonsensical (the ones apparently based on elliptic curve labels? I was not sure).

3. I think you will need to add your database name to the list in base.py, otherwise the database of genus 2 curves will not be loaded at startup.  But the curves are being found, so perhaps that does not matter. Odd.

4. Lots of missing knowls....

As I said originally, this shows a huge amount of progress!  It will certainly not be long before we can merge this into beta.  I hesitate at the moment mainly because the current beta (as running on beta.lmfdb.org) is very close indeed to being made the new prod (at www.lmfdb.org), and I think it would be better for that to happen before this is merged into beta.

jvoight:
John,

Thanks very much for this feedback and taking the time to look things over.

If we ever do this again, we'll do our best to get something viewable
by the cognoscenti with some time for feedback, so we can integrate
more readily (and satisfy the conditions necessary for merging).  As
it stands, our goal was just to get something working, with the hope
that it could be shared, commented on, and polished in due course.  I
don't know what the development/branch/merge policies are so as to
help make this possible, so just let me know what you think is best.

(Personal comment: it's important to be rigorous about the code, but
not overly rigorous; my general feeling working with the existing code
base this past week is that it is spaghetti held together with duct
tape in many places.)

> 1. I was pleased to see that you included a test_genus2_curves.py file,
> but much less impressed that all the tests in there are
> EllipticCurves/Q tests, in fact it is the same file renamed!!! Please can
> you put in some genuine tests? I think this is rather important, and
> have been trying to get into the habit of adding a test every time
> I add some functionality. It also means that if others make changes
> which break your code, they will notice!

Sorry, we ran out of time before doing this.  I added a few.  Do they
pass for you, too?

> 2. It also appears that there are still functions in the file
> genus2_curve.py which have been copied from the elliptic curves
> file with no sensible changes. Some of the route-ings seem
> nonsensical (the ones apparently based on elliptic curve labels?
> I was not sure).

OK, there were a few auxiliary functions we forgot to delete.

I'm not sure which route-ings are nonsensical--could you be more
specific?  (I just went through and checked that all "@...route"
references make sense.)  I don't know exactly what's happening in all
cases, so we're happy to hear suggestions on this (with the proviso
that comments of the form "it would be nice if" will be added to a
queue to be worked on at some point in the future).

> 3. I think you will need to add your database name to the list in base.py,
> otherwise the database of genus 2 curves will not be loaded at startup.
> But the curves are being found, so perhaps that does not matter. Odd.

OK, added 'genus2_curves' to the list of readonly_dbs.  But this
didn't seem necessary.

> 4. Lots of missing knowls....

Yes, many!  Is this a block from merging, or a desiderata?

> As I said originally, this shows a huge amount of progress! It will
> certainly not be long before we can merge this into beta. I hesitate
> at the moment mainly because the current beta (as running on
> beta.lmfdb.org) is very close indeed to being made the new prod
> (at www.lmfdb.org), and I think it would be better for that to happen
> before this is merged into beta.

No problem; this is all spanking new, so it makes sense not to make it
public for awhile.

If indeed you're going to move beta.lmfdb.org -> www.lmfdb.org in the
near future, then you should just do that then only after merge in the
genus2 code.  (I understand from Fredrik that the classical=elliptic
modular forms code underwent and is undergoing substantial
modification, so much so that working outside his branch was
suboptimal.  Do you intend to wait for this development before moving
beta -> www?)

But I have a feeling we'll get more and better feedback if I can ask
people to "go to beta.lmfdb.org and try things out", rather than
"install git, download and run the lmfdb on your website, send your
public ssh key to...".  So we shouldn't wait too long.

JV

jwj61:
I haven't had a chance to look at the code.  From the conversation it sounds to me like there was excellent progress made with some things which would be desirable to fix/improve in the long run.  If a suggestion seems minor, it is a sign that things are looking good on the whole.

There are at least two levels of merging: merging the code into the code base, and then putting links on the home page so the public will start looking at it.  Unless the new code breaks existing parts of lmfdb, I don't see a problem with merging it even if it is not ready for the public.  We have other sections of lmfdb which are in this state (merged into the code, but no links to it from the homepage/sidebar), and some are probably less well developed than this new code.

If it this material is not ready for the homepage, the next large meeting is not that far away (Oregon in July).  That would be an opportunity to demo to a larger group and get feedback.

JohnCremona:
Good points both.  My suggestion to add the genus2_curves database to base.py was misguided though, since it does not work with that, and I had to comment it out.  Does anyone know why?  AFter that fix, the tests do work -- thanks a lot for adding these.

Apart from that JJ's suggestion is a good one, and consistent with what we currently have in beta for elliptic curves over number fields, namely that they are not navigable from the home page but are there if you click on certain inner links or put in a suitable URL.  So despite what I suggested earlier I would be happy for an early merge into beta.

JohnCremona:
By the way, about the knowls:  a good and easy easy thing for an LMFDB beginner to do.  As long as they have a login on the website they can create and edit knowls -- and of course to see the new knowls needed they will need to run your branch.  Apart from that no git knowledge is needed, and they can always be edited later.

Some other comments / suggestions (not meant to be done before we merge, just ideas):
    . You count the curves every time the root page is displayed.  Would be better to have a simpler form of the stats class I put in for elliptic curves.
    . Sato-Tate groups need to be knowl'd,
    . Impressive that when the Jacobian is a product that is shown!  But why not give the factors as elliptic curves instead of isogeny classes?  (Same in the "related objects" box).
    . In the Search Results table you have a column for GL2-type.  Other data could be listed here?

jvoight:
Hi John,

> By the way, about the knowls: a good and easy easy thing for an LMFDB beginner
> to do. As long as they have a login on the website they can create and edit knowls
> -- and of course to see the new knowls needed they will need to run your branch.
> Apart from that no git knowledge is needed, and they can always be edited later.

Not sure who to task with this, but maybe ask for volunteers at the
Oregon meeting?  There were several other places in the LMFDB where
these could use an update (particularly for stuff in development).

> Some other comments / suggestions (not meant to be done before we merge, just
> ideas):

Thanks!  Forwarding these suggestions to the group, so they are not forgotten.

> . You count the curves every time the root page is displayed. Would be better to
> have a simpler form of the stats class I put in for elliptic curves.

Yes, it would be better if this count is expensive to do.  It seemed
fast, so I assumed mongodb was caching this--is that wrong?

> . Sato-Tate groups need to be knowl'd,

"Sato-Tate group" needs a knowl, yes--or do you mean each kind of
Sato-Tate group (e.g. N(G_{3,3}) needs a knowl?)?  If the latter, I
worry this goes beyond "knowl", and actually requires real knowl-edge.
To explain all of the symbols involved one should really just read the
relevant paper (by Fite-Kedlaya-Rotger-Sutherland) and refer to the
right table.

Do we have a method for putting in bibliographic references for
technical bits like this?

> . Impressive that when the Jacobian is a product that is shown! But why not give
> the factors as elliptic curves instead of isogeny classes? (Same in the "related
> objects" box).

The Jacobian of a nice genus 2 curve is never isomorphic (over QQ) to
the product of elliptic curves--the decomposable Jacobians are
precisely those that fail to be in the image of the Torelli map.
(They are instead the Jacobians of two elliptic curves attached at
their origins, but then that's almost a tautology.)  We list when
there is an isogeny over QQ to a product of elliptic curves--but then
it doesn't make sense to list a particular pair of curves, as only the
isogeny classes are uniquely identified.

(We really do need to write things like this down somewhere...  maybe
they'll go in the paper that BSVY is writing.)

> . In the Search Results table you have a column for GL2-type. Other data could be
> listed here?

Anything in particular you think that belongs there?  We were trying
to keep it simple for reasons of speed, assuming one usually navigates
to a homepage for an object...  but maybe more is more?

JV

JohnCremona:
On 30 March 2015 at 22:09, jvoight <notifications@github.com> wrote:

> Hi John,
>
> > By the way, about the knowls: a good and easy easy thing for an LMFDB
> beginner
> > to do. As long as they have a login on the website they can create and
> edit knowls
> > -- and of course to see the new knowls needed they will need to run your
> branch.
> > Apart from that no git knowledge is needed, and they can always be
> edited later.
>
> Not sure who to task with this, but maybe ask for volunteers at the
> Oregon meeting? There were several other places in the LMFDB where
> these could use an update (particularly for stuff in development).
>

For these new knowls, it's also quite an easy job to be done bit by bit
between meetings.  ANyone with the branch can run it locally and (provided
they have a login on the website) can create/edit the knowls just by
clicking and typing.


>
> > Some other comments / suggestions (not meant to be done before we merge,
> just
> > ideas):
>
> Thanks! Forwarding these suggestions to the group, so they are not
> forgotten.
>
>
Of course.


> > . You count the curves every time the root page is displayed. Would be
> better to
> > have a simpler form of the stats class I put in for elliptic curves.
>
> Yes, it would be better if this count is expensive to do. It seemed
> fast, so I assumed mongodb was caching this--is that wrong?
>
>
I think that is indeed wrong.  You call

curve_count = db_g2c().curves.count()

and pass the value to the template, but next time that function is run it
will count again.  In this case I expect that it is very quick for mongo to
count *all* the records, so perhaps it is not serious -- for the elliptic
curves, I do some other couts and stats such as the max condictor and rank.



> > . Sato-Tate groups need to be knowl'd,
>
> "Sato-Tate group" needs a knowl, yes--or do you mean each kind of
> Sato-Tate group (e.g. N(G_{3,3}) needs a knowl?)? If the latter, I
>

It's tricky since in practice it will depend on how many different ones
there are.   Compare
http://beta.lmfdb.org/EllipticCurve/Q/11/a/1
under Galois Representations,  where these are encoded using Sutherland's
scheme;  we want at xome point to have "dynamic knowls" for each type, but
at presene there is just one knowl
http://beta.lmfdb.org/knowledge/show/ec.q.galois_rep_image which explains
the whole lot in one.

Another perhaps similar example is the new data about image of 2-adic reps
(from Rouse, not yet in beta) where there are hndreds of possibilities and
our current version links each label to Rouse's page.(e.g.
http://users.wfu.edu/rouseja/2adic/X150.html).  This is something I want to
ask others about in Oregon.


> worry this goes beyond "knowl", and actually requires real knowl-edge.
> To explain all of the symbols involved one should really just read the
> relevant paper (by Fite-Kedlaya-Rotger-Sutherland) and refer to the
> right table.
>

Agreed; rather like the Rouse-Zureick-Brown paper for 2-adic images.


>
> Do we have a method for putting in bibliographic references for
> technical bits like this?
>

No, I don't think so.  To be discussed!


>
> > . Impressive that when the Jacobian is a product that is shown! But why
> not give
> > the factors as elliptic curves instead of isogeny classes? (Same in the
> "related
> > objects" box).
>
> The Jacobian of a nice genus 2 curve is never isomorphic (over QQ) to
> the product of elliptic curves--the decomposable Jacobians are
> precisely those that fail to be in the image of the Torelli map.
> (They are instead the Jacobians of two elliptic curves attached at
> their origins, but then that's almost a tautology.) We list when
> there is an isogeny over QQ to a product of elliptic curves--but then
> it doesn't make sense to list a particular pair of curves, as only the
> isogeny classes are uniquely identified.
>
> I see, so J is never actually isomorphic to E1 x E2, just isogenous, and
then both E1 and E2 can be arbitrary members of their class.  I still think
I would prefer to see the statement that J is isogenous to such a product
(taking a curve in the class, say the optimal one, agreeing that this is
rather arbitrary) rather than saying that it is isogenous to a product of
isogeny classes.  Well, perhaps that does make sense, I am not sure.  Good
for a workshop discussion perhaps.


> (We really do need to write things like this down somewhere... maybe
> they'll go in the paper that BSVY is writing.)
>
>
Good idea.  (Can you get uniqueness by taking curves in the classes for
which the degree if isogeny from J is minimal?)


> > . In the Search Results table you have a column for GL2-type. Other data
> could be
> > listed here?
>
> Anything in particular you think that belongs there? We were trying
> to keep it simple for reasons of speed, assuming one usually navigates
> to a homepage for an object... but maybe more is more?
>
>
I'm not sure.  Similar search result tables for e.c.s list rank and torsion
(at least), but that is a little arbitrary.  ALso if the search asked for
curves whose mod p reps were surjective at (say 2 and 3) then the tables
gives the list of non-sirjective primes for each curve in the table, which
I think is a nice touch!

davidfarmer:
I think we should be able to finish the pending changes to the beta
site (mostly sidebar issues, yes?) relatively soon, and then move it
to the main site.  Then the genus 2 material can go on beta, at which
time editing knowls dosn't require running your own copy.

JohnCremona:
merged and pushed to beta, beta webserver restarted and indeed shows hyperelliptic curves in the left sidebar!

PR #218, opened 2015-03-31
commented out db authentication which did not work anyway

JohnCremona:
After email exchange with Harald.  The authentication process was originally intended to make most databases read-only except to users who could authenticate, but it did not work.  Also we cannot add new databases to this scheme since no-one remembers the password which was originally used.  One day this could be re-done, but for the moment this just removes a lot of the verbose start-up messages.


haraldschilly:
ok, looks good to me :smile_cat:

PR #219, opened 2015-04-01
Adding interface for data on class groups of quadratic imaginary fields.

jwj61:
To find the page, you have to go to a number field home page, and the link is in the "learnmore".

To actually download files from an account other than the server, you need the files and a link from your home directory set to match the lmfdb server.

Finally, moving files into place is in progress.  At this instant cl3mod8 and cl4mod16 are in place so you can request those files.  The rest should be in place in 48 hours.

### DIFF
--- a/lmfdb/number_fields/number_field.py
+++ b/lmfdb/number_fields/number_field.py
@@ -36,6 +36,8 @@ nfields = None
 max_deg = None
 init_nf_flag = False
 
+# For imaginary quadratic field class group data
+class_group_data_directory = os.path.expanduser('~/data/class_numbers')
 
 def init_nf_count():
     global nfields, init_nf_flag, max_deg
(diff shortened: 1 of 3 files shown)


JohnCremona:
It works for me!   I run my own lmfdb from atkin on which the data directory is mounted, which helped.
Perhaps we should add a warning about how large the files are?  I downloaded cl3mod8.77.gz which has about 150MB, quite a lot from home...
The "learn more about entry is on every number field page, correct?  Should it only be on pages of imaginary quadratic fields?  And from the number field search page possibly.  I will not mention the sidebar menu.

jwj61:
I added a statement about the sizes of the files (just after mentioning that they are in compressed format), and added a link to the number field front page (i.e., search page as suggested).  It seemed to work well at the end of the first paragraph.

This would put them in the learnmore of every number field homepage, but that can be tweaked later.  

It would be nice to get it onto the beta server soon (not to be too pushy).  I think the people who computed the data have been waiting on me for ages to get this in place so they can mention it in their paper.

JohnCremona:
That looks good -- thanks for the quick response to my suggestions!  I'll merge it now and push it to the beta branch so it will be available from beta.lmfdb.org.  An extra test would be nice too at some point, probably just to check that NumberField/QuadraticImaginaryClassGroups loads.

JohnCremona:
Unfortunately now beta.lmfdb.org now gives errors on (for example)  http://beta.lmfdb.org/NumberField/2.0.4.1
I will check the logs since we must fix this!

JohnCremona:
It's a problem with the .coefficients() method for polynomials, which is from a change in Sage.  It will go away if we run the server with Sage-6.5.  I can fix this tomorrow morning, but not right now... sorry.

PR #224, opened 2015-04-24
updated acknowledgements page

JohnCremona:
The main point was to correct the Hosting section which still said that everything was hosted in Seattle, but while we were at it we tidied up the lower half of the page, updated links, and added some workshops.


JohnCremona:
Reviewed by Aurel, Nicolas and Samuele.

PR #229, opened 2015-06-24
Fix issue from a feedback comment: the user supplied a polynomial

jwj61:
to look up in the number field database which contained non-ascii characters, causing an internal error.  Since the user should not need non-ascii characters to type their polynomial, we strip them.


jwj61:
It now does the substitution, and then uses decode/encode to strip characters.

PR #230, opened 2015-06-30
Fix reported by in computing Dirichlet characters for a C_16 field

jwj61:
by keeping integers sent to Conrey character package reasonably sized.

The bug was noticed by John C. from server logs.  When asking for characters, we just reduce the index modulo the character modulus first.

### DIFF
--- a/lmfdb/WebNumberField.py
+++ b/lmfdb/WebNumberField.py
@@ -497,7 +497,7 @@ class WebNumberField:
         while p in pram:
             p = P.next(p)
         fres = K.factor(p)[0][0].residue_class_degree()
-        a = p ** fres
+        a = (p ** fres) % f
         S = set(G[a].kernel())
         timeout = 10000
         while len(S) != self.degree():
@@ -505,7 +505,7 @@ class WebNumberField:
             p = P.next(p)
             if p not in pram:
                 fres = K.factor(p)[0][0].residue_class_degree()
-                a = p ** fres
+                a = (p ** fres) % f
                 S = S.intersection(G[a].kernel())
             if timeout == 0:
                 raise Exception('timeout in dirichlet group')


JohnCremona:
I have forgotten exactly what the error was.  Can we record here the URL which did fail before?  I would also like it to be normal practice for us, as with bugfixes in Sage, to add a test when a bug is fixed.
About the patch: it is fine, of course, though in both places it would be simpler to write

a = K.prime_above(p).norm() % f

If I were (re)writing this I would use "for P in K.primes_of_bounded_norm_iter(xxx):"   with some value of xxx instead of the "timeout" variable, set a=P.norm()%f and skip if gcd(a,f)>1.
OK, so I just did and will push to a branch at JohnCremona/lmfdb shortly.

JohnCremona:
By the way, when I searc for number fields with Galois group C16 it gives a not very nice error message about C16 being an invalid group label.  But C15 works fine, so I huess that it is just that there are no C16 fields in the database?  This should be imporved; shall we open a new Issue?

jwj61:
The bug you reported to me was

http://www.lmfdb.org/NumberField/16.16.48023489818559305679372288.1

Since you have a pull request for the same problem, I am not sure of the protocol here.

The question on C16 being a valid group label is a different issue.  If you look in the list of group labels, you see that they currently stop at degree 15.  At the time, we did not have higher degree fields.

We do have C16 fields (as above), but as the message says.  It won't be hard to add to the list of labels -- they are simply typed into the end of transitive_group.py in the dictionary "aliases".  I will add it as an issue.  When this gets settled, I can extend it.

JohnCremona:
I didn't want to tread on your toes!  So I did not make a pull request from my branch.  Since your fix does fix it I will merge it now -- my branch was based off that, so could be merged on top (if anyone like the new code more than the old).

PR #233, opened 2015-07-01
Added some more group aliases in degrees 16 and 17 (and fixed D7)

jwj61:
This is more like added data than new code.  Possibly, we should consider whether we prefer this in the database now that it is growing further.


JohnCremona:
I can confirm that putting C16 into the Galois group box on the number field search now works.
I agree that having a database instead of the a long list of code commands is the way to go, but not urgent.
I get the following odd behavious running locally (no problem on beta.lmfdb.org):  click on the knowl for "group labels" in the Number FIelds browse & search page;  the knowl opens, with title ""Galois group of a number field - naming convention".  Clicking on that title gives me a run-time error instead of displaying the full table (as happens on beta).  I think the line causing this is
a = users.find({"$or": a_query}, fields=["full_name"])
and the reason is that "fields" was renamed in newer pymongo to "projection" (or vice versa).  But this has nothing to do with the current patch, so I am merging it.

PR #235, opened 2015-07-03
Ecnfplot

nmascot:
Added code to plot elliptic curves defined over number fields. There is one curve drawn for each real embedding of the base field, all on the same picture.

Note: The code is a bit ugly as a workaround has had to be used so as to force Sage to display the legends. When the bug which prevents Sage from normally displaying legends is fixed, this code should be cleaned up.


davidfarmer:
Everything looks okay, but I didn't know how to find a page that actually shows the plots.

PR #237, opened 2015-07-20
Ecnf: better browse & search pages and enhanced pages for elliptic curves over number fields

JohnCremona:
The search & browse page for elliptic curves over number fields reflects the larger range of fields now included, and includes some highlights as direct links.
The search results page for ec/nf is improved.
For individual ecs, the web page gives better info about whether the model is minimal.
The code in kraus.py includes some function needed for the above, which have since been merged into Sage-6.8.beta? (and are now in the 6.8 release candidate) so after Sage-6.8 is released and we switch to it, that file will be removed and the imports from it currently in lmfdb/ecnf/WebEllipticCurve.py adjusted accordingly.


JohnCremona:
I am about to resolve the conflicts here so we can merge it: after this it will be much easier & better to ask for feedback on the current state of play with ecnf (elliptic curves over number fields).

PR #243, opened 2015-07-26
made previous maassdb changes work also on pymnogo<3

JohnCremona:
After merging the recent pull request on fixing maassdb errors I pushed to beta but then found that the fix worked only with pymongo >=3, while both beta and prod websites are still running pymongo-2.8 (and Sage-6.5!) so that for example http://beta.lmfdb.org/ModularForm/GL2/Q/Maass/ gives as error.

I will merge this and push to beta right away as it will fix that.


JohnCremona:
Despite what I wrote about this still gives an error: http://beta.lmfdb.org/ModularForm/GL2/Q/Maass/

To test this properly I will have to make myself a version of Sage with pymongo-2.8 installed, what a nuisance!

PR #245, opened 2015-07-27
G2plot

nmascot:
New code to plot hyperelliptic curves


nmascot:
My favourite one so far is 847.d.456533.1 
<http://beta.lmfdb.org/Genus2Curve/Q/847/d/456533/1>

PR #256, opened 2015-07-28
Support

JohnCremona:
New form of the contacts page (click on "contact" at the left of the very bottom line on every page" which no longer links to the feedback form but instead mentions the lmfdb-support mailing list.


jwj61:
This should be discussed with the group before merging.

Personally, I wonder if it sets the bar too high for a casual user who wants to give some sort of feedback.

Also, the old feedback form had the virtue that it automatically recorded user feedback and triggered e-mail to developers to look at it.  Here, we would have to log new issues.  This may be a problem since between workshops, developer activity seems low, so I don't know if feedback will get responses and the issues logged in a timely way.

JohnCremona:
On 28 July 2015 at 10:44, John Jones <notifications@github.com> wrote:

> This should be discussed with the group before merging.
>
Agreed!  I did this as a trigger & prompt for such discussion.

John

PR #257, opened 2015-07-28
Createit

davidfarmer:
See
/Genus2Curve/Q/169/a/169/1
for an example of the change.


JohnCremona:
I like this very much, but have not merged it yet since I think we want a discussion, e.g. about the colour (red? blue?).

JohnCremona:
OK == the change of colour is being dealt with elsewhere.

PR #258, opened 2015-07-28
Tors(U) replaced with \mu

nmascot:
In the description of the Galois structure of the unit group of a number field, replace the cumbersome "U/Tors(U)" with "U/\mu". Also, updated the knowl accordingly.

### DIFF
--- a/lmfdb/number_fields/templates/number_field.html
+++ b/lmfdb/number_fields/templates/number_field.html
@@ -184,7 +184,7 @@ function show_code(system) {
         <p><h2> {{ KNOWL('nf.multiplicative_gal_module', title='Multiplicative Galois Module Structure') }}</h2>
             <table>
             <tr><td>
-              $U_{K^{gal}}/\textrm{Tors}(U_{K^{gal}}) \cong$ {{nf.unit_galois_action_show()|safe}}
+              $U_{K^{gal}}/\mu_{K^{gal}} \cong$ {{nf.unit_galois_action_show()|safe}}
            </tr>
     {% if nf.unit_galois_action_type_knowl() %}
             <tr><td>


pascalmolin:
Shouldn't this be done at the macro level: a \tors actually rendered by \mu?

nmascot:
Pascal, what do you mean ? My point is that the torsion of the group of units U is the roots of unity, which I denote by \mu; but in general, the torsion of an Abelian group is not denoted by \mu.

pascalmolin:
Using macros makes it easier to have consistent notations between
knowls. The torsion of an elliptic curve as well as that of a ring of
integers may be denoted the same way, so I am asking if it is a good
idea to introduce the macro (by the way, this helps changing the
notation in the future, if one wants to revert to a more explicit one).
Anyway this was a question, the right answer should become obvious if
one lists the torsion groups occuring in lmfdb and their notation. 

Le 2015-07-28 23:02, Nicolas Mascot a écrit : 

> Pascal, what do you mean ? My point is that the torsion of the group of units U is the roots of unity, which I denote by mu; but in general, the torsion of an Abelian group is not denoted by mu. 
> 
> --
> Reply to this email directly or view it on GitHub [1].

 

Links:
------
[1] https://github.com/LMFDB/lmfdb/pull/258#issuecomment-125753177

nmascot:
Ah, I think I see what you mean ; a macro called "tors", which, given a generic Abelian group name "A", would return something like "A^{tors}", but, given "U", would return "\mu". I do not know how to do that.

pascalmolin:
No, I do not mean any complicated macro, sorry if this is confusing. I wanted to review the torsion notation currently used in knowls to see if there is a point in making them consistent (I object to changing the notation in one knowl if the same notation is used somewhere else, a macro helps in this case), but connecting to the lmfdb is too heavy for my connection here (China) and I won't manage!
Keep going with \mu, this is standard.

jwj61:
I honestly think the current notation will be understood by more readers.

JohnCremona:
I agree with John.  Can we close this (without merging)?

PR #263, opened 2015-07-28
Strudiness improvement

nmascot:
Made the plotting code resistant to the case where the curve is almost singular under some real embedding. This could lead to Sage believing that the curve is indeed singular, thus preventing the page from being displayed.

### DIFF
--- a/lmfdb/ecnf/WebEllipticCurve.py
+++ b/lmfdb/ecnf/WebEllipticCurve.py
@@ -261,7 +261,7 @@ class ECNF(object):
             ('Label' , self.label)]
         
         # Plot
-        if(E.base_field().signature()[0]):
+        if E.base_field().signature()[0]:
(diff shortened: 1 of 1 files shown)


lmfdb/ecnf/WebEllipticCurve.py
```diff
@@ -249,8 +261,7 @@ def make_E(self):
             ('Label' , self.label)]
         
         # Plot
-        n1 = len(E.base_field().embeddings(RR))
-        if(n1):
+        if(E.base_field().signature()[0]):
```
haraldschilly: no need for brackets around if's :blush:

nmascot: FIxed :) Now guess which is the programming language I learnt when I was in high school...

nmascot:
For instance, .../EllipticCurve/2.2.89.1/81.1/a/1 now renders properly.

PR #268, opened 2015-07-29
Genus 2 invariants: merging, normalization and factorization

JRSijsling:
This includes a number of changes to pages of genus 2 curves, such as

http://beta.lmfdb.org/Genus2Curve/Q/12500/a/12500/1

(1) The sections on invariants have been merged; as in the case of automorphisms and endomorphisms, the upcoming knowl will have to mention that there are two flavors of these, namely geometric and arithmetic.

(2) The Igusa-Clebsch invariants are now displayed with a colon to indicate that these are homogeneous invariants. Moreover, these invariants have been normalized by dividing out common integral factors (of course respecting the weights of these invariants). This proceeds by an intermediate Sage script; the actual Igusa-Clesch invariants of the given polynomials are therefore not lost and can still be recovered by modifying the source.

(3) Finally, the normalized Igusa-Clebsch invariants are also presented in factored form, as in the case of elliptic curves.

### DIFF
--- a/lmfdb/genus2_curves/templates/curve_g2.html
+++ b/lmfdb/genus2_curves/templates/curve_g2.html
@@ -52,29 +52,29 @@ src="http://code.jquery.com/jquery-latest.min.js"></script>
         </tr>
 	
         <tr>
-          <td> \( I_4 \) </td>
-	  <td>&nbsp;:&nbsp;</td>
+          <td> \( I_2 \) </td>
+	  <td>&nbsp;=&nbsp;</td>
 	  <td>\({{ data.data.invs[0] }}\)</td>
           <td>&nbsp;=&nbsp;</td>
 	  <td>{{ data.data.invs_factor_latex[0] }}</td>
         </tr>
 	<tr>
-          <td> \( I_6 \) </td>
-	  <td>&nbsp;:&nbsp;</td>
+          <td> \( I_4 \) </td>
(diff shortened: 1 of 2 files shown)


JohnCremona:
I am not sure about the notation I4 : -60 = −1⋅2^2⋅3⋅5 and similar.  Can we show this tomorrow?

There are some inefficiencies in the python code.  For example, scalar_mult(c,P,W) is one line as

return [c**w * p for p,w in zip(P,W)]

and the main normalize_invariants() function can also be streamlined a bit.  Good use of list comprehension in python is actually more efficient.

haraldschilly:
Technical comment, since you are working on this, maybe this should be done quickly here (it's just a small detail, but makes working with the template easier).

Goal is to avoid those double `data.data....` prefixes. Instead, a single `data.` should be enough (or even better, calling `data.` actually `curve.`)

To do this, at the bottom of genus2_curve (?):

```python
def render_curve_webpage_by_label(label):
    credit = credit_string
    data = WebG2C.by_label(label)
    return render_template("curve_g2.html", credit=credit, **data)
```

The `**` untangles the dictionary into keyword arguments.

JohnCremona:
I think Harald's suggestion will apply in many places as well as this -- these genus 2 pages were modelled on the elliptic curve ones, and use similar functions to the ones I wrote there.  We just fon't know python well enough!  Thanks.

PR #270, opened 2015-07-29
For number fields, use an integral basis which is closer to a power basis

jwj61:
You can test with number field 3.1.23.1.  Currently, the basis is not a power basis.  With this, it is.  Of course, lots of number fields get simpler integral bases with this change.

This also moves generating the integral basis to the webnumberfield class.  This should give more consistency to other objects, and it will first check the database to see if it has an integral basis prestored.  That data needs to be added by an external script (but it has been tested).


JohnCremona:
I had a quick look but will need to look more carefully in the morning.

jwj61:
Oops, forgot to branch on second change.  This now contains a second feature: addition of class numbers to number field knowls.

PR #271, opened 2015-07-29
browse search

danyasaki:
Fixed functionality on browse search page.  Uses next and previous buttons like ecnf pages.  Fixed gl2type.  Added search for 2-selmer, torsion, wpts.  added refined search feature on search results.


lmfdb/genus2_curves/genus2_curve.py
```diff
@@ -123,6 +165,7 @@ def genus2_curve_search(**args):
             query[fld] = info[fld]
     for fld in ['aut_grp', 'geom_aut_grp']:
         if info.get(fld):
+            print "*********************", info[fld]
```
haraldschilly: there is a leftover print statement

JohnCremona:
I think Dan is working on a follow-up commit, so wants us to wait until he is done with that before merging.

PR #272, opened 2015-07-29
Styling of non-existent knowls, such as those at the top and bottom

davidfarmer:
of most elliptic curve pages.

### DIFF
--- a/lmfdb/templates/knowl-defs.html
+++ b/lmfdb/templates/knowl-defs.html
@@ -71,7 +71,7 @@
   {# knowl does not exist. wrong knowlid or opportunity to create one! #}
     {% if user_is_authenticated -%}
       <div class="knowl knowl-error">
-      <a href="{{ url_for('knowledge.edit', ID=knowlid) }}">'{{knowlid}}' Create it</a>
+      <a href="{{ url_for('knowledge.edit', ID=knowlid) }}">'{{knowlid}}' [create knowl]</a>
       </div>
     {%- else -%}
       {% if show_missing -%}<span class="knowl knowl-error">'{{knowlid}}'</span>{%- endif %}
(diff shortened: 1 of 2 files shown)


JohnCremona:
Any reason for keeping in the string "Create it"?  Otherwise good, I like the smaller size & colour!

jwj61:
I think keeping "create it" sounds like a good idea -- clearer to someone who someone working on lmfdb who may not figure out the color convention.

JohnCremona:
So for consistency should we reinstate "Create it" in the normal situation which we changed yesterday?  e.g. on http://beta.lmfdb.org/Genus2Curve/Q/169/a/169/1 ?

JohnCremona:
Perhaps have '[create knowl]' instead of 'Create it', i.e. put the words in brackets and be more specific that clicking will create the knowl.

davidfarmer:
I am okay with '[create knowl]' instead of 'Create it'.

What is the right way for me to make that change?

davidfarmer:
I think we need the "Create it" in the case of top and
bottom knowl which have no useful title.

But in the case of knowls in regular text, a good example is
at the end of the Conductor knowl on

/Genus2Curve/Q/
,
the words "Create it" are obtrusive.  I don't think the
learning curve is too steep for seeing what knowls need
to be created -- assuming the new purple color is clear
enough.

JohnCremona:
Make a new commit on top of the current one and push to your github, same branch name createit2.  Then github will automatically update the pull request to include this additional commit.

JohnCremona:
PS I'm OK with not having the extra words normally, you are right that these can occur in the middle of text and the extra words would distract.

davidfarmer:
I pushed the update.

JohnCremona:
Good -- tested and about to merge...

PR #282, opened 2015-07-29
Download elliptic curves

adeines:
Adds functionality to download searches of elliptic curves.


JohnCremona:
Much faster now -- great!  I would suggest simplifying the output as followed.  For magma and sage the backslash characters are simply redundant;  for gp they are also provided that you put { on the line before and } on the line after the multi-line expression.  I just tried all three and they do then read in fine.

But after talking it over with JJ we decided that it is fine as is -- merging!

PR #283, opened 2015-07-29
Streamlining display of isogeny classes

JRSijsling:
The conventions on isogeny matrices and graphs are now no longer mentioned by default on every individual page, but rather mentioned in the knowl, thus streamlining the display of the pages.

Also, a link to the knowl on isogeny matrices has been added to the pages of elliptic curves over the rationals.


JohnCremona:
Tested, looks good.

PR #286, opened 2015-07-30
Added lines to copy-paste defining equations into Sage and Magma

JRSijsling:
The genus 2 pages (of specific curves) can now display lines of Sage and Magma code that enable copy-pasting of the curve in question into the user's sessions.

The current code snippets can probably be improved; on the other hand, I see no way to avoid defining a variable in Sage, since the HyperellipticCurve command there requires polynomials. Also in Magma this seems tricky; the coefficient lists of the constituent polynomials can be empty, add Polynomial([]) then gives an error. I have therefore defined a base polynomial ring first in both cases. Any suggestions to work around this are very welcome.

Other small changes:
(1) Changed KNOWL_ID to g2c instead of ec.q;
(2) In the list of downloads in the sidebar, changed the name to suggest downloading all stored data. Note that regardless of the name this is actually not possible, since the relevant code is not yet available.

### DIFF
--- a/lmfdb/genus2_curves/templates/curve_g2.html
+++ b/lmfdb/genus2_curves/templates/curve_g2.html
@@ -26,12 +26,20 @@ function show_code(system) {
 src="http://code.jquery.com/jquery-latest.min.js"></script>
 <script src="http://aleph.sagemath.org/embedded_sagecell.js"></script>
 
-    {% set KNOWL_ID = "ec.q.%s"|format(data['label']) %}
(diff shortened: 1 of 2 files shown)


JRSijsling:
Extra comment for clarification: the Sage snippet now typically looks like

R.<x> = PolynomialRing(QQ); C = HyperellipticCurve(R([1, 0, -2, 0, 2, 0, -1]), R([]))

whereas the Magma one then becomes

R<x> := PolynomialRing(Rationals()); C := HyperellipticCurve(R![1, 0, -2, 0, 2, 0, -1], R![]);

PR #287, opened 2015-07-30
Move label search

sanni85:
the box for searching by label has been moved before the block with search queries all over the website (hopefully all occurrences.....)

### DIFF
--- a/lmfdb/ecnf/templates/ecnf-index.html
+++ b/lmfdb/ecnf/templates/ecnf-index.html
@@ -38,6 +38,18 @@ Curves with interesting properties:
 </p>
 
 
+<h2> Find a specific curve by label</h2>
+<form>
+<input type='text' name='label' size=25 example="2.2.5.1-31.1-a1">
+<button type='submit' name='jump' value='label' >label</button>
+<br><span class="formexample">e.g. 2.2.5.1-31.1-a1</span>
+{% if err_msg %}
+<br>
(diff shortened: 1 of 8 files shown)


JohnCremona:
David volunteered to go through this one, and I suggest that we don't merge others until that is done.

davidfarmer:
At /EllipticCurve/,  the following appears at the bottom of the page
 Raw field data: [['the rational field', (('1.1.1.1', ['/EllipticCurve/Q/', 'Q']),)], ['real quadratic fields', <generator object <genexpr> at 0x11197f3c0>], ['imaginary quadratic fields', <generator object <genexpr> at 0x11197f410>], ['cubic fields', <generator object <genexpr> at 0x11197f460>]]

davidfarmer:
At /Genus2Curve/Q/, the "Find a specific curve by label" appears twice.

davidfarmer:
Except for the two comments above, everything looks okay for this merge.

If this is holding up other things, it might be reasonable to fix those issues later?

Note that this pull request does not take care of the left/right justification of the search
tables: it only puts the "find" above the "search".

sanni85:
for EllipticCurve/, that part of code appears because in the html is under the check
{% if DEBUG %}, so if you run sage -python start-lmfdb.py --debug -t you will see it while if you run sage -python start-lmfdb.py -t  you will not see it. Probably it needs an edit but that was not the aim of this pull request.

Fixing the genus 2 now.

PR #297, opened 2015-07-30
For6.8

fredstro:
This pull request contains a few small patches which should make the old version of modular forms run under Sage 6.8.

### DIFF
--- a/lmfdb/modular_forms/elliptic_modular_forms/__init__.py
+++ b/lmfdb/modular_forms/elliptic_modular_forms/__init__.py
@@ -6,7 +6,7 @@ import flask
 default_prec = 10   # The default number of terms in a q-expansion
 default_bprec = 53  # The default number of bits of precision to compute for floating point data
 default_display_bprec = 26  # The default number of bits of precision to display for floating point data
-emf_version = 1.1
+emf_version = 1.2
 EMF_TOP = "Holomorphic Modular Forms"  # The name to use for the top of this catergory
 EMF = "emf"  # The current blueprint name
 emf = flask.Blueprint(EMF, __name__, template_folder="views/templates", static_folder="views/static")
--- a/lmfdb/modular_forms/elliptic_modular_forms/__init__.py
+++ b/lmfdb/modular_forms/elliptic_modular_forms/__init__.py
@@ -6,7 +6,11 @@ import flask
 default_prec = 10   # The default number of terms in a q-expansion
 default_bprec = 53  # The default number of bits of precision to compute for floating point data
 default_display_bprec = 26  # The default number of bits of precision to display for floating point data
-emf_version = 1.2
+from sage.all import version 
+if float(version().split(',')[0].split(' ')[-1]) >= 6.8:
+    emf_version = 1.2
+else:
+    emf_version = 1.1
 EMF_TOP = "Holomorphic Modular Forms"  # The name to use for the top of this catergory
 EMF = "emf"  # The current blueprint name
 emf = flask.Blueprint(EMF, __name__, template_folder="views/templates", static_folder="views/static")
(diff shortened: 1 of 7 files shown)


JohnCremona:
Fredrik, can you give some URLs which did not work but now do?  It is still the case that not all the tests pass.

fredstro:
All tests now pass. 
Example of urls which exists and did not work are
http://localhost:37777/ModularForm/GL2/Q/holomorphic/1/12/0/a/
where I tested for a larger number of digits in the Stake parameters than what I now displayed. 
Otherwise the main problems were that the urls in the tests used the new character label format while the pages used the old (Sage) numbering.

JohnCremona:
lmfdb.test_lfunction.LfunctionTest  fails for me!

fredstro:
The problem here was that the previously computed WebModforms worked with my patch only for the non-trivial character (there was another incompatibility with the cyclotomic fields in Sage 6.8) and I therefore deleted these from the database to force them to be recomputed.  
However, I didn't think about that as soon as someone with an older sage version viewed the website then these spaces were recomputed (and the new copies incompatible with 6.8)... 
I now solved that (in the way I should have from the beginning) by incrementing the version number (from modular_forms.elliptic_modular_forms.version) from 1.1 to 1.2. in the case sage is running 6.8 and this way allowing the two versions running without interference. 
After these changes all  tests now pass again. 
It would be good if this could be tested by someone else and then merged into beta at least and preferably alpha soon.

JohnCremona:
Testing again -- the sooner this gets in the better.  Once this is merged into beta then no more problems should be caused by people causing modular forms to be computed & stored on the beta server even though it runs 6.5, and similar for the prod ('alpha') server, I think.  Testing...

JohnCremona:
Great -- all tests pass (except one about dbnames which has already been fixed in the upstream master).  I am about to merge this and push to beta.

JohnCremona:
I did that but then saw that Fredrik's pull request was into beta not master -- Fredrik, that was not the way we normally do things but I should have seen it.  Anyway all I had to do was locally pull from upstream/beta and push to upstream/master, and the end effect was the same.
The beta server has been restarted with this.

JohnCremona:
I forgot to mention one thing:  Fredrik, your test for Sage version >= 6.8 needs fixing, since you convert the version to a float which (as I just verified) fails on 6.9.beta0:
ValueError: invalid literal for float(): 6.9.beta0
It's easier to get the version string from sage.env.SAGE_VERSION, and in a similar circumstance I used
v = sage.env.SAGE_VERSION.split(".")
if int(v[0])<6 or int(v[1])<8:
   # old version
else:
  # new version

I will make aptch for this right away in case anyone else is using a Sage version which does not look like a float!

fredstro:
> On 4 Aug 2015, at 09:29, John Cremona <notifications@github.com> wrote:
> 
> I did that but then saw that Fredrik's pull request was into beta not master -- Fredrik, that was not the way we normally do things but I should have seen it. Anyway all I had to do was locally pull from upstream/beta and push to upstream/master, and the end effect was the same.
> The beta server has been restarted with this.
> 
> —
> Reply to this email directly or view it on GitHub <https://github.com/LMFDB/lmfdb/pull/297#issuecomment-127524257>.
> 
If that is normally not the way we do things then you really need to explain *the way we do things* better since I have been around a long time and I thought that that was exactly how we did things. 
You were talking the whole time about it first being merged into beta (and switch beta to 6.8) and then to master so I assumed that the natural thing would first be to first make a pull request to beta and if that works there then we would take it to master. 

Fredrik

JohnCremona:
Our version/branch names -- which were not invented by me! -- are certainly confusing.  

prod is the branch with the code running on www.lmfdb.org
beta is the branch with the code running on beta.lmfdb.org
master is the development branch.

A better name for prod would be master and a better name for master would be develop.  I agree!  And the current state of things needs to be explained better if we want developers to make pull requests to the (de facto) develop branch.  However, github seems to expect by default that pull requests are to master, so that every other pull request from last week was indeed to master...luckily this did not matter.
I just made the small change in Pull Request #311 so if you can check that over I will merge it...

PR #301, opened 2015-07-31
Added conversion of geometric invariants in genus 2

JRSijsling:
The pages for individual curves in genus 2 now have a clickable conversion for geometric invariants. The default choice for these, which is displayed when entering the page, are the Igusa-Clebsch invariants. Clicking one of the other choices (Igusa and G2) gives the other commonly used invariants. As before, the Igusa-Clebsch and Igusa invariants are normalized before being displayed.

### DIFF
--- a/lmfdb/genus2_curves/templates/curve_g2.html
+++ b/lmfdb/genus2_curves/templates/curve_g2.html
@@ -68,6 +68,7 @@ src="http://code.jquery.com/jquery-latest.min.js"></script>
           <td>{{ data.data.disc_factor_latex }}</td>
         </tr>
         </table>
+        <br>
         <div class='igusa_clebsch'>
             Igusa-Clebsch invariants:
         </div>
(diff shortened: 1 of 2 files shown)


davidfarmer:
I suggest changing "Convert to preferred geometric invariants" to
"Available geometric invariants".
And also remove "(default)".

Try replacing the comma "," between the different forms by a slash "/" to see
if it looks better, and maybe put a bit of extra space around the slashes.

It would be really good if the invariants being displayed had their name in a
different color. (So initially, " Igusa-Clebsch" would be highlighted, and if I
click on G2 then "G2" becomes highlighted.)  If that is too difficult, then we
can leave it as-is and submit the improvement as a feature request issue.

JRSijsling:
All right, I updated the layout. The final point is something that I cannot do in the time remaining in Oregon. I will keep trying though, and we could indeed make this a feature request.

JRSijsling:
The chosen invariants are now preceded by a line telling the user what type follows. After this, another line follows where alternative choices are clickably presented.

PR #307, opened 2015-08-01
L-function home pages in arithmetic normalization

davidfarmer:
For L-functions of genus-2 curves, the only L-functions currently stored in the database, we make two versions of the L-function homepage.  One is in analytic normalisation (default when not logged in and not on BETA); the other is in arithmetic normalisation.  There are pseudo-buttons at top to indicate which.  Also, the Euler product information has been changed to include more specific data in table form and minor changes to the presentation of other elements on the page.  These include the functional equation, the colour of zeroes, particular values, and the Dirichlet series.

Sample pages:
    http://localhost:37777/L/Genus2Curve/Q/202720/a/  
              (looks different whether or not logged in)
    http://localhost:37777/L/EllipticCurve/Q/60420.b/
              (this is always in analytic normalisation, but buttons on top appear only in BETA version)

The L-functions code is still messy.  We plan to work on that next, along with other improvements to the L-function homepage.


JohnCremona:
I have read through the code and tested the relvant pages (L-functions of genus 2 curves), and checked that tests pass.  Here are some comments for future iterations, not required to be done before this is merged:

   * In the code: (1) where you have two (or more) fields such as self.diriechlet_coefficients_arithmetic and self.dirichlet_coefficients_analytic, or self.texname and self.texname_arithmetic, I suggest that (e.g. in the second case) it would be tidier for self.texname to be a dict with keys 'analytic' and 'arithmetic', so that in the template one would refer to <object>.texname.arithmetic or <object>.texname.analytic. (2) There's at least one left-over print statement from debugging. (3) There's a comment / question "DF: why is this needed if pulling from database?" which needs to be answered!

   * On the web page: (1) The Euler factors only show up in the Arithmetic view.  Is that intentional? (2) I would prefer the normalization buttons to be right-shifted so they do not come above Dirichlet Series, but are still near the top;  and their appearance could make clearer than at present that these are buttonds for selecting the desired normalisation, and also give an indication of the current normalisation.

About merging this:  my only hesitation is that there is a lot more work to be done before all the L-function pages are rewritten in this way, and it is fine for beta to be in this half-way stage, but becuase of what Fredrik is doing we would like to push what is currently running on beta to the production branch as soon as possible, since the current production branch under 6.5 has nasty side-effects when it stores old-style modular forms -- though from what Fredrik said on the other pull request this may not be an issue.  If others think that there is no great hurry to push current beta soon to prod then this is not an issue.

davidfarmer:
Things like coefficients_arithmetic and coefficients_analytic
are artifacts from the old way of doing the L-function home
pages.  Recent changes have focused on making the home pages
nicer for the L-functions of genus 2 curves, and not breaking 
any of the old code.  We will fix things like the names of      
those fields when we start refactoring (which is going to take
a bit of effort).  Similarly for DF's question in a comment.

The Euler factors only show up in the arithmetic view because
in that case they look nice.  We haven't figured out how to 
best present the Euler factors in the analytic normalization.
Maybe we will show the Satake parameters?  There is also the
issue of whether that information is computed or stored in
the database.

Normalization buttons are now centered at the top.

All the arithmetic pages are BETA only, so this can be
pushed to production.

All other suggested changes were made, in addition to some
other improvements: show all bad local factors, fix bug that
missed the factor at 2 when 2 was good.
http://localhost:37777/L/Genus2Curve/Q/301945/a/

JohnCremona:
That all sounds (and looks) good but I may have misunderstood something: when run in non-beta mode (i.e. leave out --debug on the command line) you still see the buttons for normalization since the template's else clause includes them.  Not intentional, I think?

davidfarmer:
Sorry.  Will wrap those in BETA, too.

JohnCremona:
OK, I'll merge when you have.  (There should be an easier way for reviewers to make small additional changes themselves.  I could have taken my local copy of your branch, added a commit, pushed to my github and then made a pull request to your github fork with that, then if you agreed to it the original pull request -- this one -- would be updated.  But that seems like a lot of effort!)

davidfarmer:
The tabs now only show in beta when the user is logged in.

JohnCremona:
Hmmm only 75% on that:  fine with no beta;  under beta and logged in, I see Arithmetic in green, but under beta and not logged in I see no buttons and the old analytic display, while I thought I would see the buttons, Analytic in green and theanalytic display.

Feel free to say that this is work in progress, or that I habe understood your intentions!

davidfarmer:
Our intention when not logged in was to show no buttons.

In a future iteration, when not logged in we will show
the analytic button but *not* the arithmetic button,
because it is not really a button and there is no
arithmetic option for most pages.

davidfarmer:
Maybe the buttons are okay now?

JohnCremona:
Yes!

JohnCremona:
... and pushed to beta server.

PR #310, opened 2015-08-03
Simplify html, and some styling

davidfarmer:
The main thing I did was replace the code for switching between invariants,
which repeated everything twice, by simpler code which does not repeat anything.

And since I was working on that page, I implemented some of the layout suggestions
which were lingering from the workshop.

http://localhost:37777/Genus2Curve/random
The code I changed involves the options after "Alternative geometric invariants"

### DIFF
--- a/lmfdb/genus2_curves/templates/curve_g2.html
+++ b/lmfdb/genus2_curves/templates/curve_g2.html
@@ -1,7 +1,12 @@
 {% extends 'homepage.html' %}
 {% block content %}
 <style>
+/* seems to be unused?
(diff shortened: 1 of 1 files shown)


JohnCremona:
Looks good and test pass.
Two things for the future: (1) when I put a label into the box "find a specific curve by label" and the label has a trailing space, it gives an error, as also when entring any invalid label.  Better error checking is needed. (2) On the page in question (for an individual curve) I don't like the blanks in the endomorphism block: if not known the text "not known" / "not computed" should be put, not just blank space.  And I would put "not of GL2-type after the table, not before.  But this is for another ticket.

PR #311, opened 2015-08-04
fixed Sage version code in emf initialization

JohnCremona:
Small left-over from pull request #297: now handles Sage versions more robustly (and does not crash on version 6.9.beta0, for example).


JohnCremona:
Thanks -- now I can do my development with Sage-6.9.beta0 ;)

PR #313, opened 2015-08-04
Random

danyasaki:
Now the random curve page shows the curve label in the url, like elliptic curves page.
See 
http://127.0.0.1:37777/Genus2Curve/Q/random


JohnCremona:
Code looks good and I'll test...OK!
A question for the genus 2 team would be:  if and when genus 2 curves over other (number) fields are added, will they be handled by the code already in lmfdb/genus2_curves, or by a new module, in the same way as elliptic curves over Q are handled separately?  Either way, there's a case for the /Q being inserted in the prefix for this module in the file __init__ instead of in each route tag separately.

PR #314, opened 2015-08-04
mostly cosmetic fixes (including some knowl id changes) and one bug f…

AndrewVSutherland:
cosmetic fixes, including some knowl id changes, and a small bug fix (no longer attempts to factor zero)

### DIFF
--- a/lmfdb/genus2_curves/genus2_curve.py
+++ b/lmfdb/genus2_curves/genus2_curve.py
@@ -178,7 +178,7 @@ def genus2_curve_search(**args):
             query[fld] = info[fld]
     for fld in ['aut_grp', 'geom_aut_grp','torsion']:
         if info.get(fld):
-            query[fld] = eval(info[fld])
+            query[fld] = map(int,info[fld].strip()[1:-1].split(","))
     if info.get('ic0'):
         query['igusa_clebsch']=[info['ic0'], info['ic1'], info['ic2'], info['ic3'] ]
         
--- a/lmfdb/genus2_curves/genus2_curve.py
+++ b/lmfdb/genus2_curves/genus2_curve.py
@@ -17,19 +17,21 @@ from lmfdb.genus2_curves.web_g2c import WebG2C, list_to_min_eqn, isog_label, st_
 import sage.all
 from sage.all import ZZ, QQ, latex, matrix, srange
 q = ZZ['x'].gen()
-# credit_string = "KNOWL('g2c.credit', title='The Genus 2 Team')" 
 credit_string = "Andrew Booker, Andrew Sutherland, John Voight, and Dan Yasaki"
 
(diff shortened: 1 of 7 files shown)


JohnCremona:
Some comments:
 (1) st_group_dict is defined as a set, but used as a dict.  That does not look right. 
 (2) applying eval() to a string entered into a search box is a bad idea since this can cause arbitrary code to be run but the server!  I think that info['fld'] is already a string in the place this is used, so why is any conversion needed?
(3) I see the point of the zfactor() function, but it does not actually factor n when non-zero!  Ias that intentional?
I have not yet looked at the pages to judge the cosmetic element...

AndrewVSutherland:
I'll fix these (some were remnants from changes Dan Yasaki had in progress that he transferred to me).  Should I submit a new pull request once I have done this?

Let me know if you spot anything else!




On August 4, 2015 2:50:06 PM EDT, John Cremona <notifications@github.com> wrote:
>Some comments:
>(1) st_group_dict is defined as a set, but used as a dict.  That does
>not look right. 
>(2) applying eval() to a string entered into a search box is a bad idea
>since this can cause arbitrary code to be run but the server!  I think
>that info['fld'] is already a string in the place this is used, so why
>is any conversion needed?
>(3) I see the point of the zfactor() function, but it does not actually
>factor n when non-zero!  Ias that intentional?
>I have not yet looked at the pages to judge the cosmetic element...
>
>---
>https://github.com/LMFDB/lmfdb/pull/314#issuecomment-127712698

-- 

JohnCremona:
A new pull request is not needed, just update your branch and the pull
request will notice.  I'll come back to this tomorrow...

AndrewVSutherland:
Hi John,

(1) st_group_dict is in fact a dict, it's just a silly one (the key and 
value are always the same).  This is as it is in the current version, I 
just changed the name "st_temp" to the more informative "st_group_list". 
  I believe it was originally implemented this way just to match the 
other drop down lists: the key is the database query value and the value 
is the string displayed -- in the case of the Sato-Tate groups these two 
things happen to coincide but in general they don't.

(2) I got rid of the eval().  It is necessary to parse the string 
because these fields are all stored in the database as arrays of ints, 
not strings.  The search for these items in the current version of 
beta.lmfdb.org does not work at all because of this mismatch.

(3) Fixed zfactor() which should indeed factor nonzero inputs -- thanks 
for noticing!

The changes to (2) and (3) have been committed, I don't think (1) 
requires any change but let me know if you do.

Cheers,

Drew

JohnCremona:
Sounds good.  I had notes that in general, input parsing was rather
primitive here (doing a search by label crashes if no object with the label
exists, even if you put in a trailing space!) but that should be fixed
independently once David Roe's general-purpose parsing module is complete.
I'll have another quick check and then expect to merge this -- thanks.

JohnCremona:
OK!

PR #315, opened 2015-08-05
Getting started -- updated instructions

JohnCremona:
I have updated the Getting Started instructions after various experiences at the Oregon workshop.  You can see the changes without getting the branch by going to https://github.com/JohnCremona/lmfdb/blob/GettingStarted/GettingStarted.md

### DIFF
--- a/GettingStarted.md
+++ b/GettingStarted.md
@@ -2,13 +2,16 @@ Installation
 ============
 
 * To develop and contribute new code, see below on Sharing Your
-  Work. If you only want to run a copy of the site move into a new
-  directory and type
+  Work. If you **only** want to run a copy of the site, move into a
+  new directory and type
+
   ```
      git clone git@github.com:LMFDB/lmfdb.git lmfdb
   ```
-  but follow the instructions below (under Sharing Your Work) instead
-  to set up your own fork on github and clone from there instead.
+
+  but **follow the instructions below instead** (see *Code development
+  and sharing your work*) to set up your own fork on github and clone
+  from there if you want to carry out development on the code.
 
 * Make sure you have sage (>=6.8) installed and that
   `sage` is available from the commandline.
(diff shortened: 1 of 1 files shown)


davidfarmer:
I find that the instructions are growing too detailed.  Maybe it would be good to have a streamlined
set of instructions specifically for new developers, which makes a clear distinction between the one-time
work of setting things up, and the every-time work of contributing new code.

If that sounds reasonable, my suggestion would be that I accept this pull request, and then I
start working on the next iteration.

JohnCremona:
You are right.  I do tend to be verbose.  We could restrict this particular document to literally getting started, and move some of the rest to the Development page?  That one is bit of a mix between things everyone should know and technical stuff only "release managers" need to know.

jwj61:
I don't know an easy way to do this, but it seems like this document could benefit from knowls.  There are several parts which apply to a small percentage of people.  Putting them in knowls would make things simpler for most people, but it would still be available for those that need it.

davidfarmer:
My plan was to give the minimal instructions needed by
someone who wants to become a developer.  The other items
can go elsewhere.

jwj61:
A reasonable idea.  My candidates for elsewhere:

  memcache
  running your own mongo server
  running off some other port (for multiple people on the same server)
  lmfdb on windows

JohnCremona:
I agree with the above, and was about to suggest at least 3 of them myself (the ones I never use!).

Is there actually any point in someone running the website locally and *not* setting themselves up with git etc to do development?  They don't need to do that to edit knowls, so it is not clear to me why anyone would want to do that.  This would be a simplification.

JohnCremona:
Another possibility would be to move these documents to https://github.com/LMFDB/lmfdb-collab.  And there, we could start to use the Wiki more.  I can see that possibly being useful at ICERM (and before)

jwj61:
I agree that there is not much point in explaining how to run a local copy of the main code rather than a fork, while having the instructions may mislead some people into making the first clone.  Your instructions are an improvement, but taking out the first clone instructions may be better still.

JohnCremona:
David has volunteered to make the next version, so let's see what he comes
up with!

John

PR #317, opened 2015-08-05
Magma

danyasaki:
Added magma commands to curve pages.  To test, go to a specific curve page, e.g., 
http://localhost:37777/Genus2Curve/Q/4096/a/65536/1
and click magma at the top.

### DIFF
--- a/lmfdb/genus2_curves/templates/curve_g2.html
+++ b/lmfdb/genus2_curves/templates/curve_g2.html
@@ -51,13 +51,17 @@ src="http://code.jquery.com/jquery-latest.min.js"></script>
     </div>
     
     <h2> {{ KNOWL('g2c.minimal_equation', title='Minimal equation') }} </h2>
-    <p> ${{ data.data.min_eqn_display }}$ </p>
-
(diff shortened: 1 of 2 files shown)


JohnCremona:
Dan -- could you take a look at the branch "code" in my github fork and see what you think?  It is unfinished (I was working on it this afternoon) and I think it is a better approach to adding code to home pages than what we have done in the past, and will also make it very much easier to provide a download of all the commands in magma (or sage or gp).
If you like the concept you might want to make similar changes to this branch...

danyasaki:
Hi John,

I agree that your way is a better approach to adding code.  I would suggest not using LaTeX \tt for the prompts as it does not align with the code.  You can just do  Magma > instead of $ \tt Magma > $ and it will line up.  

I will mimic your code in this branch.  I am closing this pull request, and will put in another one once everything is set up.

JohnCremona:
OK -- of course my new way is still experimental!  There is still quite a
lot of repetition in the html, but I have not yet worked out a way of
avoiding it.

John

JohnCremona:
I agree about the prompts.  Sage is easiest and I now just have "sage:".  For the others I now have "gp:" and "magma:", although the actual prompts are "?" and ">".

JohnCremona:
Dan, see pull request #318 for a great model to follow!

PR #318, opened 2015-08-06
Code snippets for sage/pari/magma for elliptic curve pages

JohnCremona:
I have completely rewritten the way code snippets for sage/pari/magma are put into elliptic curve home pages (over Q).  All the actual code strings for all 3 are now defined in the python file, and there is a new macro in the html template which makes adding each code snippet just one very simple line.

I am quite proud of this and think it could be a model to be used elsewhere; it will also make it very much simpler to provide the facility to fownload all the code from a home page -- but I left that for another ticket.

Logos: currently not used, but there is provision to add them again if we find suitable ones.
Names: "Sage" is now officially "Sagemath" so I used that; and we used "pari" where we really meant "gp" so I have put "pari/gp" -- this is in the selction buttons at the top of the page.
Prompts: Sage has "sage: ", gp just has "? " and Magma just has "> ".  The latter two don't look good, and neither did "gp? " and "magma> ", so I went for consistency and have "gp: " and "magma: ".   These can be changed by changing *just one* place in web_ec.py.

The code snippets themselves are essentially unchanged, except that in all three cases I showed how to construct the curve from its label.  (This only works in gp and Sage if an optional package is installed, but anyone using this would surely know that.)

Reviewers, feel free to make comments about the aesthetics, which can be changed, but don't let such issues distract you from the beauty of the new code!

### DIFF
--- a/lmfdb/elliptic_curves/templates/curve.html
+++ b/lmfdb/elliptic_curves/templates/curve.html
@@ -24,6 +24,7 @@ function show_code(system) {
         src="http://cdn.mathjax.org/mathjax/latest/MathJax.js"></script>
 <script type="text/javascript"
 src="http://code.jquery.com/jquery-latest.min.js"></script>
+{#
 <script src="http://aleph.sagemath.org/embedded_sagecell.js"></script>
 
 <style type="text/css">
@@ -40,10 +41,8 @@ src="http://code.jquery.com/jquery-latest.min.js"></script>
 
   .sagecell_sessionContainer {
       margin-bottom:1em;
-  
   }
-
-  </style>
+</style>
 
 
 <script>
@@ -58,6 +57,7 @@ src="http://code.jquery.com/jquery-latest.min.js"></script>
 //      }
 //   ))
   </script>
+#}
 
     <div align ="center">
       Show commands using:  <a onclick="show_code('sage'); return false" href='#'>sage</a>,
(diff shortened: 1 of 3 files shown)


danyasaki:
I like this new way of handling code snippets.  I tested it, and it works beautifully.  I copied the syntax and used it on the genus 2 curve pages.

jwj61:
This now has conflicts

davidfarmer:
In make_code_snippets, I think it would be better to make a dictionary containing all of the
items on which set_code is called, and then call set_code once in a loop over that dictionary.

I also preferred the old display, where the code just looked like code and did not call so much
attention to itself in a purple box.  Maybe some highlighting, but more subtle, would be best.

Those two items can be addressed in the future.

davidfarmer:
I resolved the conflict (which was trivial) and will push to my fork and accept the pull request from there.

JohnCremona:
I'll look into the dictionary idea, and of course the styling should be more subtle -- I found it useful while testing to have a bright colour...

haraldschilly:
Instead of this dictionary: I once again advocate to use this yaml file format and store it in a separate file. this "contributors.yaml" is an example where there is a list of dictionaries. In this case (after having a very brief look), it seems to be more suitable to have a `dict(id-key -> dict("code-types" -> "code"))`. I.e.:

```yaml
tors:
  sage: 'E.torsion_subgroup().gens()'
  magma: 'elltors(E)'
  pari: 'TorsionSubgroup(E);'
reg:
  sage: 'E.regulator()'
  magma: 'Regulator(E);'
```

Note, that I've conveniently left out pari in the second example. It's trivial to check if such a key exists. What might look a bit cumbersome are the repetitions of the keys. I think that's fine and helpful in general for debugging and extending this. (and yaml parsers this as a list of dictionaries, which would be exactly what is going on in the code)

If there is ever the need for multi-line code snippets, that's also easily done using yaml's "`|`" pipe operator:

```yaml
key:
    what: |
        first line of code
        second line of code
        third line of code
```

JohnCremona:
yaml looks perfect for this.  After the weekend I'll make that change
(unless someone else does it first ;) ).  I think Dan Y's examples include
2-line snippets.

Thanks, Harald!

John

PR #319, opened 2015-08-06
Magma

danyasaki:
added magma code to genus 2 curve pages.  Followed Cremona's code.  Left most sage and pari empty.  See 
http://localhost:37777/Genus2Curve/Q/12500/a/12500/1
for example.


JohnCremona:
Great!  I will check it out...

davidfarmer:
Please look at how I edited the previous comment.  The < p> and < /p> tags were taken
as HTML and thus not displayed.  The same would have happened here if I had not
pus spaces after each "less than" sign.

There are other ways to indicate that a string should appear literally, but I forget how
to do it and I just put in the space.

JohnCremona:
Thanks -- clearly I did not look at  the post after posting.

danyasaki:
addressed John's comments and pushed again.

JohnCremona:
Looks good to me.  Now I hope that someone will merge #318!

PR #322, opened 2015-08-07
when data not in database, don't write that row in end alg table

danyasaki:
clean up tables on genus 2 curve pages so that if data for end alg is not in database, do not write that row.  See 
http://127.0.0.1:37777/Genus2Curve/Q/169/a/169/1
http://127.0.0.1:37777/Genus2Curve/Q/23552/a/23552/1


JohnCremona:
Looks good to me -- thanks for dealing with the things I raised,

When testing this with the "rankdom" option I came across one which fails (also on beta):  Genus2Curve/Q/1152/a/147456/1
but I don't think that has anything to do with this patch.

PR #323, opened 2015-08-09
Adding central character info for the L-functions coming from Artin representations in the degree 2 and 3 tables

patrickkuehn:
Thanks to John Jones, the central characters for L-functions coming from Artin representations have been added in the degree 2 and 3 tables.
I also added the yaml and pl files for generating such tables.


JohnCremona:
This look sreally good.  I spent a while looking through both degree 2 and 3 pages, testing as many links as I could, and found nothing wrong.
I'll let David (at least) make further comments.
One technical question: is seems a little inconvenient to have to manually run Perl scripts to convert the yaml files to html.  I may be biassed as I don't know Perl and so I do not know if this could have been done on python instead (reducing the number of languages needed for maintenance!), or whether it would be possible to process the yaml files on the fly as we do with the sidebar yaml and some others,  If that could be done, I think it would be better.  But that should not stop these excellent improvements going on.

jwj61:
Looking at it again, I think I am off on the central character of the tensor product (going back and forth between degree 2 and degree 3, I squared the Dirichlet character instead of cubing it).  I will send a correction.

Merging the processing of the yaml files in python seems like a better approach.  A reason to use a standard format is to let a library handle the parsing.  Perl has such a library, but it isn't being used.  On the other hand, we already require the python library to parse yaml, so could be done in one line.  We should probably migrate toward python parsing.

This can be merged once I sort out my own error here.

davidfarmer:
The perl script running on the yaml file was just a hack I did a few years ago, and I never got around
to doing it better.  We are in the midst of substantially changing the way L-functions and their
home pages are handled, and we also have many new objects which give rise to degree 4 L-functions.
So I think it makes sense to leave things as they are for now, and re-do how we handle those tables
when we make the degree 4 browsing table.

Presumably the yaml files are okay, and it is just an issue of how we process them?

JohnCremona:
On 12 August 2015 at 12:06, davidfarmer <notifications@github.com> wrote:

> The perl script running on the yaml file was just a hack I did a few years
> ago, and I never got around
> to doing it better. We are in the midst of substantially changing the way
> L-functions and their
> home pages are handled, and we also have many new objects which give rise
> to degree 4 L-functions.
> So I think it makes sense to leave things as they are for now, and re-do
> how we handle those tables
> when we make the degree 4 browsing table.
>

Fair enough.  No hurry.


> Presumably the yaml files are okay, and it is just an issue of how we
> process them?
>

Certainly!

John

jwj61:
In degree 3, the zero 1.36848 which comes from the tensor product should have central character chi_1145(228, -).

PR #324, opened 2015-08-11
EC code snippets now use yaml file

JohnCremona:
As suggested by Harald: the code snippets for sage/pari/magma which appear on elliptic curve home pages now have the code for each language in a separate yaml file.  This is a lot easier to read, and maintain!  Also, the python function make_code_snippets() is almost trivial;  and the html template is simplified -- each code item now "knows" which languages are implemented, so there is no need to pass a list of languages with each call to the code() macro.
The appearance of the web page is unchanged (I hope), this is just a back-end simplification.  Once merged, the same should be done for the genus 2 pages, and number fields -- and independently, we can make a global decision on the style of these "code boxes" whose appearance should be controlled entirely by the codebox css style.
I included the not-implemented and logo items in the yaml file, but they are not used in the web pages,

### DIFF
--- a/lmfdb/elliptic_curves/code.yaml
+++ b/lmfdb/elliptic_curves/code.yaml
@@ -1,0 +1,110 @@
+prompt:
+  sage:   'sage:'
+  pari:   'gp:'
+  magma': 'magma:'
+
+logo:
+  sage: '<img src ="http://www.sagemath.org/pix/sage_logo_new.png" width = "50px">'
+  pari: '<img src = "http://pari.math.u-bordeaux.fr/logo/Logo%20Couleurs/Logo_PARI-GP_Couleurs_L150px.png" width="50px">'
+  magma: '<img src = "http://i.stack.imgur.com/0468s.png" width="50px">'
+
+not-implemented:
+  sage:  '# (not yet implemented)'
+  pari:  '\\\\ (not yet implemented)'
+  magma: '// (not yet implemented)'
+
+curve:
+  sage:  'E = EllipticCurve(%s)    # or: E = EllipticCurve("%s")'
+  pari:  'E = ellinit(%s)       \\\\ or E = ellinit("%s")'
+  magma: 'E := EllipticCurve(%s); // or: E := EllipticCurve("%s");'
+
+gens:
+  sage:  'E.gens()'
+  magma: 'Generators(E);'
+
+tors:
+  sage:  'E.torsion_subgroup().gens()'
+  pari:  'elltors(E)'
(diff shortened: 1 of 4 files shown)


JohnCremona:
Sorry, I don't think the multi-line snippets work yet, hold it....

lmfdb/elliptic_curves/code.yaml
```diff
@@ -0,0 +1,110 @@
...
+  magma: 'OrderTorsionSubgroup(E);'
+
+sha:
+  sage:  'E.sha().an_numerical()'
+  magma: 'MordellWeilShaInformation(E);'
+
+qexp:
+  sage:  'E.q_eigenform(10)'
+  pari:  |
+    'xy = elltaniyama(E);'
```
haraldschilly: there is no need for the "`'`" quotes. See http://learnxinyminutes.com/docs/yaml/ -> literal block explanation. It's like an indentation in python, where it ends when you indent less than in the first indented line.

JohnCremona:
OK, I removed the quotes.  One thing which needed attention then is that yaml treats a colon ':' as a special character when followed by a space, which I solved by putting such lines (in Magma code) into a literal block.
On the multi-line issue: some code snippets are multi-line but I just put a semi-colon ';' after each line since they get concatenated in the html (with only one prompt) anyway.  Anything more clever than that is beyond me (today!).
Ready for review again...

haraldschilly:
regarding the multilines, there is a comment about single quotes, but I assume your problem is with the actual linebreak in the HTML? If so, usually, a "pre" tag is used, but the "div" in your case can easily be tuned to preserve linebreaks by adding this to the CSS: `white-space: pre;`.

JohnCremona:
On 11 August 2015 at 14:11, Harald Schilly <notifications@github.com> wrote:

> regarding the multilines, there is a comment about single quotes, but I
> assume your problem is with the actual linebreak in the HTML? If so,
> usually, a "pre" tag is used, but the "div" in your case can easily be
> tuned to preserve linebreaks by adding this to the CSS: white-space: pre;.
>

OK, will try that....

John

JohnCremona:
The latest commit adds Harald's suggestion, and indeed the mulitline bits of code now appear on separate lines.  It is not perfect  though since we only see the prompt on the first line.  Any suggestions about that?

JohnCremona:
See previous comment: multiline snippets now get the prompt on each line.  I used split() to split a multiline code string at newlines, and made the macro loop over the resulting list of strings, with a prompt before each,
I will now do the same for the number field equivalents.

PR #325, opened 2015-08-11
fixed bug.  made end_alg table in isogeny class look like curve page.

danyasaki:
Update: Fixed bug John found.  Changed \mathrm{End} to \End.

Fixed the bug John found.  Modified the code so that the behavior of the end_alg table on the isogeny class page mimics the curves page.  Code for curve and isogeny class is kept separate for now, but might be better to consolidate later.
See 
http://127.0.0.1:37777/Genus2Curve/Q/1152/a/147456/1
http://127.0.0.1:37777/Genus2Curve/Q/1152/a/


lmfdb/genus2_curves/isog_class.py
```diff
@@ -106,27 +106,31 @@ def make_class(self):
         self.curves = [ {"label" : c['label'], "equation_formatted" : list_to_min_eqn(c['min_eqn']), "url": url_for_label(c['label'])} for c in curves_data ]
         self.ncurves = curves_data.count()
         self.bad_lfactors = [ [c[0], list_to_factored_poly_otherorder(c[1])] for c in self.bad_lfactors]
-        for endalgtype in ['end_alg', 'rat_end_alg', 'real_end_alg', 'geom_end_alg', 'rat_geom_end_alg', 'real_geom_end_alg']:
+        ## duplication of get_end_alg.  need to clean up!
+        end_alg_title_dict = {'end_ring': '\\mathrm{End}(J)', 
+                              'rat_end_alg': '\\mathrm{End}(J) \\otimes \\Q',
+                              'real_end_alg': '\\mathrm{End}(J) \otimes \R',
```
jwj61: Should this be \\otimes and \\R?

Alternatively, you could use the python construct r'\End(J) \otimes \R' (the r in front lets you single backslash.

We have a macro for End, but I see it needs fixing ... coming in a pull request soon.

danyasaki: Yes, you are right.  The analogous change needs to be made in web_g2c.py since I copied code.  I will fix it soon if I don't have git trouble.  (I am currently away from my laptop.)

davidfarmer: This knowl lists the "official" LaTeX macros currently
supported for use anywhere in the LMFDB:

http://beta.lmfdb.org/knowledge/show/doc.macros

For example, the operator \End .

PR #326, opened 2015-08-11
Conversion of number field code snippets to use YAML file

JohnCremona:
Note that this includes the same changes to the templtes/style.css as in the latest version of the elliptic curve patch.  With luck, github will be happy whichever is merged first.

### DIFF
--- a/lmfdb/WebNumberField.py
+++ b/lmfdb/WebNumberField.py
@@ -4,6 +4,7 @@ from sage.all import *
 import re
 import pymongo
 import bson
+import yaml
 from lmfdb.utils import *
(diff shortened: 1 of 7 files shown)


jwj61:
I was going to look at this this afternoon.  Thanks for beating me to it.

JohnCremona:
The similar elliptic curve branch / pull request ( my branch "code") now
deals better with multi-line code.  If no-one else gets there first I'll do
the same for number fields tomorrow.  I had split some multi-line stuff
(e.g. Invariants) into separate pieces, which appear in a separate column
in the table, and it would be good to have a comment on how that looks.

JohnCremona:
I rebased on my elliptic curve similar branch, so that one should be merged first.  This one also now handles multi-line code properly (each line gets its own prompt).

PR #331, opened 2015-08-25
Latticeinfo

jwj61:
The main change is the addition of information on when integral representations are not unique, which is now shown on the appropriate Galois group page, e.g., on 

  http://beta.lmfdb.org/GaloisGroup/3T2

at the very bottom of the page.

This pull request also picked up addition/modification of import scripts which should be there, but should have been a different pull request.  In any case, they don't affect how the site runs.

### DIFF
--- a/lmfdb/WebNumberField.py
+++ b/lmfdb/WebNumberField.py
@@ -77,6 +77,18 @@ def do_mult(ent):
         return ent[0]
     return "%s x%d" % (ent[0], ent[1])
 
+# input is a list of pairs, module and multiplicity
+def modules2string(n, t, modlist):
+    C = base.getDBConnection()
+    modlist = [[galois_module_knowl(n, t, z[0], C), int(z[1])] for z in modlist]
+    ans = modlist[0][0]
+    modlist[0][1] -= 1
+    for j in range(len(modlist)):
+        while modlist[j][1]>0:
+            ans += r' $\oplus$ '+modlist[j][0]
+            modlist[j][1] -= 1
+    return ans
+
 def nf_display_knowl(label, C, name=None):
     if not name:
         name = "Global Number Field %s" % label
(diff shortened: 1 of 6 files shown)


davidfarmer:
Line 77 of lmfdb/import_scripts/number_fields/import_galois_modules.py
says
         line.strip()  

Do you mean line = line.strip(), or do you not actually need to strip the line?
(Just because I started looking at the pull request does not mean I will
feel competent to approve it!)

JohnCremona:
On 27 August 2015 at 21:11, davidfarmer <notifications@github.com> wrote:

> Line 77 of lmfdb/import_scripts/number_fields/import_galois_modules.py
> says
> line.strip()
>
Perhaps the author of that line of code thought that strip() affected the
string in place, like sort() on a list; but it does not (as I just checked).

jwj61:
Strip problem fixed.  The author was thinking along the lines of perl's "chomp", which acts on the given string.

davidfarmer:
Everything looks good to me, but I couldn't actually test it because I haven't been able to
install the gap packages on my laptop.

This is a good example of where it would help if someone else pulled and checked that the output
looks okay.

JohnCremona:
On 7 September 2015 at 12:21, davidfarmer <notifications@github.com> wrote:

> Everything looks good to me, but I couldn't actually test it because I
> haven't been able to
> install the gap packages on my laptop.
>
> This is a good example of where it would help if someone else pulled and
> checked that the output
> looks okay.
>

I am doing that... watch this space.

John

JohnCremona:
screenshot from 2015-09-07 12 33 16

Looks OK to me (screen shot attached)

JohnCremona:
Assuming that taking a screenshot of the window is as easy for others as it was for me, perhaps we should consider recommending adding one to pull requests wherever relevant.

davidfarmer:
I don't think the screenshot is necessary in general, especially if it stops the reviewer from
actually pulling and running the code!

A screeenshot is helpful, but again not really necessary, when one person checks the output and another
reviews the code.  It would have been adequate for me if you just said "yes, looks good".

PR #333, opened 2015-09-03
G2 layout

JRSijsling:
This is to improve the layout of the genus 2 pages. Changes in order of importance:

(1) Things that should be on one line because of their short nature (like number of rational Weierstrass points) are now on one line.

(2) Invariants are now under a header of the correct font size (h3).

(3) Sato-Tate groups are now laid out in table form, just as in the case of automorphism groups.

(4) Removed an empty line before the geometric invariants that seemed strange.

(5) Despaghettified the code somewhat by using indents.

Note the order of the Magma code and the one-line items; the code is always before the result, which as a consequence of the one-line modification means that in those cases they come before their green header. Since this happens throughout (for example with Aut (X)) I have found this to be logical. But the previous commit in this push is a version where the code is always under its header in the one-line cases; this means that they then follow their results.

### DIFF
--- a/lmfdb/genus2_curves/templates/curve_g2.html
+++ b/lmfdb/genus2_curves/templates/curve_g2.html
@@ -224,24 +224,23 @@ function show_invs(invstyle) {
     </tr>
 </table>
 
-<br>
-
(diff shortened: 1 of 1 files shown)


davidfarmer:
The style="display: inline" would be better to do with a css class.

I will make those changes and submit as a separate pull request.

davidfarmer:
Taken care of by another pull request.

PR #335, opened 2015-09-07
Arithtoggle

davidfarmer:
Toggle between arithmetic and analytic normalization on L-function home pages
(assuming the page is available in the arithmetic normalization, currently only true for
L-functions of genus 2 curves).  Also toggle to show more local factors in the Euler product,
when viewing the arithmetic normalization.  Also make the Dirichlet series display nicer by
putting the whole thing in html.

http://localhost:37777/L/Genus2Curve/Q/411062/a/

http://localhost:37777/L/Genus2Curve/Q/1152/a/

### DIFF
--- a/lmfdb/lfunctions/Lfunction.py
+++ b/lmfdb/lfunctions/Lfunction.py
@@ -1569,7 +1569,9 @@ class Lfunction_genus2_Q(Lfunction):
         # Need an option for the arithmetic normalization, leaving the
         # analytic normalization as the default.
         self.texname = "L(s,A)"
+        self.htmlname = "<em>L</em>(<em>s,A</em>)"
         self.texname_arithmetic = "L(A,s)"
+        self.htmlname_arithmetic = "<em>L</em>(<em>A,s</em>)"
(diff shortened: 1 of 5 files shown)


JohnCremona:
Looks good.   For the looks, we might consider centring the Dirichlet series as with other formulas, since when it is short it looks funny over on the left.  Also the graphs are to the left?  But I will not ask that this happens before merging.
On the code, I am not a great fan of python code which puts together long html strings, but I am a fan of stuff that works (and do not have a better solution!).  I like the show more/less functionality and would like to put something like that onto elliptic curve pages (where you can see the q-expansion of the modular form, but if you ask to see more it goes off and recomputes the modular form from scratch to hiugher precision!).
I checked to see that L-function pages for objects other than genus 2 curves still look OK, and they do: they show that the current normalization is analytic without (yet!) given the arithmetic choice.  Good: better than a butting which does not function.
One small thing before a demo:  The knowl lfunction.normalization is not yet written, and if can be before the demo that would be useful.
Merging now...

JohnCremona:
..and pushed to beta.lmfdb.org.

PR #337, opened 2015-09-08
Hmf l2

davidfarmer:
The zeros and plots of the L-function of a Hilbert modular form were broken. This seems
to be due to the url parsing expecting a different form that we currently use.

The Z-function plot and initial zeros should now appear on L/HMF pages:

http://localhost:37777/L/ModularForm/GL2/TotallyReal/2.2.5.1/holomorphic/2.2.5.1-31.1-a/0/0/

PR #338, opened 2015-09-08
parsing of URL for L-function of HMF

JohnCremona:


JohnCremona:
See pull request #337: this only includes the last commit on that branch.  If this one is merged then that one can be closed.

JohnCremona:
Testing, I see that http://localhost:37777/L/ModularForm/GL2/TotallyReal/2.2.5.1/holomorphic/2.2.5.1-31.1-a/0/0/ looks fine including the zeros and graph, while http://beta.lmfdb.org/L/ModularForm/GL2/TotallyReal/2.2.5.1/holomorphic/2.2.5.1-31.1-a/0/0/ does not get either zeros or graph.
Positive review!

JohnCremona:
I pushed this commit to the beta branch, but:  the URL http://beta.lmfdb.org/L/ModularForm/GL2/TotallyReal/2.2.5.1/holomorphic/2.2.5.1-31.1-a/0/0/ does not diplay zeroes of graph for me despite taking 45s doing something.  Running locally works OK.  This suggests to me that there is something not installed on lmfdb.warwick.ac.uk which should be.  I will investigate further.

JohnCremona:
Checking the flask log (for those that can, log into lmfdb@lmfdb.warwickac.uk, and it is the file lmfdb-git-beta/flasklog) I see various errors.  I found one problem -- the version of conrey-dirichlet characters on the server was not the latest one!  I pulled the latest version of that and installed it in the sage copy which runs beta, stopped and restarted beta, but those Hilbert L-function graphs are still not appearing.  The flasklog shows lcalc errors.

1. There is a bug in the code at line 206 of https://github.com/LMFDB/lmfdb/blob/master/lmfdb/modular_forms/elliptic_modular_forms/backend/emf_classes.py since xc is referred to when not defined;  this sees to be because of the change made in line 199.  This will effect elliptic modular forms, not Hilbert.

2. The object returned by the constructor Lfunction_HMF is not correct, so when this happens:

    File "/home/lmfdb/lmfdb-git-beta/lmfdb/lfunctions/main.py", line 781, in getLfunctionPlot
    L = pythonL.sageLfunction

we get

    AttributeError: 'Response' object has no attribute 'sageLfunction'

but I do not know why this causes a failure on beta.lmfdb.org but not when I run the same branch locally.

davidfarmer:
Previously the problem was that Lfunction_HMF was not being called
(because of the incorrect parsing of the url), and the symptom was
the same you just quoted:

AttributeError: 'Response' object has no attribute 'sageLfunction'

Now you are saying that on beta it does call Lfunction_HMF, but
results in the same error?

I tried to investigate further, but if I go to a HMF on
beta.lmfdb.org I just get a proxy error.

JohnCremona:
As far as missing zeros are concerned, I think it is just that in line 818 of lmfdb/lfunctions/main.py there is a timeout of 10 (seconds?).  I think it would be better if would catch that timeout and replace the strings to be displayed normally with a message such as "(timeout computing zeros)".

JohnCremona:
OK, so I also get a proxy error from the URL http://beta.lmfdb.org/L/Plot/ModularForm/GL2/TotallyReal/2.2.5.1/holomorphic/2.2.5.1-31.1-a/0/0/ (with Plot in it).

I am looking at lines 792-796 of lfunctions/main.py.   We create a plot object p, save it to a .png file whose name is created on the fly, then read that in and return the result.  What is the reason for the output-then-input?

davidfarmer:
I don't recall the details of making the L-function plot.  Probably it 
made sense at the time to save the file and then serve it?
We are in the midst of changing how we do the L-function home pages,
so it is best not to "fix" that now.

I still can't figure out why I don't get a proxy error for that URL
on my laptop, but I do on beta.  Unless it is some other type of error
which is being incorrectly reported as a proxy error?

JohnCremona:
No, I don't understand the proxy error thing either.

PR #342, opened 2015-09-14
fix small bug in emf tables

JohnCremona:
This fixes an error which causes errors in the flasklog, but I do not yet have a URL which triggers that.
Fredrik is looking into it too.  He says " I think this was not intended to be part of that pull request as I was trying to work on some simpler fix to make the dimension tables quicker."  Referring to #297.

Hoping to have a test case before this is merged...


fredstro:
An example of URLs which failed before is: http://localhost:37777/ModularForm/GL2/Q/holomorphic/ranges?level=2&weight=&character=0&Submit=Search
(obtained by typing 2 in the level search box in the main browsing page) 

After this fix the above URL is now working and gives information on Gamma_0(2)
(there should be more information here, but that is another issue...)

JohnCremona:
Confirmed.  I'll merge now and push to beta server after testing on there too.

PR #343, opened 2015-09-16
Adding ranks and generators to eliptic curves over number fields

JohnCremona:
The file point-find.py contains scripts to go through the curves already in the database and compute their ranks and generators.  This involved developing new code for Sage which is currently under review there, and it is not intended that reviewing this file is done before merging this pull request!

The code changes to be looked at are quite small: depending on whether the database has the rank, or only rank bounds, or neither, and whether it contains any generators, appropriate information is displayed on an individial curve home page: these are the only pages where anything looks different.  Useful examples are:

  - a curve of rank 0: http://localhost:37777/EllipticCurve/2.2.5.1/36.1/a/1
  - a curve of rank 1 with generator: http://localhost:37777/EllipticCurve/2.2.5.1/209.2/c/1
  - a curve of rank 2 with generators: http://localhost:37777/EllipticCurve/2.2.5.1/1831.2/c/1
  - rank bounds 0..2 and no generators: http://localhost:37777/EllipticCurve/2.2.5.1/2929.3/a/1
  - rank bounds 0..1 and no generator: http://localhost:37777/EllipticCurve/2.2.5.1/3599.4/a/1
  - no rank information yet: http://localhost:37777/EllipticCurve/2.2.24.1/10.1/a/1

I am still filling the database and these are the only possibilities currently in there.  As the last example shows, there is no need to wait until I have processed all the curves before merging this.  When all curves have rank or rank bounds in the database I will add rank to the search fields, but that is a separate task.

### DIFF
--- a/lmfdb/ecnf/WebEllipticCurve.py
+++ b/lmfdb/ecnf/WebEllipticCurve.py
@@ -237,7 +237,11 @@ class ECNF(object):
 
         # URLs of self and related objects:
         self.urls = {}
-        self.urls['curve'] = url_for(".show_ecnf", nf = self.field_label, conductor_label=self.conductor_label, class_label = self.iso_label, number = self.number)
+        # It's useful to be able to use this class out of context, when calling url_for will fail:
+        try:
+            self.urls['curve'] = url_for(".show_ecnf", nf = self.field_label, conductor_label=self.conductor_label, class_label = self.iso_label, number = self.number)
+        except RuntimeError:
+            return
         self.urls['class'] = url_for(".show_ecnf_isoclass", nf = self.field_label, conductor_label=self.conductor_label, class_label = self.iso_label)
         self.urls['conductor'] = url_for(".show_ecnf_conductor", nf = self.field_label, conductor_label=self.conductor_label)
         self.urls['field'] = url_for(".show_ecnf1", nf=self.field_label)
--- a/lmfdb/ecnf/WebEllipticCurve.py
+++ b/lmfdb/ecnf/WebEllipticCurve.py
@@ -218,6 +218,19 @@ class ECNF(object):
             self.rk = web_latex(self.rank)
         except AttributeError:
             self.rk = "not recorded"
+        try:
+            self.rk_bnds = web_latex(self.rank_bounds)
+        except AttributeError:
+            self.rk_bnds = "not recorded"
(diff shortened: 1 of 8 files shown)


davidfarmer:
Some comments on the Python style:

In  lmfdb/ecnf/point-find.py

line 61:      if HeightBound == None:  
should be     if HeightBound is None:
See:
https://www.python.org/dev/peps/pep-0008/#programming-recommendations

line 74: put spaces around the " > " sign:
          if index > 1:

line 128: same for " == ":
          if nfill == ncurves:

(Other examples in the same file)

Note that you do not put spaces around the "=" in the arguments of a function.
But everywhere else you do.

Spacing recommendations are here:
https://www.python.org/dev/peps/pep-0008/#other-recommendations

JohnCremona:
Thanks for the style comments.  After fixing a few manually I also ran all the .py files in lmfdb/ecnf/ through the codestyle.sh script.  No other changes!

jwj61:
I looked at the first page and there is tons of diagnostic output at the bottom.  That should probably go before going to beta.

JohnCremona:
The diagnostics stuff only appears when running in debug mode (and is there already, not added by this).  See the bottom 10 lines of the template: https://github.com/LMFDB/lmfdb/blob/master/lmfdb/ecnf/templates/show-ecnf.html

(To run in beta-but-not debug mode. leave out the --debug from the command line but set the environment variable BETA to something first.  Or trust me.)

jwj61:
OK.

The last example is supposed to be no rank information available, but it comes up as rank 0.  Is this a bug or just because of updated data?

Over number fields, generators (for infinite part or torsion) are given in homogeneous coordinates whereas over \Q they are given in affine coordinates.  Since we give the affine equation and none of these things are at infinity, affine seems like a better choice.

JohnCremona:
OK, I started computing some more ranks &  generators today.  Try http://localhost:37777/EllipticCurve/2.2.24.1/45.1/e/1 instead.

Good point about using affine notation -- I will change that.  I may also add the rank info on the isogeny class page as we do over Q.
I'll report here when it is "ready for review" again.

JohnCremona:
You were too quick for me -- I changed gens (including torsion gens) to affine but you had already merged.  Never mind, there will be more pull requests!

PR #350, opened 2015-09-17
Fix issue 341, missing artin representation

jwj61:
There was a simple off by one error.  This fixes the corresponding test as well.

http://beta.lmfdb.org/ArtinRepresentation/?dimension=3&conductor=1988-2015

should now produce a representation.


JohnCremona:
Looks good, the URL works and there are now _no_ test failures.

PR #352, opened 2015-09-21
fixed bug in user authentication

davidfarmer:
user authentication was messed up: anonymous users had the authority to add new users.
(Just look at the upper right corner of beta.lmfdb.org when not logged in).

I think I fixed all the errors.

This was brought to my attention by Samuele Anni.

### DIFF
--- a/lmfdb/users/main.py
+++ b/lmfdb/users/main.py
@@ -51,11 +51,12 @@ def get_username(uid):
 @app.context_processor
 def ctx_proc_userdata():
     userdata = {}
-    userdata['userid'] = 'anon' if current_user.is_anonymous else current_user._uid
-    userdata['username'] = 'Anonymous' if current_user.is_anonymous else current_user.name
-    userdata['user_is_authenticated'] = current_user.is_authenticated
-    userdata['user_is_admin'] = current_user.is_admin
+    userdata['userid'] = 'anon' if current_user.is_anonymous() else current_user._uid
+    userdata['username'] = 'Anonymous' if current_user.is_anonymous() else current_user.name
+    userdata['user_is_authenticated'] = current_user.is_authenticated()
+    userdata['user_is_admin'] = current_user.is_admin()
     userdata['get_username'] = get_username
+    print "user data", userdata
     return userdata
 
 # blueprint specific definition of the body_class variable
--- a/lmfdb/users/main.py
+++ b/lmfdb/users/main.py
@@ -53,7 +53,7 @@ def ctx_proc_userdata():
     userdata = {}
     userdata['userid'] = 'anon' if current_user.is_anonymous() else current_user._uid
     userdata['username'] = 'Anonymous' if current_user.is_anonymous() else current_user.name
-    userdata['user_is_authenticated'] = current_user.is_authenticatedi()
+    userdata['user_is_authenticated'] = current_user.is_authenticated()
     userdata['user_is_admin'] = current_user.is_admin()
     userdata['get_username'] = get_username
     return userdata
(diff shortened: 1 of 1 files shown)


lmfdb/users/main.py
```diff
@@ -51,11 +51,12 @@ def get_username(uid):
...
-    userdata['userid'] = 'anon' if current_user.is_anonymous else current_user._uid
-    userdata['username'] = 'Anonymous' if current_user.is_anonymous else current_user.name
-    userdata['user_is_authenticated'] = current_user.is_authenticated
-    userdata['user_is_admin'] = current_user.is_admin
+    userdata['userid'] = 'anon' if current_user.is_anonymous() else current_user._uid
+    userdata['username'] = 'Anonymous' if current_user.is_anonymous() else current_user.name
+    userdata['user_is_authenticated'] = current_user.is_authenticated()
+    userdata['user_is_admin'] = current_user.is_admin()
     userdata['get_username'] = get_username
+    print "user data", userdata
```
haraldschilly: did the api change? bad :-1: 
also, the print statement in the last line should go away

davidfarmer:
I removed the print statement I was using for debugging.

JohnCremona:
Typo on line 56?

jwj61:
Has this been tested with sage 6.8?

JohnCremona:
I am about to test on 6.8 (actually 6.9.beta6)...

JohnCremona:
OK, works fine on 6.9.beta7 after correcting that typo -- David can you do that?

    -    userdata['user_is_authenticated'] = current_user.is_authenticatedi()
    +    userdata['user_is_authenticated'] = current_user.is_authenticated()

davidfarmer:
done

jwj61:
Running the website is now broken for me (again) on this code.  Here is the end of the backtrace:

File "/Users/jj/Lmfdb/lmfdb/lmfdb/users/pwdmanager.py", line 136, in is_anonymous
     return not self.is_authenticated()
TypeError: 'bool' object is not callable

JohnCremona:
Sounds like we have a method and an attribute with the same name but that
doesn't explain why it worked for me.  I'll look at it after dinner .

JohnCremona:
I see the problem (but not why our testing was inconsistent).
The class LmfdbUser has an attribute _authenticated which is True/False but the methods which retrieve it are is_authenticate() [sic!] and is_anonymous() which returns its negation BUT wrongly is coded as 
`return not self.is_authenticated()`.

New patch coming up...

JohnCremona:
When testing I think it is important to shut down the server and restart rather than rely on debbug mode to restart itself.  Perhaps.

JohnCremona:
Spent some time on this and fixed some obvious errors but it is still
broken and I am tired.  Tomorrow is another day.

John

jwj61:
It seems that the problem stems from a change in Flask-Login from 0.2.11 to 0.3.0.  Googling this shows several pages indicating that it is breaking things, although I am not sure of the right fix.

JohnCremona:
Referring back to the original report: although anonymous users (in fact anyone) see the RegisterTokens link at the top right, it does not actually work (for anyone).

I corrected two things which seemed certainly wrong, though they did not fix the issue (see by usertokens branch on github).  (1) the function is_authenticate() was corrected to is_authenticated() -- note that this overrides a function in the base class so the name matters, but we were calling is_authenticated() with the correct spelling, so were in fact using some default from the base class for the last several years and not our own version.  (2) in the function ctx_proc_userdata() I changed current_user.name to current_user.name() since we need the value returned by the function here, not the function itself.

JohnCremona:
A new pull request is coming up which reverts some previous changes, the result does work with Flask-Login 0.2.11 (but not the new 0.3.0).

PR #354, opened 2015-09-24
no redirect of URLs for Cremona label curves

JohnCremona:
When a Cremona label for an elliptic curve or isogeny class is entered as a URL we used to use redirect to display the curve / class with its LMFDB URL.  Now we do not.

Example: http://localhost:37777/EllipticCurve/Q/338a displays "Elliptic Curve Isogeny Class 338.c (Cremona label 338a)" without changing the URL.  Similarly, http://localhost:37777/EllipticCurve/Q/11a1 displays "Elliptic Curve 11.a2 (Cremona label 11a1)" without changing the URL.

I think this is what was greed at the meeting at ICERM today.


davidfarmer:
Looks okay, but I will check further once it is in master.

PR #355, opened 2015-09-24
improved parsing of torsion structure search box entries

JohnCremona:
As requested at today's meeting at ICERM.  Now you can enter [], [n], [n,m] as before but also n and n,m without the brackets.  It also catches more invalid inputs (more than two integers, integers >1, or n,m where m is not a multiple of n.


davidfarmer:
When I enter "2 4" (without the quotes) then I don't get an error message, and on the page
returned there are no results found, and just the number "2" (without the quotes!) is in the search box.

If I had put in just the number 2 in the search box, then of course I would have found a lot of curves.

And if I put "2, 4" (no quotes, space after the comma), then the return search results seem to be
correct, but only "2," appears in the search box.

I think the parsing before the search is okay, and the problem is when the search box gets
filled out again.  Maybe it is better to use clean_input to change the global value, not the value
used at the point where the torsion structure is parsed? (Ask me if that suggestion is unclear.)

JohnCremona:
I'll look at it again.  And I do need an explanation of the last suggestion!

davidfarmer:
You clean up the user's input torsion_structure using parse_torsion_structure.

But on the page where you return the results, the search box has the string entered by the
user, not the corrected version from parse_torsion_structure.

At the point where you convert the entered string to the correct syntax, you should
also change the dictionary that is later used to fill out the search form.

JohnCremona:
Thanks.  I will clean this up on Wednesday.

JohnCremona:
OK, it is cleaned up -- parse_torsion rewritten, will correctly handle '[]', 'm', '[m], 'm,n', 'm n', '[m,n]', '[m n]' with whitespace anywhere, and will put the cleaned up input into the search-again box.

PR #356, opened 2015-09-26
Siegel Modular Forms completely rewritten: object oriented, using mongodb, amount of data easy to scale

nilsskoruppa:
This is the completely rewritten Siegel modular forms part. The objectives were:

   * It should be easy to add new data of whatever kind (the 'whatever' was a lomg lasting problem wıth    the SMFs) 
   * the SMF webpages should comfirm to the paradigma we agreed on in the last years
   * we want to use the mongodb
   * the code shoule be easy to scale (again because new data may again be very different)

The goals were achieved. However. there are some tiny corners where we still have to file in the future. Also, we did not yet put all available data to Warwick (too many Gigs - next week?).


davidfarmer:
The changes to
lmfdb/users/main.py
have to be reverted.  We have been dealing with that code for the last week or so,
because there is a problem depending on your version of the software.
(Actually, all I am certain is that this change breaks it for me.  It is possible that I
need to update something.)

davidfarmer:
There are some issues, but (except for the one mentioned previously) they all seem to only involve
Siegel modular forms, so I am merging and we can sort out those items later.

PR #363, opened 2015-09-30
Another major step for leaving beta

nilsskoruppa:
Implemented missing dimension formulas and added missing data


JohnCremona:
I think it is excellent that the functions have documentation, and this is something we should all do.  (But why have you used html markup in there?)

nilsskoruppa:
That is quite subtle/clever :-):

Here is the answer: The html code in the "documentation" is included in the
respective webpages displaying the queried dimensions. This makes
maintaining the SMF pages in fact simpler: implementing dimension formulas
for new collections requires only corresponding code in dimensions.py and
works then out of the box via a generic page displaying results for
dimension queries. Needed specifications of the result can be put as
"documentation of the function" and will then also be automagically
included in the generic page.

Best,
---Nils


2015-09-30 19:26 GMT-04:00 John Cremona <notifications@github.com>:

> I think it is excellent that the functions have documentation, and this is
> something we should all do. (But why have you used html markup in there?)
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/LMFDB/lmfdb/pull/363#issuecomment-144573029>.
>



-- 
---
Prof. Dr. Nils Skoruppa
http://w3.countnumber.de
Algebra and Number Theory - Department Mathematik
Universität Siegen, Germany

JohnCremona:
Very subtle / clever!  I'll ask you to explain to me how to do that.  I also like the way that if you enter incorrect syntax in a search box it just pops up a little window without leaving the page.

PR #364, opened 2015-10-01
Major step: bugs left the Siegel modular forms pages 

nilsskoruppa:
Last bugs removed. The SMF pages work now technically as planned, all data were reviewed, cleaned up and refilled into the mongodb.


davidfarmer:
I don't see this change:

list of l should be list of \ell
+    http://beta.lmfdb.org/ModularForm/GSp/Q/Kp.2_PY2_587_plus/?dets=8

davidfarmer:
When I try this:
 R: Every input I try on ModularForm/GSp/Q/Sp8Z/ ends in an error message.
+  DONE

If I put in 6+20  (which is in the box in grey)  I get an error.
If I put in 6+4 then it works.

PR #368, opened 2015-10-07
Refactor

fredstro:
Initial pull request for the new refactored classical (elliptic / holomorphic) modular forms.
The main intention for having this pull request is to obtain feedback and help on issues regarding the lmfdb interface, browsing and displaying functions and stuff etc. 

For the initial review of this request please pretend that all data that is shown is correct :)
We are working on the data generation / validation...  In case you wonder how the data is populated see 
the classes wmf.WebNewForm_computing and wmf.WebModFormSpace_computing in the separate repository https://github.com/sehlen/modforms-db (there is no need to look at this I only mention it for reference)

The current version has been tested on my laptop with sage 6.8. The modular forms pages works but there are some issues with the L-functions. Presumably they are accessing the coefficients in a way which is no longer supported. It should be easy to fix but I need to go to bed now and promised to make this pull request today... 
Fredrik


JohnCremona:
Thanks for your Herculean efforts !  We will have a good look tomorrow and
report back.

fredstro:
Thanks! I look forward to some good feedback (and maybe even some brave
volunteer that wants to get into it and help out!).
Fredrik

davidfarmer:
I think this is all it takes to fix the L-functions home pages (I'll do this as a separate pull request later.
Now I am just leaving a reminder to myself).

In lmfdb/lfunctions/Lfunction.py, at the end of line 30 add
, WebNewForm_cached
and on line 364 change WebNewForm to WebNewForm_cached.

That seems to throw away the "precision" argument, so more investigation may be needed.

fredstro:
There is no need for the WebNewForm_cached there. 
In fact the L-functions would work if you just remove the prec argument from the constructor. 

However, I removed this argument because of some problems and we were still figuring out how to make it work properly so I prefer to add it back in to the WebNewForm.__init__ 
At the moment 'prec' doesn't do anything but we should make it work in the way it is intended: that is, to only return a form with at least that number of coefficients stored.. 
I pushed a version with this fix, but I am not sure how to put it into the pull request.

davidfarmer:
I am having trouble locating the change you just made.

I merged your big pull request.  Can you just make another pull request?

Otherwise we can just use the temporary fix I proposed.

davidfarmer:
Okay, I found your changed file and I made the same change.
I'll merge and then you pull.

fredstro:
Ok. Good. I did that.
From my understanding I though that my new commit to the branch with the pull request should have updated the pull request automatically, i.e. it should have appeared here in this thread.

davidfarmer:
That is what usually happens.
Maybe I had already merged?

fredstro:
It seems like you were merging at more or less the same time as I pushed the change so probably you were a little earlier (when I wrote my comment I didn't see your merge, it only appeared after I clicked the 'Comment' button)

PR #372, opened 2015-10-07
Fix dimension search from front page for Siegel Modular Forms

wbl:


JohnCremona:
Watson -- thanks!  Good practice is to give here the URL to go to and instructions on what to enter (etc) to see the bug and how it has been fixed.  See my pull request #371

JohnCremona:
Nils says this is good so it's good!

PR #373, opened 2015-10-07
Fixed header of the Fourier coefficient download for elliptic modular forms

jvoight:


JohnCremona:
http://localhost:37777/ModularForm/GL2/Q/holomorphic/1/12/1/a/

I get a file whose header starts
```
 and character \( \chi_{1}\)
## level=1, weight=12, character=1,label=a 
## coefficient field : Rational Field
```
so there should be more stuff in there at the beginning?

PR #374, opened 2015-10-07
Groupnames

jwj61:
Fixes the display of group names.

 - if you search for Galois groups, only display the "nice" name in the name column if we have one
 - on the page for an individual Galois group
    - if there is a nice name, show it in both properties and on the page
    - call the "gap" name the CHM label, and show it in the list of information
    - make knowls for the two naming conventions
  - when searching for number fields, local fields, or Artin representations, use the nice name for a Galois group if it exists, otherwise use the style nTt
  - in group display knowls, include the nice name if it exists, and always have a line for the CHM label


JohnCremona:
http://localhost:37777/GaloisGroup/6T2  gives an error for me.  I got to it from a search results  page and clicking on the label in the left column.

PR #381, opened 2015-10-07
Refactor without experimental code

fredstro:
I removed the experimental directory and all inclusions of code from there. 

This includes the "browse range" box which was shown below the "main table" of dimensions. The functionality was not there anyway and I was planning to rewrite it completely so you might as well try to write this from scratch.


fredstro:
The non-trivial character now shows but there are still problems. For instance, how should they be browsed? How to care about embeddings of characters? How to display (if at all) embeddings of coefficients?

PR #383, opened 2015-10-08
improved torsion search for elliptic and genus 2 curves

JohnCremona:
As suggested in the wiki:  genus 2 torsion structure search now allows more possible inputs, e.g. '29', '2 2 2', '[2,2,2]'.    Modelled on recent changes for elliptic curve torsion search, but allowing up to 4 factors.  The sample input is changed to [2,2,2] (from [2, 4]) to emphasize that the number of factors can be up to 4!
Try your luck at http://localhost:37777/Genus2Curve/Q/ but beware that this search page has no error handling yet.  In the case of torsion structure, if the input cannot be correctly parsed it is quietly ignored.  While if you put 'x' into the torsion order box you get a server error.  This also needs improving!

### DIFF
--- a/lmfdb/ecnf/main.py
+++ b/lmfdb/ecnf/main.py
@@ -7,7 +7,7 @@ import pymongo
 ASC = pymongo.ASCENDING
 from lmfdb.base import app, getDBConnection
 from flask import render_template, render_template_string, request, abort, Blueprint, url_for, make_response, redirect
-from lmfdb.utils import image_src, web_latex, to_dict, parse_range, parse_range2, coeff_to_poly, pol_to_html, make_logger, clean_input
+from lmfdb.utils import image_src, web_latex, to_dict, parse_range, parse_range2, coeff_to_poly, pol_to_html, make_logger, clean_input, parse_torsion_structure
 from sage.all import ZZ, var, PolynomialRing, QQ, GCD
 from lmfdb.ecnf import ecnf_page, logger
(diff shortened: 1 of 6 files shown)


JohnCremona:
Hold it -- instead of having 2 almost identical functions (here and for elliptic curves) I will write one which can handle arbitrary torsion ranks (i.e. elementary divisors of finite abelian groups) for both.

JohnCremona:
OK for review again:  the parsing function is now in lmfdb/utils and is used for ec/Q, ecnf and genus_2.  It has an extra parameter which is the max number of invariants allowed (2 for elliptic curves, 4 for genus 2).

davidfarmer:
If I enter "2 3 4" or "[2,3,4]" I get all genus 2 curves, and the corresponding field in the returned
form is empty.

Shouldn't it return no results for that search?

JohnCremona:
On 14 October 2015 at 12:06, davidfarmer <notifications@github.com> wrote:

> If I enter "2 3 4" or "[2,3,4]" I get all genus 2 curves, and the
> corresponding field in the returned
> form is empty.
>
> Shouldn't it return no results for that search?
>
It should detect that the list fails the condition that each number divides
the restl, and ideally should then raise a (nice) error message.  But there
is no error handling at all on that page so there is no mechanism to handle
it.

I will take another look but it may be that we just want to create a new
Issue for the error handling.

davidfarmer:
Accepting, but have added issue  #398 concerning invalid torsion structures.

PR #388, opened 2015-10-09
Homepage tests

JohnCremona:
Completed the home page test suite with tests for boxes 4, 5, 6.  Currently skipping the Box 6 tests as they are to external links and do not work -- I don't know if that is fixable.  It would be good to automatically test that they at least work, never mind the content.


JohnCremona:
To test: ./test.sh lmfdb/test_homepage.py  and ignore the debugging messages from Fredrik on Box 2.

JohnCremona:
Several improvements -- now ready for review.  Harald showed how to test links to external URLs so these are there (except Magma which does not allow this).  Also added tests that the paths are on the homepage, and refactored the code to b a lot shorter!

I have finished with this now.

PR #391, opened 2015-10-09
added tests for ec/Q browse & search page

JohnCremona:
As an example I added a test file for the elliptic curve (over Q) browse and search page, so check that the page itself and hard links on it work, as well as all the various browse and search functions.  Obviously not all combinations could be included but I put in the ones suggested on the page, at least.

No changes except for the addition of this file.


JohnCremona:
Since this pull request only adds one test file -- and the tests work!  I am merging it.

PR #394, opened 2015-10-14
Two HMF changes

jvoight:


JohnCremona:
The "random" function will always return the same form (whichever mongo thinks is the first, as that is what find_one() will return!  I will fix that.

JohnCremona:
This request is pending an update which currently exists at https://github.com/jvoight/lmfdb/pull/36

jvoight:
Merged, I think.  JV

PR #399, opened 2015-10-15
Fix more tests following refactoring

JohnCremona:
After this there are still 2 FAILs (both in lmfdb/test_tensor_products.py) and 2 ERRORs (both in lmfdb/test_lfunction.py) but progress is progress.

### DIFF
--- a/lmfdb/lfunctions/Lfunction.py
+++ b/lmfdb/lfunctions/Lfunction.py
@@ -9,6 +9,8 @@
 import math
 import re
 
+from flask import url_for
+
 from Lfunctionutilities import (p2sage, seriescoeff,
                                 compute_local_roots_SMF2_scalar_valued,
                                 compute_dirichlet_series,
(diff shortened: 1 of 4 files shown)


davidfarmer:
I get:
FAILED (SKIP=3, errors=3, failures=2)
Is that okay?

JohnCremona:
I'm sure I only had 2 errors.  Perhaps something new has happened.   I'll look again.  One day we'll have to see about the 3 which are being skipped too.

davidfarmer:
I don't have gap installed on my laptop, so that may be the difference.  Will merge now.

JohnCremona:
For the record, I get (using Sage 6.9 with all the recommended extras):
FAILED (SKIP=3, errors=2, failures=2)
I am trying again after merging the current master into this branch to make sure that nothing new has cropped up.

JohnCremona:
Same result using the merged version.

PR #400, opened 2015-10-16
fixed two bugs which resulted in slow ecnf search results and ...

JohnCremona:
& isogeny cllass pages.

Nicolas observed that ec/nf isogeny class pages loaded very slowly and I fixed that by doing two things, one of which also speeds up search results:  there was a cache of web number fields, used for displaying nice lists of Weierstrass coefficients, but nothing was ever put into the cache! so WebNumberFields were repeatedly being created for the same field.  Secondly,  when preparing the list of W. coeffs to display for the isogeny class, it was constructing the full ECNF object, and this is not necessary since all we need is the label and url and W.coeffs (constructed as for search result lists).

### DIFF
--- a/lmfdb/ecnf/WebEllipticCurve.py
+++ b/lmfdb/ecnf/WebEllipticCurve.py
@@ -8,7 +8,6 @@ from kraus import (non_minimal_primes, is_global_minimal_model, has_global_minim
 ecnf = None
 nfdb = None
 
-
 def db_ecnf():
     global ecnf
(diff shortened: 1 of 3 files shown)


nmascot:
I confirm the problem seems to be fixed: classes which used to take ages to display (often resulting in a server timeout) are now displayed in a few seconds. Tired several times on several isogeny classes.

JohnCremona:
I regard the above comment as a positive review so am merging.

Issue #308, opened 2015-08-01
Error displaying kronecker symbol for dirichlet character


jwj61:
Go to http://beta.lmfdb.org/Character/Dirichlet/?modulus=68&conductor=&order=2 and click on the third character (-68 / *) and on the character's home page it says (68 / *).

haraldschilly:
fixed

PR #404, opened 2015-10-21
Fixing Kronecker Symbol Bug

alok-shukla-math:
Fixed the issue https://github.com/LMFDB/lmfdb/issues/308 . 

Additionally it was decided in the LMFDB meeting last week ( 14 Oct) on the suggestion of Dr. John Voight  that the Kronecker Symbol Representation should be displayed on Dirichlet Character page only if it is primitive. This accounts for the change in the file - Character.html

### DIFF
--- a/lmfdb/WebCharacter.py
+++ b/lmfdb/WebCharacter.py
@@ -858,6 +858,7 @@ class WebDirichletCharacter(WebChar, WebDirichlet):
 
     @property
     def symbol(self):
+
         """ chi is equal to a kronecker symbol if and only if it is real """
         if self.order != 2:
             return None
@@ -866,8 +867,10 @@ class WebDirichletCharacter(WebChar, WebDirichlet):
             if cond % 4 == 1: m = cond
             else: m = -cond
         elif cond % 8 == 4:
-            if cond % 16 == 4: m = cond
-            elif cond % 16 == 12: m = -cond
+	    # Fixed : For Q(sqrt(d) the discr. = 4 d if d = 2,3 (mod) 4  and discr = d if d = 1 (mod) 4. 
+            #Please refer to eq 3.49 and section 3.8 of "Analytic number Theory - Henryk Iwaniek & Emmanuel Kowalski" for its connection with Kronecker symbol. 
+            if cond % 16 == 4: m = -cond
+            elif cond % 16 == 12: m = cond
         elif cond % 16 == 8:
             if self.chi.is_even(): m = cond
             else: m = -cond
--- a/lmfdb/WebCharacter.py
+++ b/lmfdb/WebCharacter.py
@@ -923,7 +923,11 @@ class WebDirichletCharacter(WebChar, WebDirichlet):
              \chi_{%s}(%s,r) e\left(\frac{%s r + %s r^{-1}}{%s}\right)
         = %s. \)""" % (a, b, modulus, number, modulus, modulus, number, a, b, modulus, k)
 
+
     def symbol_numerator(self): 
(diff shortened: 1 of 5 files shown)


davidfarmer:
There is an inconsistency between the modifications to the code and the parallel code
beginning at line 280 in the file  lmfdb/characters/main.py

Either that other function needs to be fixed, or better yet, the code should be refactored
properly by relegating all the if-then cases to a new function, and use that new function
to clean up the two existing functions.

Also, correct the spelling of "Iwaniec" in a comment.

alok-shukla-math:
I wanted to update the code today itself to make it cleaner as you had suggested, but trying to figure out the logic of Dirichlet  character and Kronecker symbol took more time than I expected ( even after a very generous help from Anna ) . I made the pull request so that at least we could have a working code now. ( I will work on the final fix of replacing the parallel functions with a new one to make the code cleaner.) . I will also fix the spelling mistake in the comments. 
Thank you.
Alok

JohnCremona:
I agree with David.  It is bad enough having duplication, but inconsistent duplication is worse.

The kroencker_symbol function should removed from the file it is in, any place where the function is called should be changed into a request of the appropriate WebCharacter instance for its symbol.

I don't know why WebCharacter.py is not in the characters/ directory, but some other Web*.py files are also in lmfdb/ for some reason -- ideally they should be moved.  Not important now.

About the mathematical comment -- I really don't think we need a comment reminding us what is the discriminant of a quadratic field!  This fact does not tell me why the change is correct.  It seems that the onlt substantive change is to switch the cases of 4 and 12 mod 16.  As we all said at the group meeting, this needs to be better organised and explained: in WebCharacter.py add a new member function which computes the integer m, perhaps call that symbol_numerator();  and add documentation to say what it actually *is*.  If it only makes sense for primitive characters, say so (and raise an error if called with a non-primitive character).  Then the method symbol() becomes a one-liner.  This was we separate the mathematical computation of an integer associated with the character from the formatting of some latex.

alok-shukla-math:
Thanks for your suggestion. I have changed the code accordingly, deleted
the kronecker_symbol function and replaced its usage with the one
in WebCharacter.  The code seems to work fine. However, I want to test it
more at my end before making a pull request.

davidfarmer:
Once you push to your github repository, the pull request will be automatically updated.
(Also, next time you probably want to start a new branch instead of working in your master branch.)

alok-shukla-math:
I ran the test script ./test.sh and it returned with 4 errors and 5
failures. Only one failure was related to Dirichlet Characters (please see
below). However, it appears that the assertion in the test script is
incorrect and the test script itself has to be modified i.e. it should be
\\frac{-4}{\\bullet} in the assert statement below. ( It relates to the
existing display on the page http://beta.lmfdb.org/Character/Dirichlet/4/3
). So, I have corrected the test script test_characters.py.


FAIL: test_dirichletchar43
(lmfdb.characters.test_characters.DirichletCharactersTest)
  File "/home/alok/lmfdb/lmfdb/characters/test_characters.py", line 113, in
test_dirichletchar43
    assert '\\left(\\frac{4}{\\bullet}\\right)' in W.data
AssertionError

alok-shukla-math:
All the 5 files are now ready for review and to be pulled.

davidfarmer:
Looks good.

Am I correct that we should always have had (-4/.) for the nontrivial character mod 4,
and somehow nobody noticed that we had (4/.) ?

Merging.

alok-shukla-math:
Yes, that is correct. Somehow it was missed. Thanks.

PR #407, opened 2015-10-22
Endomorphisms and decomposition for genus 2 curves

JRSijsling:
This adds more information on the endomorphism rings of genus 2 curves and the field of definition of these endomorphisms. Endomorphism rings over subfields of that full field of definition are returned as well.

If the Jacobian decomposes as a product or a square of elliptic curves, then the user will either get a link to these curves (if they exist in the LMFDB) or an equation. They are always returned over the smallest field where a decomposition exists.

There remains much to improve, of course, but this version provides the basic functionality. The corresponding database is being uploaded as I write this (at 90%).

### DIFF
--- a/lmfdb/genus2_curves/genus2_curve.py
+++ b/lmfdb/genus2_curves/genus2_curve.py
@@ -54,6 +54,7 @@ geom_aut_grp_dict = {
 #########################
 #   Database connection
 #########################
+
 g2cdb = None
 
 def db_g2c():
@@ -62,6 +63,13 @@ def db_g2c():
         g2cdb = lmfdb.base.getDBConnection().genus2_curves
     return g2cdb
 
+g2endodb = None
+
+def db_g2endo():
+    global endodb
+    if g2endodb is None:
+        g2endodb = lmfdb.base.getDBConnection().genus2_endomorphisms
+    return g2endodb
 
 #########################
 #    Top level
--- a/lmfdb/genus2_curves/templates/curve_g2.html
+++ b/lmfdb/genus2_curves/templates/curve_g2.html
@@ -260,267 +260,28 @@ function show_invs(invstyle) {
 
 <h3 style="display: inline"> {{ KNOWL('g2c.jac_endomorphisms', title='Endomorphisms') }} </h3>
 
-<!-- Old functionality:
-<p>{{data.data.endomorphism_statement}}</p>
+<p>{{data.endodata.gl2_statement_base}}</p>
 
-<table>
-    {% for endalgtype in ['end_ring_name', 'rat_end_alg_name', 'real_end_alg_name', 'geom_end_ring_name', 'rat_geom_end_alg_name', 'real_geom_end_alg_name'] %}
-        {% if data.data[endalgtype][1] != '' %}
-            <tr><td>\({{data.data[endalgtype][0]}}\)</td><td>\(\simeq\)</td><td>\({{data.data[endalgtype][1]}}\)</td></tr>
-        {% endif %}
-    {% endfor %}
-</table>
--->
+<!-- Description over QQ: -->
+<p>{{data.endodata.endo_statement_base|safe}}</p>
 
-<!-- GL_2 statement (only over QQ for the moment): -->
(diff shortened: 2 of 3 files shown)


davidfarmer:
There is a lot of logic in a template which would be easier to maintain/debug/modify if it were
in python.  But that can be addressed later.

Also, I think things like end_alg_name would be better called end_alg_name_tex, since there
are several possibilities for the name, and this function outputs TeX.  Again, can be changed later.

JRSijsling:
Yes, I noticed the first part in the code. It became difficult to put in links in that way though. At least the code runs relatively parallel for these lines, so if there is a good way to link directly via the Python code it should be doable to move it over to webg2c.py. As for the end_alg_name bits, those are not mine.

JRSijsling:
I have turned this over in my head in the meantime. First of all, I noticed a mistake; there is no well-defined minimal splitting field for the Jacobian. We can still return a field of minimal degree though.

Second, I can definitely streamline this using Python. Most stuff with links is at the end of a line, so we can just define statements in the Python code and append these snippets if necessary. That will still make the code relatively ugly because of the conditionals involved; I would love to hear whether Python can return statements that are automatically converted into hyperlinks.

JohnCremona:
On 4 November 2015 at 15:16, Jeroen Sijsling <notifications@github.com>
wrote:

> I have turned this over in my head in the meantime. First of all, I
> noticed a mistake; there is no well-defined minimal splitting field for the
> Jacobian. We can still return a field of minimal degree though.
>
> Second, I can definitely streamline this using Python. Most stuff with
> links is at the end of a line, so we can just define statements in the
> Python code and append these snippets if necessary. That will still make
> the code relatively ugly because of the conditionals involved; I would love
> to hear whether Python can return statements that are automatically
> converted into hyperlinks.
>

There should be any hard-coded links in the html template.  Rather, you can
call a python function from in there (e.g. {{something.function()}} ) which
produces the url.  Usually this is via the url_for() command.

John

JRSijsling:
I took the alternative way of making the link url a part of the data dictionary, so certainly nothing is hardcoded. I was just wondering if it is possible to have Python return a string that becomes a text with a hyperlink here and there when inserted in the appropriate environment.

I have tried {{data.some_statement}} where some_statement is generated in Python. However, I do not know how to insert a hyperlink into that statement directly. If I try, then I just see "<a href=bla1> bla2 </a>" on the page instead of a link.

So currently the links are inserted in the template file, from the data dictionary. I can live with this, and I have not seen different ways of proceeding in the other pages, but it would just clean up everything considerable if there were a way around.

haraldschilly:
after a quick peek into the diff of this PR:

1. for URLs, always use the `url_for` function. That avoids dead links and typos. quickstart doc: http://flask.pocoo.org/docs/0.10/quickstart/#url-building or just search the lmfdb source with `git grep url_for` and you'll find plenty of examples.

2. an import of "sage_eval" happens. Never do that. This might lead to some security problems, because without additional protection any code might execute. I don't see where it is used, sometimes a literal evaluation is good enough.  e.g.

        from ast import literal_eval
        x = literal_eval("[1,2,3]")

JRSijsling:
Thanks, I will start using url_for in a future improved version of this pull request. The sage_eval is indeed nowhere needed; it is a legacy from a time before I realized that this was an extremely bad idea.

JRSijsling:
All right, I updated my remote branch, and with it, the pull request (hopefully).

The majority of the code is now created in a Python file; hyperlinks are also created there and then used by enabling the {{bla|safe}} option in the Django template. Since this code cannot be modified except by those working on the LMFDB, I have reason to hope that this does not introduce a security risk. The result seems much more maintainable than the former version where the logic was in the template.

Obviously there are still enough questions about display and formatting, but to me this looks okay as a zeroth version.

davidfarmer:
The pages look good.

Please look at the pull request in github and check if your recent changes are there.

JRSijsling:
It seems correct and has all the relevant (and the most recent) changes. The date of the second commit is November 4 because I amended it; this made for a simpler history, and I was on a private branch anyway.

PR #408, opened 2015-10-22
added comments to a modular forms file

christellevincent:
to help when people go over it in december


JohnCremona:
Thanks -- since this only adds comments I am merging it right away, and assuming that the comments are accurate!

PR #413, opened 2015-10-26
Neighbors

davidfarmer:
Bringing back "previous" and "next" links for Dirichlet characters and their L-functions,
but this time as a "Nearby objects" section of the properties box.

Visit the home page of any Dirichlet character or its L-function.

### DIFF
--- a/lmfdb/WebCharacter.py
+++ b/lmfdb/WebCharacter.py
@@ -302,10 +302,6 @@ class WebDirichlet(WebCharObject):
             return 1, 1
         if n > 2:
             Gm = DirichletGroup_conrey(m)
-    #    else:
-    #        m = m - 1
-    #        Gm = DirichletGroup_conrey(m)
-    #        n = m - 1
         while True:
             n -= 1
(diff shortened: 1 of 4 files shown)


davidfarmer:
There are other use cases.  For example, the home page of an elliptic curve does not have links
to isogenous curves.  We can replace "Nearby objects" by "Isogenous curves" and then list the
corresponding numbers.  By only listing the numbers, it would be possible to accommodate a long list.
(But note that the code would have to be modified to handle that use case.)

JohnCremona:
Elliptical curves do link to the isogeny class.  I think that makes links
to isogenous curves redundant but of course that is possible.

davidfarmer:
It is a question of what are common use cases.

Just yesterday I was looking at an elliptic curve, and I wanted
to look at the isogenous curves and compare how the BSD invariants
were rearranged.  Yes, I could get to those by first going to the
isogeny class and then opening new windows from there.  It would
have been slightly more convenient to have direct links.  But on
the other hand, I had never needed to do that previously, and it
makes no sense to clutter the page in support of rare activities.

PR #468, opened 2015-10-28
l-functions: adding page about history

alinabucur:
please review and merge


davidfarmer:
I do not understand this pull request.  I was expecting to see something on the main L-functions page
(like a properties box) with a link to L-functions history.

I also think the knowl should be labeled lfunctions.history, not lfunctions.intro .

Issue #410, opened 2015-10-24
Dirichlet character breadcrumbs


davidfarmer:
The breadcrumbs on this page tend to overlap the "Feedback" link.
It is worse if you are logged in.

http://beta.lmfdb.org/Character/Dirichlet/5/4

Things would be slightly improved if the text in the breadcrumbs were shortened to
"mod 5" and "#4".

nmascot:
Working on it

nmascot:
Submitted pull request #487.

AndrewVSutherland:
This appears to have been fixed by pull request #487.

PR #487, opened 2015-10-28
Short char bc

nmascot:
Fix to ticket Dirichlet character breadcrumbs #410. Also changed breadcrumbs for Hecke character since they are even longer, cf. Character/Hecke/3.1.44.1/61.-22a0+1a1/1.

### DIFF
--- a/lmfdb/characters/main.py
+++ b/lmfdb/characters/main.py
@@ -113,7 +113,7 @@ def render_Dirichletwebpage(modulus=None, number=None):
         m = info['modlabel']
         info['bread'] = [('Characters','/Character'),
                          ('Dirichlet','/Character/Dirichlet'),
-                         ('Modulus %s'%m, '/Character/Dirichlet/%s'%m)]
+                         ('Mod %s'%m, '/Character/Dirichlet/%s'%m)]
         #logger.info(info)
         return render_template('CharGroup.html', **info)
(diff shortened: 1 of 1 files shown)


nmascot:
Made the modulus more readable in breadcrumbs, e.g. 61.-22a0+1a1 now reads <61,a-22>.

NB Hecke characters are still buggy and uncomplete so I'll stop working on them now and concentrate on stuff for 1.0.

davidfarmer:
Why was this line added?
 
from lmfdb.WebNumberField import WebNumberField 

I don't see that the other changes use it.

JohnCremona:
The redundant line needs to be removed.  Otherwise it's an improvement, I think, though it should be discussed more widely (unless this was already in response to a discussion which asked for it!).  On my laptop the page http://localhost:37777/Character/Hecke/3.1.44.1/61.-22a0+1a1/1  has breadcrumbs which take up too much space:  after Mod the rest is overwritten by the search box at the leftextremity of what is in the top right corner.  (Only when logged in since my full name takes up space, so perhaps that is oK).

JohnCremona:
Nicolas, if you remove the redundant line and update your branch then I will merge this.

nmascot:
Had to delete this branch (little problem with github). Sending a new pull request now.

PR #518, opened 2015-10-28
Siegelfix

wbl:
Contains example of the kinds of UI changes we want to make for all tables of Siegel Modular Forms and reorganization of higher level collections.  Removes some functionality we already have, far from complete overhaul.


davidfarmer:
This will be reviewed after additional work next week.

wbl:
It's now at feature parity with what we have now.

PR #523, opened 2015-10-28
Slopecontent

jwj61:
Do a search of local fields, see slope content in the search results table.  Fixes issue 489.


jwj61:
It looks like this has changes from a previous pull request.  One should consider that one first :).

JohnCremona:
Hasn't the first of these commits already been merged?  in which case why are those changes appearing here?
Perhaps you could use cherry-pick to make a branch containing only one of these two commits?

jwj61:
I don't understand why github did not pick that up.  I rebased the branch from upstream/master, pushed, and now it only shows the relevant commit.

JohnCremona:
Just don't give me a closed book exam on what the knowl actually means!

Issue #521, opened 2015-10-28
Title for L-function of genus 2 curve


jvoight:
We say "L(A,s), where A is a genus 2 curve of conductor N", which is funny; we probably mean A is an abelian surface or C is a genus 2 curve.

kedlaya:
Not just funny, but also mathematically inaccurate: the conductor we are reporting is the conductor of the Jacobian = the conductor of the L-function, whereas the "conductor of the curve" (which I don't actually know how to define) would be divisible by possibly a larger set of primes. (The difference would be the primes where the reduction of the Jacobian leaves the Torelli locus, by degenerating to a product of elliptic curves.)

nmascot:
Working on it

nmascot:
I suggest "L(A,s), where A is the Jacobian of a genus 2 curve and has conductor N".
I do realise it is a bit long, so please make suggestions :)

nmascot:
Submitted pull request #528.

davidfarmer:
This will be changed again as a result of changing how the L-function page titles are specified.
Work in progress.

davidfarmer:
In the next pull request it is changed to 
"...where $A$ is the Jacobian of a genus 2 curve with label..."

PR #528, opened 2015-10-28
Changed to Jac of g2 curve and has cond N

nmascot:
Ticket #521 : Changed to "L(A,s), where A is the Jacobian of a genus 2 curve and has conductor N".


davidfarmer:
Will be handled by a more general overhaul of L-function home pages.

PR #535, opened 2015-10-29
the new universe lmfdb map picture

alinabucur:


davidfarmer:
Do you want this in the code base now because you expect other people to work on it?
If not, it would maybe be best to leave it as a pull request which will be updated automatically
as you make and push changes.

JohnCremona:
We can actually merge this right now because the link "Map" in the sidebar still goes to the old /bigpicture, while the new one is at /universe.
That way more people could look at it, changes to the universe picture could be made without affecting the website and when we like the new version better we just make a small change to sidebar.yaml line 13 changing "bigpicture" to "universe"

PR #545, opened 2015-11-01
generalized database query api

haraldschilly:


haraldschilly:
In case someone wonders: I've mainly merged it to test the database upgrade. These pages aren't linked from anywhere, so it's in beta status right now. Also, since the important databases are filtered out, there is no security concern and it's also impossible to delete or modify anything.

PR #564, opened 2015-11-04
Added tests for ecnf pages and search

nmascot:


davidfarmer:
How should this pull request be tested?

JohnCremona:
Presumably ones tests that the tests introduced do pass when run ;)  and
that the examples used are reasonable (in case some URLs take along time,
they would not make the best tests).

You could look to see that there is at least one test for each template.
But even if there were not, having more tests is good.

JohnCremona:
Looks good to me, and tests pass.

PR #572, opened 2015-11-06
Fixes to GL2 Maass form homepage

lemurell:
Added previous and next links for GL2 Maass forms and their L-functions. Added image for select GL2 Maass forms (the looks of these will be improved).


davidfarmer:
The plot is large, so it would look better to put the plot below the coefficinets instead of above,
and not include the heading "Plot of the Maass form in the hyperbolic plane" when
there is no plot.

And maybe a more accurate title would be "Plot of the Maass form in a fundamental domain"

Issue #456, opened 2015-10-27
Maass forms coefficients


davidfarmer:
Under displaying coefficients for a form, there is a search box without any explanation of what an acceptable entry is or submit button

http://beta.lmfdb.org/ModularForm/GL2/Q/Maass/4f5558cb88aece2646000014

lemurell:
This has been fixed. No search button is needed since it's filtering automatically. An explanation has been added, but it would be good if the explanation was reviewed so that it's informative enough. This is the same as #497 .

PR #574, opened 2015-11-09
Maass coeff

lemurell:
Added previous and next links for GL2 Maass forms and their L-functions. Added image for select GL2 Maass forms (the looks of these will be improved). Added explanation to filters for Maass forms (both filtering coef for individual forms and lists of forms) that resolve issues #497 and #456 (which are identical).


davidfarmer:
The picture here is just a dark blue rectangle:

/ModularForm/GL2/Q/Maass/5411905bacf756021f2c698a

lemurell:
I fixed that in the next push by hiding the graphs for level>1 because they don't render properly.

PR #577, opened 2015-11-11
Authentication update for import scripts ec, ecnf, hmfs

JohnCremona:


JohnCremona:
I put too much into the commit and will reissue the pull request after fixing.

PR #580, opened 2015-11-12
updated authentication code for elliptic_curves, ecnf and hmfs

JohnCremona:
Note: for the authentication to work you need to have a yaml version of the password info.  This is not part of the commit!  (but .gitignore is slightly changed to ignore that as well as other password* files).
Email john.cremona@gmail.com for a copy of that file.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,4 @@
-password
-password.py
+password*
 config.ini
 
 *.pyc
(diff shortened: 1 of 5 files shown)


JohnCremona:
I am merging this myself since it does not affect the webserver at all, only import & related scripts which I use all the time.

Issue #562, opened 2015-11-04
Label for ECNF isogeny class is incorrect (gives label of an ECNF) 


sibilant:
The label of isogeny classes of ECNFs is incorrect, at least in certain places: instead of the label of the isogeny class, we get the label of the first elliptic curve in this isogeny class.

Example: http://beta.lmfdb.org/EllipticCurve/2.2.5.1/31.1/a/
First header on the page should say "Elliptic curves in class 31.1-a" rather than "Elliptic curves in class 31.1-a1". Same for the label in the properties box. 

Note that this apparently never happens for elliptic curves over Q (Example: http://www.lmfdb.org/EllipticCurve/Q/11/a/).

JohnCremona:
You are right, my fault.  The naming of the data fields is different for
ec/Q.   Will be a trivial thing to fix.

JohnCremona:
Working on this now.

JohnCremona:
Fixed by pull request #582

PR #582, opened 2015-11-12
fix for issue #562 (heading for ecnf isogeny class page)

JohnCremona:
This fixes the issue, and also makes a similar fix for the Properties box and includes nf knowls for the base field.  All in 3 lines -- a bargain!


jwj61:
I think we had a general preference to not use knowls in the property box.  The item will appear in the main part of the page as a knowl, but the properties box gets just text.

JohnCremona:
OK, I will make that change.  Thanks for looking.

JohnCremona:
OK now?

jwj61:
Not working for me.  I think you need 

self.ECNF.field.field_pretty()

instead of field_pretty(self..)

JohnCremona:
That's funny, I'm sure it worked for me.  For example 
http://localhost:37777/EllipticCurve/2.2.5.1/31.1/a/1
gives no problems?

Meanwhile I added another small fix for issue #589 to this p.r.

PR #590, opened 2015-11-13
Short char bc

nmascot:
Fix to ticket Dirichlet character breadcrumbs #410. Also changed breadcrumbs for Hecke character since they are even longer, cf. Character/Hecke/3.1.44.1/61.-22a0+1a1/1.

This request is a new copy of #487, I had to remove the branch corresponding to #487 due to a problem with github.

### DIFF
--- a/lmfdb/characters/main.py
+++ b/lmfdb/characters/main.py
@@ -14,7 +14,7 @@ from lmfdb.utils import to_dict, parse_range, make_logger
 from lmfdb.WebCharacter import *
 from lmfdb.characters import characters_page, logger
 import ListCharacters
-from lmfdb.WebNumberField import WebNumberField
+##from lmfdb.WebNumberField import WebNumberField
 
 try:
     from dirichlet_conrey import *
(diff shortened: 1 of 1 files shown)


JohnCremona:
Why is the last part of the breadcrumb for

http://localhost:37777/Character/Hecke/3.1.44.1/61.-22a0+1a1/1

(for example) #[1] and not just #1?  I don't know how Hecke chars are labelled... but this is not just cosmetic, it results in the URL linked to by that breadcrumb to have [1] in it and that crashes.  So this *is* something to be fixed.

In the python code which defines these breadcrumbs there are many hard-coded URLs.  These should be replaced by calls to url_for().   This would avoid errors such as the previous one.

I hope it is not unreasonable to delay merging this until those more serious issues (broken links!) are fixed -- even though they were already broken.

nmascot:
Hi John,

Yes, I now the links are borkoen, actually the last thing I did on this 
branch was an attempt to fix them. Howver, I was then told (I cannot 
remeber by whom, but it coild be Brian Conrey) that the Hecke characters 
are only half finished and still buggy, and I was encouraged to focus on 
something else.

If you think it is important, I could try to see what I can do.

Best,
Nicolas

jwj61:
Yes, Hecke characters are only partly done, hidden under beta, and not expected to make the May release.  Fixing things related to them is still good, but not a priority.  In any case, this pull request should not be held up by Hecke characters.

JohnCremona:
Ok then please merge!

PR #591, opened 2015-11-13
change source link at foot of page to point to beta or prod branch la…

JohnCremona:
…test commit list depending on environment BETA

(it used to point to the diffs of the latest commit on master, which was not at all helpful especially on the production server).

Ideally it would also point to beta when running locally in debug mode, but we (JJ & JC) could not get that to work.


JohnCremona:
Pushed to beta so can already be seen at http://beta.lmfdb.org/ (click on the date)

PR #593, opened 2015-11-16
Added degree 4 L-functions browsing page

patrickkuehn:
The degree 4 browsing page has been updated with a description of the L-functions of degree 4 and a sample of degree 4 L-functions in a table.

### DIFF
--- a/lmfdb/lfunctions/main.py
+++ b/lmfdb/lfunctions/main.py
@@ -42,7 +42,7 @@ def l_function_history():
     bc = [('L-functions', url_for('.l_function_top_page')),
           (t, url_for('.l_function_history'))]
     return render_template(_single_knowl, title=t, kid='lfunction.history', body_class='', bread=bc)
-    
+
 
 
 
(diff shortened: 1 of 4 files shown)


JohnCremona:
I have been taking a look at this, to check that links work, and so on.

In the first line of the table, should "Symmetric cube of rank 0 elliptic curve" be a link to the same place as in the column to the left, as in the other rows?

The "Maass for GL(4)" link does not work for me.  And should that say "Maass for for GL(4)"?

Same in the line below, and in the bottom line.

The html has actual URLs in it, which should not be, and makes broken links such as the ones above more likely.  These should all be done using url_for().

I would like to see an example of an elliptic curve over a number field, in the table and also in the Browse list 

I think the Browse by underlying object should a list (down the page, not across).

davidfarmer:
I will look at this carefully, and make a few changes before
it gets pushed to beta.

davidfarmer:
Merging, and will submit an issue for items that need to be addressed immediately.

PR #600, opened 2015-11-18
Reorganized Siegel main page.

wbl:


davidfarmer:
Probably we will not merge this because it does not address the issue that the
links are not comparable: some have actual data, and some just tables of dimensions.

davidfarmer:
Will not merge. Currently focusing of changes to the home pages, in another pending pull request.

PR #601, opened 2015-11-18
Ack

swisherh:
I updated the workshops list and added a test file.  The test file tests the ack. page, the worksh. page, and all the links on the workshop page except ICERM which I couldn't get to work for some reason.


JohnCremona:
Good job!   Does this link:  http://indico.ictp.it/event/a13219/  not work for you?  It does for me.
Also you should include ICTP as a sponsor for that meeting, since they did pay for a lot of it.
Otherwise a great improvement.

Issue #599, opened 2015-11-18
breadcrumbs


davidfarmer:
are needed for:

1.  Top page
2. Modular forms top page

JohnCremona:
I am working on this.

JohnCremona:
Since p.r. #602 has been merged, I will close this.

PR #602, opened 2015-11-18
Breadcrumbs

JohnCremona:
Added & made more consistent breadcrumbs everywhere.

Also replaced Top with a house symbol (including for Home page).  This can be changed but all the rest should be merged as soon as possible to avoid future conflicts.

Fixes Issue #599

### DIFF
--- a/lmfdb/characters/main.py
+++ b/lmfdb/characters/main.py
@@ -125,10 +125,10 @@ def render_Dirichletwebpage(modulus=None, number=None):
         except IndexError:
             info['navi'] = [("next",) + navi([info['next']])[0]]
         m,n = info['modlabel'], info['numlabel']
-        info['bread'] = [('Characters','/Character'),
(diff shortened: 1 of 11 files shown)


davidfarmer:
I don't think the home house is big enough to tell that is what it is supposed to be.

Try putting this in lmfdb/templates/style.css at line 228:

#header .bread a:first-child {
    font-size: 150%;
}

and also comment out line 225, which says
    padding-top: 2px;

Even with those changes, it is hard to tell it is a house.

JohnCremona:
I know that the house symbol is not good enough.  When I have time (tomorrow) I can try other things.  But all the other breadcrumb changes are surely OK and need to be merged soon, so perhaps it is better for now to reinstate "Top" and not worry about the symbol.

davidfarmer:
I will merge and then make those CSS changes

JohnCremona:
Thanks!

PR #605, opened 2015-11-19
Fix issue 570, missing gamma factors in some L-functions.

jwj61:
The bug was on the L-function side.  Any L-function with a mu_i of multiplicity bigger than 1 which is not the largest mu_i would not get an exponent on the corresponding gamma factor.  This happens a lot for higher degree Artin representations.  An example which is now fixed is

  http://localhost:37777/L/ArtinRepresentation/4/1609/1/

### DIFF
--- a/lmfdb/lfunctions/Lfunctionutilities.py
+++ b/lmfdb/lfunctions/Lfunctionutilities.py
@@ -686,6 +686,8 @@ def lfuncFEtex(L, fmt):
                 curr_nu_exp += 1
             else:
                 old_nu = nu
+                if curr_nu_exp > 1:
+                    ans += "^{" + str(curr_nu_exp) + "}"
                 curr_nu_exp = 1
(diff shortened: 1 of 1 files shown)


JohnCremona:
I am probably convinced that the fix is correct!  Though something along the lines of

sage: mu_list = [1,1,1,2,2,3,3,3]
sage: dict((m,mu_list.count(m)) for m in set(mu_list))
{1: 3, 2: 2, 3: 3}

might end up a lot simpler....

Also it seems as if the same change needs to be made for the nu exponents.

jwj61:
I added the logic for nu's, but don't know of any L-functions where this would come up (at least degree 6 with two nu's and the smaller one repeated).

I didn't use the simplification since people closer to this code should deal with that.  For example, maybe they would want them sorted in a particular way which ends up different from how "set" sorts them, since they can be complex numbers.

JohnCremona:
Ok.  Shows how much I know about L-functions.  I'll  let David merge it.

Issue #603, opened 2015-11-19
Search elliptic curves by j-invariant


JohnCremona:
This is possible at present over Q only.  Since the input is a number field element, it needs to be parsed properly, and the fiekd also specified.

JohnCremona:
See pull request #612 for a solution.

PR #612, opened 2015-11-20
added j-invariant search to ec/nf

JohnCremona:
As it says:  go to http://localhost:37777/EllipticCurve/ and put in a field label and a j-invariant (a rational number is OK) and see what happens.  It will complain if you enter j but no field.

Fixes issue #603


jwj61:
This is not working as advertised.

If I put in just a j-invariant (no field), I get curves if I pick "1728" and no fields if I pick something with an "a" in it, but no error messages.

I then found a curve over 2.0.11.1, it had j-invariant 842425675/7921 * a -1150047337/7921.  If I enter the j-invariant and the field, I get nothing (and should have gotten the curve I took the j-value from).  If I enter just the j-invariant, I get nothing, including no error message.

I also tried the sample j-invariant (provided by the "e.g.") with no field, and got no curves and no error message.

JohnCremona:
Thanks for looking, and sorry it appears not to work.  I did test it!  I will investigate.

JohnCremona:
I checked  and cannot reproduce the problems you found.  The new commit is to make sure that this branch is up-to-date with master, and also adds the relevant field (2.2.89.1) to the j-invariant in the hint (but note that if you cut and paste from the hint text you need to replace the minus sign with a simple '-' character -- to be fixed).

davidfarmer:
I will see if this branch works for me.

davidfarmer:
When I enter only a j-invariant, I get an error message that I also need to enter a field.

If I enter a j-invariant and the field 2.2.5.1 then it seems to work properly.
(However, I can't cut-and-paste the sample j-invariant because the minus sign is not
really a minus, as previously noted.)

If I enter a j-invariant and the field Qsqrt5 then it says

Error
unrecognised field: Qsqrt5

JohnCremona:
On 1 December 2015 at 10:16, davidfarmer <notifications@github.com> wrote:

> When I enter only a j-invariant, I get an error message that I also need
> to enter a field.
>

That is as intended.  Is there a better way?  Perhaps if the j entered is
actually rational then it makes sense for the field to be blank but I would
need to work out how then to do a search covering all fields.



> If I enter a j-invariant and the field 2.2.5.1 then it seems to work
> properly.
> (However, I can't cut-and-paste the sample j-invariant because the minus
> sign is not
> really a minus, as previously noted.)
>

I don't know how to fix that.  The minus sign I get when I cut/paste is a
string of length 3 in python.  A solution would be rather welcome (here and
elsewhere).


> If I enter a j-invariant and the field Qsqrt5 then it says
>
> Error
> unrecognised field: Qsqrt5
>

I will fix that, it just needs some processing of the input.  Watch this
space...

John

JohnCremona:
I fixed the parsing of Qsqrt5 and similar.  Really stupid: I called the parse_field_string function but did not use the result...

davidfarmer:
When I edit
lmfdb/ecnf/templates/ecnf-index.html
and search for the "-" character, then I find a match on line 79 (in "1-100"), but not in line 99
(in "(818172355/1024)*a−(2133314821/512)").

So it looks like the source has a different character, which happens to look like "-".

JohnCremona:
On 1 December 2015 at 13:55, davidfarmer <notifications@github.com> wrote:

> When I edit
> lmfdb/ecnf/templates/ecnf-index.html
> and search for the "-" character, then I find a match on line 79 (in
> "1-100"), but not in line 99
> (in "(818172355/1024)*a−(2133314821/512)").
>
> So it looks like the source has a different character, which happens to
> look like "-".
>

Well spotted!  Of course I got that example from the home page of some
curve, where I found it very hard to cut/paste for the j-invariant as
displayed on that page (via MathJax)....fixed now.

JohnCremona:
Well spotted!  Of course I got that example from the home page of some curve, where I found it very hard to cut/paste for the j-invariant as displayed on that page (via MathJax)....fixed now.

PR #621, opened 2015-12-02
Newdq

jwj61:
Add dynamic data quality knowl for extent of data for global number fields.

Added data quality menu to home page of number fields.  Contents of "how computed" needs improvement, but that is inside a knowl.  This and the last are accessible from main number field page.

The learnmore entry only now appears on main number field page and pages for any quadratic imaginary field, but no where else.

Removed extaneous closing html tags from some templates -- no visible change.


JohnCremona:
I'll take a good look at this as a model for the elliptic curves pages.

JohnCremona:
The knowl nf.label contains almost the same (but not quite!) as http://beta.lmfdb.org/NumberField/FieldLabels -- wherever that comes from.  Presumable the latter is, or should be, a static version of the same knowl?

PR #622, opened 2015-12-02
Filling in text in the general information pages.  

jwj61:
Click on Fields, Groups, and Representations in the sidebar to view.

This is all text -- no programming.


JohnCremona:
Good start -- should we not have this information in knowls though?  It would certainly make it easier to edit.

PR #625, opened 2015-12-02
changes to sidebar

JohnCremona:
Lattices now appear in the sidebar (under beta only).  Also added (tagged "future" so not visible except under beta):  Local zeta-functions (under L-functions), and Sato-Tate groups (under groups) and some small tidying up.

Not that the design was supposed to be that this sort of thing *only* required changing the sidebar.yaml file;  however it also always seems to also requre changes to the html template too, whose logic is quite baroque.

### DIFF
--- a/lmfdb/sidebar.yaml
+++ b/lmfdb/sidebar.yaml
@@ -61,7 +61,7 @@
           status: beta
         - title: dummy
           status: future
-        - title: Local $\zeta$-functions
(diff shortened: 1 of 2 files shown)


JohnCremona:
I corrected "Local zeta functions" to "Local L-functions".  Also, Samuele said he is happy for Lattices to be visible in beta.  So this can be merged now.

PR #629, opened 2015-12-02
added data quality knowls and learn-more links for ec/Q

JohnCremona:
Go to any elliptic curve page (over Q only so far), e.g. http://localhost:37777/EllipticCurve/Q/  and look for the Learn More About box.


JohnCremona:
This had positive feedback at the session today.

PR #633, opened 2015-12-03
New artin representation code

jwj61:
Sorry for the big commit, but there did not seem to be a good way to do this piecemeal.

There are still improvements which can be made, but I don't think there are any regressions.

New features are
 * artin rep'ns now have labels
 * on the page of an artin rep'n, we give the size of the Galois conj class
 * on the page of an artin rep'n, we give the "first" permutation representation containing it.  First means smallest degree, and then smallest t-number.
 * artin rep'n search results have paging

Some things to check (if you want):
 * search for an artin rep by label, like 3.3e4_19e2.12t33.1c2
 * search for degree 2 artin rep'ns and page back and forth
 * go to http://127.0.0.1:37777/NumberField/4.0.592.1 , click on the lines at the bottom for the two 3-dimensional representations and notice that they have different "smallest permutation representations", the one which has the star on the number field page will be a degree 4 version of S_4, and the other will be a degree 6 version

### DIFF
--- a/lmfdb/WebNumberField.py
+++ b/lmfdb/WebNumberField.py
@@ -454,7 +454,7 @@ class WebNumberField:
             else:
                 if "nfgg" not in self._data:
                     from math_classes import NumberFieldGaloisGroup
-                    nfgg = NumberFieldGaloisGroup.find_one({"label": self.label})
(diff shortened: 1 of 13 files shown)


JohnCremona:
I am looking at this now...

Issue #636, opened 2015-12-03
Invalid factors of End x RR


kedlaya:
Consider the page:

http://beta.lmfdb.org/Genus2Curve/Q/11664/a/11664/1

Some of the endomorphism data includes the line:

End(J_K)⊗R	≃ 	Invalid factors of End x RR... please report this

Duly reported!

JRSijsling:
Will be fixed in an upcoming pull request. The issue was (fortunately!) only that I had forgotten to take into account the obviously possible case where this ring is CC x CC.

JohnCremona:
Fixed in #639

PR #637, opened 2015-12-03
corrected polynomial display on genus 2 curve pages (issue #634)

JohnCremona:
Fix for issue #634

PR #639, opened 2015-12-03
Debugging of genus 2 endomorphism functionality

JRSijsling:
Fixed three small issues:

(1) Monomials of high degree were represented incorrectly as "x^{1}0" instead of "x^{10}", fixed by using latex() instead of str();
(2) Cases with endomorphism ring tensor RR isomorphic to CC x CC were for some reason not accounted for and resulted in a debug statement, so this possibility has been added now;
(3) Cremona and LMFDB labels were confused since my database did not take into account that the "label" field in the LMFDB database for a rational elliptic curve is in fact the Cremona label, not the LMFDB label. Accounted for by an imported conversion function, and will update the database later.

### DIFF
--- a/lmfdb/genus2_curves/templates/curve_g2.html
+++ b/lmfdb/genus2_curves/templates/curve_g2.html
@@ -258,6 +258,14 @@ function show_invs(invstyle) {
 
 <br>
 
+<h3 style="display: inline"> Decomposition </h3>
+
+<!-- Description of a splitting field of minimal degree: -->
+<p>{{data.endodata.spl_fod_statement|safe}}</p>
+
+<!-- Description of the splittings themselves. -->
+<p>{{data.endodata.spl_statement|safe}}</p>
+
 <h3 style="display: inline"> {{ KNOWL('g2c.jac_endomorphisms', title='Endomorphisms') }} </h3>
 
 <p>{{data.endodata.gl2_statement_base}}</p>
@@ -275,13 +283,5 @@ function show_invs(invstyle) {
 <h3 style="display: inline">{{data.endodata.lattice_statement_preamble}}</h3>
 
 {{data.endodata.lattice_statement|safe}}
-
-<h3 style="display: inline"> Decomposition </h3>
-
-<!-- Description of a splitting field of minimal degree: -->
-<p>{{data.endodata.spl_fod_statement|safe}}</p>
-
-<!-- Description of the splittings themselves. -->
-<p>{{data.endodata.spl_statement|safe}}</p>
 
 {% endblock %}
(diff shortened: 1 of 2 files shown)


JohnCremona:
Note that (1) and (3) are fixed in my two pull requests.  Jeroen, you might like to take a look at those!  Otherwise this could be merged and mine ditched.  I'll look at your fixes now.

jvoight:
Note this will resolve issue #636.

JRSijsling:
Sorry about the doubling, I had just heard about the issues this morning and did not even know that they were already on the Issues list. As for #637, I did not copy that but I guess that we thought alike. I have will update the pull request with the changes in #636 so as to have the best of all worlds, also in such a way that it is easier to get right when I finally update the database.

Love the url_for('ec.by_ec_label', label = label) thing instead of the verbose thing that I put in. It would be great to have for elliptic curves over number fields as well, but I am not quite competent enough to copy this correctly.

davidfarmer:
It might be nice to put the "Decomposition" section earlier,
since it is short, interesting, and gets lost at the bottom when
the Endomorphisms section is long.

JohnCremona:
On 3 December 2015 at 15:45, Jeroen Sijsling <notifications@github.com>
wrote:

> Sorry about the doubling, I had just heard about the issues this morning
> and did not even know that they were already on the Issues list. As for
> #637 <https://github.com/LMFDB/lmfdb/pull/637>, I did not copy that but I
> guess that we thought alike. I have will update the pull request with the
> changes in #636 <https://github.com/LMFDB/lmfdb/issues/636> so as to have
> the best of all worlds, also in such a way that it is easier to get right
> when I finally update the database.
>
> Love the url_for('ec.by_ec_label', label = label) thing instead of the
> verbose thing that I put in. It would be great to have for elliptic curves
> over number fields as well, but I am not quite competent enough to copy
> this correctly.
>

That should be possible:  use show_ecnf(nf, conductor_label, class_label,
number)

JRSijsling:
David: good point, I will reorder it and we can see how that looks.

John: That works, yes, but I was wondering if there was something that just accepts the label plain and simple like in the case of rational elliptic curves. The function by_ec_label is very difficult to break as it tries many possible interpretations and permutations of the label. (In fact, even putting in the Cremona label directly would have worked!) I was just wondering if a similar thing existed in general.

JohnCremona:
On 3 Dec 2015 17:45, "Jeroen Sijsling" <notifications@github.com> wrote:
>
> David: good point, I will reorder it and we can see how that looks.
>
> John: That works, yes, but I was wondering if there was something that
just accepts the label plain and simple like in the case of rational
elliptic curves. The function by_ec_label is very difficult to break as it
tries many possible interpretations and permutations of the label. (In
fact, even putting in the Cremona label directly would have worked!) I was
just wondering if a similar thing existed in general.

No, at least not yet.

Issue #635, opened 2015-12-03
Decomposition of genus 2 curves show wrong elliptic curve labels


kedlaya:
Consider the page:

http://beta.lmfdb.org/Genus2Curve/Q/11664/a/11664/1

At the bottom, there are references to elliptic curves "432b1" and "27a3" (Cremona-style labels) but the links are to "432.b1" and "27.a3" (LMFDB labels), which are different curves entirely. Within LMFDB I think we should be consistently using the LMFDB-style labels, but in any case we need to eliminate the confusion about whether we are linking to the correct curve.

JohnCremona:
I'll fix that too,

JohnCremona:
Fixed  in #639

PR #640, opened 2015-12-03
Fixed ec labels on g2 curve pages for Issue #635

JohnCremona:
Fixed Issue #635 -- the database stores Cremona labels and these had to be converted.


JohnCremona:
Closing without merging as #639 took over the same fix.

PR #641, opened 2015-12-04
Make number field label page use the corresponding knowl 

jwj61:
(which has been brought up to date).


jwj61:
Forgot the testing url:

  http://127.0.0.1:37777/NumberField/FieldLabels

although the change is internal.  The url should look the same before and after.

JohnCremona:
OK, merging -- but edit the knowl where it says "cases where there are more than field with" !!

Issue #645, opened 2015-12-14
Trivial Dirichlet Character


sanni85:
The page of the trivial character needs to be fixed:
http://beta.lmfdb.org/Character/Dirichlet/1/1

JohnCremona:
2/1 also, since again the trivial group has no generators.

PR #647, opened 2015-12-14
Do not display values on gens when (Z/NZ)* is trivial

nmascot:
When N<=2, (Z/NZ)* has 0 generators, and so for Dirichlet characters with modulus N the "Values on generators" section does not look good (just a lonely arrow). This pull request prevents the section from being displayed when N<=2.


nmascot:
I forgot to mention: this fixes #645 .

PR #692, opened 2015-12-16
Refactor

fredstro:


JohnCremona:
Individual commits were reviews over at fredstro.

Issue #726, opened 2015-12-18
Allow defining polynomial to wrap


davidfarmer:
When this page narrow, the defining polynomial is forced to be below the properties box:
ModularForm/GL2/Q/holomorphic/11/5/2/a/

JohnCremona:
Currently (21/12/15) ModularForm/GL2/Q/holomorphic/11/5/2/a/ crashed on beta.  Fixing...

davidfarmer:
After pulling upstream master (and John and Fredrik's refactor branches, which gave nothing new)
the page 
ModularForm/GL2/Q/holomorphic/11/5/2/a/
still fails for me.  The error is in
lmfdb/modular_forms/elliptic_modular_forms/backend/web_newforms.py
    line 541, in WebNewForm_cached
return M.hecke_orbits[label]
    KeyError: 'a'

Also, if I make the url shorter (ModularForm/GL2/Q/holomorphic/11/5/2/) then it also fails, with error
lmfdb/modular_forms/elliptic_modular_forms/views/emf_render_web_modform_space.py
    line 106, in set_info_for_modular_form_space
WMFS_rep = WebModFormSpace_cached(level = level, weight = weight, cuspidal=True, character = c.number(),update_from_db=true)
NameError: global name 'true' is not defined

Probably I am just missing some changes made in the past day as I travelled, but I have not been
able to unravel all the emails.

JohnCremona:
Both those URLs work OK on beta (did you try that?) and for my local copy.

On the other hand, "update_from_db=true" should surely be
"update_from_db=True" and will cause trouble some time.

I put in a test (normally skipped) to load all pages of the form

ModularForm/GL2/Q/holomorphic/N/k/?group=i

for i=0, 1; 1<=N<=40; 2<=k<=20 (k even) to make sure those pages all load
without crashing, but without checking the page's contents.  Something
similar to loop over individual newforms spaces
like ModularForm/GL2/Q/holomorphic/11/5/2/a/
but that would take more effort: the character number should be in [1,N]
and coprime, though the others do not cause an error;  the form letter will
cause an error if out of range though (e.g. 11/5/2/b) so such a test would
have to know how far to go with each (N,K,chi) triple.

John

davidfarmer:
They work fine on beta for me. 

It looks like I can't get to any modular forms on my local version.  This is after checking out master
and then pulling upstream master.

Not sure if these logging messages are relevant:
[...]

PR #739, opened 2015-12-22
Better splitting of long series and polynomials

davidfarmer:
Fixes issue #726 .

Example:

ModularForm/GL2/Q/holomorphic/11/5/2/a/

or any other modular form with coefficients in a higher degree field.

### DIFF
--- a/lmfdb/genus2_curves/genus2_curve.py
+++ b/lmfdb/genus2_curves/genus2_curve.py
@@ -8,7 +8,7 @@ from flask import Flask, session, g, render_template, url_for, request, redirect
 import tempfile
 import os
 
-from lmfdb.utils import ajax_more, image_src, web_latex, to_dict, parse_range2, web_latex_split_on_pm, comma, clean_input, parse_torsion_structure
+from lmfdb.utils import ajax_more, image_src, web_latex, to_dict, parse_range2, comma, clean_input, parse_torsion_structure
 from lmfdb.number_fields.number_field import parse_list, parse_discs, make_disc_key
 from lmfdb.genus2_curves import g2c_page, g2c_logger
 from lmfdb.genus2_curves.isog_class import G2Cisog_class, url_for_label, isog_url_for_label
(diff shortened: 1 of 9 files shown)


JohnCremona:
David, what does your Sage give for this:

sage: from sage.env import SAGE_VERSION
sage: version_major, version_minor = [int(x) for x in SAGE_VERSION.split('.')[:2]]
sage: (version_major, version_minor)
sage: (version_major, version_minor) >= (6,9)

-- since this PR includes the commenting out of the code which checks the Sage version and sets emf_version accordingly?  SImilarly, when you start lmfdb locally what do you see in the startup messages similar to this:

INFO:emf@2015-12-23 10:24:53,171: Initializing elliptic modular forms blueprint with Sage version 6.10, emf version 1.3 [__init__.py:22] ?

davidfarmer:
I have Sage version 6.7. (so false on that last line)
On startup the following appears twice
(not sure why it happens twice)


INFO:emf@2015-12-23 05:52:36,910: Initializing elliptic modular forms 
blueprint with Sage version 6.7, emf version 1.3 [__init__.py:23]


The reason I commented out that code and set the emf version to 1.3
is that (if I understand correctly) the available data has that version.
So you don't have access to any modular forms if you use another
emf version, because the version is used when looking for data in
the database.

I could have changed 1.1 to 1.3 in the final else, but then we would have
an if-then-else block that always did the same thing.

SAGE_VERSION isn't used anywhere else in the code.

JohnCremona:
Ok fair enough, this was only meant to be a temporary transitional switch.

davidfarmer:
If the pull request that fixes this is okay, then I think we
are ready to update beta and production.  And email the google
group to let them know that modualr forms are in better shape.

fredstro:
the reason for that version switch is that there was some changes in sage
pickles (as usual) in 6.8 so if you are  using 6.7 you might still
experience some problems with emf_version  1.3 so please update your sage.
Fredrik

JohnCremona:
I'll test, pull and push later today family duties permitting.

PR #742, opened 2015-12-23
This fixes #741

### LINKED ISSUE #741: Server error
http://www.lmfdb.org/ModularForm/GL2/Q/holomorphic/7/7/5/ causes an Server Error.
In fact, all pages of spaces that are not in the db but have Galois conjugates in the db, cause this error.

sehlen:


sehlen:
This fixes an internal server error when navigating to spaces that are not in the db but have Galois conjugates in the db. Moreover, it includes a small sanity check that sometimes caused an error (not on a page as fr as I can see).

Issue #749, opened 2016-01-04
Genus 2 curve twists should be identified using g2inv rather than igusa_clebsch


AndrewVSutherland:
Search for twists should use g2inv rather than igusa_clebsch

On Sat, Jan 2, 2016 at 7:20 AM, Andrew Sutherland
<drew@math.mit.edu> wrote:
While working on the ANTS paper I noticed the following issue in the LMFDB.

The two curves in isogeny class 1088b:
http://www.lmfdb.org/Genus2Curve/Q/1088/b/2176/1
http://www.lmfdb.org/Genus2Curve/Q/1088/b/2176/2
are twists of each other (isomorphic over Q(i)), but the LMFDB does not recognize this fact. I believe the issue is that the Igusa invariants are not exactly the same (there are some sign differences), but the relevant ratios that determine the Qbar-isomorphic class are the same, as can be seen by looking at the G2Invariants, which do coincide.

Magma code demonstrating the problem:
R<x> := PolynomialRing(Rationals());
C1 := HyperellipticCurve(R![34, 0, 24, 0, 4], R![0, 1, 0, 1]);
C2 := HyperellipticCurve(R![-14, 32, 1, -13, -1, 1], R![0, 0, 1,1]);
IgusaInvariants(C1),IgusaInvariants(C2);
G2Invariants(C1),G2Invariants(C2);
Qi:=NumberField(x^2+1);
IsIsomorphic(ChangeRing(C1,Qi),ChangeRing(C2,Qi));

To fix this in a way that keeps the twist lookup efficient, I think we are going to need to add the G2Invariants to the database (I believethey are currently computed on the fly).

Let me know what you think -- if you agree I can go ahead and put this in (and add an index for the new attribute). [THIS HAS NOW BEEN DONE]
The code that searches for twists will then need to be changed
accordingly.

Drew

JohnCremona:
I agree -- as with j-invariants of elliptic curves, once we decisded to presomute them and put them into the database several things (such as searching for twists) became a lot simpler.  You have already done the one-off work of computing and storing the relevant data, so go ahead with the code changes which follow.  Thanks!

JRSijsling:
This has been dealt with by #751.

PR #750, opened 2016-01-06
Genus 2 twists

JRSijsling:
Precomputed invariants used; finding twists now uses G2 invariants. Deals with Issue #749.


JRSijsling:
(Messed up history by abortive commits and amendments: this is now #751.

PR #751, opened 2016-01-06
Genus 2 twists

JRSijsling:
Precomputed invariants used; finding twists now uses G2-invariants.
Fixes Issue #749.


JRSijsling:
Related: fixed sign issues when normalizing Igusa and Igusa-Clebsch invariants. These are now normalized in such a way that the discriminant always has positive sign, which makes the normalization unique (which it had better be).

PR #752, opened 2016-01-07
Genus 2 label bug

JRSijsling:
Fixed a bug where a variable called lmfdb_label was not defined, resulting in
errors for splittings over quadratic fields


JohnCremona:
Can you include in pull requests like this an example of a URL which shows the problem, so people can see the before & after?
The previous few pull requests which were merged are already running on beta, by the way.

JRSijsling:
Good point, I will do that henceforth. A problem page for this particular case used to be http://www.lmfdb.org/Genus2Curve/Q/1088/a/1088/1.

PR #758, opened 2016-01-26
Issue755

JohnCremona:
This fixes Issue 755: 
   1.  The modular form URL in the related Objects list for an elliptic curve used the old-style numbering; now corrected
   2. if a complete and valid  URL for a classical modular form is given for which that form is not in the database, for example 
ModularForm/GL2/Q/holomorphic/201/2/1/b/
from
EllipticCurve/Q/201/b/1
then instead of crashing, a reasonable error message appears explaining why the modular form could not be displayed.


fredstro:
This indeed seems to fix issue 755 and does not cause any other issues

Issue #764, opened 2016-01-27
Add an acknowledgement to OpenDreamKit


JohnCremona:

JohnCremona:
See PR#765 which also adds an acknowledgment to MathJax.

JohnCremona:
Fixed in #765

Issue #740, opened 2015-12-23
Thank MathJax


davidfarmer:
Add MathJax to the list of software we use.

JohnCremona:
See PR #765

PR #765, opened 2016-01-27
added acknowledgements to ODK and MathJax

JohnCremona:
Not sure about the wording -- ODK (=OpenDreamKit) is huge and LMFDB is only involved with a small part of that.
Addresses Issues #764 and #740.


JohnCremona:
I had only positive feedback so am merging.

Issue #767, opened 2016-02-05
Weight 24 classical modular form error


jvoight:
http://www.lmfdb.org/ModularForm/GL2/Q/holomorphic/24/4/1/a/

gives an error.  This is a link from

http://www.lmfdb.org/ModularForm/GL2/Q/holomorphic/24/4/1/

which is a link from

http://www.lmfdb.org/ModularForm/GL2/Q/holomorphic/24/4

JohnCremona:
Fixed by #768

PR #768, opened 2016-02-05
changed the collection to the correct collection as well as the corre…

fredstro:
This should fix issue #767  
Changed the collection to the correct collection as well as the correct key for the records.


JohnCremona:
Thanks -- it works.

PR #776, opened 2016-02-18
Switch to larger database in genus 2

JRSijsling:
Recently we switched to a larger database for genus 2 curves. A new curve is for example

Genus2Curve/Q/847/d/847/1

As you can see this works without any need for this pull request. This does the following:

(1) Most importantly, we switch to a new endomorphism collection, which is now in genus2_curves, not in genus2_endomorphisms. The old one will be axed should this pull request be accepted.
(2) Removed all extraneous references and dirty hacks needed to make the previous database work.
(3) Fixed 6 more curves in the new database for which a bug occurred; one of these is Genus2Curve/Q/900601/a/900601/1.
(4) Some small layout and wording fixes.

### DIFF
--- a/lmfdb/genus2_curves/genus2_curve.py
+++ b/lmfdb/genus2_curves/genus2_curve.py
@@ -31,16 +31,6 @@ def db_g2c():
         g2cdb = lmfdb.base.getDBConnection().genus2_curves
     return g2cdb
 
-# TODO: Remove database below when switched to use of genus2_curves only
-
-g2endodb = None
-
-def db_g2endo():
-    global endodb
-    if g2endodb is None:
-        g2endodb = lmfdb.base.getDBConnection().genus2_endomorphisms
-    return g2endodb
-
 ###############################################################################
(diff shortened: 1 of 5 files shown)


JohnCremona:
Testing now.  To be doubly sure:  once this is merged and running on both prod and beta, we can use C.drop_database('genus2_endomorphisms') where C is the mongo client to delete that database?  At that point of course anyone running the old code would see errors.

I can delete that database but would prefer it if you did, Jeroen, after I post here that the above mergings have taken place.  And *please* go to https://github.com/LMFDB/lmfdb-inventory, delete the mention of genus2_endomorphisms, and fill in the data for genus2_curves.

JohnCremona:
All done -- I ran the test suite and tried some pages including the ones you mentioned above.  Branches master, beta and prod now include this and both webservers updated.

JRSijsling:
All right, I will drop the database, perhaps after a week to allow everyone to incorporate the new code. That said, those that would usually run it are in on this. After that I will also update the inventory.

Issue #790, opened 2016-03-15
crash on display of isogeny class of elliptic curve over Q(i)


JohnCremona:
Go to http://www.lmfdb.org/EllipticCurve/2.0.4.1/%5B225%2C0%2C15%5D/a/6  (which is the base cghange to Q(i) of 15a1 (a.k.a. 15.a5) and click on the link to its isogeny class in the Related Objects column.  Boom.

I actually wanted to see the isogeny graph for something I was working on!

JohnCremona:
I have fixed this; there were two different things to do:
   1. conductor labels over imaginary quadratic fields were designed by me but I really hate them and they look like [225,0,15].  The character "[", ",", "]"  need to be quoted using urllib in some places (and unquoted again in others.
   2. Then the isogeny graph I was interested in looking at was a shape which cannot occur over Q and that broke the special graph code.

I have fixed both these and will put in a pull request after more testing.

JohnCremona:
Fixed and pushed to beta & prod.  See http://homepages.warwick.ac.uk/staff/J.E.Cremona/isogeny_graphs.html for the start of a gallery.

PR #791, opened 2016-03-15
issue 165: quote ecnf labels over imag quad field, and fix 2 isogeny …

JohnCremona:
…graphs

Fixes for issue #790 -- it uses urllib.quote/unquote on conductor labels, and also fixes an isogeny graph over Q(i) which crashed (it's a graph shape which can only occur over a field containsing 4th roots of unity).

The test suite passes.

Test URL:  EllipticCurve/2.0.4.1/%255B225%252C0%252C15%255D/a/  and also the ones in the Issue ticket.

### DIFF
--- a/lmfdb/ecnf/WebEllipticCurve.py
+++ b/lmfdb/ecnf/WebEllipticCurve.py
@@ -1,4 +1,5 @@
 from flask import url_for
+from urllib import quote
 from sage.all import ZZ, var, PolynomialRing, QQ, GCD, RealField, rainbow, implicit_plot, plot, text, Infinity
 from lmfdb.base import app, getDBConnection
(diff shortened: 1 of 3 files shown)


JohnCremona:
Holdit -- I have been systematically going through all the possible isogeny graphs and there are some more which crash.  A new patch will appear in due course.
On the bright side there are some fantastic isogeny graphs which do work!  I'll make a gallery.

JohnCremona:
OK now.   An example which did not work wit the original patch is EllipticCurve/2.0.3.1/%255B196%252C0%252C14%255D/a/ (and generally any curve with full 3-torsion and also 2-torsion).

Some of the new graph shapes could do with being hand-crafted to look optimal, but  that can be another Issue; this fixes two different bugs so should not be held up.

PR #805, opened 2016-03-21
Dataquality

jwj61:
Add data quality for local fields, Galois groups, and Artin representations.


jwj61:
Fixed glitch in tests

jwj61:
Updated to not show some of the completeness links

Issue #804, opened 2016-03-21
allow line breaking on defining equation


davidfarmer:
Try this in a narrow window:
EllipticCurve/4.4.2000.1/239.1/d/2

Issue #806, opened 2016-03-22
Weierstrass equation layout


davidfarmer:
The qquad that makes it look better for large equations, makes it look worse when
the equation is small.  For example

EllipticCurve/2.2.5.1/31.1/a/4

davidfarmer:
Also see issue #804

PR #810, opened 2016-03-22
ECNF scroll long W equation

davidfarmer:
See the Weierstrass equations on these pages, in both a wide and a
narrow window:

EllipticCurve/4.4.2000.1/239.1/d/2

EllipticCurve/2.0.3.1/[2268%2C36%2C18]/a/1

Both have problems in the current version on beta.


davidfarmer:
Fixes issues #804 and #806

PR #844, opened 2016-03-24
Add pretty name for quartic fields with Galois group V

jwj61:
As requested by JEC.

Sample pages http://127.0.0.1:37777/NumberField/4.0.2304.2 and http://127.0.0.1:37777/NumberField/4.0.225.1


JohnCremona:
Fails on 4.4.2000.1 (a C4 extension).  I think you need to count the subfields?

jwj61:
You are right (clearly).  It is now fixed

Issue #833, opened 2016-03-24
Don't show Galois groups of degree 2 L-function local factors


davidfarmer:
Because it is always C_2 at the good primes.  For example:

L/EllipticCurve/Q/11.a/

davidfarmer:
Also fix the C_1 \times Trivial issue when a bad factor splits completely:

L/Genus2Curve/Q/15475/a/

PR #847, opened 2016-03-25
Epfix

davidfarmer:
Don't show the Galois group of the local factors if the degree is 1 or 2

And use the Lfunctions collection instead of LfunctionsECtest


davidfarmer:
Fixes #833

PR #853, opened 2016-03-25
replaced Cremona's curve with 5077.a1 in the Hall of Fame

alinabucur:
I called it the BGZG curve, from Buhler-Gross-Zagier (who proved BSD for it in the rank aspect) and Goldfeld (who applied to the class number problem).


JohnCremona:
Oops, the L-function link is wrong, it should be to L/EllipticCurve/Q/5077.a/

JohnCremona:
I will fix that and also make the necessary corresponding changes to lmfdb/test_homepage.py (which checks the URLs on the home page are as expected and all work).

Issue #62, opened 2013-09-14
1-dimensional Artin representations friend link to Dirichlet characters


davidfarmer:
And similarly on the L-functions pages

jwj61:
Done

PR #867, opened 2016-03-27
Central characters for Artin representations

jwj61:
For a given Artin representation, we compute the Conrey label of its determinant (= its central character).  The character is displayed on the home page of the representation.  

- If the representation has dimension 1, we call it the corresponding Dirichlet character since that is what it is (which resolves issue #62).
- if the representation has dimension > 1, then we label it the determinant

If the modulus is less than 100000, then we put in a correspondingly named friend link.  If the modulus is too large, then the Dirichlet character's home page would not render.

To test, search for Artin representations of dimension 1 and of higher dimension to see the two behaviors, and you can use conductor ranges to see the friend link, or lack thereof.

### DIFF
--- a/lmfdb/artin_representations/main.py
+++ b/lmfdb/artin_representations/main.py
@@ -256,10 +256,11 @@ def render_artin_representation_webpage(label):
     if nf_url:
     	friends.append(("Artin Field", nf_url))
     cc = the_rep.central_character()
-    if the_rep.dimension()==1: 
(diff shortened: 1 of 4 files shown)


davidfarmer:
Line 58
quick = p2==2
of
lmfdb/math_classes.py
seems to not do anything.  Delete in a future pull request.

davidfarmer:
When I go to
ArtinRepresentation/1.7.3t1.1c1
and then follow the link to the L-function, it seems to have forgotten that the character is (7,2).

I'll submit an issue.

jwj61:
I think none of the Artin L-function pages know their character

PR #884, opened 2016-03-29
Tidy up L-function index pages (for different degrees)

jwj61:
This updates the yaml files for the degree n L-function home pages making them more uniform and adding some central characters, updates the yaml conversion scripts so that column headers will be knowls, and gives the new html files after running the scripts.  It also removes an unused line in another file.

To check, go to the summary pages for degree 2, 3, 4 home pages for L-functions.


jwj61:
More tidying up done (use varepsilon for the sign), and make degree 4 appear in the sidebar for the release version.

Issue #522, opened 2015-10-28
Genus 2 curve missing elliptic curve friends?


jvoight:
For the curve
  http://www.lmfdb.org/Genus2Curve/Q/720/b/116640/1
it says it is not simple over QQ, but it has no elliptic curve friends.

jvoight:
Same for http://www.lmfdb.org/Genus2Curve/Q/614656/a/614656/1

davidfarmer:
That information doesn't seem to be in the database.

JohnCremona:
Do you know what the conductors of the elliptic curves would be?

John

steffenmueller:
For 720.b.116640.1 the curves are 24.a4 and 30.a6.

Steffen Mueller

steffenmueller:
For 614656/a/614656/1 it's the square of 784.g2.

steffenmueller:
Whoops, apparently that information has already been added...

JohnCremona:
On 25 March 2016 at 15:24, Steffen Müller <notifications@github.com> wrote:

> Whoops, apparently that information has already been added...
>

I seem to remember there was an issue with Cremona / LMFDB labels, and we
fixed it.  So perhaps this issue can be closed.

jvoight:
This is not ready to be closed yet: although the elliptic curves appear in the main body, they are not listed in the "Friends" tab on the right-hand side.  This should be quick to add, and they belong there.

AurelPage:
working on it.

AurelPage:
Fixed by PR #895. Ready to be closed?

PR #895, opened 2016-03-29
g2: add EC from splitting as friends

AurelPage:
Ticket #522 
When a genus 2 curves splits, the corresponding elliptic curve is now in the "related object" box.

Examples:
/Genus2Curve/Q/614656/a/614656/1
/Genus2Curve/Q/720/b/116640/1

Question: should the related object be the elliptic curve or the isogeny class of EC?

### DIFF
--- a/lmfdb/genus2_curves/web_g2c.py
+++ b/lmfdb/genus2_curves/web_g2c.py
@@ -761,7 +761,7 @@ class WebG2C(object):
 
         # Curve
         set_code('curve',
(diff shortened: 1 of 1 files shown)


AurelPage:
Hold it, there are bugs. I will fix them and update.

AurelPage:
Should now work properly.

Example that was bugged with the previous commit:
/Genus2Curve/Q/249/a/249/1

PR #897, opened 2016-03-29
Standard colors

adeines:
Addes a file color.css in lmfdb/templates that defines, sets, and standardizes the colors.  Then removes all hex and rgb color definitions from all other files and sets colors to those in color.css.  Finally, in color.css other color schemes are added.


jwj61:
Looking at this

PR #901, opened 2016-03-29
Errors in parsing Hilbert Modular Forms searches

sanni85:
Typing hello in any search field for Hilbert modular forms gives a server error even if the parsing of the input is done correctly. This pull request fixes the html.


davidfarmer:
Okay, but it would be good if there were some indenting of levels in the template,
since it was hard to parse the new level of ifs in this pull request.

PR #905, opened 2016-03-30
Bigdirichlet

pascalmolin:
The main change is the introduction of a WebSmallDirichletCharacter class to be used when one only wants the properties of a character but no computations (this class does not precompute discrete logs). This allows to have homepages of much larger Dirichlet characters.

(maybe Small is misleading, I could rename it Lazy...)


pascalmolin:
fixed sage 7.0 issue

jwj61:
Checking this now

PR #915, opened 2016-03-30
Ecnf download

adeines:
Allows users to download ecnf searches.


JohnCremona:
I would still prefer it if the length of the array "data" was equal to the number of curves, with each entry being of the form [poly_coeffs, a_invs]!

adeines:
Ahh!! I new I forgot to do something.

adeines:
Should be fixed.

PR #932, opened 2016-03-30
Change labels

fredstro:
I have simply changed the labels for modular forms from the 'old Cremona': e.g. 11.2.1a to 'new Cremona': 11.2.1.a   and similarly for the embeddings: 11.2.1a.1 is now 11.2.1.a.1 
I.e. as requested by many (but not put into any issue as far as I could see) I am now using  dots as separator between all fields.
No URLs will be broken since we always use the / as separator in urls and the quick-search parses both the old and new style labels correctly.


JohnCremona:
I am worried about all the places where links are made to modular forms as Friends, where as well as the correct URL we also display the label.  For example EllipticCurve/Q/389/a/1 in the Related Objects box still shows "Modular Form 389.2.1a" though when you follow the link it now shows the modular form in its new name 389.2.1.a.

I don't think there are too many such places -- perhaps only EC/Q and isogeny classes/Q,  Better to fix them now than wait for someone to notice and object!

fredstro:
The pages in question should  use the function newform_label from
modular_forms.elliptic_modular_forms.emf_utils which is the "correct" way
to display a newform label. (although I have to admit that not even all
modular forms pages used this function everywhere )

Fredrik

JohnCremona:
That is excellent -- I strongly approve of having such a function as newform_label!  I will use it in the elliptic curve pages, and prove such a function for elliptic curves too.

Issue #702, opened 2015-12-16
Distinguish between even and odd characters


davidfarmer:
For a given weight, the space of cusp forms is automatically empty if the character
has the wrong parity.  But there are cases where it is 0 but that 0 is not forced by parity,
for example:
ModularForm/GL2/Q/holomorphic/11/2/

It would be nice to separate the characters by parity, so one could more easily see
the "interesting" empty spaces.

AurelPage:
Other example where a 0 is not forced by parity: (5,4).

fredstro:
I have made an addition to the table in the Gamma1 listing to show the parity of the characters. 
In addition I am running computations which adds this information to the database of dimensions to make it easier to fetch for this page.
See e.g.
ModularForm/GL2/Q/holomorphic/42/2/ 
ModularForm/GL2/Q/holomorphic/42/2/ 
ModularForm/GL2/Q/holomorphic/100/2/
have a look  at my and give me feedback

davidfarmer:
What do you think of this:
then parity forces the dimension to be 0, the entries in the right column are grey instead of black?
(Black means there is nothing to click to, so this gives a bit more information.)

If you like that idea then I could implement it.

fredstro:
I think that sounds good. 
In particular it is good that it makes it clear why some of the dimension 0 spaces can be clicked on and some not. 

Fredrik

PR #937, opened 2016-03-30
If a space is trivial because of parity, show it in grey

davidfarmer:
I think all possibilities occur on these two pages:

ModularForm/GL2/Q/holomorphic/5/4/
ModularForm/GL2/Q/holomorphic/42/2/

Also fixed a color problem with included knowls.

### DIFF
--- a/lmfdb/modular_forms/elliptic_modular_forms/views/emf_render_web_modform_space_gamma1.py
+++ b/lmfdb/modular_forms/elliptic_modular_forms/views/emf_render_web_modform_space_gamma1.py
@@ -90,7 +90,6 @@ def set_info_for_gamma1(level,weight,weight2=None):
             orbit = r['character_orbit']
             k = r['weight']
             parity = r.get('character_parity','n/a')
-            print "parity",parity
             if parity == 'n/a':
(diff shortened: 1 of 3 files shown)


davidfarmer:
Fixes the last part of issue #702

JohnCremona:
How about explaining this grey convention on the page, wehere we already have an explanation in small print of what "n/a" means?

Also wasn't there an intention to sort the spaces by parity?   So that (e.g.) ModularForm/GL2/Q/holomorphic/42/2/
would have the nontrivial even ones first, followed by the trivial grey odd spaces?

davidfarmer:
I added the sentence (after the n/a explanation):
Spaces in grey are trivial because the character and the weight have opposite parity.

I think the spaces should be in a natural order, perhaps by the label of the character,
and it would lose information if you grouped them by parity.  I assume there is some
logic to the current order (but I am not sure because I didn't implement it), so I would
rather leave the order as-is, and discuss that in a separate issue.

fredstro:
The order comes from (or at least it used to come from) the ordering of the
galois orbits as returned by the dirichletgroupconrey

Fredrik

Issue #956, opened 2016-03-31
crash: Sym^2 L-function of CM curves


AurelPage:
Sym^2 L-function of CM curves crashes.

/L/SymmetricPower/2/EllipticCurve/Q/32.a/
or click Symmetric square L-function on /EllipticCurve/Q/32/a/3

See also /roadmap: Recognize non-primitive L-functions and provide links to the primitive components. (sym^2 of CM curves, Dedekind zeta-function in the nonabelian case, etc)

davidfarmer:
There is a message given:

This Elliptic curve has complex multiplication and the symmetric power of 
its L-function is then not primitive. This has not yet been implemented

which should be shown, and not crash the page.

JohnCremona:
I will leave out the symmetric power links for CM curves.

JohnCremona:
This is fixed in PR #980

PR #980, opened 2016-03-31
EC/Q & isogeny class pages only show newform as friend when the newfo…

JohnCremona:
…rm is in the database.

This uses Fredrik's new useful function is_newform_in_db().


JohnCremona:
I will add a fix for #956 to this as it is similar and affects the same places in the same files.

JohnCremona:
Added the fix for #956

PR #988, opened 2016-03-31
Search by label HMF

sanni85:
On beta if you search by label "2.2.5.1-76.1-a", it returns a messy URL and says Bad Request Request Line is too large (7817 > 4094) (noticed by John Voight).

This pull request goes back to the inefficient way and solve the issue.


JohnCremona:
OK, that was my fault -- I had no idea that putting the big dictionary into the args list would have this effect (it *all* appears in the URL!!!).  There has to be a different way to do this.

Issue #585, opened 2015-11-12
Search instrucitons for lattices


davidfarmer:
http://beta.lmfdb.org/Lattice/

Search box lists options (E_8, Leech) which do not work.

kedlaya:
In fact none of the given options 

3.4.16.1.1, D3, E8, Leech lattice

seems to work. The first one is a label that doesn't occur. In the case of E8, the lattice exists (8.1.1.1.1) but the name doesn't find it.

kedlaya:
This is related (and trivial), so I'll add it rather than making a new issue: the explanations for the individual search boxes are not formatted consistently.

Dimension: e.g. 2, 3, 5 or a range of integers 1-5
Determinant: e.g. 1, 2, 3, or a range of integers 1-5
Level: e.g. 48, 336, 456 a range of integers 48-150

There is an inconsistency between whether or not the list example ends with a comma, and whether or not it is followed by the word "or".

sanni85:
This is fixed in #1017 #1115

PR #1017, opened 2016-04-01
Lattice8

sanni85:
Just added some root lattice to the database
A1, A2, ..., A14

I am still missing the theta series of A10, ..., A14 and we are planning to add higher dimensional lattices too, so I modified the code in order to deal with the case where no theta_series is stored.

Import script are fixed too (also for half integral weight forms) to remove explicit passwords.....(sorry)

Partially addressed issue #585 (at least the last remark)


sanni85:
wait, I should add the index creation in there

PR #1021, opened 2016-04-01
Lattice fixes high dimension additions

sanni85:
Just added some root lattice to the database
A1, A2, ..., A14

I am still missing the theta series of A10, ..., A14 and we are planning to add higher dimensional lattices too, so I modified the code in order to deal with the case where no theta_series is stored.

Import script are fixed too (also for half integral weight forms) to remove explicit passwords.....(sorry)

Partially addressed issue #585 (at least the last remark)

### DIFF
--- a/lmfdb/import_scripts/half_integral_weight_modular_forms/half_int_weight_mf.py
+++ b/lmfdb/import_scripts/half_integral_weight_modular_forms/half_int_weight_mf.py
@@ -32,8 +32,14 @@ import json
 import sage.all
 from sage.all import os
 
-from pymongo.connection import Connection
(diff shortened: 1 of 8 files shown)


sanni85:
Wait I am implementing the suggestions during the show and tell.

sanni85:
Now the Gram matrix is removed from the property box, the table for the genus representatives displays better

see
http://localhost:37777/Lattice/A10
http://localhost:37777/Lattice/3.209088.792.1.1

sanni85:
Sorry, forgot to check the tests, now everything pass: it was tested if "Normalized minimal vector list" was shown

PR #1034, opened 2016-04-01
Add dynamic knowls for local fields.

jwj61:
Now local number fields will have dynamic knowls.

This will be useful if/when we address the issue on local decompositions of global number fields at bad primes.


jwj61:
To test, try

 /knowledge/show/lf.field.data?label=5.5.5.4

PR #1038, opened 2016-04-02
general code cleanup, added statistics, fixed twists

AndrewVSutherland:
Final cleanup from the Bristol LMFDB workshop of March 2016 -- cleared out some unused code and fixed some inconsistencies and duplication in the code resulting from historical cutting and pasting (so lots of changes, but very few make any functional difference).  Added statistics page, fixed a bug in searching for twists that was introduced when the new parsing validation was added in the first week of the Bristol workshop, and fixed direct searching for isogeny class labels (which was also broken when the new parsing validation was introduced).

I've done a fair amount of testing and am unable to find any obvious bugs.  Once this pull request is merged I can get others to help with further testing in preparation for LMFDB 1.0.

### DIFF
--- a/lmfdb/genus2_curves/genus2_curve.py
+++ b/lmfdb/genus2_curves/genus2_curve.py
@@ -12,29 +12,34 @@ from markupsafe import Markup
 import tempfile
 import os
 
-from lmfdb.utils import ajax_more, image_src, web_latex, to_dict
+from lmfdb.utils import ajax_more, image_src, web_latex, to_dict, comma
 from lmfdb.search_parsing import parse_bool, parse_ints, parse_signed_ints, parse_bracketed_posints, parse_count, parse_start
 from lmfdb.number_fields.number_field import make_disc_key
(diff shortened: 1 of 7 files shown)


JohnCremona:
Great stats page!  I really like the way that each count is a link to a search for the corresponding curves, and how you have the stats for such a lot of attributes.  Some of that can be used for elliptic curve stats too.

AndrewVSutherland:
Thanks, it was actually pretty easy to do (about 20 lines of python and 
20 lines of html -- see the function init_g2c_stats in genus2_curve.py 
and statistics_g2.html).

I made it completely data driven so the code to generate and display the 
statistics actually knows nothing about the particular attribute.  
Adding a new one is just a matter of adding a single record with 4 
attributes to the stats_attribute_list in genus2_curves.py (in fact, 
this could even be put in the LMFDB so that adding more stats would not 
require any code changes, but that seemed like overkill).

See you tomorrow,

Drew

PR #922, opened 2016-03-30
initiate pages for higher genus with automorphisms

jenpaulhus:
Adding new files for new section on higher genus curves with automorphisms.  

See for example:
/HigherGenus/C/aut/
(The search is not fully functioning, if you click on "Search" with empty boxes, you will get a list of all pages.)

/HigherGenus/C/aut/2.12T5.0.3-4-4
/HigherGenus/C/aut/2.4T2.0.2-2-2-2-2

PR #1043, opened 2016-04-04
created browse and search functionality on index page for Higher Genus curves

jenpaulhus:
1. Made a series of small changes in files /lmfdb/higher_genus_w_automorphisms/_init_.py
 and /lmfdb/higher_genus_w_automorphisms/main.py to better conform to Python style guides
(see notes from pull request #922).

2. On page:  /HigherGenus/C/aut/   the "Browse" now allows browsing by genus (although there is only currently genus 2 data in the database), and the "Search" works now. There is now a test for the search functionality in the test_hgcwa.py file.

3. Uniformized names in "Learn More About" box across files, also made several other small word choice changes to clarify meaning on the pages or make consistent my capitalization usage. See for example:
/lmfdb/HigherGenus/C/aut/Completeness
/lmfdb/HigherGenus/C/aut/
/lmfdb/HigherGenus/C/aut/?genus=2

The next two comments are for any of the individual pages. Here is one example for both:
/lmfdb/HigherGenus/C/aut/2.12T13.0.2-4-6
4. The "Automorphism Group" link now actually connects to the corresponding transitive group and not just the index page for /GaloisGroups.
5. The signature data is now always in LaTeX mode, so I changed the test which was changed for pull request #922 back to "[0;2,4,6]". (This involved creating a function called sign_display to turn a list of integers into the signature notation--that ";" between the first and second number.)

### DIFF
--- a/lmfdb/higher_genus_w_automorphisms/__init__.py
+++ b/lmfdb/higher_genus_w_automorphisms/__init__.py
@@ -3,7 +3,9 @@ from lmfdb.base import app
 from lmfdb.utils import make_logger
 from flask import Blueprint
 
-higher_genus_w_automorphisms_page = Blueprint("higher_genus_w_automorphisms", __name__, template_folder='templates', static_folder="static")
+higher_genus_w_automorphisms_page = Blueprint("higher_genus_w_automorphisms",
+                                       __name__, template_folder='templates',
+                                       static_folder="static")
 logger = make_logger(higher_genus_w_automorphisms_page)
 
 
--- a/lmfdb/higher_genus_w_automorphisms/main.py
+++ b/lmfdb/higher_genus_w_automorphisms/main.py
@@ -39,7 +39,7 @@ def groupid_to_meaningful(groupid):
         return groupid
 
 def tfTOyn(bool):
-    if bool == True:
+    if bool:
         return "Yes"
     else:
         return "No"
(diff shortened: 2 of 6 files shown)


davidfarmer:
Small things to address in a future pull request:
1. On HigherGenus/C/aut/ it says "Maximum number of fields" but you don't mean "fields".
2. All the "Learn more about" links need content (no need to do that until the pages are linked for the public).
3. When I search on the sample label, I don't actually see "6T3" anywhere on the page.  I understand
that the main users of the page expect to see the group described as an abstract group, but there
should be some way to also give the name of the permutation group.  (In fact, "Permutation group"
appears, but it is labeling the abstract group.  And the "Generating vector" refers to the abstract group,
but I thought that was for the permutation group?)
4. Un-necessary parentheses on right side of defining equation.
5. Clicking on "Download this example" just reloads the page.

davidfarmer:
The test test_url_label  fails for me.  If I look at the page source I see:
[ 0; 2, 4, 6 ]
where there are several spaces in that string, but the test has no spaces.

If the test passes for you but not for me, then we need to figure out why.
Can you confirm that
./test.py lmfdb/higher_genus_w_automorphisms/
runs 3 tests and they all pass?

jenpaulhus:
Right, so I clearly don't understand how the testing works. (I hoped blindly copying from local_fields would be good enough!)  I tried running the command 
./test.py lmfdb/higher_genus_w_automorphisms/

at a command line prompt inside the first lmfdb folder and I get an error that the file test.py doesn't exist.  There is a test.sh file, but I'm not sure what to do with that either. Can you please elaborate on where I should be running that command?

davidfarmer:
Sorry, I meant to write

./test.sh lmfdb/higher_genus_w_automorphisms/

If you do

./test.sh

it runs all tests.

If you do

./test.sh "something"

it runs just those tests that fall below "something" in the directory 
tree.

jenpaulhus:
When I run ./test.sh lmfdb/higher_genus_w_automorphisms  (or with other folders as well) I get:

Using Sage command sage
./test.sh: line 45: sage: command not found

I'm not sure what this means. I can run sage directly in a command line.  I've installed "nose" "unittest2" and "coverage" using the pip install command found in the test.sh file and got the message "Successfully installed argparse coverage linecache2 nose six-1.4.1 traceback2 unittest2"

Regardless, I suspect if I ran the test it would fail for me too. When I wrote the tests I didn't take into account that MathJax will remove spaces.  So what I wrote in that particular test is how it appears on the screen, post MathJax processing, not what is in the source.

davidfarmer:
Do "view source" in your browser and cut-and-pase from there into
the test.

I had the same trouble with sage not being found.  I made a work-around
(which actually was the cause of my typing the wrong test command 
earlier).
I will email you my test file.

jenpaulhus:
I've now pushed a additional change which should fix the test error.

davidfarmer:
About to merge.

Isn't the test
assert '2 matches' in L.data
going to fail as soon as more data is uploaded?
If you know which result is likely to show up first, the test could look for that.

jenpaulhus:
That test won't fail because there won't be any more genus 2 data added, and the search is for genus 2 with a particular group.  However, I will now test all my tests before submitting pull requests.

PR #1045, opened 2016-04-06
Cleaned up downloading code

AndrewVSutherland:


JohnCremona:
Drew and I worked on and tested this together.  I plan to make similar simplications to the download for elliptic curves over Q and over number fields.

jwj61:
Does it let you download lots of data from beta?

Code which worked last week on local versions for downloading say all elliptic curves over number fields fails now on beta because it takes a long time.  On a local version, you can wait.  On the warwick machines, it times out.

I have not looked to see if you found a solution to this yet, but I think a different approach may be needed, either

1. check the count and only offer download from search results if the count is relatively small.  Otherwise tell the use to see the api page
2. always direct the user to the api page.
3. see if the timeout can be avoided by saving the results to a file on the disk and give a link to it.  There are still timing issues: you initially have to say the file is being prepared and javascript then switches when the file is ready.  Also, after certain web calls finish, you trigger a function which checks that last access time of download files, and deletes ones which are too old (say 30 minutes).

In options 1 and 2, we currently have a limit on the total you can download via the api, although that can be changed.  Also, the api provides certain general formats, whereas the search results downloads provide "custom" formats.  Option 3 is what I use for downloading with the AZ number field database.  It has worked well there, but I don't have a timeout to work around.

JohnCremona:
Drew and I tested downloading all genus 2 curves, with no trouble (66,000
curves).  But trying all the EC/NF -- about 330000 curves -- times out.
Smaller numbers (e.g. only for field 2.2.5.1) is OK.

The code is written so that the entire output is a single string.  There is
probably a better way, for a large download.  It might be worth changing
this (after defining the huge string s):

 strIO = StringIO.StringIO()
    strIO.write(s)
    strIO.seek(0)

with individual write() calls of shorter chunks.  I don't have time to try
this now.

PR #1050, opened 2016-04-07
switch Conenction to MongoClient, and new defaults for the DB connection

edgarcosta:
I did a couple of things:

1- move some of the global variables to the right spot, or to the beginning of the file
2- got rid of the:
```
if pymongo.version_tuple[0] < 3:
            from pymongo import Connection		
            _C = Connection(port=dbport)
```
3- added three defaults, that will be useful in the feature:
```
DEFAULT_DB_HOST = 'localhost'
DEFAULT_DB_REPLICA_SET = None
DEFAULT_DB_READ_PREFERENCE = ReadPreference.NEAREST 
```

4- made those defaults, defaults arguments for base.makeDBConnection and base._init

### DIFF
--- a/lmfdb/base.py
+++ b/lmfdb/base.py
@@ -20,6 +20,19 @@ from werkzeug.contrib.cache import SimpleCache
 # logfocus
 logfocus = None
 
+# global db connection instance (will be set by the first call to
+# getDBConnection() and should always be obtained from that)
+_C = None
+
+
+DEFAULT_DB_PORT = 37010
+DEFAULT_DB_HOST = 'localhost'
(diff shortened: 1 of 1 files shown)


JohnCremona:
The reason for this sort of mess is some people's unwillingness or inability to upgrade, so there are bits of code intending to work with more than one version (e.g. of pymongo).  Personally I am happy with insiting on later versions of Sage, pymongo, and anything else.  But don't break things for hard-pressed developers in the next month if at all possible.

edgarcosta:
Okidoki, I will split the pull request into 2 or 3 different ones.

Issue #396, opened 2015-10-14
Sign of HMF L-function


davidfarmer:
There is some problem computing the Atkin-Lehner eigenvalue,
resulting in a "sign" which does not have absolute value 1.

L/ModularForm/GL2/TotallyReal/2.2.104.1/holomorphic/2.2.104.1-5.2-c/0/0/

jvoight:
This is actually a fairly serious issue, and I am resolving it with the help of Naomi Tanabe.  The problem is that, as it turns out, some of the forms we have in the database do *not* have trivial central character when the narrow class number is even: sometimes, a character of the Hilbert class group (trivial on the class group itself) intervenes, coming about from how the quaternionic way these forms were computed.  I am investigating.  

At this point, I'm not sure if it makes sense to delete these forms from the database or to add functionality for nontrivial central character--I'm leaning toward the former.

JohnCremona:
Might that include some dimension 1 newforms, for which I have not been
able to find an elliptic curve?

jvoight:
Yes, it certainly would!!  Can I look at an example?  

You can read off the central character from the Brandt matrices, but I don't immediately see how to do it from the eigenvalues themselves.  Which means this could be some work.

jvoight:
Actually, I take it back: when it applies, the Eichler-Shimura construction says that if you have a dimension 1 newform so rational Hecke eigenvalues, then you associate an elliptic curve.  This comes from taking a quotient of the Jacobian of a Shimura curve, and the central character does not figure into this.  Although the Eichler-Shimura construction fails to apply to HMFs over even degree fields with squarefull level, the answer should be uniform and so won't depend on this.

jvoight:
Update: the space of Hilbert modular forms computed using quaternionic methods is in fact a direct sum of spaces according to characters of the narrow class group.  One can compute this decomposition (as mentioned above), but we did not do this when we computed forms--and it is not clear how to do this from the eigenvalues themselves.  

So, in short, over a totally real field with narrow class number >1, the L-function of any Hilbert modular form that appears may be wrong.  This includes a large number of fields, but it looks like this affects fewer forms than it the worst case may make it seem.

To fix this, we'll need to recompute the space and then the character.  I am working now to do this, but it may take some time.

[...]

PR #1057, opened 2016-04-11
Hide HMF L-function when class number is greater than 1

jvoight:
Plus something I still don't understand about merging branches that should be harmless.


jvoight:
Addresses #396.  (Should I have made a patch there?)

Issue #983, opened 2016-03-31
Add plots to all Maass form pages


AurelPage:
From /roadmap: Add plots of the Maass form in the fundamental domain.

fredstro:
Some of the Maass forms already have plots. 
Somehow someone formatted the image wrongly with css and this made all plots except level 1 to display badly (the interesting parts landed outside the page).
I will submit a pull request where I fixe these display problems shortly.

JohnCremona:
The pull request which fixes the issue has been merged.  This could be closed, or left open until all the relevant plots have been computed and added: currently the graph is shown when already computed, otherwise not.

davidfarmer:
This work is underway, and this old issue is not helping, so am closing it.

PR #1058, opened 2016-04-11
changed formatting for the maass form plots

fredstro:
I fixed the formatting of plots of Maass forms so that all levels work. 
This fixes issue #983 . 
I will also restart the scripts to make the plots so that more forms will have them (currently we only have 619 plots)


JohnCremona:
Can you give a URL to look at and an indication of what to look for?  I tried a random level 2 form but saw no plot (ModularForm/GL2/Q/Maass/4f5556ef88aece241f000005)

fredstro:
I still have to run (and update) the script to generate plots for all Maass forms in the database.
You can see a plot for a level 2 form at:
/ModularForm/GL2/Q/Maass/54104e98acf756699fe0831f

JohnCremona:
OK thanks -- that one looks fine to me.  I assume that the plots are shown when they exist, otherwise not, so this looks good.

PR #1065, opened 2016-04-15
Cites in knowls

davidfarmer:
Convert 4 different possibilities of \cite{...} into hyperlinks:
MR
arXiv
href
doi

For an example of all four possibilities:
/knowledge/show/dq.ec.source


jwj61:
I suppose I should check the various types of cites.  Do you know of places where they occur?  If not, maybe make a knowl on doing citations which demos them.

davidfarmer:
Sorry, I should have explicitly said that every possibility is in the knowl mentioned in the issue this fixes:
http://beta.lmfdb.org/knowledge/show/dq.ec.source

PR #1070, opened 2016-04-18
now the .dat files sit on data/zeros/zeta/data/

edgarcosta:
Hello,

For the cloud solution, it is convenient for us to have all the zeros*.dat files to sit on one folder.
If you accept this pull request, then the LMFDB will look for the files zeros*.dat files in  data/zeros/zeta/data/ instead of  data/zeros/zeta/.

I tried this out in our cloud version, and everything seem works flawlessly, try it out: https://lmfdb.xyz:8080/zeros/zeta/

The change will also be seamless for Warwick, as I created as I created a symlink at warwick to take care of this:
$ ls -l /home/lmfdb/data/zeros/zeta/data
lrwxrwxrwx 1 lmfdb lmfdb 27 Apr 17 04:16 /home/lmfdb/data/zeros/zeta/data -> /home/lmfdb/data/zeros/zeta

### DIFF
--- a/lmfdb/zeros/zeta/platt_zeros.py
+++ b/lmfdb/zeros/zeta/platt_zeros.py
@@ -14,8 +14,9 @@ from math import log, floor
 import mpmath
 mpmath.mp.prec = 300
 
-data_location = os.path.expanduser('~/data/zeros/zeta/')
-
+zeta_folder =  os.path.expanduser('~/data/zeros/zeta/')
+data_location = os.path.join(zeta_folder, 'data/')
+db_location = os.path.join(zeta_folder,'index.db')
 
 def list_zeros(filename,
                offset,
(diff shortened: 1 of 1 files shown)


JohnCremona:
The link you set up means that those of us who test on machines which have the lmfdb data partition mounted find that it does not work.  I agree that this is a minority but still:  I develop on atkin.warwick.ac.uk which has  lmfdb's  /data partition mounted as /mnt/data.  When I run the website locally (on atkin)  it can see ~/data/zeros/zeta, but the link ~/data/zeros/zeta/data points to /home/lmfdb/data/zeros/data which does not exist on atkin.  Hmmm.   I don't suppose you would want to replace that absolute link to a relative one?  Otherwise I would have to create a user on atkin classed lmfdb;  which would of course be easy.

edgarcosta:
John, I have just fixed the link to a relative link.
However, if you want, I can also just copy the `*.dat` files to `data/zeros/zeta/data/`, and once this commit trickles down to to prod and beta we get ride of the `*.dat` files in `data/zeros/zeta`

JohnCremona:
That works fine.  Let's not move them around.  Thanks.

PR #1073, opened 2016-04-19
Random fix

edgarcosta:
1- I wrote a generic "get random object" with [$sample (aggregation)](https://docs.mongodb.org/manual/reference/operator/aggregation/sample/).

2- I used the new generic "get random object to replace:
```
    n = get_stats().counts()['collection name']		
    n = randint(0,n-1)		
    random_oject = getDBConnection().collection.find()[n]
```
as we should avoid at all costs anything similar to `getDBConnection().collection.find()`.

### DIFF
--- a/lmfdb/ecnf/main.py
+++ b/lmfdb/ecnf/main.py
@@ -11,7 +11,7 @@ ASC = pymongo.ASCENDING
 from urllib import quote, unquote
 from lmfdb.base import app, getDBConnection
 from flask import render_template, render_template_string, request, abort, Blueprint, url_for, make_response, redirect, flash, send_file
-from lmfdb.utils import image_src, web_latex, to_dict, coeff_to_poly, pol_to_html, make_logger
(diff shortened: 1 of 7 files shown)


JohnCremona:
Thank you for working out how to do that; I asked around but could not find or did not recognise the aggregate function.  Once I had done it once the old way for elliptic curves, others copied...

JohnCremona:
My testing was with pymongo 2.8.

edgarcosta:
> My testing was with pymongo 2.8.

Same here, but things change a bit for pymongo 3.0, I tried to take that into account

JohnCremona:
I just tested after upgrading pymongo (to 3.2.2) I get errors from the random_object_from collection function.  I will try to fix that.  It works OK with 2.8 which is the version forced by our requirements.txt but I was wondering why we put that in (I cannot remember).
Done.  Pull request coming up.

PR #1074, opened 2016-04-20
Mongo client configfile

edgarcosta:
Hello,

Now you can read, from `mongoclient.config`, how do you want the LMFDB to connect to the MongoDB.
Explicitly, you can pass: `port`, `host`, `replicaSet` and `read_preference` to `pymongo.MongoClient`.
The config file is `mongoclient.config`, and here is an example of one equivalent to the default settings:
```
[db]
port = 37010
host = localhost
replicaset =
read_preference = NEAREST
```

We need something like this to have the client to connect to something else than localhost:37010, and we will also need the other arguments in the future.

All this could have been done with command line arguments, however, those are hard to pass through gunicorn, and the solution becomes quite ugly.

Regular users, and Warwick won't notice a difference until they create `mongoclient.config` and write something nontrivial on it.

Also, at the moment the config file overruns the command line arguments, the only one is affected is --dbport.
If you want it the other way around, we can simply swap the order that they are processed.

I have also tweaked the position of some global variables in `lmfb/base.py`

### DIFF
--- a/lmfdb/base.py
+++ b/lmfdb/base.py
@@ -20,7 +20,6 @@ from werkzeug.contrib.cache import SimpleCache
 # logfocus
 logfocus = None
 
-
 def set_logfocus(lf):
     global logfocus
     logfocus = lf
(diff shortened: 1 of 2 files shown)


JohnCremona:
This looks like a very good idea to me.  I was the one who first wanted the ability to change the default port since we had several people developing on the same machine in Warwick, and the code I hacked to do that was the mess which this replaces.

I would like to suggest that in the same pull request we update the GettingStarted guide, so that they are in step.

JohnCremona:
Forget what I said about the GettingStarted guide, since nothing there needs changing.
With this patch I tried the following and all worked: (1) no mongoclient.config file; (2) the default file (as above); (3) defulat file except that I replace localhost with lmfdb (which I know is aliased in my .ssh/config to lmfdb.warwick.ac.uk).  All work fine.  Thanks.

JohnCremona:
Edgar: I thought of adding a (default) mongoclient.config file to the repository, then saw it would be a bad idea since the only people who need it are those who do not want the default, and there would always be a danger that a non-default version was committed.  Instead, shall we add a file called default_mongoclient.config, containing what you put about together with a comment "If you need to run the server with non-default settings, copy this file to mongoclient.config and edit that."  And add mongoclient.config to the .gitignore file.  What do you think?

edgarcosta:
@JohnCremona  That sounds great to me!

Here is a possible default config file with some text:
```
# This is the default mongoclient.config
# and below are the default values passed to MongoClient
#
# To change the defaults, edit the file below, and rename it to mongoclient.config
#
# settings without values are parsed as None
# the settings from port and read_preference are processed
# all the other values are parsed and passed to MongoClient as strings
[db]
port = 37010
host = localhost
replicaset =
# the options for read_preference are PRIMARY, PRIMARY_PREFERRED, SECONDARY, SECONDARY_PREFERRED, and NEAREST
# see https://docs.mongodb.org/manual/reference/read-preference/ for more details
# read_preference only takes effect if replicaSet is set
read_preference = NEAREST
```

PR #1077, opened 2016-04-20
fixed random element function for pymongo>=3

JohnCremona:
#1073 worked fine with pymongo 2.8 but not for version 3 and above.  After this fix it works on both (I tested).
I could not remember why we forced pymongo 2.8 in requirements.txt, but git blame says that it was Harald on 1 Nov 2015 and the reason was that the then current version was 3.0.3 which was broken.  Well, the current version now is 3.2.2 and it works fine, so I'll also make that change.


haraldschilly:
for the record: IIRC, I did that because those with Macs where having issues … linux was always fine. so yes, it's definitely time to upgrade or it's some other issue.

edgarcosta:
Thank you. I mentioned that pipeline need to be a list but forgot to take care of it .

PR #1078, opened 2016-04-20
add default mongoclient config file

JohnCremona:
See discussion at #1074

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,6 @@
 password*
 config.ini
+mongoclient.config
 
 *.pyc
 *.pyo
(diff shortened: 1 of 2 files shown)


edgarcosta:
👍

fredstro:
Does anyone actually manage to use the config file?
As far as I can tell the check: os.path.exists(mongo_client_config_filename) 
checks for an absolute path so we need to provide a directory as well. 
Basically I guess we want to have one level above the lmfdb module directory.

edgarcosta:
Yes, I use it all the time on my local machine, and all the webservers are
also using it.

Best,
Edgar

fredstro:
Yes, but you are probably using it with the script from the local directory. 
It doesn’t work if you either import lmfdb as a module in sage  or run the script from another location. 
See my pullrequest #1290 to add an absolute path.

Best,
Fredrik

Issue #1079, opened 2016-04-21
Need to have and implement general download policy


jwj61:
Different areas let you try to download search results.  When there are lots of results, this will fail with a proxy error which really comes from a timeout on the server.  This can easily happen with number fields (with over 5 million fields in the database), and will happen in other areas as they expand.

Options I see:
1. remove the download buttons when there are too many search results.  Possibly provide a link to the api page instead
2. put up a message saying results are being compiled, start a process on the server to make the file which is not subject to the timeout, and then provide the link when the file is ready.  Files would have to be cleaned up afterward.

A drawback of (1) is that the download links make a customized version of the results whereas the api doesn't, and the result is not as friendly to systems (sage, gp, magma) where the data may be destined.  Also, the api currently has a total limit on number of entries to be retrieved, although that can be changed.  On the other hand, I am not sure how to make the side process which is not governed by the server timeout.

JohnCremona:
I don't really understand why the server should time out, when it can do the query.  Is it caused by the fact that the query produces an iterator, and the iterator itself times out if left open for too long?  I ran in to this a while back when I was looking through elliptic curves and doing some very nontrivial computations with each one.

If that is the case, here is a solution:  (1) make sure that the query which produces the downlod data only delivers the data fields which are needed;  (2) convert the iterator into a list as soon as it is created; (3) create the download file from that list instead of directly from the iterator.

For a quick test, try changing the donwload_search function call to give it list(res) instead of res and see if it still times out.

haraldschilly:
as you already said, a proper way to really address this download question is, to use a background task manager. the most natural fit for the lmfdb setup is http://www.celeryproject.org/ 

* as a task broker, use https://www.rabbitmq.com/
* the generated data is stored in mongodb, (called "results backend")

regarding the front-end, you are then periodically polling an endpoint giving you progress information or other status updates (or none), and at some point when the async task has finished, present the link to download the result. 

drawback of all this is, that it roughly doubles the number of moving parts one has to take care of on the server. also, for local development, you need to implement some simple fallbacks, when such a task manager isn't available.

jwj61:
Thanks Harald.  It sounds like it is worth looking into.

@JohnCremona : This is not the same as a cursor going bad.  If run locally, the current code works.  The warwick servers for beta and prod have timeouts for web requests, and that is what is killing the download.

[...]

PR #1084, opened 2016-04-25
made number field search result download more efficient

JohnCremona:
See Issue #1079.   It is now possible to download 1.36 million number fields in a little over 2 minutes, no timeout.  The key is to avoid constructing the WebNumberField object for each.

### DIFF
--- a/lmfdb/number_fields/number_field.py
+++ b/lmfdb/number_fields/number_field.py
@@ -598,10 +598,17 @@ def download_search(info, res):
     else:
         s += 'data = ['
     s += '\\\n'
+    Qx = PolynomialRing(QQ,'x')
+    str2pol = lambda s: Qx([QQ(str(c)) for c in s.split(',')])
     for f in res:
-        wnf = WebNumberField.from_data(f)
-        entry = ', '.join(
-            [str(wnf.poly()), str(wnf.disc()), str(wnf.galois_t()), str(wnf.class_group_invariants_raw())])
+#        wnf = WebNumberField.from_data(f)
+#        entry = ', '.join(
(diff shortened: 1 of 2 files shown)


JohnCremona:
If you like this we can also think about doing what Drew and I did for the equivalent genus 2 curves function: more slick switching between language options, and also defined a function make_data() in each language so that (for example in Magma) after the user reads in the file they type curves:=make_data();
and then curves is a Magma list of genus 2 curves.  This saves the user from having to work out how to create the objects from what we give them and would often be useful.

jwj61:
This produces the downloads faster.  It seems stylistically wrong to have bits that need to know how items are stored in the database outside webnumberfield, although I can see how to work around that.  This version does not handle the case where the class number/group are not computed.  Again, that is not hard to fix.

But, I don't think this solves the timeout issue.  I believe beta and prod kill web requests which take more than 30 seconds, and 2 minutes is much more than that.  While it works when running your own copy, I think the same download will be killed and end with a proxy error if pushed to beta/prod.

In the next day or two I will put in a pull request doing what I allude to in the first paragraph, and remove the download links when the number of results is more than 100000.

JohnCremona:
You may be right.  The main thing I was looking for was whether WebNF created a Sage NF.  If there are slow parts of the WebNF to create they can perhaps only be done when actually needed so that here we could have a "WebNF lite".

jwj61:
I don't think webnumberfield was should not be doing anything aside from creating a copy of the data (because it was using the from_data construct).  I see now creating code snippets was added to the init routine, which means a hard disk access to initialize.  That is a waste of time here and should be done only when actually needed.

JohnCremona:
This was superseded by #1091 so I'm closing it.

PR #1085, opened 2016-04-25
changed permutation and transitive group notation in HigherGenus/C/aut/

jenpaulhus:
When I extended the group action data to higher genus, I ran into an issue with the transitive group labeling.  Many entries were of transitive groups of degree greater than 32, hence I could not write them in the “dTn” notation.  

I’ve reverted to using the GAP or Magma SmallGroup(s,n) notation, where s is the order of the group, and n is which entry of that order in the database the group is.  I have changed the generating vector from permutations to “words” on the generators of the group also.

This forced a small change in the labeling. Instead of, say  2.6T2.0… I now put 2.6-1.0… (since the transitive group 6T2 corresponds in the data to the group (6,1) in GAP/Magma.) 

Here are corrections I needed to make to the pages (for example: HigherGenus/C/aut/3.48-48.0.2-4-6)
- Change “Permutation Group” to “GAP/Magma notation”, along with a common name of the group listed as “Isomorphism Class” if this particular group is listed in the group dictionary given in “genus2_curves/data.py”
- Removed “Degree of the covering”
- Removed “Automorphism Group” friend (Related Object).
- “Generating Vector” and “Hyperelliptic Involution” not listed as permutations 
- Change “Properties” to reflect group notation

I also needed to modify the search (HigherGenus/C/aut/)
- Search the group in new notation (as a list of 2 integers).
- Updated the order which the search functionality displays entries (first by genus, second by dimension of the family)

A few other odds and ends
- Updated test_hgcwa.py to account for new labeling
- Removed a few “import” commands in main.py that are no longer being used
- Fixed /template/hgcwa-search.py so that the “Next” button properly works.
- Changed the word “fields” to “families” in HigherGenus/C/aut/ as commented on in  #1043
- Added a couple of knowl links to signature and dimension of the family.  Still need to write all of these, especially to explain notation.
- All equations should no longer have unnecessarily parentheses as commented on in #1043

### DIFF
--- a/lmfdb/higher_genus_w_automorphisms/main.py
+++ b/lmfdb/higher_genus_w_automorphisms/main.py
@@ -10,15 +10,13 @@ from lmfdb import base
 from lmfdb.base import app, getDBConnection
 from flask import render_template, render_template_string, request, abort, Blueprint, url_for, make_response
 from lmfdb.utils import ajax_more, image_src, web_latex, to_dict, coeff_to_poly, pol_to_html, make_logger
-from lmfdb.search_parsing import parse_galgrp, parse_ints, parse_count, parse_start, parse_list, clean_input
+from lmfdb.search_parsing import parse_ints, parse_count, parse_start, parse_list, clean_input
 
-from sage.all import ZZ, var, PolynomialRing, QQ, Permutation
+from sage.all import Permutation
 
 from lmfdb.higher_genus_w_automorphisms import higher_genus_w_automorphisms_page, logger
 
 from lmfdb.genus2_curves.data import group_dict
-
-from lmfdb.transitive_group import *
 
 
 HGCwA_credit = 'J. Paulhus'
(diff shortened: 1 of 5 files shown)


JohnCremona:
When I put 2 in the genus box and press Search I get an error "UndefinedError: 'dict object' has no attribute 'trangplabel'".  SImilarly with any of the "browse by genus" links.
Also the example near the "search by label" box produces no hits.

Since there is (I think?) no way of getting to those HigherGenus/C/aut/ pages without knowing that they are there, there would be no harm in merging this now, but also no rush to do so at least until these are fixed.  After the May 10 release it would be fine to have this in beta (think about where in the sidebar this would go -- under Genus 2?)

jenpaulhus:
Clarification question for you, are you working from my branch and getting those errors, or are you getting them on beta right now?  When I go to my local copy of remove_permutations, the "Browse by genus" links work just fine, as does searching by the given label.

If you are testing the current beta pages, those should fail. Perhaps I did this in the wrong order, but I changed the data in the database first, and am now submitting pages that should work with the new data. The "trangplabel" you mention above was in the old data, but not the new data and my new pages don't reference this at all. Obviously, if something else is the issue, I'm happy to fix it.

jenpaulhus:
Oh, and I think the idea back in March was that this would eventually be under the "Higher Genus" link (currently greyed out in beta) under the "Curves" section of "Varieties", perhaps with the name "with Automorphisms".

JohnCremona:
I'll check again tomorrow in case I mixed branches.  I  certainly intended
to test yours.

JohnCremona:
We can add a sidebar entry whenever you like, to appear on beta only for the time being.

jenpaulhus:
Thanks for double checking.  And I think we should wait for the moment on the sidebar entry until a few more pieces are in place (like the Download feature has not been implemented yet).

PR #1088, opened 2016-04-26
Lattice

sanni85:
In this pull request I am introducing more functionalities and changing the code.

If genus data is missing it is not displayed on the page (this will be the case for the Leech lattice).

We are also dealing with multiple name assignation for a lattice.

Download functionalities are added: download for search results and downloads for list of normalized shortest vector and genus representatives (in these last two cases, the list displayed on the page is shorten down to be more "human", again for the Leech lattice the list of normalized shortest vector contains 98280 vectors with 24 entries each).

The import script has been changed to check for isometries, in order to be sure not to have duplicates.

In the next days (while I will fill in applications....), I will upload the data that Anna computed (~30.000 lattices) and add their theta series (if not available they are already not shown on the pages). 

Tests have been added for download functionalities, all tests pass.

### DIFF
--- a/lmfdb/import_scripts/lattices/lattice.py
+++ b/lmfdb/import_scripts/lattices/lattice.py
@@ -14,7 +14,7 @@ import re
 import json
 import sage.all
 from sage.all import os
-
+from sage.all import Integer, ZZ, QQ, PolynomialRing, NumberField, CyclotomicField, latex, AbelianGroup, polygen, euler_phi, latex, matrix, srange, PowerSeriesRing, sqrt, QuadraticForm
 from lmfdb.lattice.isom import isom
 
 from pymongo.mongo_client import MongoClient
@@ -26,7 +26,7 @@ password = pw_dict['data']['password']
 C['Lattices'].authenticate('editor', password)
 lat = C.Lattices.lat
 
-saving = True 
+saving = False 
 
 def sd(f):
   for k in f.keys():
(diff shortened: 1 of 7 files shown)


sanni85:
Sorry, I forgot a list of useful pages:

http://localhost:37777/Lattice/14.15.15.9.1
http://localhost:37777/Lattice/A8
http://localhost:37777/Lattice/A3
http://localhost:37777/Lattice/D3
http://localhost:37777/Lattice/?start=0&dim=&det=&level=&gram=&minimum=&class_number=1&aut=&count=50    (try the download option)

JohnCremona:
I was looking at this with Samuele today and will review it properly tomorrow morning.

JohnCremona:
Sage download file: Need to insert # at start of first line, and delete */ at end.  On penultimate line delete comma near end (Sage handles this but the others do not).
gp: same with comment char //.
magma: same with comment //

I will fix these.

sanni85:
I am fixing one more thing.

sanni85:
Just made a last commit : in the case of unknown class number, in the property box it says not available instead of 0.

JohnCremona:
In the main page the examples under the search by label or name box do not work (I tried Leech, E8).
Lattice/E8 works in the URL but Lattice/Leech does not.

Can yo ugive an example where the class number is not available?

sanni85:
E8 and Leech will be uploaded today/tomorrow

I just added a "wrong" lattice where you can see the change:
http://localhost:37777/Lattice/ko

I will remove it before the upload of the big data.

sanni85:
the lattice ko have been removed from the database

PR #1090, opened 2016-04-27
enhanced stats and browsing for Hilbert MFs

JohnCremona:
In addition, the knowl http://www.lmfdb.org/knowledge/show/dq.mf.hilbert.extent display actual content with this code.
I modelled this in the ecnf browsing pages.  You can see differences in the top page http://localhost:37777/ModularForm/GL2/TotallyReal/
whcih shows how many fields of each degree there are, and you can browse by degree to see which fields and what level norm bound we have for eac, for example http://localhost:37777/ModularForm/GL2/TotallyReal/browse/2/
You can also get an overviewof all fields at http://localhost:37777/ModularForm/GL2/TotallyReal/browse/
I added some indexes to make the counting faster but it still takes several seconds the first time the counting is done.

### DIFF
--- a/lmfdb/hilbert_modular_forms/hilbert_modular_form.py
+++ b/lmfdb/hilbert_modular_forms/hilbert_modular_form.py
@@ -445,7 +445,7 @@ def completeness_page():
     t = 'Completeness of the Hilbert Modular Forms data'
     bread = [('Hilbert Modular Forms', url_for(".hilbert_modular_form_render_webpage")),
              ('Completeness', '')]
-    return render_template("single.html", kid='dq.mf.hilbert.extent',
+    return render_template("single.html", kid='dq.mf.hilbert.completeness',
                            credit=hmf_credit, title=t, bread=bread, learnmore=learnmore_list_remove('Completeness'))
 
 @hmf_page.route("/Source")
--- a/lmfdb/hilbert_modular_forms/hilbert_modular_form.py
+++ b/lmfdb/hilbert_modular_forms/hilbert_modular_form.py
@@ -69,7 +69,6 @@ def hilbert_modular_form_render_webpage():
                  ('Hilbert Modular Forms', url_for(".hilbert_modular_form_render_webpage"))]
         info['learnmore'] = []
         info['counts'] = get_stats().counts()
-        info['stats'] = get_stats().dstats()
         return render_template("hilbert_modular_form_all.html", info=info, credit=hmf_credit, title=t, bread=bread, learnmore=learnmore_list())
     else:
         return hilbert_modular_form_search(**args)
@@ -469,7 +468,7 @@ def labels_page():
 def browse():
     info = {
         'counts': get_stats().counts(),
-        'dstats': get_stats().dstats(),
+        'stats': get_stats().stats(),
         'extent': dict([(d,hmf_degree_summary(d)) for d in get_stats().counts()['degrees']]),
         'full_extent': hmf_summary()
     }
@@ -493,7 +492,7 @@ def statistics_by_degree(d):
             info['error'] = "The database does not contain any Hilbert modular forms over fields of degree %s" % d
         d = 'bad'
     else:
-        info['dstats'] = stats.dstats()[d]
+        info['stats'] = stats.stats()[d]
         info['degree_stats'] = hmf_degree_summary(d)
     credit = 'John Cremona'
     if d==2:
--- a/lmfdb/hilbert_modular_forms/hilbert_modular_form.py
+++ b/lmfdb/hilbert_modular_forms/hilbert_modular_form.py
@@ -9,7 +9,7 @@ from flask import Flask, session, g, render_template, url_for, request, redirect
 from sage.misc.preparser import preparse
 from lmfdb.hilbert_modular_forms import hmf_page, hmf_logger
 from lmfdb.hilbert_modular_forms.hilbert_field import findvar
-from lmfdb.hilbert_modular_forms.hmf_stats import get_stats
+from lmfdb.hilbert_modular_forms.hmf_stats import get_stats, hmf_summary, hmf_degree_summary
 
 from lmfdb.ecnf.main import split_class_label
 from lmfdb.ecnf.WebEllipticCurve import db_ecnf
@@ -69,6 +69,7 @@ def hilbert_modular_form_render_webpage():
                  ('Hilbert Modular Forms', url_for(".hilbert_modular_form_render_webpage"))]
         info['learnmore'] = []
         info['counts'] = get_stats().counts()
+        info['stats'] = get_stats().dstats()
         return render_template("hilbert_modular_form_all.html", info=info, credit=hmf_credit, title=t, bread=bread, learnmore=learnmore_list())
     else:
         return hilbert_modular_form_search(**args)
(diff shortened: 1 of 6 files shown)


JohnCremona:
I am upgrading this so that the time-consuming counting only happens when needed -- as it is even displaying the main HMF page takes ages.  Revised branch up later...

JohnCremona:
I separated out the simpler counts so that the main page displays quickly.  But the stats still take about a minute, despite my best attempts -- you see that if you browse by degree or "all available fields".
I also sort the fields of fixed degree by discriminant which looks better.
Back open for review...

jwj61:
http://localhost:37777/ModularForm/GL2/TotallyReal/browse/

sat for a long time, and eventually gave me an error.

Other comments: the completeness knowl does not assert any completeness results (like the database contains all forms of norm conductor <= X for fields ...).  It is really a repeat of the extent (so many forms over this many fields).

There is a link to a knowl for Source shown, but it seems to be empty.

JohnCremona:
Thanks for the detailed comments -- I really wanted someone to look at this closely.

I spent half a day trying to get the /browse to be faster.  For me it normally takes 42s -- too long but not failing.  (But trying just now is failing)  I could disable the home page link which trigger that, or store today's counts in the db itself.  But it is a mystery to me sine the similar ecnf counts on about the same number of objects only takes 15s.

The source and completeness knowls are going to be done by John Voight.  I could have put in a placeholder for the first saying "Computed by John Voight using Magma" but I did not know that he would take so long.

The extent knowl is OK.  The completeness link is *wrongly* displaying the extent knowl, and that can be fixed -- but there is no completeness knowl, so JV needs to write that too.

If I cannot get the browse counts to work faster tomorrow I'll remove those links or something, since otherwise the extra information on the home page for HMFs is certainly better.

jwj61:
In many cases, extent and completeness are the same.  I put the short statement in the source knowl.  If we are freezing things tomorrow, it is best to have something there.  Copying the database to the cloud version seems like it would be a long process, and knowls are in the database, so it seems worthwhile to have something there now.

On the browse page, even if the error is fixed, remember that 42s on the warwick server is timed out after 30s (which of course could be changed).  Is the cloud version going to have a similar timeout?

jvoight:
I'm sorry, I did not have time during the week to do this.  I'm sorry to keep you waiting, I planned on doing it this weekend.

JohnCremona:
OK so John V: could you edit http://www.lmfdb.org/knowledge/show/dq.mf.hilbert.source to what you would like it to say, and also create and write a dq.mf.hilbert.completeness knowl?
I will remove the links to the slow browse pages and fix the links to the completeness statement which currently wrongly shows the extent knowl.

It may be that Edgar is planning to use tonight's database dump as the base for the cloud version; if so then any knowl editing needs to be done in the next 4 hours.  Of course, further dumps are possible.

jvoight:
OK, I wrote the knowl dq.mf.hilbert.source.

The knowl dq.mf.hilbert.extent used to be

{% if hmf_summary is defined %}
  {{ hmf_summary() }}
{% else %}
  This information can only be displayed with a more recent version of the LMFDB.
{% endif %}

I changed it instead to say "The database contains 368,356 Hilbert newforms over 400 totally real number fields of degree up to 6."  This is what you wanted me to do @JohnCremona , right?

There are no other "completeness" knowls, only "extent" knowls.  There was one knowl dq.g2c.extent which was named "Completeness...", so for consistency I renamed it "Extent...".

JohnCremona:
Well no,  the reason you did not see the real content of the extent knowl (which actually looks much the same) is that it is a dynamic knowl which is only visible after the PR is merged.
I seem to be confused between Extent and Completeness.  See JJ's comments above.  We can easily talk about the Extent (either using the dynamic version I had, or your replacement static one), but users might ask more questions about the data.  Perhaps all we can say is that for all the fields for which we have any data, the data is complete up to a certain level norm.  That would have been a more sensible thing to say if we had the max level norms for each field, but (as noted above) this was taking too long to obtain dynamically as there are so many fields.  And then there is still the question of why certain fields have no data (I know the reason since you explained it to me but the user will not).

The latest version will satisfy for now, though perhaps the title of that knowl should have the word Completeness replaced by Extent.  And after merging and pushing this PR you could reinstate the original dynamic content.

jvoight:
OK, I added a sentence to dq.mf.hilbert.extent that says the data is complete up to a norm level depending on the field.  

BTW, this maximum norm level for each field I computed to be:
[levels.txt](https://github.com/LMFDB/lmfdb/files/243810/levels.txt)

@jwj61, are we ready to merge this and then decide what to statically display?

JohnCremona:
When this is merged (and pushed to beta & prod) you'll see the label change form the other ticket, and also the URL ModularForm/GL2/TotallyReal/browse has all the level ranges for all fields -- but it takes too long to display so will timeout which is why I took out the link to that from the HMF home page.  the functionality is still there.

JohnCremona:
I am going to merg & push this myself now.  Then JV can see how it looks.  We can change more tomorrow if you want.

jwj61:
A few additional comments.

For most areas Extent=Completeness, since for example, we have all elliptic curves over Q of conductor <= B for some B, and nothing more.  If these two things are different, then there should be two different knowls.

Ultimately, the information in levels.txt should be in the completeness knowl.  Maybe I am missing something, but I don't see how a user would know what is in levels.txt.

Currently the home page for HMF has the completeness information and a link to it.  That learnmore link should not be there.

The special Browse page times out on beta (as expected).  Is this linked to from somewhere?  If not, that's ok.  If there is a link, it should be removed.

JohnCremona:
JJ: (1) don't worry there is no link to the slow /browse page.  There will not be unless  / until it is fast (I have some new ideas while I am about to try out). (2) I'll remove the link from the HMF home page to the "completreness" which just shows info already on that page. (3) JV's levels.txt was, I think, just to show me that he knew what those numbers are -- these are the same numbers as computed (with more) in my slow code.
In future we might need more "completeness" info but here it can still be summarised, as we now do, by saying that for any field for which there is any data at all, the data is complete up to a certain level norm.  This logically includes fields which one might expect to see (smaller discriminants than the ones we have) but for which that bound is 0 as we have nothing.   We may upload more data later and then the dynamic knowls will show what we have.

JohnCremona:
I am working on a successor to this and almost have it running perfectly.  Separating the counting function by degree (of the base field), degrees 2, 3, 5, 6 are almost instantaneous while degree 4 takes a long time.  Clearly degree 4 was causing the problem before.  The numbers of newforms for degrees 2-6 are 188459, 45880, 102780, 16226, 15011 so degree 4 has fewer than degree 2.  So I do not know why degree 4 takes so much longer.
I will make a new pull request just so people can look at the result, which certainly works nicely for degrees 2, 3, 5, 6!

jvoight:
@jwj61: Is "extent" supposed to be the pithy one liner in a knowl, "The database contains 368,356 Hilbert newforms over 400 totally real number fields of degree up to 6" and the completeness something displayed on a separate page?  That seems to be what is done for many pages, but not all: for genus 2, there is a short line at the top of the browse page which is not a knowl, for example.

There is currently no knowl that ends in ".completeness", so what do you mean by "completeness knowl"?  Even for number fields (http://www.lmfdb.org/NumberField/Discriminants), the completeness is displayed on a page and not editable by knowl technology.  Also, that page has a link to itself on the right-hand side, is that intentional?

(And yes, I'm sorry, the levels.txt was just to confirm the data and put something here in our thread that anyone can access to see what we're talking about.)

I'm sorry for being slow on what these mean and how we're supposed to do this.

I'll play with the browse #1116 now.

jwj61:
It is confusing since we made a plan, and then people did various things which were slightly different.  Number fields is different because it already had a complicated completeness page.  At some point, we will convert that to a knowl, but doing that was not a priority since we already had the information.

I think extent is meant to be pithy.  It turns out that in many cases, this completely described the completeness, hence the double duty of the extent knowls.

It sounds like HMF may ultimately want 2 different statements because the completeness statement may be complicated.

Do you mind if I adjust the wording of the current knowl.  I can imagine someone reading it and not knowing what could be counted on (maybe there are some examples which are past the unstated bounds).

JohnCremona:
Go ahead, though JV is the boss for this.

jvoight:
@jwj61 Please, edit away!  And where is the plan so I can read about what I'm supposed to do?  Maybe it doesn't matter, and we'll just do something for HMF that makes sense.  I'll wait while @JohnCremona works on this.  (And thank you!)

JohnCremona:
the work is done, see #1116

jwj61:
I wrote something which can probably be improved upon.

The plan was never written out aside from notes on the blackboard at icerm.  It may have been erased by now.

jvoight:
OK, this has been merged in #1116 and worked fine for me.  Whatever minor tweaks could be done to this, I'm happy with it.

AndrewVSutherland:
If/when you figure this out, let me know if you think we should have 
separate extent/completeness links in the learn more about box on the 
genus 2 pages.

Drew

jwj61:
The genus 2 situation looks ok to me.

PR #1091, opened 2016-04-27
John cremona issue1079

jwj61:
This builds on John Cremona's download fix.  It deals with the case where a field in the download does not have its class group computed, and limits the number of fields in a download to 100,000


jwj61:
To see the class group change, search for fields with discriminant 77323367588329968832 (there are two), and then download.

JohnCremona:
Looks good to me.  I tested three files (not mathematica) and all is well (though for Magma you have to define x before reading the file).

PR #1095, opened 2016-04-28
Mf searching

fredstro:
This fixes problem nr. 3 in issue #1094 
i.e: If you search levels 1-12 and weight 2 the table you get is fine but in the boxes underneath you see 2-36 in the weight box, where it should be what you entered earlier.
and the same problem for the level aspect. See e.g.

ModularForm/GL2/Q/holomorphic/ranges/?level=3&weight=2-12&group=0
and
ModularForm/GL2/Q/holomorphic/ranges/?level=3-20&weight=2&group=0

For lmfdb version 2.0 (or anything > 1.0) I think we should use wtforms (or flask-wtforms) instead since this will deal with all things related to populating and displaying forms consistently.

### DIFF
--- a/lmfdb/modular_forms/elliptic_modular_forms/views/emf_render_navigation.py
+++ b/lmfdb/modular_forms/elliptic_modular_forms/views/emf_render_navigation.py
@@ -325,8 +325,12 @@ def render_elliptic_modular_form_navigation_wp(**args):
         weight_range = filter(is_even,weight_range)
     if len(weight_range)>1:
         info['weight_range']="{0}-{1}".format(limits_weight[0],limits_weight[1])
+    elif len(weight_range)==1:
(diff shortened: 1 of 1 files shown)


JohnCremona:
Testing right away...

fredstro:
Ok. and is someone dealing with the general search parsing? Otherwise I am happy to check for nonsense input as well (I think I wrote a function to do it while in Bristol but was told off by you or Samuele to not do it)

JohnCremona:
I think everyone who might do that is waiting for someone else to.  I hope nobody "told off" anyone else!  It is true that a new system for parsing search boxes was introduced.  I myself have not used it yet, and it did introduce several bugs.
If you can deal with parsing in this case -- whether or not you use the new system -- that will be an improvement and I'll accept it.  Meanwhile, this one passes.

PR #1097, opened 2016-04-28
Add error message in browse and search

fredstro:
This fixes problem nr. 4 in issue #1094 
I added safe error messages for wrong input in the 'search boxes' for weight, level and label.
See e.g.
/ModularForm/GL2/Q/holomorphic/ranges/?level=1-24a&weight=2-12a&group=1
and for the 'label box', go to /ModularForm/GL2/Q/holomorphic/ and input junk in the last box.
To check the 'safety', input anything from e.g. https://html5sec.org/


JohnCremona:
OK, I will test this now.  Thanks!

JohnCremona:
Looks good with one proviso:  I never knew that one could write delta in the box and get something, but it would be even better if it worked!  But I will merge this now in case you don't have time to so that now, since 'delta' is not a suggested in put anyway and it will be harmless since that does not cause an error anyway (it just reloads the same page).

PR #1098, opened 2016-04-28
added LMFDB version to homepage template

JohnCremona:
This adds a file lmfdb/version.txt which currently contains '0.9'.  When running not in beta mode the string 'LMFDB version 0.9' is displayed at the foot of every page (before the source link and Sage version).  When running in beta mode the string is 'LMFDB version 1.0beta'.

This is relavant to the recent thread opened on lmdb@googlegroups, referring to the proposal at https://github.com/LMFDB/lmfdb-collab/blob/gh-pages/workflow.txt


JohnCremona:
Although this is a pull request I am really only putting here so people can easily try it out.  It should not be merged unless agreement is reached on whether to actually have version numbers.

davidfarmer:
If we are posting a version, shouldn't it increment every time something changes?  So after
Mat 10 it would go
1.0.0
1.0.1
etc until we decide it is worthy of 1.1.0?

Of do we only show a truncation of the "true" version?

JohnCremona:
My plan was to only increment the 3rd field for bug fixes.  So the next genuine (non bug-fix) release is always of the form a.b with b incremented.

Since it is undefined when to go from (say) 1.5 to 1.6 or to 2.0 I did not implement that except that incrementing when b=9 triggers the next one to be (a+1).0, so that at least I could get the 0.9 --> 1/0 transition correct.  The automatic increment function is only there to display (next-version)beta on the beta pages so that does not matter a lot.  It may be that after a release,  when the current release is 1.5 and the beta site shows 1.6beta (say),  actually we next go to 2.0 and not 1.6.  No problem as that will involve changing te content of the text file.

!

AndrewVSutherland:
One small concern I have about this is that the version number will only 
reflect code changes, but it will not reflect data changes, which might 
be more important.  Changes to the mongo db will not trigger a version 
number change, so people citing a specific version have no guarantee 
that someone else looking at the same version will necessarily see the 
same thing.

That doesn't mean a version number isn't a very useful to have (e.g. 
when creating bug reports), I just want to flag this concern.

JohnCremona:
I fully agree with Drew -- and I said as much in my post to the mailing list.  But I have no idea how we might ever get to having discrete releases of the data itself.  People should perhaps be asked to put in a date when they cite the LMFDB, as we tell our students to do when citing a web page in a project or essay.

JohnCremona:
It might be a good idea to restart the mongo server -- ask JB to do that or I will tomorrow.

davidfarmer:
I have re-read the comments on version numbers, and I don't see how this solves more problems than
it creates.  We already post something like
 2016-05-04 21:52:27 +0100
which contains at least as much information as a version number -- unless we increment the
version number on all changes to the database (which is impossible or at least a headache),
and make it easy to convert a version number to a source code commit (which is possibly a headache).

We can talk informally about "version 2.0" and perhaps make a wish-list and set a target date
for that.  But I don't think that will really motivate people.  The version 1.0 push
is good and is helping us to do a lot and to think differently about the project, and we also hope
it is going to help with future funding.  I don't see us being willing to endure another such push
in the near future, especially since testing and keeping things working is now being taken
much more seriously: we should strive to be continually improving, and know that things
will not go backward.

In one sentence: even in the unlikely case that we could come up with a version numbering
system that accurately described the current state of the LMFDB, I don't see that it gives any
benefits.

JohnCremona:
That is fine.  I just wanted to stimulate a discussion.  We can revisit this at some point in future especially of we ever have a way to make a versioning system which includes the data as well as the code.
I will close this (without merging).

Issue #1051, opened 2016-04-08
degree 1 L-functions - Links on graph


edgarcosta:
Many of links in the graph to "associate Dirichlet L-function." in http://beta.lmfdb.org/L/degree1/ are broken.

At the moment, all the ones in the 2nd column, i.e., order 2, are broken.

davidfarmer:
There are some entries in the database that maybe should be different.
The one that is most relevant is:
 u'root_number': u'+*I'
which I think is meant to be a "1"

Other entries which could be different
u'sign_arg': -3.123599698885378e-17
   (I would prefer to see 0)
 u'values': [[1, u'0.62322524014023051339402008025+*I']]
   (I would prefer to see u'0.62322524014023051339402008025' ,
     since I think the "+*I" means "0*I", which is 0)
 u'leading_term': u'0.37369171291254730738158695002+*I',
     (same as previous comment)
 u'coeff_info': [u'x^2 - 1',
                 u'-1.00000000000000 + 1.22464679914735e-16*I',
                 u'\\zeta_{2}'],
    (can we omit the 0 imaginary part?)
 u'a2': [1.0, -2.8163065144692913e-30],
  (again, 0 imaginary part)

With an updated database the page should work properly.

jwbober:
The issues go deeper than just malformed database entries. I removed all of the Dirichlet L-functions from the database and I've reinserting corrected entries now. I should have a PR soon to fix the other issues, but really I'm not quite sure I understand why _all_ L-functions weren't broken.

jwbober:
Fixed by #1099 and updated database entries.

PR #1099, opened 2016-04-28
call p2sage on root number, fix p2sage.

jwbober:
The plotting code was trying to multiply the values of the plot by the
root number, which was a string in some cases. Meanwhile, p2sage didn't
really do the right thing, because it was returning a list of strings
sometimes when the input was a list of strings. Additionally, it didn't
take into account that the input might expect some variables to be
defined.

### DIFF
--- a/lmfdb/lfunctions/Lfunction.py
+++ b/lmfdb/lfunctions/Lfunction.py
@@ -121,7 +121,7 @@ def makeLfromdata(L, fromdb=False):
 
     if 'coeff_info' in data:   # hack, works only for Dirichlet L-functions
         base_power_int = int(data['coeff_info'][0][2:-3])
-        print 'base_power_int',base_power_int
+        #print 'base_power_int',base_power_int
         L.dirichlet_coefficients_analytic = L.dirichlet_coefficients_arithmetic[:]
         for n in range(0, len(L.dirichlet_coefficients_arithmetic)):
             an = L.dirichlet_coefficients_arithmetic[n]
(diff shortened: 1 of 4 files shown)


jwbober:
This should fix #1051

JohnCremona:
Can I have a URL or 2 to test on?  Thanks

edgarcosta:
Could you add to the p2sage description some examples of input and output?
So that someone can replace sage_eval in the future.

jwbober:
This is supposed to fix quadratic Dirichlet L-functions, e.g.
http://localhost:37777/L/Character/Dirichlet/12/11/.

It should also not break other L-functions, e.g.
http://localhost:37777/L/Character/Dirichlet/19/11/ or
http://localhost:37777/L/EllipticCurve/Q/11.a/ or
http://beta.lmfdb.org/L/Genus2Curve/Q/42983/a/.

I don't think anything non-L-function related should be affected.

I agree that this p2sage function should have a better description of what
I expect it to do, but right now I'm just trying to quickly fix things that
are broken.

edgarcosta:
> I agree that this p2sage function should have a better description of what
> I expect it to do, but right now I'm just trying to quickly fix things that
> are broken.

I agree and I understand that.

I'm just asking for some documentation to what kind of inputs you expect for p2sage, and what is the expected to output.
This would make it much easier for someone to be able to rewrite the utility function p2sage, avoiding sage_eval. Otherwise, if someone wants to do this, they will need to familiarize themselves with the rest of the lfunctions code...

JohnCremona:
There are lots of functions in lmfdb which are not well documented, compared (say) to Sage functions where nothing get in unless it has full documentation and tests.  For utility functions this is most undesirable.  But I think this week is not the time to be fussy.

JohnCremona:
I'll look again at this tomorrow.

PR #1100, opened 2016-04-28
Lattice fixes after upload

sanni85:
Fixed lots of problems on the homepage of lattices:
http://localhost:37777/Lattice/
now the lists of determinant and class numbers is shorter, I have added a list of nice names. 
For the moment I have also removed the Leech lattice as a possibility below the label search. Once it will be in the database I will edit the change.

Add download options for A1:
http://localhost:37777/Lattice/A1

Fixed the code for displaying A5 and other lattices:
http://localhost:37777/Lattice/A5

Changed the extent of the data file, modifying the statistics to include the  biggest class number.
http://localhost:37777/Lattice/Completeness

Added more indices in the import script, as well modified the add_q_expansion function.

Changed the tests to correct errors and added more to check high class number lattices.

At the moment I am computing theta series, for the high dimensional lattices it may require time.


JohnCremona:
Looks great.  I suppose you could have put in a test and offer Leech iff it is in the db, but that is not important.

Issue #1093, opened 2016-04-27
Genus 2 endomorphism display


JRSijsling:
If one goes to the page of a genus 2 curve, like

http://www.lmfdb.org/Genus2Curve/Q/169/a/169/1

then the endomorphism information of this curve is displayed; upon scrolling down, we see nice rings and algebras. Now we can scroll up and change the chosen invariants by clicking "Igusa" or "G2". We get new invariants. But if we scroll down, then only the skeleton of the endomorphism information remains, in the sense that only the surrounding statements are left. The meat, that is, all relevant lines on endomorphism rings and algebras, has gone.

This is somewhere in lmfdb/genus2_curves/templates/curve_g2.html. I know that since I coded the current invariant display (search for invstyle in that file to see where it comes from). I just do not know how to fix it... any help would be tremendously appreciated.

JohnCremona:
Very strange.  I am looking at it now.

JohnCremona:
Looking at line 210 in that html file: there are three items, the second of which has a double level of <span>...</span>  Why?  Is it that Igusa-Clebsch is the default?
Also one of these items have nodisplay in the class= lists (the first) but not the others.  Why?
I am just trying to understand the logic.

JRSijsling:
Your remarks are very correct. I wish that I understood the logic. The problem is that I took some of the similar code used for showing Magma and Sage code and hit that over the head until it worked. I may be able to hit it back into shape, but since I understand too little of this environment and since this is exactly how the current bug came to be, that seems like asking for trouble at some point further down the line. Perhaps someone who designed the code for showing Magma and Sage code will know a way out of this.

The nodisplay and the double level for Igusa-Clebsch are indeed both there because those are the default.

AndrewVSutherland:
I figured out how to fix this -- the problem is a name conflict (g2 is used for two different things).  I'll push the fix shortly.

JRSijsling:
Thanks Drew, that is great!

JohnCremona:
You win a prize!  Thanks

AndrewVSutherland:
Fixed in pull request #1102

PR #1102, opened 2016-04-28
Fixed g2/G2 name conflict in curve_g2.html responsible for #1093

AndrewVSutherland:


JohnCremona:
I am testing now.

JohnCremona:
This does not work properly for me.  OK if you click G2, but not if you then click Igusa, or back to Igusa-Clebcsh.

AndrewVSutherland:
Hmm, it works for me without any problems, I can click back and forth 
and cycle through all three with changing the endomorphism ring display.

Does the curve_g2.html file you are using have "G2" instead of "g2" in 
lines 210-213 (and a few other places)?

The branch I commited and pushed is called master (not genus2 which I 
used in Bristol).  I hope I didn't issue the pull request from the wrong 
branch?

Drew

AndrewVSutherland:
I've checked this on both chrome and firefox on ubuntu 14.04 and it seems OK, what browser/OS are you using?

JohnCremona:
Firefox on Suse.  I will check agan since I got mixed up with branches yesterday.

JohnCremona:
Sorry, I checked out the branch drew/minorgenus2changes but I see that it should be drew/master.  Which works!

Issue #1092, opened 2016-04-27
Help wanted


sanni85:
I just finished uploading almost all lattice data computed by Anna (total 39,292 lattices). Only one lattice is missing the Leech lattice. When I try to run the import script on it I get the following message:

> DocumentTooLarge: command document too large

Here is the file [Leech.txt](https://github.com/LMFDB/lmfdb/files/239043/Leech.txt), it contains only one line, and its size is 5.7 Mb
Do you know how to import it?

To run the import script, I usually run sage in the folder lmfdb, and in sage I use the commands:
```
%run lmfdb/lattice/isom.py
%run lmfdb/import_scripts/lattices/lattice.py /path_to_file
```

JohnCremona:
It seems thet this object is large enough to require gridFS.  Even though your text input file is <6MB and the mongo/bson limit is 16mb, it may be true that after creating the bson object to insert into the db it is > 16mb.

fredstro:
Since I suffered similar problems in the past I took a look at this and it is indeed the bson encoding of integers which takes up a lot of space. I think the default encoding is set to 64bit integers.
If you don't want to use gridfs there are essentially two options:
1) store the data as a string instead of encoding the integers (I checked that for the Leech lattice this will give a db record of about 7MB). 
2) write your own integer codec class for bson

sanni85:
Thank you Fredrik !
I will try to implement the first suggestion, convert the list to a string, it may solve the problem for now.

sanni85:
This is fixed in pull request #1115

jvoight:
Can I close this now?

PR #1103, opened 2016-04-28
Leech lattice

sanni85:
I have uploaded in the database part of the data we have about the Leech lattice.
I could not upload the list of 98280 normalized minimal vectors, but I could add all the rest of the data.
Therefore some changes were needed. All test pass, look at:
http://localhost:37777/Lattice/Leech

This does not solve Issue #1092, but we could update the record later on.


JohnCremona:
OK, good.  Can you
1. Add Leech to the "favourites list" (yopu can keep the US spelling fovorites if you must)
2. On the Leech page replace the 2-line Comment at the end with content in the top knowl saying (at least) "The Leech lattice is..."
The current text "This integral lattice is known with name Leech" is bad English.  It is not actually called "Leech" it is called "The Leech Lattice".

sanni85:
All done, i have also added a knowl in the test of the name.

JohnCremona:
That was quick -- but it still repeats "This is the Leech lattice" in the html.  (And I am not sure that anyone will notice  that the word "is" is a knowl linking to the definition of "name", but that is OK.

For simplicity I am going to merge this then edit that html file on github to remove the duplicate.

PR #1104, opened 2016-04-28
Change the sort of a number field search so that it can match an index.

jwj61:
We had problems with entry lengths in the indexes, so the index wasn't made, which meant there was no index to sort on.

Test with the url in issue 1063


JohnCremona:
Looks good.  We should look for other places where sorting is done since this works so well.

PR #1106, opened 2016-04-29
avoid duplicate comment about Leech lattice

JohnCremona:


JohnCremona:
See Lattice/Leech where the comment at the bottom had a superfluous line.  I also added a (brief) top knowl, which makes the comment itself irrelevant.
For the future I prefer having such somments as top knowls rather than storing them in the database.

Issue #936, opened 2016-03-30
Characters of unknown parity


davidfarmer:
This page has characters whose parity is n/a:

http://beta.lmfdb.org/ModularForm/GL2/Q/holomorphic/42/3/?group=1

fredstro:
At the moment the page above does not display this but the page:
ModularForm/GL2/Q/holomorphic/99/3/?group=1
does. I will see what I can do about this over the weekend... 
But I also think that someone (maybe Pascal) said that it would be possible to write a function which determine if a character is even or odd directly from the number so if someone can do that that would be even better then we could just call that function instead.

JohnCremona:
Agreed!  
I notice that the page for all Dirichlet chars of a given modulus (e.g. http://beta.lmfdb.org/Character/Dirichlet/42) does not show parities as such (they are in the last column!) , but the page for any individual one does.
Sine showing all characters for a given modulus is already computing the value at -1, surely a page such as http://beta.lmfdb.org/ModularForm/GL2/Q/holomorphic/42/3/?group=1 which shows the same characters could do also?  Perhaps I am missing the point.

fredstro:
I have fixed it now by calling the class ConreyCharacter from WebCharacter.py. I will make a pull request soon.

jwj61:
fixed by pull request #1113

PR #1113, opened 2016-04-29
Added functionality for getting the even/odd-ness of a character thro…

fredstro:
Added functionality for getting the even/odd-ness of a character through the ConreyCharacter class. I also added a test which fails without this pull request but passes after it.
While I was in the test file I also improved the long test function: 'test_many' by actually adding a test to it... 
See:
ModularForm/GL2/Q/holomorphic/999/3/?group=1

This fixes issue #936


JohnCremona:
The long test is skipped anyway.  Before, it did at least test that the pages could be created, but fine to add more.

PR #1115, opened 2016-05-01
Fix for the Leech lattice

sanni85:
this fixes Issue #1092 
for the Leech lattice shortest vectors and theta series are stored as list of strings
the download function has been modified for this lattice
all test pass

PS For all except 3 lattices A12, A13, A14, the first 150 coefficients of the associated theta series have been computed and stores (just finished)


sanni85:
http://localhost:37777/Lattice/Leech
vs
http://www.lmfdb.org/Lattice/Leech

PR #1116, opened 2016-05-01
more work on HMF stats

JohnCremona:
Successor to #1090 .  As I said there:

Separating the counting function by degree (of the base field), degrees 2, 3, 5, 6 are almost instantaneous while degree 4 takes a long time. Clearly degree 4 was causing the problem before. The numbers of newforms for degrees 2-6 are 188459, 45880, 102780, 16226, 15011 so degree 4 has fewer than degree 2. So I do not know why degree 4 takes so much longer.

This also removes redundant items in the Learn More section.  Please test, at least for degrees other than 4.

### DIFF
--- a/lmfdb/hilbert_modular_forms/hilbert_modular_form.py
+++ b/lmfdb/hilbert_modular_forms/hilbert_modular_form.py
@@ -9,7 +9,7 @@ from flask import Flask, session, g, render_template, url_for, request, redirect
 from sage.misc.preparser import preparse
 from lmfdb.hilbert_modular_forms import hmf_page, hmf_logger
 from lmfdb.hilbert_modular_forms.hilbert_field import findvar
-from lmfdb.hilbert_modular_forms.hmf_stats import get_stats, hmf_summary, hmf_degree_summary
+from lmfdb.hilbert_modular_forms.hmf_stats import get_stats, get_counts, hmf_summary, hmf_degree_summary
 
 from lmfdb.ecnf.main import split_class_label
 from lmfdb.ecnf.WebEllipticCurve import db_ecnf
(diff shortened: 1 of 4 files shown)


jwj61:
Started page, then clicked on 2 (and then 3) in the "browse by degree parts, and they gave errors.

jvoight:
I had an error from the previous merged version #1090, 
  AttributeError: 'dict' object has no attribute 'next'
in hmf_stats.py.

JohnCremona:
I am very puzzled.  First of all #1090 was merged yesterday.  This one is on top of that but you should not need to know that.  I just ran it again and get no errors. (Just rather slow in degree 4).

JohnCremona:
OK so now I am trying again from my laptop at home and get the errors you get.  (Before I was running from atkin.)  Weird.

OK I understand.  I use this aggregate function and we must be using different pymongo versions, hence different behaviour.  I'll rewrite the code not to use that (simpler anyway) and the PR will get updated.  Hang on...

JohnCremona:
Please try now.  If all works well with the possible exception of the degree 4 fields, that is a victory.

jvoight:
This works for me, but it was very slow at first--I would have assumed it was broken, if I wasn't patient.  But once it's computed, I guess it's cached, and then it's fine (even in degree 4).  

@jwj61 : Now we should add a link "Completeness of the data" on /ModularForm/GL2/TotallyReal/ and link it to /browse/?  I don't see a way to knowl this (and I also don't think everything needs to be a knowl).

JohnCremona:
Thanks.

jwj61:
I think we are ok for now.  A future thing to do is to make the completeness information more explicit.

I think knowl is better than separate page.  It can still be generated from the database/python as, for example, the elliptic curve extent knowl currently is.  The number field one was left for now because it already existed.  I expect to modify it at AIM since there are now more fields in the database, and updating it by editing the knowl is better than through a pull request.

JohnCremona:
At the risk of annoying those who have already patiently looked into this, I will make a new pull request for a version in which the time-consuming counts are read from a file, since the slowness is just not acceptable.

PR #1123, opened 2016-05-01
Fix jacobi slider

pascalmolin:
only the first digit of the argument was parsed + better error management


davidfarmer:
Looking at this

Issue #575, opened 2015-11-09
write blurbs for the objects and arrows in new picture


alinabucur:
The new picture is 
<img width="1031" alt="lmfdbuniverse" src="https://cloud.githubusercontent.com/assets/5437759/11040456/5d845d6e-86da-11e5-9141-dae268b26115.png">

JohnCremona:
See my comments at #535

alinabucur:
I created a pull request that partially solves this problem. But it is not fully functional yet.

jvoight:
@AndrewVSutherland pointed out that we are missing the relevant knowls, e.g. "automorphic form" and "motive".  (He also 'needs' them for his Sato-Tate groups.)  This should be a copy and paste from blurb to knowl, plus delta.

PR #1124, opened 2016-05-02
new map, some javascript issues

alinabucur:
I tried to put the new map into LMFDB, but there are some issues with the javascript. A stand-alone working version of the new map exists at 
http://www.math.ucsd.edu/~alina/lmfdbuniverse2/

However, when I tried to put the code into lmfdb, there are errors in the javascript. It loads, and it displays the definitions/examples on click. But the console registers errors on hoverover. If anyone knows enough javascript to fix it, please go ahead. 

This is related to #575


davidfarmer:
I don;t see where the file
lmfdbmap_code.js
is loaded in the picture.

Did you "git add" that file so it would be part of your pull request?
Even if that is done, there still needs to be a line (in the lmfdb/templates/universe.html file?)
which loads that file.

The same may need to be done for other new files.

alinabucur:
When I tried to add the lmfdbmap_code.js file, I could not figure out how to call on it correctly. What I tried to do now with the universe.html file was to include the content of the mfdbmap_code.js and the rest of the js files as inline code inside universe.html.

alinabucur:
Did git add. Are the files now part of the pull request? Or do I need to do something else to update the pull request?

JohnCremona:
After git add did you commit and push to github? If so then the pr would be
updated.   But github cannot detect changes on your own computer!

alinabucur:
i had forgotten to commit; now I can see 9 files associated with this pull request

davidfarmer:
What is supposed to happen when you click on one of the examples?
Those appear to be links, but when you click one you just get the same page again.

davidfarmer:
Some of them are links to "undefined", which gives an error when you click.

alinabucur:
Nothing should happen when you click on the examples. That was decided because we do not have incarnations of the examples in the database for most of them.

davidfarmer:
I'll try to disable those links.

davidfarmer:
I moved the js files to the appropriate place (although I don't think all of them are needed)
and made the Examples not be links (since they were not linking anywhere).  It seems to be
working and I am about to accept the pull request (actually, I will make a new pull request which
includes my changes), but the next time something needs to change on this picture, it would
be good if a knowledgeable person deleted the redundant code and checked that the code
is in the correct places.

alinabucur:
Thank you, David! And I completely agree with you about the redundant code. In fact, at least two files lmfdb.js and snap.svg-min.js already existed in the code. 

Do we want to change the sidebar to link to this new image?

PR #1125, opened 2016-05-02
Hidden Sato-Tate group implementation, not reachable from anywhere

AndrewVSutherland:
Initial implementation of Sato-Tate groups.  There is still some work to do on the data and there are knowls that need to be written, but the display code should be stable.  Currently these pages cannot be reached from anywhere, so there is no danger in merging.   I would like to get this pull request merged and pushed to beta so that people can look at them (by typing in urls directly) and we can then decide whether to link them to the genus 2 pages (which will be a very small change).

There are also a few tiny changes to some of the genus2 files that are just clean up, they are not related to the Sato-Tate group changes (I meant to push these separately but I didn't).

### DIFF
--- a/lmfdb/genus2_curves/genus2_curve.py
+++ b/lmfdb/genus2_curves/genus2_curve.py
@@ -17,7 +17,7 @@ from lmfdb.search_parsing import parse_bool, parse_ints, parse_signed_ints, pars
 from lmfdb.number_fields.number_field import make_disc_key
 from lmfdb.genus2_curves import g2c_page, g2c_logger
 from lmfdb.genus2_curves.isog_class import G2Cisog_class, url_for_label, isog_url_for_label
-from lmfdb.genus2_curves.web_g2c import WebG2C, list_to_min_eqn, isog_label, st_group_name, st0_group_name, aut_group_name, boolean_name, globally_solvable_name
+from lmfdb.genus2_curves.web_g2c import WebG2C, g2cdb, list_to_min_eqn, isog_label, st_group_name, st0_group_name, aut_group_name, boolean_name, globally_solvable_name
 
 import sage.all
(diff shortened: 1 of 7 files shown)


JohnCremona:
Do you want or need this to be merged before the 1.0 release?

AndrewVSutherland:
Yes.  We don't necessarily have to include it, but I would like to and 
it won't be an option if you don't merge it now so that people can see 
it.  We can make a decision on 1.0 later today.

Drew

JohnCremona:
Testing now.  All the URLS from your email work (after adding a missing ")") so I'll put them here:

I also put in the genus 1 Sato Tate pages, you can access them via:

http://127.0.0.1:37777/SatoTateGroup/1.2.U(1)    (for elliptic curves with CM over their field of definition)
http://127.0.0.1:37777/SatoTateGroup/1.2.N(U(1))  (for elliptic curves with CM not over their field of definition)
http://127.0.0.1:37777/SatoTateGroup/1.2.SU(2)   (for elliptic curves without CM)

You could link these into the elliptic curve pages if you wanted to by adding the Sato-Tate group to the list of invariants.

To see the the genus 2 Sato Tate pages, use URLs like

http://127.0.0.1:37777/SatoTateGroup/1.4.USp(4)

You can also access a random Sato-Tate group via

http://127.0.0.1:37777/SatoTateGroup/random

I still need to massage some of the data (adding generators and subgroup/supergroup relations), but this won't require any code changes.

There are also a handfule weight 0 examples I put in for David Farmer.  If he wants/likes them, I could put in a bunch more.  Checout

http://127.0.0.1:37777/SatoTateGroup/0.3.3T2,

for example.

JohnCremona:
OK!

PR #1130, opened 2016-05-02
improved speed of HMF stats by storing counts in a file

JohnCremona:
Here is the reason why my HMF stats page was so
slow, particularly for degree 4 fields.  In the list of fields I used
the fields_pretty() function, which was enhanced at my suggestion, in
Bristol, by John J, to display 4.4.1600.1 (say) as Q(sqrt(2),sqrt(5))
.  And this enhanced field_pretty, in degree 4 specifically,
constructs the WebNumberField object and computes its subfields.

I think that if I disable calling field_pretty on quartic fields then
I can make this fast as it should be.

Here I also have precomputed the counts and store them which also helps a lot.

This does mean that the stored file needs to be recomputed when more HMF data is uploaded.  But for now it just makes the existing pages faster.


JohnCremona:
Update PR actions JV's request to have raw field labels and not field_pretty versions for all degrees (affects quadratics).  In all cases the fields are sorted by discriminant.

jvoight:
This worked great for me.  Thanks JC!

PR #1133, opened 2016-05-02
hide bad NF L-function links

davidfarmer:


JohnCremona:
On 2 May 2016 at 15:56, davidfarmer <notifications@github.com> wrote:

> Merged #1133 <https://github.com/LMFDB/lmfdb/pull/1133>.
>

& pushed

PR #1135, opened 2016-05-02
hide HMF L-functions with d^2 N > 40,000

davidfarmer:


JohnCremona:
pushed to beta & prod

jvoight:
Thanks David!  The "first" one
  http://beta.lmfdb.org/L/ModularForm/GL2/TotallyReal/2.2.5.1/holomorphic/2.2.5.1-31.1-a/0/0/
comes up more quickly now.

There are very few HMF L-functions which meet this stringent requirement, so for 2.0 we should sit down and really spec out what we should want to have in terms of Dirichlet coefficients.

davidfarmer:
John: the required coefficients will arise naturally as the L-functions are computed rigorously.
The issue may not be the number of coefficients, but rather the other parameters in the search.

Edgar: fixed in PR #1138

PR #1136, opened 2016-05-02
Alinabucur univalmost

davidfarmer:
Alina's new universe picture.


JohnCremona:
Pushed to beta and prod.  The Map link on the sidebar still links to the old /bigpicture while this on is at /universe.  Perhaps we should publicise this new one and make subsequent changes (if any) prior to switching the link to this is linked from the sidebar (still "Map"?)

alinabucur:
I think we should change the sidebar as soon as possible. Everyone worked so hard to make the change before the press release goes out. But yes, asking people to "vet" it might be a good idea. Should we send an email around today and ask people to test it by tonight? Or should we skip the vetting entirely? It now works, the math won't change as it's been arrived at through many thorough discussions.

JohnCremona:
If you decide to do this today (fine with me) just change "bigpicture" to "universe" in lmfdb/sidebar.yaml.  And if you want the word Map in the sidebar itself to be different, change that in the line above.  Universe?

kedlaya:
I think there needs to be some text above the picture, particularly for those on touch screens who won't have the mouseover feedback to indicate that there are clickable spots. E.g., "This diagram is an overview of the interconnections predicted by the Langlands program and illustrated in the LMFDB. Click on any of the words in the diagram for more information."

sanni85:
For my screen is a bit too big... Only because I know that there are boxes appearing at the bottom I was scrolling down.

kedlaya:
On a touch screen, it is not so easy to hit some of the hot spots for the arrows. It might help if some of them were a bit "taller" (i.e., in the direction perpendicular to the text).

jwbober:
It is also too big on my (big) screen at the default zoom level. If I make the window smaller then it resizes to a better size.

Also, Chrome (on windows, at least) has terrible kerning on the text. and sometimes just makes some letters disappear. There's nothing we can do about that directly, of course, but we might want to be careful using things that have bad support in major browsers.

kedlaya:
On the other hand, when I "zoom out" in my browser, to make the text smaller, the diagram gets bigger to compensate! Confusing, but maybe not something to lose sleep over.

PR #1141, opened 2016-05-03
Add sato-tate group browsing and linked in genus2

AndrewVSutherland:
Added searching/browsing for Sato-Tate groups, and support for a (static) histogram thumbnail in the properties box.  Currently only USp(4) has a histogram, but I will add them for others tomorrow (this will not require any code it just means putting them in the database).

The Sato-Tate pages are now linked to from the genus 2 curve and isogeny class pages (but this is easy to undo if need be).

This should make Sato-Tate groups a candidate for inclusion in the sidebar -- once this is pushed people can play with it and see what they think.  I still need to add some data and write knowls, but there should be no further code changes (other than bug fixes).

### DIFF
--- a/lmfdb/genus2_curves/genus2_curve.py
+++ b/lmfdb/genus2_curves/genus2_curve.py
@@ -16,8 +16,8 @@ from lmfdb.utils import ajax_more, image_src, web_latex, to_dict, comma, random_
 from lmfdb.search_parsing import parse_bool, parse_ints, parse_signed_ints, parse_bracketed_posints, parse_count, parse_start
 from lmfdb.number_fields.number_field import make_disc_key
 from lmfdb.genus2_curves import g2c_page, g2c_logger
-from lmfdb.genus2_curves.isog_class import G2Cisog_class, url_for_label, isog_url_for_label
-from lmfdb.genus2_curves.web_g2c import WebG2C, g2cdb, list_to_min_eqn, isog_label, st_group_name, st0_group_name, aut_group_name, boolean_name, globally_solvable_name
+from lmfdb.genus2_curves.isog_class import G2Cisog_class, url_for_label, isog_url_for_label, st_group_name, st_group_href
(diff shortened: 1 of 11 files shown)


JohnCremona:
So the page http://localhost:37777/SatoTateGroup/ which already existed now has more content and links, right?  And is still not linked to from anywhere?

In the search results page I see baf html in the header to column 8.  It is from title='&nbsp;&nbsp;c' in the template.

We cannot make this public (= in sidebar on www) until all the knowls have content.  I would be quite happy to have it in the sidebar with beta flag so it would not appear in www, but it would be a big step for it to go straight into www sidebar with very little exposure and discussion on details (however good it is, and of course it is!).

Apart from the small bug mentioned above, it is hard to find a reason not to merge this now.

Is it not important to have, for as many of these groups as possible, some pages showing which objects they are attached to?  Something like the Degree 2, 3, 4 L-function pages?
Meanwhile

AndrewVSutherland:
@JohnCremona

* Yes the main search page has new content and links but is not linked 
to from anywhere.

* The heading to column 8 is intentionally written that way to improve 
the layout and renders ok on both chrome and firefox, but I am happy to 
remove the &nsbsp;&nbsp if you prefer.

* Agree 100% about the need to put in knowls and have the content 
reviewed more widely before it goes to the www sidebar, I only meant to 
signal my response to an earlier comment you made, that I think this is 
a possibility worth considering for 1.0.  If you can merge/push this 
(without putting it in the sidebar) we can get that feedback.  I will 
get all the knowls put in today.

* I agree with the idea of a page showing which objects they are 
attached to but suggest deferring this for the moment.  Once Sato-Tate 
groups are integrated into the L-function interface (in principle, every 
L-function should have a Sato-Tate group) this will happen 
automatically, but clearly we have more important things to worry about 
on the L-function side this week.  I don't see this as critical for 1.0.

* Did you mean to add anything after "Meanwhile", or was that just for 
dramatic effect?  :)

JohnCremona:
I don't know about my "Meanwhile"!

I'll merge it now so you can continue to work on it.

My comment baout column 8 did not come out as I meant in this thread.  I actually see the symbols "ampersand n b s p semicolon" etc.  But that can be fixed later.

PR #1148, opened 2016-05-03
Now you can use --enable-profiler, to profile the LMFDB

edgarcosta:
This is very helpful when trying to figure out the bottleneck on a specific webpage.


JohnCremona:
Just to confirm -- this will have no effect unless --enamble-profiler is on the command line, right?

edgarcosta:
Yes,

the line:
`    if "PROFILE" in configuration['flask_options'] and configuration['flask_options']["PROFILE"]:`
assures that.
and that key is only set in:
```
elif opt =="--enable-profiler":
                 flask_options["PROFILE"] = True
```

PR #1154, opened 2016-05-03
add Sato-Tate groups to beta sidebar

JohnCremona:
Does what it says.  No change to sidebar in non-beta mode.


JohnCremona:
No negative comments so merging.

PR #1156, opened 2016-05-03
Logo without beta tag (to be applied on May 9)

sanni85:
logo in prod with no beta tag, in beta with it

to test, run without --debug


JohnCremona:
I have confirmed that this works as it should: as now if either --debug on command line or BETA environment variable set, not otherwise.

To be merged on May 9.

PR #1159, opened 2016-05-03
Added links to Sato-Tate groups to ec and ecnf pages

AndrewVSutherland:
Small change to add the Sato-Tate group of an elliptic curve over Q or other number field as an invariant listed on the curves home page immediately below the endomorphism ring.

Over Q the Sato-Tate group is either N(U(1)) (CM case) or SU(2) (no CM).
Examples of both can be found at:

http://localhost:37777/EllipticCurve/Q/27/a/1
http://localhost:37777/EllipticCurve/Q/11/a/1

Over a number field K the Sato-Tate there are 3 possibilities: U(1), N(U(1)), or SU(2) depending on whether we have CM with CM field contained in K, CM field not in K, or no CM.

Examples of all three cases can be found at:

http://localhost:37777/EllipticCurve/2.0.11.1/%5B9%2C2%2C1%5D/CMa/1
http://localhost:37777/EllipticCurve/2.0.11.1/%5B1024%2C0%2C32%5D/c/1
http://localhost:37777/EllipticCurve/4.4.3600.1/100.1/f/1

Note that the branch is named ecstlink, not master.

### DIFF
--- a/lmfdb/ecnf/WebEllipticCurve.py
+++ b/lmfdb/ecnf/WebEllipticCurve.py
@@ -215,6 +215,12 @@ class ECNF(object):
                 self.End = "\(\Z[\sqrt{%s}]\)" % (d4)
             else:
                 self.End = "\(\Z[(1+\sqrt{%s})/2]\)" % self.cm
+            if self.signature == [0,1] and self.abs_disc == -self.cm:
+                self.ST = '<a href="%s">$%s$</a>' % (url_for('st.by_label', label='1.2.U(1)'),'\\mathrm{U}(1)')
+            else:
+                self.ST = '<a href="%s">$%s$</a>' % (url_for('st.by_label', label='1.2.N(U(1))'),'N(\\mathrm{U}(1))')
+        else:
+            self.ST = '<a href="%s">$%s$</a>' % (url_for('st.by_label', label='1.2.SU(2)'),'\\mathrm{SU}(2)')
 
         # Q-curve / Base change
         self.qc = "no"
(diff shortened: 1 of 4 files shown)


JohnCremona:
Fantastic.  I checked a variety of examples and can confirm the code's correctness.   When we have some fields of even degree > 2 not totally real represented, further testing will be needed -- perhaps a comment to that effect would be worth adding now?  But I'll merge right away -- Drew, can you add to the knowls ec.q.invariants and ec.invariants to mention the ST group in each case?  I found the knowl st.label but this will need something more.  For now it would suffice to say that ST(E) is the Sato-Tate group of E andgive an external reference.

AndrewVSutherland:
Yes, I meant to add that I will update the knowls once it is merged -- 
I'll do that now.

Regarding future work for fields of even degree > 2, where is the best 
place to note this?  Perhaps I should I open a feature request issue to 
add a boolean that indicates whether the CM field is contained in the 
field of definition (which I think is the best way to handle this, we 
probably don't want to be testing field inclusions on the fly and we 
might want to be able to search on it at some point in the future).

JohnCremona:
I was just thinking of a comment in the code you added.
When the CM is defined over the base field I call that "rational CM" and there's a Sage function for that, but I agree that it should be added to the database.  I can write a one-off script to do that to the curves we have and also change my upload scripts to include it.  Not right now.

AndrewVSutherland:
I added a comment to the code noting that it needs to change if/when we 
add elliptic curves over non-quadratic fields thar contain the hilbert 
class field of some imaginary quadratic field.

Issue #1140, opened 2016-05-03
Cannot create top/bottom knowls for objects whose labels are not valid knowl ids


AndrewVSutherland:
I was originally going to add an issue for not having top knowls for elliptic curves over number fields, but in fact, even if this functionality was added to ecnf, it wouldn't work, because many elliptic curves have labels that contain characters such as "[" and "]" that are not allowed in a knowl id.

For example,  you cannot create the knowl: "ec.nf.[11,5,1]-a1.top"

This is not specific to elliptic curves over number fields, it causes a problem with Sato-Tate groups, and is likely to cause problems elsewhere as well.  Anything character that can appear in a label should be able to appear in a knowl (in particular, we definitely need "(", ")", "[", "]", "{", "}" and possibly "^")

JohnCremona:
I know now that my labels for ideals of imaginary quadratic fields are really bad, and this is another example.  (Another is the problem which can arise having those [,] characters in a URL.)
The whole issue of labelling ideals is one which we have spent a lot of time on but never actually done anything about.  It will not happen this week.
The problem raised is only as issue for imaginary quadratics since for all other fields the conductor part of the label has the form N.i where N is the norm and i an index from 0 (and of course the issue is to know what order the ideals of a given norm are in).

AndrewVSutherland:
@JohnCremona This issue isn't about whether your ideal labels are good or bad, it is simply about the fact that we have imposed restrictions on knowl ids that are incompatible with many of our existing labels (which we are surely not going to change, especially after May 10).  We need to remove the restriction on knowl ids that prevents us from creating top/bottom knowls for objects whose labels contain characters not currently permitted in a knowl id.  This is not specific to elliptic curves over imaginary quadratic fields, that just happens to be where I first noticed it.

JohnCremona:
OK.  I was looking for the place where the exact restrictions on knowl ids is explained but could not find it.  It would actually be quite easy for this purpose to translate a label such as "ec.nf.[11,5,1]-a1.top" into (say) "ec.nf.11.5.1-a1.top"  -- though the knwl ID should include the field label too, so an example would be to go from the curve whose full label is 2.0.4.1-[65,18,1]-a1 to the knwol id "ec.nf.2.0.4.1-65.18.1-a1" by just doing id =  id.replace(",",".").replace("[","").replace("]","")

AndrewVSutherland:
[...]

PR #1161, opened 2016-05-04
Changed label format to conform to knowl id format

AndrewVSutherland:
Changed format of Sato-Tate group labels to conform to knowl id requirements (see discussion at https://github.com/LMFDB/lmfdb/issues/1140).  The new format is:

    w.d.r.c.ns

where w=weight, d=degree, r=real dim, c=components, n are all nonnegative integers and s is a lower case letter used to break ties (n is the 2nd number in the GAP id [c,n] of the component group).

This change should be merged and pushed quickly.  Despite my best efforts I managed to slightly break the current SatoTateGroups implementation in a few places.

### DIFF
--- a/lmfdb/sato_tate_groups/main.py
+++ b/lmfdb/sato_tate_groups/main.py
@@ -21,9 +21,9 @@ st0_dict = {
     'SU(2)':'\\mathrm{SU}(2)',
     'U(1)_2':'\\mathrm{U}(1)_2',
     'SU(2)_2':'\\mathrm{SU}(2)_2',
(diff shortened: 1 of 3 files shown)


JohnCremona:
In sto_dict you need to change some 'x' to '\times'.

Shouldn't the Properties box be a subset of the main page content (for an indivisual group)?  I see "components" in the former but "index" in the main page, do not see Name or label in the page (except for the name in the title).

Otherwise good!

AndrewVSutherland:
OK, these have all been addressed, should be ready to merge/push.

PR #1162, opened 2016-05-04
smaller plot and zero ranges for high degree L-functions

davidfarmer:
For things like symmetric powers and E/quadratic field, decrease the plot range to 12 and
the zero range to 10.  Should be not too bad for conductor up to 40000.


JohnCremona:
Will test in about 90 mins.

davidfarmer:
I suggest you test this and simultaneously test your related pull request, and then accept both of them
(assuming successful tests, of course).

A possible reason a test could now fail: the first zero of the Riemann zeta function will not show
up as a zero of a Dedekind zeta function, because those are only computed up to height 10.

All this is just temporary because we have not yet precomputed those L-functions.

Issue #1166, opened 2016-05-04
Explore and Learn / Knowledge


fredstro:
I think the page /knowledge/ should be re-organised (or we should include another link from the main page).

My Reasoning:
After talking to 'my' press person I was trying to look at the lmfdb  from a non-researcher's point of view.

I think that an interested person with maybe a small background in mathematics would definitely (after maybe a brief stop at 'visualise data') go to "Explore and Learn'. He or she would not necessarily start browsing from the sidebar... 
In the Explore and Learn the LMFDB map gives some information about what is included but it is too brief to provide any real information. 
Then the unsuspecting user then goes to 'Knowledge' and finds themselves directly at the interface of a large collection of information which they have to either scroll or search through. Including 10 instances named 'Delete me'.

To enhance the experience for  a random user that wants to 'Explore and Learn'  the link from the front page should go somewhere with some kind of structure, even if it is just a table of contents analogous to the sidebar but with relevant knowls for all topics.

The 'database' version should of course still be available for people who wants more detailed information.   But the 'delete me' knowls should probably either be deleted before the release or hidden from non-logged-in users.

If someone has time to look at this before the release I think it would be good  
Or if you think this is a non-issue feel free to close it or put it at low priority...

davidfarmer:
Maybe the "Knowledge" link in the sidebar should only be there if you
are logged in (just like the "Data" section of the sidebar)?

AndrewVSutherland:
@fredstro +1 for raising this issue, which is definitely not a "non-issue" or low priority.   We will hopefully get a lot of new visitors on May 10, and whether they ever visit the site again or recommend it to anyone else will depend largely on what they see in the first few clicks.  If we don't want the first click to be to the "Knowledge link" (which I agree is a natural place to go), then we should hide it.

JohnCremona:
I agree entirely and will right away make a pull request which hides Knowledge from the sidebar when not logged in.
Is the proposal also to remove the Knowledge link from the top home page?  I'll do that too.
Related note: if we switch to the new /universe then the other link in that box needs changing *as well as the image in the box*.

PR #1167, opened 2016-05-04
hide Knowledge when not logged in

JohnCremona:
This does two thongs (see discussion at #1166):
1. The sidebar does not include Knowledge at the bottom when not logged in (similar to the Data section)
2. The lower left browse box in the top home page, titled Explore and Learn, does not have the Knowledge link at all.  (It still mention knowls and the word Knowl could be a knowl pointing to http://beta.lmfdb.org/knowledge/show/doc.knowl).

### DIFF
--- a/lmfdb/index_boxes.yaml
+++ b/lmfdb/index_boxes.yaml
@@ -37,7 +37,9 @@ links:
 title: Explore and Learn
 image: map
 content: <p>The LMFDB makes visible the connections predicted by the Langlands program.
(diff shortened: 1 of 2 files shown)


JohnCremona:
I did try making the word "Knowl" in the Explore & Learn box into a knowl showing the knowl for knowl (!) and can add it to this if people want it, but it looks very odd when it opens up, and there are no other knowls at all in the top home page, so perhaps it is best without.  Either way the original version of this can be merged if people agree with the consensus as #1166

Issue #1170, opened 2016-05-04
Avoiding database_names/collection_names


edgarcosta:
Hello,

If possible the actions database_names and collection_names should be avoided in the flask app.  This is because:
0- the code for the web app, should know what it is doing, and what db/collection is trying to use
1- they need to be performed in the primary instance, which can be quite slow
2- they are slow

Should be replaced by a try:
- [x] lmfdb/modular_forms/elliptic_modular_forms/backend/__init__.py 

- [x] lmfdb/modular_forms/elliptic_modular_forms/backend/emf_utils.py 

Out of my league:
- [x] lmfdb/raw.py 

Best,
Edgar

fredstro:
I agree that these calls  should probably be replaced out of principle,
but for the purpose of optimization (user experience )  I don't think
they  are really relevant since they should only be called on module import
(unless something is wrong with the implementation )

Are you saying that a try / except instead of database_names etc. would
really speed anything up significantly? In that case these routines are
badly implemented in mongo/pymongo.

Fredrik

edgarcosta:
Yes it speeds up.

Take for example:
http://localhost:8082/EllipticCurve/Q/375893/a/1

With those calls, time to the first byte: 5.30s, 4.83s, 4.13s, 3.69s, 3.7s, 3.7s
Commenting all those calls out: 4.5s, 3.25s, 3.39s, 3.40s

So even though it caches some of those calls after the 3rd request, it still manages to waste 0.3s there.

Here you can see how expensive those calls are if you are not at warwick:
```
INFO:LMFDB@2016-05-04 18:21:53,975: Command listDatabases with request id 1361286294 started on server (u'warwick.lmfdb.xyz', 37010) 
INFO:LMFDB@2016-05-04 18:21:53,975: Command listDatabases with request id 1361286294 started on server (u'warwick.lmfdb.xyz', 37010) 
INFO:LMFDB@2016-05-04 18:21:54,127: Command listDatabases with request id 1361286294 on server (u'warwick.lmfdb.xyz', 37010) succeeded in 151747 microseconds 
INFO:LMFDB@2016-05-04 18:21:54,127: Command listDatabases with request id 1361286294 on server (u'warwick.lmfdb.xyz', 37010) succeeded in 151747 microseconds 
INFO:LMFDB@2016-05-04 18:21:54,128: Command listCollections with request id 39394527 started on server (u'warwick.lmfdb.xyz', 37010) 
[...]

PR #1174, opened 2016-05-05
avoiding database_names and collection_names -> try/except

edgarcosta:
Addressing #1170

### DIFF
--- a/lmfdb/modular_forms/elliptic_modular_forms/backend/__init__.py
+++ b/lmfdb/modular_forms/elliptic_modular_forms/backend/__init__.py
@@ -1,6 +1,5 @@
 from plot_dom import *
 
-db_name = 'modularforms2'
 
 import lmfdb
 import gridfs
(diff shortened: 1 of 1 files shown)


fredstro:
This patch seems ok.  All tests for elliptic modular forms pass.
Although for code-consistency I would have preferred new-style string formatting rather than '%s' and possibly a critical error message written to the logfile in addition to raising the exceptions but these are minor issues.

fredstro:
I just noticed one problem (which should have been a problem before this pull request as well)
When there is an error we try to write to the log with emf_logger but this is not defined. 
It might be easier if you could just add a line
from lmfdb.modular_forms.elliptic_modular_forms import emf_logger at the top of the file

jwj61:
@fredstro  Add that to which file?

fredstro:
To lmfdb/modular_forms/elliptic_modular_forms/backend/__init__.py

Issue #1169, opened 2016-05-04
Avoiding ensure_index/create_index


edgarcosta:
Hello,

The functions: ensure_index/create_index must be avoided in the flask app.
Perhaps the exception is when app saves a knowl.
The two main reasons are:
- this operations must be done in the primary instance
- and can lock writes and reads.

If we are running a client in the cloud, and warwick is the primary, the ensure_index knowles adds a 5/10s delay.

Also, since 3.0, ensureIndex()  == createIndex(), see https://docs.mongodb.org/manual/reference/method/db.collection.ensureIndex/.

From: https://docs.mongodb.org/manual/core/index-creation/
> By default, creating an index blocks all other operations on a database. When building an index on a collection, the database that holds the collection is unavailable for read or write operations until the index build completes. Any operation that requires a read or write lock on all databases (e.g. listDatabases) will wait for the foreground index build to 

For example, in these 3 examples the index should be done after a write, not before a read:
- [x] lmfdb/lmfdb/knowledge/knowl.py
- [x] lmfdb/knowledge/main.py 
- [x] lmfdb/lmfdb/users/main.py 

I'm not sure what do with this one:
- [x] lmfdb/raw.py

JohnCremona:
Agreed.  one can only do it when authenticated to write to the db anyway,
which surely the web app is not?

John

edgarcosta:
the lmfdb user can make enough damage if they desire:

```
db.system.roles.find();
{ "_id" : "admin.index_and_list_collections", "role" : "index_and_list_collections", "db" : "admin", "privileges" : [ { "resource" : { "db" : "", "collection" : "" }, "actions" : [ "createIndex", "listCollections" ] } ], "roles" : [ ] }


db.system.users.find();
...
{ "_id" : "admin.lmfdb", "user" : "lmfdb", "db" : "admin", "credentials" : { "MONGODB-CR" : "13e13d846619ccc9f49ce62b72736d2c" }, "roles" : [ { "role" : "index_and_list_collections", "db" : "admin" }, { "role" : "readAnyDatabase", "db" : "admin" } ] }
...
```


those two actions: "createIndex", "listCollections", they re always performed in the primary, and should be avoided at all costs.

I was going to open another issue regarding listCollections and listDatabases.

jwj61:
[...]

PR #1175, opened 2016-05-05
commenting out ensure_index

edgarcosta:
See #1169 

Am I breaking something in the knowls or users?

### DIFF
--- a/lmfdb/knowledge/knowl.py
+++ b/lmfdb/knowledge/knowl.py
@@ -11,14 +11,15 @@ DSC = pymongo.DESCENDING
 def get_knowls():
     _C = getDBConnection()
     knowls = _C.knowledge.knowls
-    try:
-        knowls.ensure_index('authors')
-        # _keywords is used for the full text search
-        knowls.ensure_index('title')
-        knowls.ensure_index('cat')
-        knowls.ensure_index('_keywords')
-    except pymongo.errors.OperationFailure:
-        pass
+# See issue #1169
+#    try:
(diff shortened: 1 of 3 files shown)


JohnCremona:
I cannot think of any reason to have those there.  But I also thought that the (now deprecated) ensure_index function was harmless in that if that index existed it did nothing at all.  Once created, surely any index automatically updates itself as data changes?
The tests pass ok.

jwj61:
I was just about to click merge

JohnCremona:
@jwj61 sorry to spoil your fun.  There will be a much more trivial one for you to click on in a minute.

edgarcosta:
@JohnCremona It indeed should be harmless, and instantaneous. However, with the current topology, each one of those ensure_index is quite expensive, as it needs to go through the tunnel, each createIndexes end up consuming 0.15s to 0.25s. 

For each get_knowls() there are 4 of those, and get_knowls() gets called everytime you do get_knowl(). 
Further, there are pages that do get_knowls() a lot of times, for example:  /L/Genus2Curve/Q/681115/a/
ends up doing 22 calls to get_knowls(), in the particular case I was testing it on my desktop all those added up 19.744s.

for some gory details: https://gist.github.com/edgarcosta/c138283aa0c945cc3efde0541ff235d9

JohnCremona:
I think it is a Very Good Thing that you are noticing and eliminating all these inefficiencies.  I'm sure there will be more.  Thanks.

PR #1176, opened 2016-05-05
Mongo replica set client

edgarcosta:
For pymongo 2.x it is highly advisable to use MongoReplicaSetClient instead of just MongoClient.

I tested it with pymongo 2.8 and 3.2.2

For more details see:
https://api.mongodb.org/python/2.8/api/pymongo/mongo_replica_set_client.html
https://jira.mongodb.org/browse/PYTHON-1091?focusedCommentId=1254814&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-1254814

### DIFF
--- a/lmfdb/base.py
+++ b/lmfdb/base.py
@@ -47,7 +47,7 @@ def makeDBConnection(port, **kwargs):
         logging.info("establishing db connection at port %s ..." % port)
         import pymongo
         logging.info("using pymongo version %s" % pymongo.version)
-        if pymongo.version_tuple[0] >= 3 or kwargs.get("replicaset") == '':
+        if pymongo.version_tuple[0] >= 3 or kwargs.get("replicaset",None) is None:
             from pymongo import MongoClient
             _C = MongoClient(port = port,  **kwargs)
         else:
--- a/lmfdb/base.py
+++ b/lmfdb/base.py
@@ -55,16 +55,16 @@ def makeDBConnection(port, **kwargs):
             _C = MongoReplicaSetClient(port = port,  **kwargs)
         mongo_info = _C.server_info()
         logging.info("mongodb version: %s" % mongo_info["version"])
-        logging.info("_C = %s", _C)
+        logging.info("_C = %s", (_C,) )
         #the reads are not necessarily from host/address
         #those depend on the cursor, and can be checked with cursor.conn_id or cursor.address 
         if pymongo.version_tuple[0] >= 3:
-            logging.info("_C.address = %s" % _C.address)
+            logging.info("_C.address = %s" % (_C.address,) )
         else:
-            logging.info("_C.host = %s" % _C.host)
+            logging.info("_C.host = %s" % (_C.host,) )
 
-        logging.info("_C.nodes = %s" %  _C.nodes)
-        logging.info("_C.read_preference = %s" %  _C.read_preference)
+        logging.info("_C.nodes = %s" %  (_C.nodes,) )
+        logging.info("_C.read_preference = %s" %  (_C.read_preference,) )
 
 
 
--- a/lmfdb/base.py
+++ b/lmfdb/base.py
@@ -55,6 +55,17 @@ def makeDBConnection(port, **kwargs):
             _C = MongoReplicaSetClient(port = port,  **kwargs)
         mongo_info = _C.server_info()
         logging.info("mongodb version: %s" % mongo_info["version"])
+        logging.info("_C = %s", _C)
+        #the reads are not necessarily from host/address
+        #those depend on the cursor, and can be checked with cursor.conn_id or cursor.address 
+        if pymongo.version_tuple[0] >= 3:
+            logging.info("_C.address = %s" % _C.address)
+        else:
+            logging.info("_C.host = %s" % _C.host)
+
+        logging.info("_C.nodes = %s" %  _C.nodes)
+        logging.info("_C.read_preference = %s" %  _C.read_preference)
+
 
 
 # Global to track of many auto reconnect attempts for _db_reconnect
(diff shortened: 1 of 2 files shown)


fredstro:
I tried this on lehner (which is still running pymongo 2.8) and I get an error: 
ConfigurationError: the replicaSet keyword parameter is required
First of all the 'if' test doesn't work since the parameter replicaset doesn't get set through the configuration. Maybe add an 'is  None' test also? Or use get('replicaset','') to get the default value of ''.

Another question is what should replicaset be set to if it is not empty? 
I wonder this mainly because in my routines which computes and inserts modular forms (separate from lmfdb) I now keep getting errors: '‘waiting for replication timed out’  which prevents me from uploading data... and I want to try to use the  ReplicaSetClient and see if that works better.

edgarcosta:
Thank you very much for your feedback.

for the pymongo replicaset == None or '' are equivalent, however the default is indeed None (as it should be!) and I fixed it that in `lmfdb/website.py`

I also added some logging info when the client connects to the DB.

@fredstro Could you please double check that the previous bug doesn't occur anymore.

Regarding your issues with writing to the replica set.
Unless you are logged in warwick, you cannot take advantage of the replica set.
As you can't connect to {m0, m1, warwick}.lmfdb.xyz .
Are you checking that your writes are going through, before submitting the next write?
The default write concern is: `{ w: "majority", wtimeout: 5000 }`
see: https://docs.mongodb.org/manual/core/replica-set-write-concern/
If you are connected to the replica set, you can always change that in your writes, but given the topology of the setup I would highly advise you against that, as you might make our secondaries become stale.

fredstro:
I can verify that the previous bug doesn't occur now.

I am indeed logged in to warwick (that's where lehner is)
I am not checking anything but I will take a look at the docs and see what I can do. 
Everything worked fine before today or yesterday and I haven't changed anything to my code so either the replica sets are just very slow now or something has happened.

since we don't want wrong data in the database for the release I really hope 
 I can figure it out quickly how to do this since I have a lot to do and won't have any time over the weekend...

JohnCremona:
We could easily update every warwick machine to have a more up to date
pymongo.  I have been running pymongo 3.2.2 on atkin for a while and all is
good.

Fredrik, I can easily upgrade lehner's sage-python from pymongo 2.8.

Edgar, same for the sage on lmfdb.warwick ( as is running for both www and
beta).  Then for consistency you should use 3.2.2 everywhere,

Note that I removed the 2.8 from lmfdb/requirements.txt but of course that
will not make any difference to people's existing setups unless they
upgrade pymongo.

Let me know if you want those updated,

John

fredstro:
I think it would be good to upgrade pymongo on leaner as well.
I just am wondering though why it seems that the database connection from leaner is so slow that the replica timesout.
Mybe the problem might be that the ssh port 37010 (presumably opened by you or John Jones) is getting busy because one of you is uploading a lot of data or something. 

Fredrik

jwj61:
I have a connection open, but I am not uploading data.  It is just there for development.

edgarcosta:
@fredstro I'm sorry for the confusion. 
In all the above comments please read warwick = lmfdb.warwick.ac.uk, in other words, if your external ip is not `137.205.56.254`, then you can't connect to the replica set, as you don't have access to the port warwick.lmfdb.xyz:37010 (which is a ssh forward to lmfdb.warwick.ac.uk:37010).
However, you can connect directly to lmfdb.warwick.ac.uk:37010, without being aware that it is in a replica set, as it's the Primary member of the replica set, and the writes will follow the default write concern of the replica set.

AndrewVSutherland:
www.lmfdb.org has been intermittently very slow all morning (with 
occasional timeouts) and still is (http://www.lmfdb.org/EllipticCurve/Q/ 
just took 20s to load).

I assumed someone was uploading or replicating data.  If this is not the 
case we should figure out what the problem is.

edgarcosta:
@AndrewVSutherland, @jwbober is adding Lfunctions :)

edgarcosta:
@JohnCremona I will look into that. I would only update it before the release if there is a good reason, like the webpages magically load much faster.
If I update it in the cloud I will let you know.

@fredstro at the moment there is one extremely slow ssh tunnel between lmfdb.warwick.ac.uk to warwick.lmfdb.xyz, which makes all the up arrows in this diagram:
https://docs.mongodb.org/manual/_images/crud-write-concern-w2.png
extremely laggy.

I'm aware that I need to do some drawing to explaining the topology of our setup to everyone and its caveats, but this solution is supposed to be temporary.

JohnCremona:
I am upgrading one of the currently unused Sage builds on lmfdb@warwick from 7.0 to 7.1 and then will upgrade its pymongo also.  I don't think that will add much to the load on that machine but I can stop it if wanted.

jwj61:
Trying this on lehner, lmfdb won't start with:


INFO:LMFDB@2016-05-05 18:50:26,442: establishing db connection at port 37010 ... 
INFO:LMFDB@2016-05-05 18:50:26,442: using pymongo version 3.2.2 
INFO:LMFDB@2016-05-05 18:50:26,451: mongodb version: 3.2.1 
INFO:LMFDB@2016-05-05 18:50:26,451: _C = MongoClient(host=['localhost:37010'], document_class=dict, tz_aware=False, connect=True, replicaset=None, read_preference=Nearest(tag_sets=None)) 
    import lmfdb
  File "/mnt/data/jj/lmfdb/lmfdb/__init__.py", line 6, in <module>
    from website import main
  File "/mnt/data/jj/lmfdb/lmfdb/website.py", line 376, in <module>
    _init(**configuration['mongo_client_options'])
  File "/mnt/data/jj/lmfdb/lmfdb/base.py", line 114, in _init
    makeDBConnection(port = port, **kwargs)
  File "/mnt/data/jj/lmfdb/lmfdb/base.py", line 62, in makeDBConnection
    logging.info("_C.address = %s" % _C.address)
TypeError: not all arguments converted during string formatting

edgarcosta:
@jwj61 Thank you. It is fixed

PR #1178, opened 2016-05-05
Minor formatting changes to Sato Tate groups

AndrewVSutherland:
The only changes are in the formatting / sort order of the search results and some minor tweaks to the properties box layou.


JohnCremona:
Can you make a test file for Sato-Tate?  It would increase confidence if we could run the tests after these changes and know that everything still worked.
Ideally there should be a lot of tests on the ST home page which includes a test of all links from there and a variety of searches, as well as some tests for the individual pages.

jwj61:
Checking ...

AndrewVSutherland:
@JohnCremona added test_st.py

JohnCremona:
Thanks for the tests.  I'll let JJ merge this one as I'm off to dinner.

PR #1180, opened 2016-05-05
remove one test after #1167

JohnCremona:
Trivial -- after #1167  we don't test for "knowledge" in the top home page.

### DIFF
--- a/lmfdb/test_homepage.py
+++ b/lmfdb/test_homepage.py
@@ -69,7 +69,8 @@ class HomePageTest(LmfdbTest):
         """
         homepage = self.tc.get("/").data
         self.check(homepage, "/bigpicture", 'some varieties are modular')
-        self.check(homepage, "/knowledge/", 'Recently modified Knowls')
+        # removed in PR #1167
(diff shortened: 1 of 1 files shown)


JohnCremona:
Comment (noticed when testing this with ./test.sh lmfdb/test_homepage.py) -- testing Box 2 now takes longer than it did, and also produces  a lot of emf debugging output.

@fredstro was this something you did?  Try the above test to see what I mean.

fredstro:
I see what you mean but I don’t know why the debugging info appears there.
As far as I can see the log level for emf_logger is only set at two places:

1) In the test file I set the logging level to 100 (this was done by me in December) to avoid displaying anything
2) in modular_forms/elliptic_modular_forms/views/emf_main.py the log level is indeed set to 10 which is debugging 
but according to ‘blame’ that has been there since september 23 2014 so it shouldn’t suddenly start printing stuff. 
I thought, however that you or someone else removed all the cases where I had set the log level to debug a long time ago.

Fredrik

PR #1182, opened 2016-05-05
raw, Issues #1167 #1170 (partially)

edgarcosta:
Commented out the ensure_index, and replaced a colletion_names() call with a try and except.

Also, I don't see a way to really fix the function 'def database_list()'

Best,
Edgar

Issues #1169 #1170


JohnCremona:
I would not worry about /raw.  I never even knew it existed.  Surely /api does all that and more.

PR #1189, opened 2016-05-05
Put Sato Tate groups on production side bar

AndrewVSutherland:
I think Sato-Tate groups is now ready for the production side bar.  For Release 1.0 only weight 1 Sato-Tate groups will appear, but the code supports other weights.  There are currently a handful of weight 0 groups in the database for those who want to see how they look when reviewing this pull request.


jwj61:
Waiting until we get the database side sorted out a bit better since at the moment, there is no production server.

PR #1191, opened 2016-05-05
Fixes #1185

### LINKED ISSUE #1185: Problem with knowls in developer versions
If a developer is running their own copy of lmfdb and clicks on a knowl which doesn't exit yet, it crashes on a permissions error.  Recently, this would be ok, although trying to create the knowl would fail silently.  

Crashing seems like the wrong behavior.

edgarcosta:
I added try and except to g/setting locks

I'm not sure this is the ideal solution, but here it is.


jwbober:
Please don't put in try/except blocks that accept every exception.

PR #1194, opened 2016-05-06
replaced bigpicture with universe in the sidebar and homepage explore …

alinabucur:
in the sidebar and homapge "explore and learn" box; and on templates/index.html. I also changed the knowl 'Introduction' 
I hope I caught everything.


AndrewVSutherland:
It looks like smalluniverse.png cuts off the edges of the motivic and automorphic worlds, is that intentional?

AndrewVSutherland:
How does this differ from http://www.lmfdb.org/universe?

sanni85:
Alina could you also change the file test_homepage.py
it is enough that you replace the line 
>  self.check(homepage, "/bigpicture", 'some varieties are modular')
 with
> self.check(homepage, "/universe", 'universe')
so the test will pass

alinabucur:
@AndrewVSutherland: smalluniverse png is cutoff to be square (otherwise the code changes proportions and it looks horrid). If you look at the current version on prod and beta, that box has a cutoff of David's bigpicture. One could change and choose another square part of the picture, but it has to fit the shape reserved for in the box. Not sure I understand your second comment, I mean I have no idea what object/page you are referring to. 

@sanni85: done

PR #1202, opened 2016-05-06
Tweak universe

davidfarmer:
Does this fix the visibility problems with the universe picture?

### DIFF
--- a/lmfdb/static/images/lmfdbmap.svg
+++ b/lmfdb/static/images/lmfdbmap.svg
@@ -7,10 +7,9 @@
    xmlns:svg="http://www.w3.org/2000/svg"
    xmlns="http://www.w3.org/2000/svg"
    xmlns:xlink="http://www.w3.org/1999/xlink"
-   height="100%"
-   width="100%"
    version="1.1"
    id="svg3114"
+   class="universe"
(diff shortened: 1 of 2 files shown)


AndrewVSutherland:
It helps!  But it's still a bit too big, let me play with it on my machine and I may post another pull request shortly.

davidfarmer:
For those people who have a narrow browser, it is already on the small side.  (For example, if
I put two browsers side-by-side.)

Let's not ruin it for everyone else just to accommodate those people who maximize their browser.

AndrewVSutherland:
If I change the svg.width to 72% it just fits.  At 88% clicking on the L-function button puts all the text below the bottom of the screen except the word "Example".   I think running browser maximized is a more common use case (and at least it would still works at 72% when not maximized, where as it is not usable at 88% when maximized).

davidfarmer:
It seems like you are testing on a screen that is unusually wide compared
to its height.

AndrewVSutherland:
It's a standard 16:9 aspect ratio, which is also the standard ratio for projectors.

sanni85:
One of the problem is that the information displayed on hovering are at the bottom of the picture, so it would be great not to have to scroll, and have it all on one screen.

PR #1208, opened 2016-05-06
Only show L-function friend if conductor & dimension make it reasonable.

jwj61:
The cutoff is the analogous to the one used for Dedekind zeta-functions.

In dimension 3, this allows the link for the first few Artin reps, but not ones near the bottom of the first page of search results.

### DIFF
--- a/lmfdb/artin_representations/main.py
+++ b/lmfdb/artin_representations/main.py
@@ -206,7 +206,8 @@ def render_artin_representation_webpage(label):
                 friends.append(("Determinant character", url_for("characters.render_Dirichletwebpage", modulus=cc.modulus, number=cc.number)))
 
     # once the L-functions are in the database, the link can always be shown
-    if the_rep.dimension() <= 6:
+    #if the_rep.dimension() <= 6:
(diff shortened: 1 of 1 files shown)


JohnCremona:
I checked examples from top and bottom of the dim 3 search results -- good.

JohnCremona:
Have merged and will push to prod & beta.

davidfarmer:
The Dedekind zeta functions require a restrictive cutoff because they
are not primitive L-functions.  This means that there is no zero
repulsion, so there can be zeros that are very close together.  A naive
search for zeros, which the compute-on-the-fly does, can miss pairs
of zeros.

And it does: the first degree 4 Dedekind zeta function was missing
the first zero of the Riemann zeta function.

You can probably get away with a larger cutoff (my guess: 10 times
higher) for the Artin L-functions.  Just look at some plots and see
if things look okay.

JohnCremona:
I based my merge no the fact that the code works, not on any judgement that the cutoff was appropriate!  Feel free to change.

jwj61:
I will adjust it after dealing with the dimension 1 issue

PR #1210, opened 2016-05-06
L-function friends of 1-dim Artin reps now go to Dirichlet L-functions

jwj61:
Implements result of e-mail discussion.

Note, if you search for 1-dimensional artin reps with conductor 10000-20000, then the Dirichlet character Lfunction is not there, so the link is not shown.


jwbober:
This is good, but I was suggesting something different, which was to
redirect the L-function urls themselves. That might be better (to maintain
the "add an L to the url" convention) but it would be a messier change.

jvoight:
This worked for me, so I pulled it.  @jwbober if you want to implement the messier change, we can compare.

Issue #1201, opened 2016-05-06
Internal server error produced by Siegel modular form links.


AndrewVSutherland:
I was just demoing the LMFDB and found that clicking on any of the following links (including the very first link on the main Seigel modular form page) results in an "Internal Server Error"

http://www.lmfdb.org/ModularForm/GSp/Q/Sp4Z_j/
http://www.lmfdb.org/ModularForm/GSp/Q/Gamma0_2/
http://www.lmfdb.org/ModularForm/GSp/Q/Gamma1_2/
http://www.lmfdb.org/ModularForm/GSp/Q/Gamma_2/

sanni85:
There is a test for all the links on the browse page for Siegel modular forms. 
I wrote it this week and all tests were passing, so the link were all working.

davidfarmer:
Works on my laptop, but it does take around 15 seconds for the
first one to render.

sanni85:
the only one who does not work for me is http://www.lmfdb.org/ModularForm/GSp/Q/Sp4Z_j/

AndrewVSutherland:
I just clicked on all 4 links above from (a) linux laptop, (b) windows desktop, (c) ipad mini (d) android phone and get Internal Server Error in every case.

davidfarmer:
Still works fine (if a bit slow) on my laptop.

On www and beta, I get Internal Server Error instantly.

I have not upgraded any database software, so that may be a clue?

AndrewVSutherland:
Hmm, is it possible we are getting redirected to different back ends?  Or that there are different worker threads being assigned?

AndrewVSutherland:
Here is a traceroute from my machine at home:

 1  10.0.0.253 (10.0.0.253)  0.983 ms  0.979 ms  1.271 ms
 2  96.120.66.105 (96.120.66.105)  13.681 ms  14.550 ms  18.560 ms
 3  ge-1-11-rr01.bolton.ct.hartford.comcast.net (68.87.148.177)  19.749 ms  20.079 ms  20.083 ms
 4  be-115-ar01.woburn.ma.boston.comcast.net (68.85.106.25)  21.603 ms  22.035 ms  22.030 ms
 5  4.68.71.221 (4.68.71.221)  21.022 ms  20.635 ms  23.635 ms
 6  * * *
 7  * * *
 8  212.187.173.54 (212.187.173.54)  93.795 ms  97.363 ms  94.683 ms
[...]

PR #1211, opened 2016-05-06
please do not close the connection

edgarcosta:
a bug  related to #1201


edgarcosta:
This not only adresses #1201 partially, it will also speed up the whole interaction with mongodb.

Issue #1207, opened 2016-05-06
Classical modular forms crash


JohnCremona:
ModularForm/GL2/Q/holomorphic/49/8/?group=0
(i.e. enter 49 for level and 8 for weight in the search boxes)
crashes with:
AttributeError: 'str' object has no attribute '__call__'
when evaluating
<td align="left">{{ f.coefficient_field.relative_polynomial_latex('x') }}</td>

fredstro:
I can have a look at it on Monday when 8 will be close to a computer again
but if @sehlen has time to look at it before then that would be good.

Fredrik

AndrewVSutherland:
Fixed by merged pull request #1214.  I will leave this open until the code is pushed and the fix is on lmfdb.org.

AndrewVSutherland:
I am removing the critical tag, as the page no longer crashes on www.lmfdb.org,  But I will leave the issue open, as @fredstro indicates that this crash demonstrates an underlying problem in classical modular forms that is not addressed by pull request https://github.com/LMFDB/lmfdb/pull/1214.

PR #1214, opened 2016-05-07
Fix critical issue #1207 in classical modular forms

AndrewVSutherland:
This is a simple one line fix, the code was failing because the rational coefficient field has no method .relative_polynomial_latex. 

Note: test_non_triv_character fails, but it failed before this change as well, I will raise a separate (non-criticla) issue for this.

### DIFF
--- a/lmfdb/modular_forms/elliptic_modular_forms/views/templates/display-list-newforms.html
+++ b/lmfdb/modular_forms/elliptic_modular_forms/views/templates/display-list-newforms.html
@@ -96,7 +96,7 @@ info in this page is set by set_info_for_web_newform in emf_render_web_newform.p
         <td align="left">{{ f.coefficient_field.relative_polynomial_latex('x') }}</td>
         {% elif not f.coefficient_field.lmfdb_label and f.coefficient_field.lmfdb_label != '1.1.1.1'%}
         <td align="left">$\Q(\alpha_{ {{loop.index}} })$</td>
-        <td align="left">{{ f.coefficient_field.relative_polynomial_latex('x') }}</td>
+        <td align="left">{{ f.coefficient_field.relative_polynomial_latex('x') if f.coefficient_field.degree() > 1 else '\(x\)'}}</td>
         {% endif %}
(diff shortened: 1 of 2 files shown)


fredstro:
It is,  but not necessarily as you wrote it...
The underlying problem is that the coefficient field of a webnewform should
be a WebProperty which has a value which is a usual  number field but
intrinsically is "our version of" a web number field which means in
particular that it has the property relative_polynomial_latex.
It is not a problem that the field is the rational field but rather that
the field was not initialized properly ftom the database,  most likely
because of an incomplete database record.
This means of course that the problem could also appear for higher degree
fields.
The best fix to avoid a critical fail would be to test for the property
before calling the template or by using a python getattr with a default
value.
Fredrik

AndrewVSutherland:
@fredstro Just to be clear, you are not saying that http://www.lmfdb.org/ModularForm/GL2/Q/holomorphic/49/8/?group=0 is now incorrect, rather you are saying that this bug is indicative of an underlying problem that could be causing trouble elsewhere, right?   Assuming that is the case, I think this fix does make things "less bad" for release 1.0 and should not be reverted.  Certainly the change I made is not going to break anything that is not already broken.

I don't think we should be making any further code changes before May 10 unless there are truly necessary (i.e. they correct a crash or prevent incorrect data being displayed), it is just too risky.  We may need to defer the "best fix" to a later date when we have time to thoroughly test (and ideally we would test on beta first before moving to production).

But it certainly would be a very good idea to understand exactly why the field is not being initialized properly and to know whether this affects other pages. If it does, then we will have to make a judgement call as to whether it is better to try and fix it or to simply hide those pages.

Is there perhaps an automated way you could test this?

fredstro:
Well, it is certainly no point in reverting this but since it might "mask"
other problems it shoukd be taken care of more thoroughly at dome later
point. ..
Fredrik

fredstro:
Stephan Ehlen @sehlen should probably take a look at it since the backend
is based on his code.

Issue #1206, opened 2016-05-06
Some elliptic curves over number fields crash when computing plots


JohnCremona:
EllipticCurve/3.3.1016.1/9.2/b/3
causes
raise RuntimeError("No horizontal tangent lines on bounded component")
from line 74 in WebEllipticCurve.py in function EC_nf_plot

(I found this in the server log /home/lmfdb/lmfdb-git-prod/flasklog)

AndrewVSutherland:
@nmascot are you working on this?  If not I will look at it.

AndrewVSutherland:
The problem is due to sage raising the error "No horizontal tangent lines on bounded component" in the call to e.plot in the function EC_nf_plot in line 74 of WebEllipticCurve.py.  I will add a try except to work around this (so this curve will not have a plot)

JohnCremona:
@roed314 you wrote that error message -- can you help?

AndrewVSutherland:
@JohnCremona I just issued pull request #1218 to add a try/except to prevent the crash, which will at least make this a non-critical issue.

Can you review and (assuming it works) merge (FYI, I did run test.sh lmfdb/ecnf)?

JohnCremona:
Done.  Will push to beta & prod now (though there will be a delay until the
prod code gets picked up by the cloud).

JohnCremona:
Fixed on beta and prod@warwick.  Can be closed once cloud has it too and www.lmfdb.org/EllipticCurve/3.3.1016.1/9.2/b/3
works without crashing

jvoight:
Works in the cloud without crashing, displays "unable to plot".

roed314:
I'm not going to get a chance to look at this until Monday, but I can once
I get to AIM. And it sounds like it might already be resolved....

AndrewVSutherland:
@roed314 The only thing that remains unresolved is whether the error "No horizontal tangent lines on bounded component" being raised by the plot function is expected behaviour (so we really cannot plot this curve), or not.  If you can confirm that it is expected behaviour then I'm happy to close.

jvoight:
I just pinged @nmascot by email to see what he thinks.  JV

nmascot:
Hi,

I've only just received @jvoight 's ping.

I have to catch a plane early tomorrow (well, later today in fact); I'll have a closer look at it as soon as I can (= tomorrow afternoon).

[...]

PR #1218, opened 2016-05-07
Fixes critical issue #1206, crash in ecnf plot

AndrewVSutherland:
Adds a try/except to catch an apparent bug in the sage elliptic curve plot function.


JohnCremona:
Code looks good, testing now.

Issue #1217, opened 2016-05-07
Modular form error


jwj61:
From Brian via the feedback page:

"From the page ""Newforms of weight 9 for Gamma_1(23)""
http://www.lmfdb.org/ModularForm/GL2/Q/holomorphic/23/9/?group=1

if I click on S_9(\chi_{23}(22,\cdot))

I get the error ""The page you are looking for has produced a server error."""

When I tried this, I didn't see that I could click on S_9(...), but clicked on the first form 23.9.22.a and got an error.

sanni85:
The page for /23/9/22/ works for me.

On the same page I tried to click on a grey link (/23/9/2/) and I got a page which says:
"Unfortunately, we do not have this space in the database, yet."
but below the table is written:
"Spaces in grey are trivial because the character and the weight have opposite parity."

AndrewVSutherland:
The problem is http://www.lmfdb.org/ModularForm/GL2/Q/holomorphic/23/9/22/a/.  It is crashing while trying to display an error message, looking at it now.

JohnCremona:
At ModularForm/GL2/Q/holomorphic/23/9/22/ form a is displayed but in the Dimension column it has 0.  Should be 1.

AndrewVSutherland:
It's crashing because max_cn is not defined in line 12 of emf_web_newform_post_load_scripts.html where it tries to compute max_cn-1 while formatting an error message.  If you fix this it then says that the space is empty, which it obviously is not (like JC said, the dimension should be 1 not 0)

JohnCremona:
Where is it getting that dimension from?  @fredstro or @sehlen ?

AndrewVSutherland:
@JohnCremona Not from the database, there is only one record in modularform2.webnewforms with parent 23.9.22, and it has hecke_orbit_label 23.9.22.'b', there is no record with hecke_orbit_label label '23.9.22.a'.

It also seems strange that the 15-dimensional Galois orbit S_9^new(Gamma_0(23),chi_23(22,.)) is decomposing as 0+12.  Even if it were 1+12 there would still be something wrong here.

sehlen:
> On May 7, 2016, at 15:32, AndrewVSutherland <notifications@github.com> wrote:
> 
[...]

PR #1219, opened 2016-05-07
Show an error message that the Hecke orbit is not available if the db record is missing.

sehlen:


sehlen:
This is part of a fix for #1217

AndrewVSutherland:
Reviewing now.

AndrewVSutherland:
So does what we want for ModularForm/GL2/Q/holomorphic/23/9/22/, i.e. it notes that the Hecke orbit 23.9.22.a is not available.

But it does something different for the similar situation for ModularForm/GL2/Q/holomorphic/11/11/10/, where 11.11.10.b is present but 11.11.10.a is missing (and there are dozens more like this).  It says that the whole space is missing (whereas the current code on www.lmfdb.org displays the space gets 11.11.10.a wrong).  If we are going to throw out the whole space 11.11.10, then we should disable the link on 11.11 (i.e. grey it out).

sehlen:
It doesn’t do that for me…

I get for ModularForm/GL2/Q/holomorphic/11/11/10/ that 11.11.10.a is not clickable and it shows the message as it should.
I don’t understand how it could possibly do anything else (look at the code).
How did you try this? Could it be a caching issue?

(I am directly connected to warwick for mongo and am running the website locally on my laptop.)

Note that the underlying issue is a data issue.
It is hard to track that back for me because Fredrik did the computations but it is probably not very hard for me to try to recompute missing guys.

sehlen:
Sorry, I was stupid so I had to update this again to make sure the number fields really are not displayed.

JohnCremona:
I am checking (after the 3rd commit!)

fredstro:
And of course the whole 'rational coefficients ' business is because the
lmfdb_label for the number field fails to update so the default 1.1.1.1 is
used..
Fredrik

sehlen:
@fredstro Yes, I know but this is not really an issue here. The dimension was shown to be 0 and this is the default in this case. If for whatever reason this happens (that there's no db record) then we should show a reasonable error message and only block navigation to the problematic records, not to the whole space. That's what I did now.

JohnCremona:
I tried about a dozen of the problem cases in Drew's list and all don't crash, instead displaying (in some cases many times) the message "This Hecke orbit is unfortunately not available." properly in the table.

I think that is what this was supposed to fix until the missing data is inserted.  So I will merge this now but we should not close the issue until the missing data is in there.

fredstro:
@sehlen i see this reply got into the wrong thread.... I'm sorry it's a
bumpy road where we are driving now....
We should probably look over the general strategy and make sure that we
dint assign default values when we can't get stuff from the database.

sehlen:
@fredstro It took me a while to realize you are actually referring to a bumpy road you're driving on and didn't mean this as a metaphor for the lmfdb project ;-)

sehlen:
Thanks @JohnCremona for reviewing this.

fredstro:
@sehlen it was intentional :)

JohnCremona:
I took the bumpy road to be metaphorical too!

AndrewVSutherland:
:)

Issue #1215, opened 2016-05-07
Classical modular forms test_non_triv_character fails


AndrewVSutherland:
The function test_non_triv_character in lmfdb/modular_forms/elliptic_modular_forms/test_emf.py fails at line 77.

jvoight:
Is this critical?

AndrewVSutherland:
I doubt it, I just noticed that it failed when I ran the test script and wanted to flag it.  I haven't looked at it closely.

JohnCremona:
All tests in test_emf pass for me (with PR#1222 on board).

AndrewVSutherland:
Great!

AndrewVSutherland:
Closing.

PR #1222, opened 2016-05-07
Fix for critical issue #1207.

sehlen:
This should now *really* fix https://github.com/LMFDB/lmfdb/issues/1207 and related issues I discovered when looking at it. I have to say, we should probably trace back where all of these issues came from because I'm pretty sure the code was a lot better (even better than my fix now, I think) back in December when the pages got the last big overhaul. @sanni85 put a lot of work into this and it was thoroughly tested and now I see all these problems and the logic did just not do what it was supposed to do anymore.

### DIFF
--- a/lmfdb/modular_forms/elliptic_modular_forms/backend/web_object.py
+++ b/lmfdb/modular_forms/elliptic_modular_forms/backend/web_object.py
@@ -859,7 +859,8 @@ class WebNumberField(WebDict):
             if hasattr(self._value,'absolute_polynomial'):
                 setattr(self._value, "lmfdb_pretty", web_latex_split_on_pm(self._value.absolute_polynomial()))
             elif self._value.absolute_degree()==1:
-                setattr(self._value, "lmfdb_pretty", "1.1.1.1")
+                setattr(self._value, "lmfdb_pretty", field_pretty('1.1.1.1'))
+                setattr(self._value, "lmfdb_label", '1.1.1.1')
             else:
                 emf_logger.critical("could not set lmfdb_pretty for the label")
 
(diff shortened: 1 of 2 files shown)


sehlen:
Sorry for the long title in the beginning. I was using hub (command line tool for github) to issue the pull request and apparently forgot how to use it.

sehlen:
OK, sorry for blaming others... in fact it seems there were no changes to @sanni85's code after December that caused these problems so it is probably all his fault ;-) Anyway, I think my commits should do the job but someone should probably check many cases...

sehlen:
Example page: http://www.lmfdb.org/ModularForm/GL2/Q/holomorphic/49/8/?group=0

sehlen:
Please note that I added another commit because the ad-hoc fix by @AndrewVSutherland unfortunately breaks some modular forms pages, e.g. /ModularForm/GL2/Q/holomorphic/20/11/17/.

sehlen:
Maybe @AndrewVSutherland and @JohnCremona could review this as this really fixes a lot of issues at once, I think - so it would be good to have it for 1.0.

JohnCremona:
Will look in a minute

jvoight:
@AndrewVSutherland has had his hands on this, so I would like for him to give an OK.  (But just let me know if should review and pull.)

JohnCremona:
Some pointers at what to look at would be useful -- I have this branch running right now.  the systematic test has reached level 33 but is only testing that these pages load, not what their content is.

AndrewVSutherland:
I'm back at MIT and can tack a look at this as well.  @sehlen did you also look at https://github.com/LMFDB/lmfdb/issues/1215?  It may be unrelated but it would be nice to have all the modular forms tests passing.

JohnCremona:
I have not been able to verify that the changed logic in the code is correct since it is not simple and I never looked at it before, but the test pass and the spot checks I did look good.

Why don't we merge this now as we can always do more...?

AndrewVSutherland:
Sounds good, please merge and push and I can test against production.

JohnCremona:
Done and pushed -- test on beta or lmfdb.warwick.ac.uk pending the cloud server picking up the new prod branch.

PR #1226, opened 2016-05-07
Adjust bound for Artin L-function bound link

jwj61:
This works for me.  It could be expanded in some higher dimensional cases.


davidfarmer:
I was going to express skepticism that the automated calculation could handle the conductor
that large, but maybe the decreased range for zeros and plotting helps a lot.  And being
primitive helps a lot, too.

No need to change it, but if some problems appear, we should knock off a couple zeros.

jwj61:
Agreed.  For the record, here is a bit more detail.

In dimension 2, the cutoff in the L-function code seemed good enough.  This is beyond that, but it just means the person gets the L-function page with no plot and no zeros (hence, no problems).

I found problems in dimensions 3 and 4, and the bound was essentially fit to match those dimensions.

There were some higher dimensional ones which looked ok, but with a bound of this form, we don't see them.

Issue #1199, opened 2016-05-06
elliptic curve pages are slow


jwbober:
Every request for the url www.lmfdb.org/EllipticCurve/Q/1001.b/ results in two database queries (copied from the mongodb log):

`2016-05-06T15:09:51.046+0100 I COMMAND  [conn57] command elliptic_curves.$cmd command: count { count: "nfcurves", fields: null, query: { conductor_label: "1001.b" } } planSummary: COLLSCAN keyUpdates:0 writeConflicts:0 numYields:2498 reslen:44 locks:{ Global: { acquireCount: { r: 4998 } }, MMAPV1Journal: { acquireCount: { r: 2499 } }, Database: { acquireCount: { r: 2499 } }, Collection: { acquireCount: { R: 2499 } } } 287ms`
`2016-05-06T15:09:51.646+0100 I QUERY    [conn57] query elliptic_curves.nfcurves query: { $query: { conductor_label: "1001.b" }, $orderby: { field_label: 1, conductor_norm: 1, conductor_label: 1, iso_nlabel: 1, number: 1 } } planSummary: COLLSCAN, COLLSCAN ntoreturn:50 ntoskip:0 nscanned:0 nscannedObjects:639502 scanAndOrder:1 keyUpdates:0 writeConflicts:0 numYields:4996 nreturned:0 reslen:20 locks:{ Global: { acquireCount: { r: 9994 } }, MMAPV1Journal: { acquireCount: { r: 4997 } }, Database: { acquireCount: { r: 4997 } }, Collection: { acquireCount: { R: 4997 } } } 598ms`

(I don't know why this is happening.)

JohnCremona:
If you out in a URL like EllipticCurve/Q/1001.b it does a database query to see if the label exists and if it does then calls another function which gets the data.  This has been true for a very long time and similar things happen all over the place, e.g. HMFs where I tried recoding to avoid this and the results was a lot worse.

I have removed the critical label since no-one has been at all bothered by this for years.

jwbober:
I think you misunderstand the queries that I copy-and-pasted. It is doing
number field queries, looking for whether the elliptic curve label it is
looking for us also the label of the conductor of an elliptic curve over a
number field. Also, it is counting elliptic curves over number fields.

These things are really slow, and it makes no sense that they are
happening.

JohnCremona:
[...]

PR #1243, opened 2016-05-08
Minimal change, trying to speed up /EllipticCurve/Q

jvoight:
Currently /EllipticCurve/Q is slow to load.  Right now, it calls "counts" in order to get ranges for browse.  I hard coded these.  I have slightly improved performance on my local copy, but it's hard to tell if this will make a difference when it goes to the cloud/Warwick.  I'd like to try, and I'm doing this in stages with the minimal change here.  

If this doesn't work, we could then try replacing this code which computes something with instead hard coded lines in the template.  (For example, what is happening in the function zip?)  Since this page is going to be loaded a zillion times, we really shouldn't be doing anything nontrivial.

### DIFF
--- a/lmfdb/elliptic_curves/elliptic_curve.py
+++ b/lmfdb/elliptic_curves/elliptic_curve.py
@@ -89,18 +89,19 @@ def rational_elliptic_curves(err_args=None):
             for field in ['conductor', 'jinv', 'torsion', 'rank', 'sha', 'optimal', 'torsion_structure', 'msg']:
                 err_args[field] = ''
             err_args['count'] = '100'
-    counts = get_stats().counts()
+    # counts = get_stats().counts()
 
-    conductor_list_endpoints = [1, 100, 1000, 10000, 100000, counts['max_N'] + 1]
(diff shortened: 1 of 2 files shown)


jvoight:
See issue #1199.

jvoight:
I'm somewhat skeptical now that this will make a real difference, because HMF pages use counts and they're plenty zippy.

jwbober:
I thought think these counts may be cached, so that they only result in one
database query per webserver process. Then it only takes a while
occasionally.

This is different than #1199, which is really an improper url that I
accidentally accessed.

I'm somewhat skeptical now that this will make a real difference, because
HMF pages use counts and they're plenty zippy.

—
You are receiving this because you are subscribed to this thread.
Reply to this email directly or view it on GitHub
<https://github.com/LMFDB/lmfdb/pull/1243#issuecomment-217734479>

jvoight:
I'm going to close this since I don't think it will help--there's something else going on.

JohnCremona:
See my comment on another ticket. Solved by adding a new index.

jvoight:
Yup, was there with @AndrewVSutherland helping him by pointing out all of the dead ends.  :)

PR #1244, opened 2016-05-08
ecnfplot

nmascot:
Home-brewed code to plot elliptic curves over number fields. This code should be capable of coping with the ridiculously large (or tiny) real coefficients that sometimes result from embedding the ground field into R (cf #1206 ).

Please test, and also (if you like) comment on how suitable the values of xmin, xmax, ymin, ymax chosen by the code are -- this can be easily tweaked.


nmascot:
This is meant to replace #1218, however maybe #1218 should be kept for the official release as it is safer.

AndrewVSutherland:
@nmascot I agree that keeping #1218 for the official release is safer.  I'm sure you wrote your code carefully, but it is not a one line change.  I would feel a lot less nervous about it if you kept the try/except (wrapped around the call to your new plot function) with a fallback to "unable to plot" if anything goes wrong (which of course should never happen).

I'll let others weigh in with their opinion, but my understanding is that the current "unable to display plot" message will appear only for a small number of curves and in my view it probably makes more sense to wait until some time after the release.  But if we were to go ahead I would definitely ask you to modify your pull request to include a try/except wrapper (this might be prudent anyway, we should never trigger an internal server error just because we can't plot a picture).

nmascot:
@AndrewVSutherland I completely agree, in fact I wanted to wrap everything in a try / except block but I forgot. I have fixed this in a new commit.

nmascot:
A few nice examples of plots :
/3.1.23.1/115.1/A/1
/5.5.14641.1/23.5/a/3
/6.6.371293.1/79.1/c/4

AndrewVSutherland:
@nmascot Thanks!  I like your examples, but I notice that these examples all plot OK (perhaps not as nicely) with the current code.  Do you know of any besides /3.3.1016.1/9.2/b/3 that cannot be plotted with the current code but can be plotted with your change?  I ask only because this might impact the decision of whether to merge your pull request today.  My vote is to wait.

nmascot:
@AndrewVSutherland No, I did not know of any example of curve which could not be plotted with the old code until /3.3.1016.1/9.2/b/3 was reported a few days ago.
I expect that curves whose coefficients specialise to huge enough real numbers under at least one embedding will also crash the old code, but I have not been able to find examples because of #1250.

AndrewVSutherland:
#1250 is much more important than this issue.  Let's shelve this for the moment and focus on the more critical issues.

AndrewVSutherland:
I will do some more testing on this today and, assuming all is well, will merge to beta, test some more, and then merge to prod (unless anyone objects).

JohnCremona:
I do not object.  I have not looked at the code but the old failing one works (however much  zoom the curves still look singular which might explain why it was a hard case).  The degree 5 and 6 pictures are stunning!

Issue #1250, opened 2016-05-09
Ecnf search results sets include_cm_curves and include_Q_curves to only by default


nmascot:
For example, over 3.3.1016.1, there are curves of conductor norm 4 (as van be seen on /EllipticCurve/3.3.1016.1/), but searching for conductor norm 4 does not find them (it does not crash, it just finds nothing) : /EllipticCurve/3.3.1016.1/?start=0&count=50&field=3.3.1016.1&conductor_norm=4&include_isogenous=on&include_base_change=on&include_Q_curves=only&include_cm=only&torsion=&torsion_structure=&jinv=

nmascot:
Actually, even without the conductor norm condition, no curves are found : http://localhost:37777/EllipticCurve/3.3.1016.1/?start=0&count=50&field=3.3.1016.1&conductor_norm=&include_isogenous=on&include_base_change=on&include_Q_curves=only&include_cm=only&torsion=&torsion_structure=&jinv=

On the other hand, the curves over 3.3.49.1 are found.

AndrewVSutherland:
Search by conductor norm does work, see

http://www.lmfdb.org/EllipticCurve/3.3.1016.1/?start=0&count=50&field=3.3.1016.1&conductor_norm=4&include_isogenous=on&include_base_change=on&include_Q_curves=include&include_cm=include&torsion=&torsion_structure=&jinv=

The problem is that your query had include_Q_curves and include_cm set to "only".  I am looking into why this happened now.

AndrewVSutherland:
OK, problem identified (issue relabled accordingly).  Working on the easy fix now.

JohnCremona:
I'm awake and will look into this this morning PT.  It must be my fault.
John

AndrewVSutherland:
@JohnCremona The problem is in ecnf-search-results.html which sets both include_Q_curves and include_cm to "only" by default.  See lines 55-88.  I'm happy to leave it to you (but let me know if you want me to make the change).

JohnCremona:
I could, and probably should, since I wrote that.  I am looking at it know and will leave it to you since my logic after a 2-hour journey is clearly no better than it was when I wrote that code.

It is not a bug when manually entered URLs are messed up.  When I search for 3.3.1016.1, norm 4, I see results and also see the include/exclude boxes correctly set.  So I have not actually yet understood what the bug actually is.

AndrewVSutherland:
[...]

PR #1251, opened 2016-05-09
Fix to issue #1250 plus optimization

AndrewVSutherland:
This addresses issue #1250 and improves performance of field links on EllipticCurves/

To test the fix to #1250, go to www.lmfdb.org and click on one of the fields listed at the top of the page to go to http://www.lmfdb.org/EllipticCurve/3.3.1957.1/, for example.  You will see that both Q-curves and CM curves are set to only.

But with the pull request applied, e.g. go to http://127.0.0.1:37777/EllipticCurve/3.3.1957.1/, you should see that both Q-curves and CM curves are set to include.

Additionally, in order to improve the performance of these pages, the currently displayed statistics about the number of curves, isogeny classes, and conductors that occur over the number field are suppressed.  This reduce the page load time from >5s to under 0.5s.  We can put the statistics back in once they are stored in the database as discussed in https://github.com/LMFDB/lmfdb/issues/1247

### DIFF
--- a/lmfdb/ecnf/main.py
+++ b/lmfdb/ecnf/main.py
@@ -255,8 +255,9 @@ def show_ecnf1(nf):
     info['more'] = int(start + count < nres)
     info['field_pretty'] = field_pretty
     info['web_ainvs'] = web_ainvs
-    if nf_label:
-        info['stats'] = ecnf_field_summary(nf_label)
+    #don't risk recomputing all the ecnf stats just to show curves for a single number field
+    #if nf_label:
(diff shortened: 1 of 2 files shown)


JohnCremona:
Looking at this now...

JohnCremona:
OK, thanks.   It's the order in which the 3 options are looked at which matters, with the default ("include") left to last instead of being first.

On the stats, what I did for the Hilbert MFs (in order to make things faster) was -- apart from storing counts in a yaml file -- to allow a call to field_summary to return the counts for that field instead of doing all fields only to extract a couple of items.

I'll merge & push now.

Issue #1229, opened 2016-05-08
Error parsing "complex numbers" for L/Character/Dirichlet/*/*/


edgarcosta:
***1 Upvote*** Here are some tracebacks:

https://gist.github.com/edgarcosta/c9642a9940b40e3681654e4d319fd47f

jwbober:
Where are you getting these error? Those pages, e.g. www.lmfdb.org/L/Character/Dirichlet/37/25/ seem to work fine.

davidfarmer:
and I tried 
/L/Character/Dirichlet/19/16/
and it also worked fine

edgarcosta:
I looking into the access log and it has to do with downloading the Lcalcfile.
```
130.211.0.250 - - [06/May/2016:21:40:34 +0000] "GET /L/Character/Dirichlet/37/25/?download=lcalcfile HTTP/1.1" 500 10010 "-" "Mozilla/5.0 (Linux; Android 6.0.1; Nexus 5X Build/MMB29P) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/41.0.2272.96 Mobile Safari/537.36 (compatible; Googlebot/2.1; +http://www.google.com/bot.html)"
130.211.1.105 - - [07/May/2016:05:14:31 +0000] "GET /L/Character/Dirichlet/37/25/?download=lcalcfile HTTP/1.1" 500 10010 "-" "Mozilla/5.0 (Linux; Android 6.0.1; Nexus 5X Build/MMB29P) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/41.0.2272.96 Mobile Safari/537.36 (compatible; Googlebot/2.1; +http://www.google.com/bot.html)"
```

davidfarmer:
The lcalc file is a way to download the data for that object.

We now have better ways, so we should consider hiding the lcalc links.
(And then maybe re-implementing them once we completely move
to taking the L-function data from the database.)

jvoight:
+1 for hiding Lcalc files

Issue #1232, opened 2016-05-08
L/Zeros/ModularForm/GL2/Q/holomorphic


edgarcosta:
Here are the tracebacks:

https://gist.github.com/edgarcosta/aab04be528490baeba3bb43c99494c95

Further, it still uses quite a lot of resources before quitting.

jvoight:
I think the issue is that the data is wrong, and so lcalc is going berserk.  I am not labeling this as critical because we know the classical modular forms are seriously broken, and so we should revisit this issue after.

davidfarmer:
This will be fixed when issue #439 is addressed.

I am working on a temporary solution.

davidfarmer:
closing because the temporary solution is adequate for now, and issue  #439 will fix properly

PR #1253, opened 2016-05-09
Hide lcalc download for Dirichlet L-function

davidfarmer:
Fixes issue #1229


edgarcosta:
Does it also fixes #1232?

JohnCremona:
@edgarcosta I doubt it since the only effect is that Dirichlet L-functions now do not offer an lcalc file.

davidfarmer:
That is a completely different problem, but I am working to fix it
by hiding the zeros and plot in certain cases (which is better than
hiding the link to the L-function).

Issue #1258, opened 2016-05-09
Factoring j EllipticCurve/6.6.371293.1/1.1/a/{1,4}


edgarcosta:
This seem to fail to load most of the times, and when they load it takes forever
http://www.lmfdb.org/EllipticCurve/6.6.371293.1/1.1/a/1
http://www.lmfdb.org/EllipticCurve/6.6.371293.1/1.1/a/4

Here are the tracebacks:

```
Exception on /EllipticCurve/6.6.371293.1/1.1/a/4 [GET]
Traceback (most recent call last):
  File "/home/sage/sage-7.1/local/lib/python2.7/site-packages/Flask-0.10.1-py2.7.egg/flask/app.py", line 1817, in wsgi_app
    response = self.full_dispatch_request()
  File "/home/sage/sage-7.1/local/lib/python2.7/site-packages/Flask-0.10.1-py2.7.egg/flask/app.py", line 1477, in full_dispatch_request
    rv = self.handle_user_exception(e)
  File "/home/sage/sage-7.1/local/lib/python2.7/site-packages/Flask-0.10.1-py2.7.egg/flask/app.py", line 1381, in handle_user_exception
    reraise(exc_type, exc_value, tb)
  File "/home/sage/sage-7.1/local/lib/python2.7/site-packages/Flask-0.10.1-py2.7.egg/flask/app.py", line 1475, in full_dispatch_request
    rv = self.dispatch_request()
  File "/home/sage/sage-7.1/local/lib/python2.7/site-packages/Flask-0.10.1-py2.7.egg/flask/app.py", line 1461, in dispatch_request
    return self.view_functions[rule.endpoint](**req.view_args)
  File "/home/lmfdb/lmfdb-git-prod/lmfdb/ecnf/main.py", line 313, in show_ecnf
    ec = ECNF.by_label(label)
  File "/home/lmfdb/lmfdb-git-prod/lmfdb/ecnf/WebEllipticCurve.py", line 125, in by_label
    return ECNF(data)
  File "/home/lmfdb/lmfdb-git-prod/lmfdb/ecnf/WebEllipticCurve.py", line 116, in __init__
    self.make_E()
  File "/home/lmfdb/lmfdb-git-prod/lmfdb/ecnf/WebEllipticCurve.py", line 207, in make_E
    self.fact_j = web_latex(j.factor())
[...]

PR #1262, opened 2016-05-09
don't factor j-invariants over number fields

JohnCremona:
See issue #1258  -- no factorization of j-invariants for elliptic curves over number fields is done.

There should not be a similar problems for discriminants since these have very limited support, while the numerator of j is much more random.

I hoped for some discussion of this over at #1258  but soemone can vote this up my merging :)


AndrewVSutherland:
Looks good.  I can now load http://127.0.0.1:37777/EllipticCurve/6.6.371293.1/1.1/a/4, whereas I currently cannot load http://lmfdb.org/EllipticCurve/6.6.371293.1/1.1/a/4.

PR #1264, opened 2016-05-09
hide some zeros and plots

davidfarmer:
Don't show the zeros or plots for L-functions of holomorphic cusp forms, when the
weight or level are not small.

Fixes issue #1232


jwj61:
Will include this in a fixed version

PR #1268, opened 2016-05-09
Don't make code snippets (sage, magma, pari) for number fields unless needed

jwj61:
This was slowing down other pages, like a search results page.


JohnCremona:
I am looking at this and will soon make a similar PR for elliptic curves

PR #1274, opened 2016-05-10
update my affiliation

sehlen:


jvoight:
Stephan, this modifies 2 files, including a line in emf_render_navigation.py.  Please separate them out (or explain why the latter is needed).

sehlen:
I'm sorry for this.
I wanted to isolate it and on purpose cherry-picked a commit that only changed my affiliation (so I thought) but unfortunately this change came through with it. However, the change does not affect the webpages (as I learned tracing back the modular forms pages navigation) because it is unreachable (unconditional return in the beginning of the function). This obviously has to be cleaned up but at least my mistake does certainly not cause any harm!

Issue #1230, opened 2016-05-08
EllipticCurve/Q/modular_form_display/*


edgarcosta:
Here are some tracebacks:

https://gist.github.com/edgarcosta/c00d5da053a8b525f56b5031af4a24ff

jvoight:
This looks confusion between elliptic curves over QQ and elliptic curves over number fields, because it is trying to format "Elliptic Curve isogeny class %s over Number Field %s"

JohnCremona:
I don't know where that URL came from.  The confusion is that the URL /EllipticCurve/Q/modular_form_display/90.b1/ is being not handled by the ec blueprint with prefix EllipticCurve/Q with 2 other parameters, as it should, but as /EllipticCurve with 3 other parameters.  I am trying to fix it.

edgarcosta:
@JohnCremona probably a web crawler

PR #1277, opened 2016-05-10
fixing ec modular form display error

JohnCremona:
Fix for #1230 
After this I don't see how the problematical URL EllipticCurve/Q/modular_form_display/90.b  could ever be created (it works infe if you add /10 to that URL, this 10 denoting the number of coefficients, which is the default anyway.)

I also removed a redundant javascript function in show-ecnf-isoclass.html since it has no function and is left over from the elliptic curves (/Q) equivalent template.

### DIFF
--- a/lmfdb/ecnf/templates/show-ecnf-isoclass.html
+++ b/lmfdb/ecnf/templates/show-ecnf-isoclass.html
@@ -76,22 +76,6 @@ Rank not yet determined.
 <p>Not available.</p>
 {% endif %}
 
-<script type="text/javascript">
-var number_of_coefficients = 10;
-function more_handler(evt) {
-    number_of_coefficients += number_of_coefficients;
-    evt.preventDefault();
-    $("#modform_output").load("/EllipticCurve/Q/modular_form_display/{{info.lmfdb_label}}/"+number_of_coefficients,
-        function() {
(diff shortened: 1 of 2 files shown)


AndrewVSutherland:
I note that http://127.0.0.1:37777/EllipticCurve/Q/modular_form_display/90.b1/ still throws a server error, failing on line 290 of show_ecnf_isoclass (no attribute 'field'). If I add 10 to the end I get an html page with just:

\(q \) \(\mathstrut +\mathstrut q^{2} \) \(\mathstrut +\mathstrut q^{4} \) \(\mathstrut -\mathstrut q^{5} \) \(\mathstrut +\mathstrut 2q^{7} \) \(\mathstrut +\mathstrut q^{8} \) \(\mathstrut +\mathstrut O(q^{10}) \)

is this what you would expect?

JohnCremona:
Yes, as expected.  This is a little helper which populates the q-expansion part of an elliptic curve home page and handles when a user asks for more coeffs.  It will only be used in that context, so should never be used with the final number missing.

AndrewVSutherland:
OK if you are happy, I'm happy to merge, the code looks safe enough.

JohnCremona:
Thanks -- I still have no idea where that malformed URL came from in Edgar's logs!

PR #1279, opened 2016-05-10
Addded st_group knowl to roadmap

AndrewVSutherland:
This should be merged and pushed so that I can add information to the st_group.roadmap knowl.


JohnCremona:
Done.  Use beta.lmfdb.org to do the knowl editing.

PR #1281, opened 2016-05-10
Cleanup

sehlen:
This contains a bunch of commits that mostly clean up some code and remove old code that isn't use.
One exception to this is a commit that modifies lmfdb/modular_forms/elliptic_modular_forms/backend/web_newforms.py. It is a small change that I did a while a go that somehow did not make it into the lmfdb repository but does not cause any harm. It just changes the way the database is queried in one instance which is safer than what is currently used.


sehlen:
All tests pass for me.

PR #1287, opened 2016-05-11
Add news link to sidebar

AndrewVSutherland:
Added a "News" item in the "Introduction and More" section of the sidebar that links to the knowl doc.news.in_the_news that lists recent news articles that reference the LMFDB  (in the future it could link to an intermediate page that might also contain announcements, information about workshops, etc...).

This addresses a suggestion of Terry Tao.

### DIFF
--- a/lmfdb/pages.py
+++ b/lmfdb/pages.py
@@ -149,6 +149,12 @@ def roadmap():
     b = [(t, url_for('roadmap'))]
     return render_template('roadmap.html', title=t, body_class=_bc, bread=b)
 
+@app.route("/news")
+def news():
+    t = "News"
+    b = [(t, url_for('news'))]
+    return render_template(_single_knowl, title="LMFDB in the news", kid='doc.news.in_the_news', body_class=_bc, bread=b)
+
 ## INTRO PAGES END
 
 @app.route('/Variety')
--- a/lmfdb/sidebar.yaml
+++ b/lmfdb/sidebar.yaml
@@ -13,6 +13,8 @@
       url_for: "universe"
     - title: Future Plans
       url_for: "roadmap"
+    - title: News
+      url_for: "news"
 
 2L-functions:
    type: L


JohnCremona:
Right now I am not sure that we need this, though it is certainly nice to have one easy place to go for all those new reports, but in 3 months time when none of these are "new" it might not look so good.  I prefer the idea of linking to a page with sections (perhaps each a knowl as this one is) in which this list would be headed something like "News items about the LMFDB", another could hold my citations list, and others could have announcements as you suggest.

More feedback needed -- what do others think?

davidfarmer:
I'd like to see the news link on the public site right away.

The knowl with the material is doc.news.in_the_news.  I set it up that way so that we can put
something useful in doc.news, which can be shown when there is a lull in the news stories.

Perhaps the current page should be in 4 categories, where we split "News articles" into
"Press releases" and "News articles", where the former were issued by our home institutions.

There are several stories not yet on our list:
https://www.google.com/search?q=lmfdb&ie=utf-8&oe=utf-8#q=lmfdb&safe=off&tbm=nws

AndrewVSutherland:
Adding more now.  There are several from today (May 11) that are 
"secondary reports", not just direct copies of our press releases, which 
is good.

Agree that it can be organized better, I will try to separate out the 
press release from the "news articles".  But there is no reason not to 
push the sidebar pull request now.  The knowl is in a reasonable state 
for viewing and I will be careful not to break it.

Drew

davidfarmer:
I didn't mean to suggest we had to reorganize the knowl
before posting.

I think it is clear enough that we need to do this, so let's
see if I understood the new way to update the live code...

AndrewVSutherland:
@davidfarmer I'm also happy to let you take over editing the knowl if you want,we just need to coordinate to make sure we aren't overwriting eachother's edits.

davidfarmer:
I think I did the following (someone please check):

1) Merged Drew's pull request into master.

2) Made a pull request to prod from master (which I got backward
the first time), and then accepted it.

3) Made a pull request to beta from prod, and then accepted it.

A couple things that seemed odd:

For step 2), after I did the first part it instantly put me on
the page for the second part.  But for 3), after I did the first part
I had to click to pull requests to to the second part.

After doing 3) it said that the prod branch could now be safely
deleted! (I did not believe them.)

AndrewVSutherland:
The link is now visible on lmfdb.org and beta.lmfdb.org, but the lmfdb.org knowl is not current!

AndrewVSutherland:
Never mind, it is now.

PR #1290, opened 2016-05-11
Added absolute path to mongoclient configuration path

fredstro:
The current configuration only reads the configuration from mongoclient.config if the script is started from the same directory as the config file. 
To be able to change the mongoclient configuration when either the startup script is run from another directory or lmfdb is imported as a module (something I often do on the sage command line) I added an absolute path to mongo_client_config_filename in website.py 

I know that there are a multitude of ways to obtain path names in python, if my way is in anyway "bad" feel free to replace it with another method.
For reference see also pull requests #1078  #1074

### DIFF
--- a/lmfdb/website.py
+++ b/lmfdb/website.py
@@ -298,7 +298,7 @@ def get_configuration():
     replicaset =
     read_preference = NEAREST
     """
-    config_dir = '/'.join( os.path.dirname(__file__).split('/')[0:-1])
+    config_dir = '/'.join( os.path.dirname(os.path.abspath(__file__)).split('/')[0:-1])
     mongo_client_config_filename = '{0}/{1}'.format(config_dir,mongo_client_config_filename)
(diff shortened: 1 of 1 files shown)


edgarcosta:
Perhaps this is a better? 
os.path.dirname(os.path.abspath(__file__))

fredstro:
Yes, it is probably safer. 
Thanks!

JohnCremona:
The new version works for me but then I always start the server from the lmfdb root directory.  So I stated it from other places and it still works.  Good!
I will merge into master but not bother doing to so beta and prod.

edgarcosta:
Perfect.
Let me know when it gets to beta, so I can test it out.

Best,
Edgar

JohnCremona:
You can already test it using master -- master is still the main development branch.  I could push to either beta and/or prod but that seemed unnecessary since developers all work off master.

PR #1301, opened 2016-05-11
use jinv to speedup minimal twist lookup

AndrewVSutherland:
Currently elliptic curve lookups spend 2-3 seconds finding the minimal twist using a query on the unindexed field 'ainvs'.  This pull request adds 'jinv' to the query, which makes it essentially instantaneous (this is also faster than putting an index on ainvs).

This pull should be merged to master, prod, and beta asap, as it will substantially speed up all elliptic curve pages.

### DIFF
--- a/lmfdb/elliptic_curves/web_ec.py
+++ b/lmfdb/elliptic_curves/web_ec.py
@@ -216,7 +216,7 @@ class WebEC(object):
             data['minq_info'] = '(itself)'
         else:
             minq_ainvs = [str(c) for c in minq.ainvs()]
-            data['minq_label'] = db_ec().find_one({'ainvs': minq_ainvs})['lmfdb_label']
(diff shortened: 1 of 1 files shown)


JohnCremona:
Great idea -- I probably wrote the minimal twist lookup before we stored j-invariants.

AndrewVSutherland:
I see that this got merged to prod a few minutes ago but doesn't appear to be live yet, it should take 15 minutes or so, right?

JohnCremona:
Less since it is now 14.59 so wait 1 more minute!

AndrewVSutherland:
I can tell that it is now there because the TTFB  times just dropped to under 1s:

curl -silent -o /dev/null -w "Connect: %{time_connect} TTFB: %{time_starttransfer} Total time: %{time_total} \n" lmfdb.org/EllipticCurve/Q/2450/y/1
Connect: 0.025 TTFB: 0.649 Total time: 0.726

PR #1306, opened 2016-05-11
Cleaned up color.css just a bit more.

adeines:


JohnCremona:
I think has changed some sidebar colours in a way not discussed.  Was that intentional?
@alinabucur noticed this.

adeines:
Nope.  That wasn't on purpose.  Which sidebar colors?  I could get to this
tonight/this weekend (and in the meantime you could revert back.)

I was mostly trying to separate some colors so Holly could make a color
scheme with better contrast and was hopefully not actually changing
anything . . .

davidfarmer:
The changes I noticed in the sidebar colors were all corrections.  The word "elliptic",
for example, is now black, meaning that it is not a link.  The /Q below it is green,
meaning that is the link to follow.

Unless I have missed something, all is well with the sidebar colors.

adeines:
Awesome!  (Yes, Holly had me set all non-links to black.)

Issue #1311, opened 2016-05-12
CM set to "only" when browsing Elliptic curves over Q


lemurell:
When clicking any of the conductor ranges, torsion or rank numbers under Browse elliptic curves on http://lmfdb.org/EllipticCurve/Q/ the search value for CM on the resulting search page is set to "only". It seems better to have it set to "include". (At least I got confused first after clicking on torsion=16, changing that to 12 and getting zero results.)

AndrewVSutherland:
I expect the same things happening here as with https://github.com/LMFDB/lmfdb/issues/1250 (for elliptic curves over number fields).  I will take a look now.

AndrewVSutherland:
Fixed by pull request https://github.com/LMFDB/lmfdb/pull/1312.  Will close as soon as it has been merged to master/prod/beta.

PR #1312, opened 2016-05-12
Fix for issue #1311

AndrewVSutherland:
Fixes https://github.com/LMFDB/lmfdb/issues/1311, should be merged to master/prod/beta asap.

### DIFF
--- a/lmfdb/elliptic_curves/templates/search_results.html
+++ b/lmfdb/elliptic_curves/templates/search_results.html
@@ -25,19 +25,19 @@
 <td align=left><input type='text' name='conductor' size=10 value={{info.conductor}}></td>
 <td align=left><input type='text' name='jinv'size=5 value={{info.jinv}}></td>
 <td>         <select name='include_cm'>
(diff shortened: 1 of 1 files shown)


JohnCremona:
Same code as you/we fixed rather recently, right?  Can I have a test case?

AndrewVSutherland:
Right.  To test click on any static link on EllipticCurve/Q, e.g. a conductor range or specific rank.  On lmfdb.org the search results page will have CM set to "only" , with the fix it will be set to "include".

On May 12, 2016 10:56:59 AM EDT, John Cremona <notifications@github.com> wrote:
>Same code as you/we fixed rather recently, right?  Can I have a test
>case?
>
>---
>You are receiving this because you authored the thread.
>https://github.com/LMFDB/lmfdb/pull/1312#issuecomment-218783563

-- 

JohnCremona:
Thanks.  I just realised that this is not the exact code you fixed before but a clone of it with my same error.  Merging (and pushing) now.

PR #1316, opened 2016-05-12
Added search by dim and browse/search by hyperelliptic label

jenpaulhus:
On the main page for Higher Genus w/Automorphisms:
/HigherGenus/C/aut/

Added browse of hyperelliptic curves only by genus.

Added option to search by dimension of the family of curves.
Added option to specify whether results in search should be hyperelliptic curves or not.

(Also impacted search result pages.)

### DIFF
--- a/lmfdb/higher_genus_w_automorphisms/main.py
+++ b/lmfdb/higher_genus_w_automorphisms/main.py
@@ -102,7 +102,11 @@ def higher_genus_w_automorphisms_search(**args):
         parse_ints(info,query,'genus',name='Genus')
         parse_list(info,query,'signature',name='Signature')
         parse_ints(info,query,'dim',name='Dimension of the family')
-        parse_bool(info,query,'hyperelliptic', minus_one_to_zero=False)
+        if 'inc_hyper' in info:
+            if info['inc_hyper'] == 'exclude':
+                query['hyperelliptic'] = False
+            elif info['inc_hyper'] == 'only':
+                query['hyperelliptic'] = True
     except ValueError:
         return search_input_error(info, bread)
     count = parse_count(info)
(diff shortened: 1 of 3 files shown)


JohnCremona:
Looks pretty good.  In the switch to include/exclude hyperelliptics it is not clear what the default is.  I would make this a 3-way switch with options 'include', 'exclude', 'only' with 'include' the preselected default.  If you can do that I'll merge it.  As an example of getting the logic straight here take a look at the include/exclude CM options in elliptic curve search -- I got that wrong first time but Drew corrected it.

jenpaulhus:
The new commit should change from yes/no to include/exclude/only.

jenpaulhus:
Deleted a line of text from /HigherGenus/C/aut which is not relevant now, based on change to include/exclude/only.

AndrewVSutherland:
Reviewing now

AndrewVSutherland:
@jenpaulhus On the results page you should set the 3-way include/exclude/only button to the value that was actually used in the query (very confusing otherwise).  If I am on the search results page and I click search again without changing anything, I should see the same results.

You need to modify lines 92-96 of 29  lmfdb/higher_genus_w_automorphisms/templates/hgcwa-index.html, see 27-40 of https://github.com/LMFDB/lmfdb/blob/master/lmfdb/elliptic_curves/templates/search_results.html for example.  Also, I'm not sure 0, True, False is a good choice of values to use, I think "include" "exclude" "only" is clearer (and less likely to lead to mistakes).

If it's not convenient for you to make the change right now, I can merge this and then make this change on for you (I'd rather not leave the PR open indefinitely).

jenpaulhus:
@AndrewVSutherland I'll go make these changes right now and do another commit in a couple minutes.

jenpaulhus:
Ok, git and I had a bit of a fight, but I think I won.  The last commit should fix the issues you pointed out: both in the search results displaying the hyperelliptic choice from the original search and changing the "behind the scenes" values to include, exclude, only instead of 0/T/F.

AndrewVSutherland:
@jenpaulhus On the results page you should set the 3-way include/exclude/only button to the value that was actually used in the query (very confusing otherwise).  If I am on the search results page and I click search again without changing anything, I should see the same results.

You just need to modify the HTML page to set this, see

AndrewVSutherland:
@jenpaulhus Looks good now, thanks!  Merging...

PR #1320, opened 2016-05-12
Update from master

AndrewVSutherland:


JohnCremona:
Hint: when merging from a pull request into master when you confirm the merge you can ask for it to be a squash-and-merge so that if the person making the PR has made lots of little commits only one appears in master.  I tihnk that makes for a tidier history.

AndrewVSutherland:
Will do in the future, thanks for the tip.

PR #1324, opened 2016-05-13
Fix api bug #1323

### LINKED ISSUE #1323: api broken by pymongo upgrade
If you open /api and click on any collection you get an error.  I think this is caused by a change of parameter name in pymongo's find() from version 2.8 to 3.2, and will try to fix it.

JohnCremona:
The pymongo upgrade from 2.8 to 3.2 requires some changes, e.g. (as here) the 'fields' parameter for find() has been renamed 'project'.  This broke the api (when you click on any collection name in /api).  Fixed here: fixes #1323


JohnCremona:
Github assumed I wanted to pull from lmfdb:api branch since I called my branch api too.  But I meant to pull to master.  I will make a new PR.

PR #1327, opened 2016-05-13
Lattice12

sanni85:
Made changes to address the comments of today: 
- label moved in the page and the property box
- added dots
- Gram matrix moved
- import script fixed

See http://localhost:37777/Lattice/3.702.1404.12.12
for example


sanni85:
All tests pass

Issue #1223, opened 2016-05-07
errors in modular form data


jwbober:
***1 Upvote*** I have been trying to match modular form data in the LMFDB with data that I computed myself, so a few days ago I grabbed all of the complex embeddings of the q-expansions of all of the modular forms in the LMFDB. Below is a list of 1080 forms with weight between 2 and 12, and level < 100, that I have had trouble matching with my data.

This is not necessarily a comprehensive list of errors, and it is even possible that some of the forms in this list are fine, and I hope I am wrong about how much is wrong (possibly many of the cases with level < 50 have been fixed by now) but every form that I have manually examined has turned out to have problems. In some of them a9 is wrong for some reason. In at least two cases there are rational forms with the wrong coefficients. I think in another case there is a form that looks rational but shouldn't be. In other cases the page or the space gives an error, but maybe it can be clicked on. For others the page or space says it is not in the database.

Here is the list: [boberlist.txt](https://github.com/LMFDB/lmfdb/files/253614/boberlist.txt)

jvoight:
@jwbober : Can you do a first attempt to separate these out by the actual issue?  (Whatever you're doing to check, you can sort them by the problem, right?)

Starting from the top...

10.11.7.b, 10.11.7.c: this means that the form is level 10 and weight 11, right?  I agree this form is not in the database, but it's not advertised as being there.
  http://www.lmfdb.org/ModularForm/GL2/Q/holomorphic/10/11/?group=1

10.12.9.a: This one appears
  http://www.lmfdb.org/ModularForm/GL2/Q/holomorphic/10/12/9/a/
When I compute it using Magma, the coefficients a_9, a_18 are indeed wrong.  And the L-function looks ridiculous.

Let's break this up into cases.

sehlen:
Please note that Fredrik mentioned before that probably all of these that
have problems are related to an error in a previous version of the code and
it is not an accident that a(9) has a problem because it was an error in
the routine computing the coefficients recursively from the a_p's. The code
is fixed but there have been many problems with the database connection
when Fredrik tried to fix them so he was not able to do so and is now
"offline".

jwbober:
When I do the same comparison but only use a1 through a8 I get fewer errors, but I still get a list of 600. I will go through this list and try to produce examples of various problems that come up.

I got my list of forms by manually querying the database by modifying some code Fredrik sent me. I don't know why in some cases I found a form in the database but the space of forms is not available from the website.

[...]

PR #1348, opened 2016-05-14
Clean up top

davidfarmer:
Rearrange the top right of every page so that it does not overlap long breadcrumbs.
Have the Feedback link go to the contact page.
Reorganize the contact page so that a welcoming message is first, and email is before
Feedback of error reporting.

The only tricky thing was to hide the login on the production version.  (We decided to do this
because you can't edit knowls on the production version, and a random user should not
even see the option to log in.)

I commented out the suggestion to see if bugs have been fixed on beta, because we should
draw less attention to beta, and bug fixes are supposed to be pushed to production.  But I
only commented it out because it was not discusses as a group.

### DIFF
--- a/lmfdb/templates/contact.html
+++ b/lmfdb/templates/contact.html
@@ -5,8 +5,16 @@
 
 <div>
 <p>
-For questions, inquiries, feedback or any other issues, including bug
-reports, please <b>either</b>
+If you are interested in becoming involved with the LMFDB,
+please 
+<a href="mailto:lmfdb-support@googlegroups.com"
+target="_blank">email</a>
+the <a href="http://groups.google.co.uk/d/forum/lmfdb-support"
+target="_blank">LMFDB mailing list</a>.
+</p>
+
+<p>
+To report an error or ask a question about the website content, please <b>either</b>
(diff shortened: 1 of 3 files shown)


JohnCremona:
All looks good to me.  I tried with/without --debug (i.e. beta) and with/without logging in.

Best if another person also checks.

On thing I noticed while finding a page with a long breadcrumb trail -- not relevant to this PR except that I can see why it had not been noticed before:  both
http://localhost:37778/ModularForm/GL2/Q/holomorphic/12/8/1/
and 
http://localhost:37778/ModularForm/GL2/Q/holomorphic/12/8/1/a/
have the same breadcrumb trail while the latter (an individual newform) should have an extra arrow showing (I suggest) '12.8.1.a'

AndrewVSutherland:
@JohnCremona I noted this in https://github.com/LMFDB/lmfdb/issues/1223#issuecomment-217665174, but there were so many more serious issues with classical modular forms that it wasn't written up.

JohnCremona:
@AndrewVSutherland OK -- not intended to delay merging of this PR though.  If you think it is good please do that as I was happy.

PR #1350, opened 2016-05-14
update from master (can't automatically merge, creating pull request to investigate why)

AndrewVSutherland:


JohnCremona:
Some file changes here do not seem relevant eg api.
This is a good case for a squash and merge so master history does not see
100 commits but only the total week s work.

AndrewVSutherland:
Hi John,

So I haven't completely figured out exactly where the problem arose, but 
it appears there are a few merges to beta that did not come from master 
(you can see them if you compare beta and master), one from David Farmer 
and one from you.  I presume these are what is creating the merge 
conflict?  What is the best way to resolve the conflict?  Is it better 
to edit the files, or to simply revert the most recent pull request and 
do a new merge?

Drew

JohnCremona:
Funny.  Neither of us intentionally ever push to either beta or prod except
from master.   Would it be best now to merge all three so we are in sync
again?   I can do that.

AndrewVSutherland:
If you can do it that would be great!  I just want to get back to a 
clean state and I'm nervous about reverting things.  But a this point I 
think we just want beta and prod to both match master.  I don't 
understand how beta and prod could have commits ahead of master.

Drew

JohnCremona:
OK, will do.

JohnCremona:
Right so fater pulling all 3 branches locally and updatin them all (upstream.master, upstream/beta etc) when I merge beta into master there's amerge conflict.  I'll resolve that on my machine and continue.

JohnCremona:
Same with merging master and prod.  I know that David and Samuele had some merge conflicts yesterday (though I don't know how they arose).  At least prod and beta are the same.

JohnCremona:
OK I have merges all three which involved a large number of merge conflicts in files edited and added by Samuele.  We must make sure that this does not happen again.  I am not certain that out of the dozens of conflicts I chose the right choice so Samuele will have to look at that and check.  Right now I have put all three merged branches onto my github and am testing locally.  Pushing to my own prod on github required a force so I expect that might be necessary getting this update onto the main prod branch, which is worrying.  I will test to see if github is happy to merge my 3 branches with the lmfdb ones without conflicts now, and if so will do that.

JohnCremona:
Now lmfdb/master and lmfdb/beta have the 3-way merge, and tests pass, and so in 9 minutes beta.lmfdb.org will show that.  Next to look at prod...

JohnCremona:
I think this has gone OK but will not update lmfdb/prod until after checking that beta.lmfdb.org looks OK in a few minutes

JohnCremona:
I just made the 3rd and last step so in 4 minutes www.lmfdb.org will pick up the new prod.  Please monitor -- I have to check out right now but will look again from the hotel lobby.

AndrewVSutherland:
Will do.

PR #1351, opened 2016-05-14
Add random items to index pages for artin rep, galois groups, local fields

jwj61:
To test, for j in {artin rep, Galois group, local fields},
  go to index page for $j
  click on new link for random item

In case you are curious, code for Artin representations is slightly different for 2 reasons: (1) it still uses the original code for accessing the database (not written by me), but works fine, and (2) database records are for Galois orbits which then hold information about each individual rep'n, and we have pages for the individual rep'ns.  So, we randomly pick an orbit, and then randomly pick a rep'n from it.

### DIFF
--- a/lmfdb/artin_representations/main.py
+++ b/lmfdb/artin_representations/main.py
@@ -6,16 +6,16 @@ import pymongo
 ASC = pymongo.ASCENDING
 import flask
 from lmfdb.base import app, getDBConnection
-from flask import render_template, render_template_string, request, abort, Blueprint, url_for, make_response, flash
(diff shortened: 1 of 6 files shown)


AndrewVSutherland:
Do to the recent reset of the master branch in github, this pull request cannot be merged.  Please resynch your local master and resubmit.

jwj61:
Rebased off of current master

AndrewVSutherland:
I agree this looks good now.

AndrewVSutherland:
Reviewing now.

PR #1379, opened 2016-05-16
Added emf_test_pages.py verification script

AndrewVSutherland:
Adds a test script which verifies that all the classical modular forms pages of trivial character within a specified range of weights and levels loads correctly.

### DIFF
--- a/lmfdb/modular_forms/elliptic_modular_forms/emf_test_pages.py
+++ b/lmfdb/modular_forms/elliptic_modular_forms/emf_test_pages.py
@@ -1,0 +1,81 @@
+
+from lmfdb.base import LmfdbTest, getDBConnection
+
+from flask import request
+import unittest2
+
+from views.emf_main import *
+from . import emf_logger
+emf_logger.setLevel(100)
+
+def anum(a):
+    return sum([26**i*(ord(a[len(a)-i-1])-ord('a')+1) for i in range(len(a))])
+
+def check_orbit_list(a):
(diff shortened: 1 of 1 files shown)


AndrewVSutherland:
Closing temporarily, will reopen

JohnCremona:
It is important that we encourage all developers to run the full test suite often, and to this end it is a good idea not to include by default teswt which take rather a long time.  The one(s) introduced here do that!  When I have added some comprehensive tests in other places I put a SKIP around them with a comment that the test should be run at critical moments but not every time (see test_emf.py in the same directory).  Incidentally I thought that our convention was that every test file should *start* with "test_*" and that only such files would get tested but this new emf_test_pages is certainly running so all that shows is that I cannot parse the regex "(?:^|[\b_\./-])[Tt]est_"

AndrewVSutherland:
I chose the file name with the understanding that it would not be run.  
Do you now want me to:

(a) fix the regex so it matches as intended
(b) add SKIP and rename the file test_emf_pages.py
(c) both

Drew

JohnCremona:
That's funny.  If you had asked me I would have said "it's ok, it won't run since it doesn't start with "test_".
If you changed the regex, you could still get it to run when required by putting in a temporary symlink.  That is less effort than editing the file to remove the SKIP when you want to test it.
I am not sure if it is documented anywhere exactly what the pattern should be (it's not in test.sh except by implication from the regex).
No strong feelings either way.
(BTW, the Sato-Tate tests are more verbose than most but they are very fast so I don't mind!)

Issue #1332, opened 2016-05-13
Upgrading to Sage 7.3


JohnCremona:
Sage 7.2 will be released soon and I want to start testing to see what changes might be needed.  This Issue can be used to make a list.  So far:
  - the hilbert_modular_forms module (but no other) make a lot of use pf Sage's preparse function (I don not know why that is necessary), and that now has to be imported from sage.repl and not sage.preparse

jvoight:
Sorry all--my guess is that it can be removed, but back in the day I was told to use it, so I did!

To make this quicker, note that only lmfdb/hilbert_modular_forms/hilbert_modular_form.py is actually part of the LMFDB code base (the rest are import scripts), and preparse isn't used in the file so we could just remove one line.

JohnCremona:
OK let's do that.  Preparsing is designed only for the Sage command line so that users can type a^2 instead of a**2 and should not be used in library code (in my opinion).   One might think that (for example) it's a useful way to convert a polynomial as a strong from Magma output into Sage, but Sage itself provides conversion from Magma oblects to Sage objects, and can contruct polynomials from strings, e.g.
sage: Qx = PolynomialRing(QQ,'x')
sage: Qx('x^2+1')
x^2 + 1
so I would huess that every use of preparse can be eliminated.

JohnCremona:
PR #1396 (now fully merged) fixed the redundant import in hilbert_modular_forms.py.

I will record any deprecation warnings shown in the test scripts and also grep for "sage.rings.arith".  Outcome: just 4 places, LCM in HeckeCharacter.py, primes in number_field.py, binomial in symL.py, lcm and gcd in hypergm/plot.pt.  In all cases I could just switch the import to be from sage.all.  I think there is a convention that one should not import from as low down the tree as possible precisely to avoid this sort of thing, and this seemed easier than trying to import from sage.arith and falling back to sage.rings.arith on ImportError (for backwards compatibility).
Testing the edited code now.

JohnCremona:
See PR #1399

JohnCremona:
There are two tests in lmfdb/test_lfunction.py which fail with sage 7.2 (and pass with 7.1) when I run them on lmfdb.warwick.ac.uk using the current master:
test_Lsym2EC and test_Lsym3EC
though they run fine on lehner using 7.2.  I am investigating how two identical Sage builds can act differently on the same code.

JohnCremona:
[...]

PR #1394, opened 2016-05-18
master --> beta after #1392

JohnCremona:
Making the PR but not about to merge.

### DIFF
--- a/lmfdb/ecnf/main.py
+++ b/lmfdb/ecnf/main.py
@@ -8,6 +8,7 @@ import ast
 import StringIO
 import pymongo
 ASC = pymongo.ASCENDING
+from operator import mul
 from urllib import quote, unquote
 from lmfdb.base import app, getDBConnection
 from flask import render_template, render_template_string, request, abort, Blueprint, url_for, make_response, redirect, flash, send_file
@@ -369,6 +370,8 @@ def elliptic_curve_search(**args):
         parse_nf_elt(info,query,'jinv',name='j-invariant')
         parse_ints(info,query,'torsion',name='Torsion order',qfield='torsion_order')
         parse_bracketed_posints(info,query,'torsion_structure',maxlength=2)
+        if 'torsion_structure' in query and not 'torsion_order' in query:
+            query['torsion_order'] = reduce(mul,[int(n) for n in query['torsion_structure']],1)
     except ValueError:
         return search_input_error(info, bread)
 
--- a/lmfdb/elliptic_curves/elliptic_curve.py
+++ b/lmfdb/elliptic_curves/elliptic_curve.py
@@ -3,6 +3,7 @@ import re
 import time
 import ast
 from pymongo import ASCENDING, DESCENDING
+from operator import mul
 import lmfdb.base
 from lmfdb.base import app
 from flask import Flask, session, g, render_template, url_for, request, redirect, make_response, send_file
@@ -222,6 +223,9 @@ def elliptic_curve_search(**args):
         parse_ints(info,query,'rank')
         parse_ints(info,query,'sha','analytic order of &#1064;')
         parse_bracketed_posints(info,query,'torsion_structure',maxlength=2,process=str,check_divisibility='increasing')
+        # speed up slow torsion_structure searches by also setting torsion
+        if 'torsion_structure' in query and not 'torsion' in query:
+            query['torsion'] = reduce(mul,[int(n) for n in query['torsion_structure']],1)
         if 'include_cm' in info:
             if info['include_cm'] == 'exclude':
                 query['cm'] = 0
--- a/lmfdb/genus2_curves/genus2_curve.py
+++ b/lmfdb/genus2_curves/genus2_curve.py
@@ -5,6 +5,7 @@ import re
 import time
 import pymongo
 from pymongo import ASCENDING, DESCENDING
+from operator import mul
 import lmfdb.base
 from lmfdb.base import app
 from flask import Flask, flash, session, g, render_template, url_for, request, redirect, make_response, send_file
@@ -257,6 +258,8 @@ def genus2_curve_search(**args):
         parse_ints(info,query,'cond','conductor')
         parse_ints(info,query,'num_rat_wpts','Weierstrass points')
         parse_ints(info,query,'torsion_order')
+        if 'torsion' in query and not 'torsion_order' in query:
+            query['torsion_order'] = reduce(mul,[int(n) for n in query['torsion']],1)
         parse_ints(info,query,'two_selmer_rank','2-Selmer rank')
         parse_ints(info,query,'analytic_rank','analytic rank')
         # G2 invariants and drop-list items don't require parsing -- they are all strings (supplied by us, not the user)
--- a/lmfdb/search_parsing.py
+++ b/lmfdb/search_parsing.py
@@ -272,7 +272,8 @@ def parse_bracketed_posints(inp, query, qfield, maxlength=None, exactlength=None
     if process is None: process = lambda x: x
     if (not BRACKETED_POSINT_RE.match(inp) or
         (maxlength is not None and inp.count(',') > maxlength - 1) or
-        (exactlength is not None and inp.count(',') != exactlength - 1)):
+        (exactlength is not None and inp.count(',') != exactlength - 1) or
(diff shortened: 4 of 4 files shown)


JohnCremona:
I reckon that after accepting this as a normal (not squashed) merge, beta and master will have the same file content but beta will be one commit ahead of master owing to the merge commit.  Similarly with PR #1395 for prod.

AndrewVSutherland:
You can see from this e-mail the cause of my concern -- the pull request 
to master whose changes you are now trying to apply to beta did not 
modify either lmfdb/elliptic_curves/elliptic_curve.py or 
lmfdb/search_parsing.py.

So its not just a question of what you see when you look at the file 
changes in the pull request on github, the e-mails going out to all the 
RMs will have misleading information in them (unless we revert to 
pushing from the command line, in which case the RMs won't get an e-mail 
at all).

AndrewVSutherland:
Merging (without squashing) to see if J.C.'s prediction holds :)

It appears that contrary to J.C.'s prediction, it is 4 commits ahead (which is what I expected but was not so bold as to publicly predict :).  There are no file differences (as we both expected, since this was not a squash-merge).  I predict the "commits ahead" count will keep growing, but as long as we do not squash merges, there will be no file differences when you compare the branches after merging master -> beta (and ditto for prod).

JohnCremona:
I cannot see how github thinks that beta is 4 commits ahead of master.  beta is now at a commit with two parents, one of which is master and the other is the previous beta,  That sounds like one ahead to me.

AndrewVSutherland:
On the page https://github.com/LMFDB/lmfdb/tree/beta if you look at the 
grey text you will see "This branch is 4 commits ahead of master".

If you then go to https://github.com/LMFDB/lmfdb/compare/beta you will 
see a list of the 4 commits:

https://github.com/LMFDB/lmfdb/commit/d39935413e81ec9be14f6d177aa13442e4981792
https://github.com/LMFDB/lmfdb/commit/080da20d400a3ae60d7fc97bbb26b4250c659730
https://github.com/LMFDB/lmfdb/commit/e79f6d066364008f8611d4bf470064c3842114b6
https://github.com/LMFDB/lmfdb/commit/0e8c2b2cc0ae3baa8df3c071ee82404f08aabef1

Or are you seeing something different in github than I am?

JohnCremona:
The page  https://github.com/LMFDB/lmfdb/compare/beta is a compare with master as base so it is showing you what would happen if you tried to merge the other way.  By comparison if you start to do a new pull reuest from master to beta (without confirming that you want to actually do this) it shows you
https://github.com/LMFDB/lmfdb/compare/beta...master
which says that there is nothing to be done: "beta is up to date with all commits from master".

So the incomprehensible thing (to me) is why, when looking at the code page for branch beta (https://github.com/LMFDB/lmfdb/tree/beta) it tells you that beta is 4 commits ahead.  That looks like a bug to me....and it is what is confusing us all.

JohnCremona:
I am going to make a tiny new branch deleting one line from hilbert_modular_forms.py to satisfy #1332, then merge to master and then to beta and prod.  (I will test long the way!)  Then we'll see how these view change.  All merges will be non-squashed.

AndrewVSutherland:
On 2016-05-18 08:02, John Cremona wrote:
> The page  https://github.com/LMFDB/lmfdb/compare/beta is a compare
> with master as base so it is showing you what would happen if you
> tried to merge the other way.

Right, it is showing the 4 commits (listed in my previous e-mail) that 
are in beta but not master (and which would be applied if you merged 
from beta to master).

These are the 4 commits by which beta is ahead.

>  By comparison if you start to do a new
> pull reuest from master to beta (without confirming that you want to
> actually do this) it shows you
> https://github.com/LMFDB/lmfdb/compare/beta...master
> which says that there is nothing to be done: "beta is up to date with
> all commits from master".

Right, beta is not behind master at all (it has every commit master 
does, as a result of merging the pull request from master), to that is 
why it says beta is up to date with all commits from master.

But notice that it does not say that the branches are identical (which 
it would if neither was ahead of the other). I guess the only thing we 
disagree on is the number of commits by which github thinks beta is 
ahead, I believe github thinks 4 (the 4 I listed in my previous e-mail), 
while you say 1.

> 
> So the incomprehensible thing (to me) is why, when looking at the code
> page for branch beta (https://github.com/LMFDB/lmfdb/tree/beta) it
> tells you that beta is 4 commits ahead.  That looks like a bug to
> me....and it is what is confusing us all.

I don't think this is a "bug", it is consistent with everything else you 
see in github (beta is 4 commits ahead and 0 commits behind master and 
their files are identical).  If you compare in one direction (imagining 
a pull request from beta to update master) you will see 4 commits (and 
no file changes), if you compare in the other direction (imagining a 
pull request from master to beta), you will see 0 commits because beta 
is not behind master at all.

Perhaps it would be better to characterize this as a "feature" that we 
do not particularly like (caused by the fact that every merge creates a 
commit that is then present in the target but not the source -- without 
squashing this commit will involve no file changes).  Subsequent merges 
in the same direction will not change this.

JohnCremona:
OK, I misled you: I did not mean that guthub said that beta was 1 ahead of master, only that if you actually look at the commit beta is pointing to (0e8c2b2) then it has master (51ebe3e) as one of its immediate parents, so to go along the graph from master to beta is only one step.  Since the beta commit at 0e8c2b2 is a merge commit (from master and its previous state) it certainly follows that beta contains all changes in master, we are agreed on that.   There is still a mystery as to why there would be anything to do to merge beta onto master: their files are the same, and 3 of the 4 commits you list are merge commits while the other is not.  I think that the one which is not a merge might be the result of a (misguided) squash operation.

I will still do what I said above (merge a very small patch and see how things change).  If I am right that it was the sqush-merge which made the only non-merge commit in those 4, then we could improve the situation by (just this once) updating beta and prod on github  from master using the command line (as we always used to do) to have a fresh start where all three are literally at the same point (same commit), and then even though we would see these "n commits ahead of master" messages they would all be merges (and visibly so) so we would know that it was OK to ignore them.

AndrewVSutherland:
On 2016-05-18 08:25, John Cremona wrote:
> OK, I misled you: I did not mean that guthub said that beta was 1
> ahead of master, only that if you actually look at the commit beta is
> pointing to (0e8c2b2) then it has master (51ebe3e) as one of its
> immediate parents, so to go along the graph from master to beta is
> only one step.  Since the beta commit at 0e8c2b2 is a merge commit
> (from master and its previous state) it certainly follows that beta
> contains all changes in master, we are agreed on that.   There is
> still a mystery as to why there would be anything to do to merge beta
> onto master: their files are the same, and 3 of the 4 commits you list
> are merge commits while the other is not.  I think that the one which
> is not a merge might be the result of a (misguided) squash operation.

Right, if we avoid squashes this will not happen and none of the commits 
by which beta (or prod) is ahead of master will include any file 
changes, they will just be "empty" merge commits.

> 
> I will still do what I said above (merge a very small patch and see
> how things change).  If I am right that it was the sqush-merge which
> made the only non-merge commit in those 4, then we could improve the
> situation by (just this once) updating beta and prod on github  from
> master using the command line (as we always used to do) to have a
> fresh start where all three are literally at the same point (same
> commit), and then even though we would see these "n commits ahead of
> master" messages they would all be merges (and visibly so) so we would
> know that it was OK to ignore them.

I agree 100% with everything you say in this paragraph!

JohnCremona:
OK I did what I said and did all the merges and (as @AndrewVSutherland predicted) both beta and prod are 5 commits ahead of master, exactly one more than before and that's the new merge commit.

Shall I rest this counter (as it were) with a clean push?  So that these bogus extra commits in future are all merges with no code changes?

AndrewVSutherland:
Yes, please do!  We should then all expect to see the "commits ahead" 
counter to get incremented by 1 with each merge from master to beta or 
prod (which we are agreeing to never squash, right?), and none of these 
commits will have any file changes in them.

JohnCremona:
Done.  Yes, we are agreed to never use squash when merging from master to weither beta or prod.  It matters less when merging a PR into master but given the confusion it might cause I suggest that we do not do that either.

For the record: as of 14:00 BST on Wednesday 18 May all three branches master, beta, prod are idential and at the commit  9755a2efa91edd36ce29d46086f16f4a040beb79

PR #1399, opened 2016-05-18
adapted imports in 4 places for compatibility with Sage-7.2

JohnCremona:
All I did was import from sage.all instead of lower down the tree so that it waas not necessary to use try/except for the new/old places where a few functions were imported (LCM, binomial, lcm, gcd, primes) in 4 files.
I tested this on both Sage 7.1 and 7.2 and so should you ;)

### DIFF
--- a/lmfdb/HeckeCharacters.py
+++ b/lmfdb/HeckeCharacters.py
@@ -1,7 +1,7 @@
 # -*- coding: utf-8 -*-
 # HeckeCharacters.py
 
-from sage.all import gp, xmrange, Integer, pari, gcd
+from sage.all import gp, xmrange, Integer, pari, gcd, LCM
 from sage.misc.cachefunc import cached_method
 from sage.groups.abelian_gps.abelian_group import AbelianGroup_class
 from sage.groups.abelian_gps.abelian_group_element import AbelianGroupElement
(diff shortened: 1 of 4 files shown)


AndrewVSutherland:
I don't have Sage 7.2 installed on my machine at the moment, and I don't want to take the time to do this right now, perhaps someone else can review?

sehlen:
Reviewing now

JohnCremona:
Thanks @sehlen -- better than trusting me!

sehlen:
All tests pass except the two mentioned errors in modular forms which I will deal with now (but which are completely independent of this issue) and I don't see any reason not to merge this.

JohnCremona:
By the way, the system-wide Sage on lehner is now 7.2 while that on atkin is 7.1.  I will leave atkin like that for now since it enables testing with both.  So far I have no evidence that there are any more problems with 7.2, and the next step will be to switch beta.lmfdb.org to run 7.2.  If any problems ensue it is easy to switch back (on lmfdb.warwick.ac.uk I have 3 separate Sage builds to that prod and beta can use different versions and there's still a 3rd for building new releases and testing).

AndrewVSutherland:
I suggest keeping Atkin (or whatever machine beta.lmfdb.org is pointing 
to) on Sage 7.1 for some time.  We want to be able to test things on 
beta using the same version of Sage that is running in the cloud.

JohnCremona:
beta.lmfdb.org is running on lmfdb.warwick.ac.uk but (see above) I have several different versions on there anyway.  atkin is separate.   For testing before merging a pull request, release managers can use either atkin for 7.1 or lehner for 7.2.  ( @davidfarmer now also has accounts on both).

edgarcosta:
Should I start to build sage in the cloud?

AndrewVSutherland:
I would not be in a rush to do this.  I'd rather get the production 
environment completely stable first (this will surely take another week 
or two), then updgrade to Sage 7.2 on atkin and run it there for a few 
weeks, and only then consider upgrading to Sage 7.2 in the cloud.

Or is there are reason we are in a rush to move to Sage 7.2?

edgarcosta:
Perfect! Please annoy me whenever you want Sage 7.2 in the cloud.

Issue #1405, opened 2016-05-18
Fix name of space for classical modular forms


sehlen:
Go to http://www.lmfdb.org/ModularForm/GL2/Q/holomorphic/7/9/3/
and observe that the character is not mentioned in the name S^new_9(7) ;-)
This is due to some errors in the template and I'm fixing it right now.

AndrewVSutherland:
Fixed by pull request #1409 which is now live on www.lmfdb.org

Issue #1385, opened 2016-05-17
Display partial information on classical modular form pages


AndrewVSutherland:
_Edited per the discussion below_

In cases where the coefficients of the Q-expansion are too large to reasonably display, as on

http://beta.lmfdb.org/ModularForm/GL2/Q/holomorphic/23/38/1/ http://beta.lmfdb.org/ModularForm/GL2/Q/holomorphic/23/38/1/a

for example, don't display the q-expansion but just give information about the coefficient field (a link to the field if it is in the LMFDB, and if not, at least its degree and perhaps a defining polynomial for it).

 We can instead provide a link for downloading whatever coefficient data we have available for those who want it.

For cases where we only know the dimensions of the Hecke orbits that make up the decomposition of the space and perhaps some other partial information (e.g. the coefficient field or its degree), we could display this and note that the coefficients are not yet available.  This would be friendlier then a long list of "Hecke orbit space not in database" entries (we currently hide these space to avoid displaying them).

I am marking this as critical because, as it stands, many of the current pages (not just the two examples above) cannot be reasonably rendered and take very long to load.  This is something we could easily fix even before filling out and verifying all the modular form data we would like to have available (see the discussion at https://github.com/LMFDB/lmfdb/issues/1248).

This more flexible approach would also allows us to have modular forms pages in cases where we have only numerical data but still want to include the L-function in the LMFDB (as noted in https://github.com/LMFDB/lmfdb/issues/1237#issuecomment-217815479).

sehlen:
It does not make sense to restrict to a_2 as this might vanish. Moreover,
it might not be of the form alpha where Q(alpha) is the coefficient field
in the data that we computed. We have to distinguish several cases and I
would rather vote for hiding the q-expansion by default in those cases.

sehlen:
But there is another issue with this, namely that pages take long to load
(before they get the chance to render) because the record is large. Please
assign me and I can offer a fix for both issues today or tomorrow for
people to test.

AndrewVSutherland:
@sehlen (https://github.com/LMFDB/lmfdb/issues/1385#issuecomment-219710846) Good point, and I think you are right that we don't want to complicate things.  So for the moment we could just have two cases

- Display everything as we do now (coefficients and information about the coefficient field)
- Display no coefficients (with a download button if they are available) and give information about the coefficient field (dimension and a link to its LMFDB page if there is one or a defining polynomial if not)

Does that sound reasonable? (If so I can amend the issue description)

[...]

Issue #1237, opened 2016-05-08
Display of Hecke eigenvalues of classical modular forms


jvoight:
The current way of displaying forms, e.g.
http://www.lmfdb.org/ModularForm/GL2/Q/holomorphic/17/12/1/?group=0
http://www.lmfdb.org/ModularForm/GL2/Q/holomorphic/17/12/1/b/
leaves much to be desired. We should think of a better way: for HMFs, it seemed better to put the eigenvalues in a table.

I think there are probably better ways, and here are two possibilities:
(1) Write in terms of a ZZ-basis of the Hecke ring;
(2) For slightly more 'geometry', generate the Hecke ring by a finite subset of a_2, a_3, a_5, ... and write the Hecke eigenvalues as ZZ-linear combinations of these (small) elements.
This is definitely needed for large dimension Hecke spaces; possible for quadratic (or cyclotomic) Hecke fields, the current way is fine.

arbooker:
More generally, beyond some point we should stop displaying exact algebraic numbers and show only embedded data. (Probably this means no longer attempting to sort things into Galois orbits also.) Insisting on exact numbers is a severe limitation; we have millions of modular form L-functions computed but no way to access most of them on the site because of this.

AndrewVSutherland:
See related issue #1385

AndrewVSutherland:
Perhaps this should be merged with #654?

AndrewVSutherland:
I think the immediate issue here has been resolved by the series of recent PRs for classical modular forms (we now simply suppress the display of q-expansions that are too unwieldly to fit on the screen in a reasonable way).  Certainly we need to think about how to handle this issue more generally, especially in cases where we might have only numerical approximations for the q-expansions (per remark https://github.com/LMFDB/lmfdb/issues/1237#issuecomment-217815479).  For that it probably makes sense to create a new issue that we might schedule into a future release.

Issue #720, opened 2015-12-17
Wrong description of complex embeddings


nmascot:
On the newform page, it says that:
"The different embeddings of the first few Fourier coefficients **that preserve the character** are shown in the table below."

The coefficients lie in an extension of a cyclotomic field; and the phrase "preserve the character" seems to imply that we only list embeddings that induce the identity on the cyclotomic field. In particular, the number of these embeddings should be the relative degree; however, it is the coefficient field.

We should either change the text or only display the embeddings that are compatible with the choice of character. I personally prefer the latter solution.

nmascot:
Cf. /11/7/2/a/ for an example.

sehlen:
I would prefer the latter, as well and you can assign me.

PR #1409, opened 2016-05-19
Modular forms improved 1

sehlen:
This pull request changes a lot in the "backend" and "frontend" code for modular forms
and has several purposes:

1) Cleaning up the code (at least a bit, it is a start)
2) New features in web_object.py, in particular support for multiple gridfs files per mathematical object to support having records with different amount of data about the same object available
3) Some performance improvements and stuff related to how we do computations of modular forms right now, there will be more changes soon be more flexible and be able to handle data from various sources, not only sage
4) Providing fixes or at least temporary solutions to issues
- https://github.com/LMFDB/lmfdb/issues/1385
- https://github.com/LMFDB/lmfdb/issues/1237
- https://github.com/LMFDB/lmfdb/issues/720
- https://github.com/LMFDB/lmfdb/issues/1405

Some of the changes are also related to the critical issue https://github.com/LMFDB/lmfdb/issues/1248 but most of that had to be done in our computational code (by adding a lot of checks and making sure we replace wrong data and also making things more reliable and improving performance in general in order to be able to now(!!) quickly address remaining data issues.

There is a lot more in the pipeline, the code is prepared to give much better search options etc but now I will first fix and extend the data and then return to the list of "features".


sehlen:
Wait a second before checking this out, I will add some examples to look at and make one more smaller modification.

sehlen:
OK, so I just increased a parameter (`default_max_height` in __init__.py of the elliptic_modular_forms module) that controls which q-expansions are hidden. I essentially use the length of the height (so the parameter is somewhat misnamed) of the rational numbers appearing in the Fourier expansion (expressing the coefficients in terms of a power basis) times the degree of the field as a measure.

You can see the effect of the current cutoff (90) here:

ModularForm/GL2/Q/holomorphic/25/36/1/

The first two are shown, the remaining ones are hidden. Compare to beta.
Also note that decreasing the `default_max_height` to e.g. 80 would also hide the second one but still show the first one if you prefer that. Maybe we have to look at many examples to figure out what is right.

Moreover, if you go to one of the forms with hidden q-expansion, e.g. /ModularForm/GL2/Q/holomorphic/25/36/1/c/, then you'll learn just a little bit more about it, namely the trace and the approximate norm of the first nonvanishing Fourier coefficient of index >1.
Maybe we should always show this information or the traces and norms of the first few Fourier coefficients...

JohnCremona:
There is a lot to look at here!
Without wishing at all to doubt the perfection of the changed code, this could be a pull request which we want to put onto beta for a while before putting it onto prod.  Our current workflow is not yet flexible enough for prod to get changes not in beta though.  Perhaps @AndrewVSutherland and I (who are preparing both a Release Manager's Guide and a Developer's Guide (mainly on dealing with git and branches etc) should think about evolving the workflow to allow this.
Sorry this remark should probably be in a new Issue...

sehlen:
I completely agree, this should be on beta for a while for testing. I
forgot to mention (shame on me, it was late!) that it depends on a change
in the data that I performed on the database so before this can go to prod,
the database has to be copied as well. The database is in this case
backwards compatible, the code isn't. It could be made backwards compatible
in theory but I think copying over the data first is the preferred solution
here. I added a field "prec" to the metadata of the gridfs files of a
webnewform to realize the feature I described above. Moreover, I started to
create "smaller records" fot everything that's in the database but haven't
finished because I actually would like to measure the performance of some
of the pages on beta and then tweak the parameters maybe a bit.

JohnCremona:
That is a crucial point to have remembered.  If we update the prod branch then within 15 minutes the code would run on www.lmfdb.org and we do not want to break that.  As I understand it the existing code will work fine on the updated data -- right?
According to our current workflow we will probably not push to beta until we can also push to prod, i.e. after the prod database has been updated.  Right now I do not know when that happens, or how to tell.

sehlen:
I am sorry I didn't mention it before but I wanted people to look at it and get feedback.
I also want that it gets tested on beta, that's what I had in mind all the time and I should have been clearer on this. But I think we should be able to push to beta completely independent from updating the database on prod to allow for testing *on* beta as soon as possible.

There is a separate issue about copying the database to the prod system for modular forms and I was holding that back but I think this is essentially ready now. I only did not finish creating duplicate, smaller records for all newforms yet because I wanted to measure the performance on beta a bit but this is not essential. The current code works without having multiple versions available and we can copy the db afaik now as well if you think it is too dangerous not to do it much in advance because someone might forget.

JohnCremona:
I am about to try out your code.  Remember that several of us (e.g. me ad @AndrewVSutherland ) can change the code branches which get pulled to the servers, but that up to now, only @edgarcosta has had control over copying and updating the cloud databases.

sehlen:
I can also make it backwards compatible and I think I'm just going to do
it. However, this is something we need to think about how to deal with in
the future as there is now no way to test the code with the production
data. There almost has to be an alpha and a beta website running different
versions of the data.

JohnCremona:
Anyone who has a login on lehner can add an ssh tunnel from a local port at home to lehner's port 37702 and they will get to see your branch up and running right now.  I can keep that up for a while.

AndrewVSutherland:
@JohnCremona Please hold of merging anything until we agree on the 
migration plan.

@sehlen Can you please explain in more detail the new data setup; in 
particular are you creating new records for everything (and if so, in 
the same collection or a new collection)?  Exactly which collections are 
affected, and what will be read from where by the old and new software?

fredstro:
I have looked this PR and it looks very good so far, I really appreciate all the effort you have put in!
Some comments:

1) all ‘standard’ tests pass (I am still running test emf_test_pages.py and haven’t yet run the long test which is skipped by default)
2) the cleanup is good. A lot of stuff that was effectively moved to the computing class had been left here and is now removed. Also a lot of unused / testing routines has been removed (somehow it seems that stuff which I thought was removed during the autumn cleaning had crept back in...)
3) the new code seems consistent and the logic seems sound.
4) I like the way the large q-expansions are dealt with in e.g. 25.36.1.c although I noticed a slight problem with the downloaded coefficients: when we return them as texts we should probably put parentheses around exponents in case someone wants to insert them into sage or latex (at the moment we have things like q^10 but having q^(10) or q^{10} would probably be better).

5) I am not sure how the ‘available_precs’ method is supposed to work but for ‘1.12.1.a’ it only returns 10 and 100 while we in fact have 100000 coefficients for download. 
   to make it consistent probably the web eigenvalues should also have _file_key_multi=['prec’] so the max_coefficient_in_db() could be rewritten simply as taking max of ‘available_precs()’ 
  alternatively we could make this more efficient by not retrieving the files but only the distinct values of ‘prec’ (and possibly even speed this up by using a multi index ‘hecke_orbit_label_prec’ or something like that)

JohnCremona:
- testing #1385 on URL ModularForm/GL2/Q/holomorphic/23/38/1/a/ : look pretty good.  I would have the text about coefficients being too big in normal size, currently it is very small.   I had to think for a second what you meant by "of index > 1", and of course you have to exclude the coefficient of q and allow for the possibility of a(2)=0.  Is "... coefficient a(n) for n>1" clearer?  It also implicitly defines a(n) so the meaning of a(2) is obvious.
Someone could argue that seeing the huge polynomial is unnecessary and you could just say (of degree 33".  People can download it if they want.  I did try the download (of the q-expansion for 200 terms) and duly received an 18MB file (which would be a lot more useful if it was in Sage or Magma or pari format but perhaps we don't want to triplicate the download options and people can do the minor editing themselves).

What criterion are you using to decide on how large is "too large to display"?  (Yes, I know I could look for it in the code...!)

- #1237 : for ModularForm/GL2/Q/holomorphic/17/12/1/b/ the above comments apply. For ModularForm/GL2/Q/holomorphic/17/12/1/?group=0 I would suggest that the footnote "...available for download" added something so prompt the user to click on the form's label to see how to do that.  Perhaps add "Click on the label in the table above for more information about each newform"?

- #720 You changed the text to make it clear that the table includes embeddings which do not preserve the character.  At #720 @nmascot preferred the other option, to exlcude embeddings which don't -- and of course it is easy to suggest that without implementing it.  How hard would that be?

- #1405 The bug is fixed, good!

I have not done a lot of experimentation, or looked at the code in any details,  but this clearly is a big step towards fixing these issues.  Thanks!

JohnCremona:
OK so three of us are looking at this now, which is good.  I am happy that @fredstro likes the code itself.
@AndrewVSutherland -- agreed.

fredstro:
I used the new revamped computing code (also by @sehlen ) to compute one of the spaces I have long struggled with 56.10.27 (in particular the 'b' newform)
The page ModularForm/GL2/Q/holomorphic/56/10/27/ 
seems to display the whole q-expansion of 56.10.27.b and not make use of the truncated one. (In fact mathjax doesn't even manage to render the whole q-expansion..)
@sehlen: Does the truncated display only work for forms computed with something on modforms-db which you haven't yet pushed?

sehlen:
Thank you all for the feedback, this is exactly what I intended.

@AndrewVSutherland I explained it above but maybe I should have given more details. In the collection `webnewforms.files`, which is kind of the "meta" collection for the gridfs files containing the newforms, there is now a field prec which contains the number of coefficients this record holds. Previously this was only in the mongodb record in webnewforms because there was only one file. I left the (unused) field prec in webnewforms for now but will delete it later. The new code for webnewform essentially looks up the record matching the query {'prec': {'$gt': self.prec}}, I will add a new commit that tries to find a record if this fails which would make it backwards compatible. (I added this field to all records, though).  (And the old code is not affected at all)

I did create records with smaller precision for some forms but no for all, because, as I wrote, I would like to test the performance on beta as I don't really know how large a record should be to make the page load fast and I thought this would be easier to test on beta than on my machine with a port forward to lehner which has a huge delay. I can see performance improvements when adding such a smaller record and they're quite noticeable though. Yesterday I tried 25/38/1/ and got (freshly restarted webserver on my machine) 21.177s time passed to when the webserver started the transfer and this went down to 12.52s when adding a smaller record. But you would like to see "real" numbers that don't involve any port forwarding over the sea because the pages should be much faster when running on beta!

Let me mention that it might be worth merging webnewforms and webnewforms.files, i.e. to only use webnewforms.files to store the data that's not stored in chunks but that's a different issue and I'm not sure which is best without testing it (probably it doesn't matter).

JohnCremona:
I agree with @fredstro about exponents and suggest () rather than {} since it is more likely these will be read into a program than into tex.

The URL @fredstro put in above almost killed my browser to render!

I ran both Drew's emf_test and the long one in test_emf normally commented out and all worked fine.

AndrewVSutherland:
@sehlen Having backward compatibility is always a good idea (e.g. imagine a scenario where we have to restore the database from backup for some reason).  If also means we don't have to be paranoid about the order in which we move things (data first vs code first).

It sounds like what you would like to do is get the code pushed to beta, possibly update the data after testing performance, then have the data migrated to lmfdb.org, then have the code pushed to production.  The only downside to this is that with our current workflow it prevents us from getting any other code changes into production until this process is complete, but it sounds like something you could do today, is that correct?

I have the scripts in place to update the data on the cloud; it takes about an hour (and is done in a way that ensures no downtime for www.lmfdb.org).

@sehlen perhaps you can let us know once the backward compatibility commit is in place (and any other changes you want to make based on the feedback from @fredstro).

JohnCremona:
I'm about to swtch my branch on lehner back from this to something else, in case anyone was looking at it to test this.

sehlen:
@fredstro There are still problems left with re-computing somehow and I'm looking at it now.
The record you computed is definitely not correct/complete afaik. I computed in independently on another machine and the result looks much better although I see several potential problems as new computational code does not make use of all the changes yet.

sehlen:
@JohnCremona Better download options will be created soon. I will increase the font size and follow your suggestion for reformulating the "hidden q-expansion text".

sehlen:
@JohnCremona regarding #720, I mentioned that this is only a temporary fix. I want the better solution but this involves computing the "correct" embeddings. It's not hard I think but just not top priority right now for me.

sehlen:
I fixed several smaller bugs I found and I adjusted the pages according to @JohnCremona's suggestions.
Moreover, the backwards compatibility should be in place now which I tested with one manually created record ("old" style).

sehlen:
One thing that I noticed is that it can sometimes take quite a while to compute the norm of the first a(n) with n>1 that does not vanish and we can't just multiply the embeddings as this gives too low precision and we can end up with a wrong result, it seems. I think we should save this in the db and it shouldn't take very long to add this to all records. Maybe I should add the code for this to this pull request?

JohnCremona:
With the update I went to ModularForm/GL2/Q/holomorphic/23/38/1/a/ but it has disappeared!

JohnCremona:
Precomputing anything we display is a good idea, but if it will take a long time we should not wait with this PR.  If you make the changes in such as way that it uses a stored norm when available and otherwise computes it that would be better (perhaps what you intended anyway).

AndrewVSutherland:
@sehlen Yes, I think it would be good to have all code changes that depend on data changes in this pull request (and of course don't break backward compatibility).

sehlen:
@JohnCremona it does load for me.

AndrewVSutherland:
@JohnCremona I was operating on @sehlen's assertion that it would not take long.  I agree that if it will take a long time we should not wait.

sehlen:
@JohnCremona I also created smaller records for ModularForm/GL2/Q/holomorphic/23/38/1/a and b now so these should load faster in any case.

AndrewVSutherland:
@sehlen So first let me say that this is a huge improvement!  I'm definitely keen to see it running on www.lmfdb.org as soon as we can (but of course only after you finish any further changes to the code and data, we test more thoroughly on beta, and decide on the order in which we should push the updates to the data and code to www.lmfdb.org).

I did notice one issue while testing your PR that is not necessarily a reason to delay merging it, as I believe everything I am about to say applies to the existing code as well.  I noticed that the first time you access a page is much slower than the second -- this suggests to me that you are caching information that gets reused, and it is clear that it is happening at the level of individual forms/pages.  For example, if I load the pages

/ModularForm/GL2/Q/holomorphic/23/36/1/a
/ModularForm/GL2/Q/holomorphic/23/38/1/a
/ModularForm/GL2/Q/holomorphic/23/40/1/a

each of these pages takes anywhere from 30 seconds to 2-3 minutes to load the first time, but after this it takes only 1.5 seconds for any of them (and this is true now matter how I interleave multiple requests to each, which makes me believe you are caching information for each independently).
Of course this behavior is massively exaggerated because I am running locally and pulling data across the Atlantic, I know it won't take 2-3 minutes running on bea.lmfdb.org, probably more like 2-3 seconds (see below).

But I do want to say that I think caching like this is a bad idea for two reasons: (1) in most cases it will not actually help, and (2) I worry about the growing memory usage.

The key point is that when a request for a URL comes in to beta.lmfdb.org (or www.lmfdb.org) it is dispatched to a pool of workers, and with our current setup, these workers are automatically reset every 100 requests or 5 minutes, whichever comes first (this actually means (2) is not such really much of a concern right now, but we would like to increase these parameters; they are set so low precisely so that we don't have to worry about (2)).  The odds that a given user who visits the same page twice is going to have that request handled by the same worker is very low (especially on www.lmfdb.org where there are more workers), so the performance they see is almost always going to be the same as the first request.

You can see this behavior with the existing code running on beta.lmfd.org, below are TTFB (time-to-first-byte) timings for a series of 100 consecutive requests to the same page: beta.lmfdb.org/ModularForm/GL2/Q/holomorphic/23/40/1/?group=0.

2.3, 2.0, 1.9, 0.5, 2.1, 2.2, 0.5, 2.2, 0.6, 2.2, 1.9, 0.5, 1.9, 0.5, 2.3, 0.5, 0.5, 2.0, 2.0, 0.5, 2.2, 0.6, 0.5, 2.2, 0.5, 0.5, 1.9, 0.5, 0.5, 0.5, 0.5, 0.5, 0.5, 0.6, 2.1, 0.5, 0.5, 0.5, 0.5, 0.5, 0.5, 0.5, ....

You will notice that at the beginning most of the response times are around 2 seconds (I believe these correspond to workers getting this request for the first time), whereas after a while they all drop to 0.5 seconds (these correspond to workers that have already seen this request).  Most of our users are not going to refresh the same page 100s of times, so they will almost always see the 2.0 TTFB, not the 0.5 TTFB.

As I said, this isn't necessarily relevant to this PR, other than that it might suggest that you might want to store in the database information that you are currently caching.

sehlen:
@AndrewVSutherland It is true that we use a cache which was introduced
quite a while ago but it is only caching information stored in the db.
There are virtually no computations involved but the record is large and
that makes the first request slow. However, it does make sense to do this
even if the user does not refresh pages because it makes if very fast to go
from the space of a page, which currently needs to load all records for the
orbits, to each of those orbits!

sehlen:
@AndrewVSutherland PS: but the record size and what needs to be loaded is
being improved and this PR should be a first step for this. Also, the
parameters can be tweaked...

AndrewVSutherland:
@sehlen But when the user clicks on the newform link that URL is likely to be handled by a different worker who does not have access to the information you cached when the user visited the space page, so I don't see how this helps (yes it will help you when you are running locally, but it won't help on www.lmfdb.org and it will use up memory).

I'm also surprised at the amount of network traffic I see, which is a lot more data than what is being sent back in the URL (i.e. the amount of data being read from the database is much larger than the amount of data being displayed on the page).

sehlen:
> On May 19, 2016, at 12:05, Andrew Sutherland <notifications@github.com> wrote:
> 
> @sehlen <https://github.com/sehlen> But when the user clicks on the newform link that URL is likely to be handled by a different worker who does not have access to the information you cached when the user visited the space page, so I don't see how this helps (yes it will help you when you are running locally, but it won't help on www.lmfdb.org <http://www.lmfdb.org/> and it will use up memory).
> 
> 

Of course. I think this has been discussed many times but the setup used to be different. 

> I'm also surprised at the amount of network traffic I see, which is a lot more data than what is being sent back in the URL (i.e. the amount of data being read from the database is much larger than the amount of data being displayed on the page).
> 
> 

It is due to the fact that the data is stored poorly in a sense. Even a few coefficients can take up a lot of space and the record being fetched contains all of the embeddings for the coefficients in the range. Actually, I realize that this might be an easy fix, even without changing the data structure. I will take a look at it. But you will now also see an improvement for the URL you tested /ModularForm/GL2/Q/holomorphic/23/40/1/ only because I inserted those smaller records which, by the sorting of the index, get fetched even on beta by default.

But all of these issues are due to a poor design in the basic classes that we use (and which I invented, so shame on me here) which don’t implement lazy loading really and we should use mongoengine I guess which I was not aware of at the time I wrote these classes and when I thought it would be a good idea to have such general purpose classes.

AndrewVSutherland:
@sehlen For the moment let's not worry too much about the caching, since this isn't specific to this PR, I just wanted you to be aware of my concern.  Can you let me know when you are happy that there are no more commits you want to include in this PR?  I can then merge/push to beta so that we can do more testing there.

sehlen:
@JohnCremona OK, something strange might be going on. I now had "Problems with Hecke orbits in the database" for ModularForm/GL2/Q/holomorphic/25/36/1/ and just a minute later everything was fine again. Somehow I'm unable to find what exactly happened in the logs but this is really weird and seems to be a database issue. Anyway, it's also weird that I don't see this in the logs but then, there are so many debug messages, it is quite hard to search them in an intelligent way.

sehlen:
OK, so I think I'm happy with this for now and maybe it would be good to see it on beta.
However, it would be good if we could somehow test it with the old data or otherwise make sure that we have the new data on prod before we push this to prod, just keep this in mind.

I have one failing test but I don't understand why and maybe it is related to the strange problem I mentioned above in https://github.com/LMFDB/lmfdb/pull/1409#issuecomment-220493443.
It is the one for hiding the large Fourier coefficients and when I go to the page, I can see the text I'm testing for **and** it used to work! Any idea?

sehlen:
Please note that this now also deals with the download part of LMFDB#1334. I will stop now and we'll make sure that everything in this PR works and then we can look forward to computing more data and implementing new features.

sehlen:
PS: I will also work on documentation, in particular the inventory soon.
But we have to think about how to offer better download options because the files get very large and some conversions take rather long. I have ideas but let's discuss them somewhere else. Everything included in this PR should at least be an improvement, albeit not perfect at all. For instance, the download will still fail (timeout) if you request too many coefficients from newforms with large degree number fields.

AndrewVSutherland:
@sehlen Your last two comments make me very happy!  Let me know when you are happy for me and/or @JohnCremona to review again and merge.  With any luck we can get both this PR and the new data onto www.lmfdb.org tomorrow.

sehlen:
Please review and let me know if you can figure out why the one test fails...

AndrewVSutherland:
@sehlen I just checked http://beta.lmfdb.org/ModularForm/GL2/Q/holomorphic/25/36/1/ (with the old software) and it looks fine.  I will check again tomorrow morning with the new software also, and can run some automated repeated tests to see if it is some sort of non-deterministic/intermittent problem.

sehlen:
Sorry for adding some last minute commits - but now I'm done for now.

sehlen:
The problematic record /ModularForm/GL2/Q/holomorphic/56/10/27/ now also seems to be good. Did you fix it Fredrik or did I and forget about it? Anyway, if you want to just browse the pages with the latest version, I'm running the server on lehner and will just leave it running for a while; the port is 37754.

sehlen:
@JohnCremona I think I haven't answered your question yet how I determine which Fourier expansions to hide. I invented the following ad-hoc method: I look at the first 3 a(n) such that n>1 and $a(n) \neq 0$ (just made that number up). Then I look at the maximal height of the rational numbers appearing when a(n) is expressed in terms of a power basis and multiply it by the degree. If this number is >90, I hide the Fourier expansion. This gave me reasonable results experimentally. You and maybe some other should compare pages where Fourier expansions are hidden with those on www.lmfdb.org and tell me if you think it was the correct decision that my great algorithm made ;-) (maybe the average height would be better but I somehow thought that taking the max is a simpler operation and so I took that).

JohnCremona:
That sounds like a reasonable ad hoc approach.  3 is a good choice, though 2 would alsobe OK since that looks at the first nonzero coefficient after the one which is usually just "alpha".  Instead of heights and degrees you could just count the length of strings.
If we knew before fetching from the db which forms were not going to have their q-expansions displayed then we need not include those in the database fetch for the newform, and that might save time.  Does the database contain information at the modular form *space* level which could tag those whose q-expansions were fit for view?  Then, only when a user asks for a download would the coefficients actually be fetched from the db.  Just an idea.

sehlen:
@JohnCremona Sorry, my information was incorrect. I edited my answer to reflect this. I take the length of the height which seems better than just the length because an integer n takes up roughly the same space as a rational number of height n. I took 3 coefficients because I saw some examples where a(3) was sth like (a^2-1234567) but then a(4) went crazy.

AndrewVSutherland:
@sehlen I've been reviewing for the past 30 minutes and found only one tiny problem.  If you go to

http://127.0.0.1:37777/ModularForm/GL2/Q/holomorphic/5/4/

you can click on the zero-dimensional space for chi_5(4,.) to get to

http://127.0.0.1:37777/ModularForm/GL2/Q/holomorphic/5/4/4

which shows a blank page, where as it should say "There are no newforms of this weight, level and character." as it does on

http://beta.lmfdb.org/ModularForm/GL2/Q/holomorphic/5/4/4/

But otherwise things look good!  I updated emf_test_pages.py (I merged it to master so you can grab a copy if you want to run it, but please don't rebase your branch, leave it as is, there is no conflict) so it will work correctly with your new code which doesn't always display the Fourier coefficients and it runs clean with no problems (and takes about half as long as it used to).

So if you can fix this one small issue and commit the fix, I'll be happy to merge this PR to beta where we can all do more testing.

sehlen:
@AndrewVSutherland so you did not have a failing test as I had (the hidden
q-exp)? I'll look at the problem later but I'm now "trapped" as an examiner
for a while ;-)

AndrewVSutherland:
I modified emf_test_pages.py so that it won't complain about hidden 
q-exp's (is that what you were asking?).  So now all the tests triv-char 
(N,k) in [1-24]x[2,40] tests pass when I run emf_test_pages.py, and it 
is  also faster, so all is good.   The only issue is that it displays a 
blank page for some of the spaces with non-trivial character that have 
dimension zero (as explained below).

sehlen:
No, I meant one of the tests in test_emf.py. I added test_hide() and it fails:

FAIL: Test that very large Fourier coefficients are not shown.
  File "/home/ehlen/lmfdb/lmfdb/modular_forms/elliptic_modular_forms/test_emf.py", line 137, in test_hide
    assert 'The Fourier coefficients of this newform are large.' in page.data
AssertionError

———————————————————————————————————

But I don’t understand why because when I check the page by hand, I do see that text.
And the test also used to work yesterday at some point and then it just stopped working.

JohnCremona:
Is a redirect occurring?  That can happen even with removing trailing/
Test by printing the string in which you search for the sfring.

sehlen:
@JohnCremona So, the page the test sees contains in fact "Problem with Hecke orbits in the database". Huh, I don't understand. If I go to that page, then it just works and does what it is supposed to do.

AndrewVSutherland:
@sehlen After several hours of digging I found the problem.  The reason test_emf.py is failing on the call to test_hide is because the test_download function leaves your cache in an inconsistent state.  You can in reproduce this manually by running sage start-lmfdb.py locally and then in your URL bar typing:

http://127.0.0.1:37777/ModularForm/GL2/Q/holomorphic/Download/?level=25&weight=36&character=1&label=f&number=10&format=sage&download=coefficients

which will download  a coefficients file.  Then type

http://127.0.0.1:37777/ModularForm/GL2/Q/holomorphic/25/36/1/

and you will see the "Problem with Hecke orbits in the database!" error message.

Fortunately this bug has an easy fix: in lmfdb/modular_forms/elliptic_modular_forms/__init__.py just set

    use_cache = False

and the problem will go away.

We are **very lucky** to have noticed this truly insidious bug.  In live operation on the web server it would only show up if a request for a particular space happened to be handled by a worker that just happened to have been asked (possibly by another user) to download the coefficients for that form before it was ever asked to display that space or form.  You might suppose that this can't happen, since the user has now way to download the coefficients before displaying the space (short of typing in the URL), but of course it can, because the workers that handle the request to display the space or the form need not be the same as the one handling the download.  In live operation this bug would be completely non-deterministic and very difficult to reproduce, it is only because when we test locally we feed URLs to a single worker that is processing them serially.
    
This highlights another reason not to use caching.  Not only does it not help (as discussed earlier in this thread), but it can lead to subtle bugs that are almost impossible to track down.  So please ,just set use_cache = False (with a comment to warning future generations of programmers not to change it) rather than wasting time and energy trying to track down exactly what the download code is doing wrong  -- I'm, sure you can fix this particular bug, but there could well be others of a similar nature and these types of bugs are very difficult to find.  It is preferable for server-side code to be as stateless as possible.

In the longer term, I would suggest removing the caching code entirely and explicitly cleaning up any non-local objects created while processing a request; as it stands we have a potential memory leak (our current settings kill workers after 100 requests, so this shouldn't be a problem for the moment).

I am very glad that you thought to implement the test_hide function, and I'm also very glad that the word "hide" happens to come after the word "download" in the dictionary, otherwise we surely never would have noticed this.

sehlen:
Wow, thanks, @AndrewVSutherland for digging into this and I'm very sorry that you had to spend so much time on this when I didn't have enough time to do it myself...

I did change change the caching setting and also found the bug that prevented the pages of dimension 0 to render correctly. This is now corrected and afaik this PR can finally go to beta. (I also just merged with lmfdb/master and pushed the result)

AndrewVSutherland:
@sehlen Thanks for the updates, everything looks good, I'm going to go ahead and merge.

Regarding the data, is it safe to copy the 11 collections listed in https://github.com/LMFDB/lmfdb/issues/1376#issuecomment-219427012 to the production system now (I just want to make sure you are not updating them right now, it will take about an hour during which time they should not be updated).

sehlen:
@AndrewVSutherland Please wait a few minutes, I'll run a check and update a
few records and then we can go.

AndrewVSutherland:
OK, I'll standby until you give the word.

sehlen:
OK, you're good to go. 

I remembered that when I tested the function that creates the smaller records, I ran it on level 1 and the parameters were far from optimal. So I reran it but discovered two bugs in convenience functions (which don't affect the production system at all but it's still good to merge the fixes so I included these commits in this PR).

sehlen:
We will work on the data again next week and then I want to finally use GCE to recompute the data for nontrivial characters and extend the ranges we have so far in the lmfdb. Btw: I completed the records for trivial character and weight<=12, level <=100 in the meantime and also extended the range for which we have weight <=40 up to 25 instead of 24. Not that this takes very long but I just took these computations as test cases to see if now everything runs smoothly. It is possible that there are some cases of "wrong data" left, I have to test for that again more systematically next week (or delete everything that's unverified and recompute, it should not take too long, I hope).

AndrewVSutherland:
@sehlen, @edgarcosta OK, data is being copied now

edgarcosta:
Roger that

AndrewVSutherland:
@edgarcosta 

Need more disk space, 100G on /mnt/tmp is not enough

edgarcosta:
2min.

edgarcosta:
I believe that 250GB should be enough, if it is not I can easily resize it.

AndrewVSutherland:
OK, restarting (I agree 250GB should be just enough)

AndrewVSutherland:
Data is now on www.lmfdb.org (still running old software).  Will confirm new software is working on beta and then push to prod.

AndrewVSutherland:
This PR has now been pushed to both beta and prod.  It looks good!
Will do more extensive testing tomorrow.

PR #1347, opened 2016-05-14
updated mod\ell gal rep homepage

alinabucur:
added knowls, changed knowl identifiers and made small stylistic changes

PR #1330, opened 2016-05-13
Mod l modular forms more fixes

sanni85:
Fixed urls and some more data formatting

http://localhost:37777/ModularForm/GL2/mod_l/11.2.9.8.1.1

PR #1414, opened 2016-05-19
1330+1341+1347

JohnCremona:
This single pull reinstates the commits which had been merged from PRs #1330, #1341, #1347 (two from @sanni85 and one from @alinabucur , a total of 5+7+1 commits).

SInce these had already been merged the only issue is whether merging them will cause any problems.  I cannot see how it can.  Apart from an import line added to website.py everything else affected here is in one of the two directories modlmf or rep_galois_modl.

### DIFF
--- a/lmfdb/import_scripts/rep_galois_modl/rep_galois_modl.py
+++ b/lmfdb/import_scripts/rep_galois_modl/rep_galois_modl.py
@@ -1,0 +1,117 @@
+# -*- coding: utf-8 -*-
+r""" Import mod l modular forms.  
+
+Note: This code can be run on all files in any order. Even if you 
+rerun this code on previously entered files, it should have no affect.  
+This code checks if the entry exists, if so returns that and updates 
+with new information. If the entry does not exist then it creates it 
+and returns that.
+
+"""
+
+import sys, time
+import re
+import json
+import sage.all
+from sage.all import os
+from sage.all import Integer, ZZ, QQ, PolynomialRing, NumberField, CyclotomicField, latex, AbelianGroup, polygen, euler_phi, latex, matrix, srange, PowerSeriesRing, sqrt
+from lmfdb.base import getDBConnection
+
+C= getDBConnection()
+import yaml
+pw_dict = yaml.load(open(os.path.join(os.getcwd(), "passwords.yaml")))
+username = pw_dict['data']['username']
+password = pw_dict['data']['password']
+C['mod_l_galois'].authenticate('editor', password)
+reps = C['mod_l_galois'].reps
+
+
+saving = True 
+
+def sd(f):
+  for k in f.keys():
+    print '%s ---> %s'%(k, f[k])
+
+def makels(li):
+  li2 = [str(x) for x in li]
+  return ','.join(li2)
+
+def string2list(s):
+  s = str(s)
+  if s=='': return []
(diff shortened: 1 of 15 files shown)


JohnCremona:
The new pages are invisible (no links from anywhere else) but can be seen at 
ModularForm/GL2/ModL/
and
/Representation/Galois/ModL
and accepting this PR does *not* require that we think that these are ready for release!

AndrewVSutherland:
Reviewing now, looks good so far.

AndrewVSutherland:
OK, I don't see anything obviously broken.  The mod-l Galois representation pages are obviously still in a very early stage, but what is there seems to work (and I'm psyched to see it, this will be very cool when it is finished).  The mod-l modular forms pages are further along (they clearly still need some more work, but I'm just checking that nothing that is there is broken).

I noticed one small bug which I am noting for @sanni85's  benefit;  there is no reason this needs to be fixed before merging this pull request (which I would like to do just to get us completely caught up).

If you go to http://127.0.0.1:37777/ModularForm/GL2/ModL/ and set the group to Gamma_1 on the results page it gets reset to Gamma_0.

JohnCremona:
Sanity check: I did a diff between (1) this commit and (2) master with three branches merged (samuele's mod_l_galrep and mod_l_modularforms, and alina's modlerp) and there are no diffs (files identical).

AndrewVSutherland:
Great!  Should we still wait for Samuele to check in?

JohnCremona:
I now don't think we need to, since I have checked what I would have expected him to.

AndrewVSutherland:
Merged to master and beta, will merge to prod in 15 minutes

JohnCremona:
Now showing at a beta near you!

sanni85:
For me it looks exactly at the state it was before I opened the last pull request, so it is great!

There is lot of work to be done, but I do not want to it all myself, and this way other people can help.

The mod-l Galois representation pages are in a very early stage (the version you see was made between 6am and 10am on Friday, including breakfast time and shuttle from the hotel to AIM), but I agree with @AndrewVSutherland they will be very cool when finished. Soon or later we should discuss about classification of subgroups in $GL_n(\mathbb{F}_\ell)$ since it could be very useful in this context. For sure I want to use the classification for subgroups of $GL_2(\mathbb{F}_\ell)$ which already appears in the elliptic curves pages. There is already a discussion going on on characters modulo $\ell$.

I have worked more  on the mod-l modular forms pages. The switch Gamma_0 Gamma_1 does not work, it is only in the html and not implemented in the code, the same is true for the search for q-expansion coefficients. All the other searches work. For the download options: the one after search is not working yet, the one on the single page does work, but it needs a couple of extra lines for the definition of the finite field.

JohnCremona:
WHile we are happy that these exciting new sections are in the pipeline I
agree with Drew that it would be good for many more people to take an
interest in improving what we already have (including bug-fixes) and less
on starting up new things.  David was very keen that the AIM workshop
should all be about starting new things despite the fact that it was held
in release week.

Since starting elliptic curves over number fields 2 years ago I have not
taken any time at all to work on new things which I want to have in (as
does David) such as Bianchi modular forms which have been in limbo for >2
years) since Ispend *all* my LMFDB time making improvements and bug-fixes.
It is less exciting but someone has to.

John

PR #1415, opened 2016-05-19
master -> beta

AndrewVSutherland:


JohnCremona:
Fantastic.  In 10 minutes we'll see this at beta.lmfdb.org.

Issue #1412, opened 2016-05-19
HMF bug in Sage download code


JohnCremona:
From sage-support (David Loeffler):
-  if you try to download HMF data to Sage, you get an error, because the Sage script that's created contains the lines

hecke_pol = x^2 - x - 1
K.<e> = NumberField(heckePol)

and there is no such variable as "heckePol". Evidently it should be "hecke_pol" in the second line.

JohnCremona:
Fixed in PR #1416

JohnCremona:
Closing since #1416 is merged.

jvoight:
Sorry, you're right, I only tested this for dimension 1.  Thanks for fixing this JC!

PR #1416, opened 2016-05-19
Hmf sage

JohnCremona:
Fixes for #1412 -- after fixing the first one, others were revealed.  To test this you need an example where the Hecke field is not Q and where there are Atkin-Lehner eigenvalues, such as ModularForm/GL2/TotallyReal/2.2.5.1/holomorphic/2.2.5.1-121.3-a

I did run the Sage code and it looks fine (clearly no-one else ever had for dimension>1).  The Magma code also looks good (no changes there).


AndrewVSutherland:
Reviewing now.

Issue #1411, opened 2016-05-19
HMF pages should show narrow class number not just class number


JohnCremona:
From sage-support (David Loeffler):

-  the page for each Hilbert modular form specifies the class number of the field; but for HMF's it is the size of the *narrow* class group that is important (e.g. the narrow class number tells you how many components the Hilbert modular variety has), while the usual class number is less relevant. Can we have narrow class numbers on HMF pages somewhere?

@jwj61 Do number fields in the LMFDB store a narrow
class number?  We avoid computing such things on the fly.  If so
it's a trivial change.  If not, could we add them?

jvoight:
Right now, number fields do not have narrow class numbers.  I added them just for the fields in hmf_fields, so as a temporary measure we could display those!  But they won't appear in the knowls, or when you go to browse the corresponding number field page.  See #994.

JohnCremona:
Actually that need not be a temporary measure, since I had (temporarily) forgotten that HMFs had their own number fields collection and these are all they need.  I'll implement what was requested using the values you added which I are 1, 2, 3, 4, 5, 6 for 171, 183, 1, 40, 1, 4 fields respectively:

sage: [(n,fields.find({'narrow_class_no' : n}).count()) for n in range(1,10)]
[(1, 171), (2, 183), (3, 1), (4, 40), (5, 1), (6, 4), (7, 0), (8, 0), (9, 0)]

JohnCremona:
I have done this (it was easy) but before I push, do we want to show only the narrow class number on each HMF's page or both?
We'll need to add knowls for "narrow class group" and "narrow class number".  That will be easiest when this is on beta.

JohnCremona:
David L says "I'm not sure whether the usual (wide) class number is useful, but it can't hurt to provide both, I guess." so I will add it.

JohnCremona:
Pull request #1417 is now up for review.  I added the knowls required.

JohnCremona:
Closing since #1417 is merged

jvoight:
Thanks JC!  You rock.

PR #1417, opened 2016-05-19
show narrow class no on HMF home pages

JohnCremona:
As requested by a user, see issue #1411.   This version shows the ordinary class number as well as the narrow one.  Two new knowls will be required (narrow class group and narrow class number).


AndrewVSutherland:
Reviewing now

Issue #958, opened 2016-03-31
Classical mf: sage/magma/gp code


AurelPage:
From /roadmap: Show sage/magma/gp commands for classical modular forms.

AndrewVSutherland:
A basic implementation of this is now up (for sage, and on the space pages, magma).

PR #1433, opened 2016-05-21
Sage and magma code

sehlen:
I added some basic commands for sage and magma to the modular forms pages, 
fixing https://github.com/LMFDB/lmfdb/issues/958.
I have no idea about gp, although I've heard that they do have an awesome implementation now but it is not in the stable branch if I see it correctly, so we should wait.

See for instance
- ModularForm/GL2/Q/holomorphic/14/8/1/ (triv char)
- /ModularForm/GL2/Q/holomorphic/14/8/9/ (non-triv char)
- ModularForm/GL2/Q/holomorphic/14/8/1/c/ (an individual form)

A few remarks:
- Commands for magma are limited to the generation of the list of newforms because afaik the ordering in magma ob the galois orbits can sometimes differ from our convention (always use the traces of the ap's) and we should compute the magma order probably and store it in the db to offer commands for identifying the individual orbit
- I used the builtin Dirichlet characters and not the Dirichletgroup_conrey because I think the user profits more from not having to install a package. I identify the character using the exponents. See any example for non-trivial character.
- Sage often does not use the exact same number field as we do (well, we do use sage but we use the eigenvalues coming from compact_system_of_eigenvalues() and they often use a different generator for the number field than the output of the Newforms function but I didn't want to go through all of those commands. Maybe my warning is a bit verbose on the page, feel free to make a proposal how to change that.


AndrewVSutherland:
Reviewing now

JohnCremona:
@nmascot @AurelPage You are our bridge to the new pari capabilities.  I know it is not yet released but could you at least see if with the new code some equivalent functionality could be provided?  It would be fine if (for now) the code downloaded started with a comment to the effect that it required <something> to run, just as Aurel did for the ecnf code (which requires some new pari version).

AndrewVSutherland:
@sehlen Not related to this PR, but while reviewing I noticed a typo on line 533 of lmfdb/modular_forms/elliptic_modular_forms/backend/web_object.py; I believe the variable r being passed to ValueError should be m.  I don't want to fix this, since I know you are making other changes to the code (adding random_object).  Can you fix this in whatever PR you submit next (no need to update this PR, since it doesn't touch this file anyway).

sehlen:
Will do. Thanks for noticing.

AndrewVSutherland:
@sehlen Here is some feedback:

- center "Show commands for: sagemath, magma" and put it at the top of the page (as on the ellitpic curve and genus 2 curve pages).
- on the space pages (e.g. http://127.0.0.1:37777/ModularForm/GL2/Q/holomorphic/13/4/1/?group=0) you do not give sage/magma code to get the coefficient field
- in most places you do not give magma code; I presume this is due to difficulty in matching the forms (and characters).  In principle this should be doable, but I don't think it is a priority right now, I suggest removing the magma option for the moment rather than putting in partial support.
- The download option is currently for sage; it should be labelled as such (as on the elliptic curve and genus 2 curve pages, for example).  Here it would not be hard to also give a magma option (but again I don't think this is a priority, getting the data cleaned up is more important)

sehlen:
Thanks for the feedback!

> On May 21, 2016, at 11:29, Andrew Sutherland <notifications@github.com> wrote:
> 
> @sehlen <https://github.com/sehlen> Here is some feedback:
> 
> center "Show commands for: sagemath, magma" and put it at the top of the page (as on the ellitpic curve and genus 2 curve pages).
I copied it from number fields which uses a right aligned layout. I actually like it but I can change it if this is the “new” standard.

> on the space pages (e.g. http://127.0.0.1:37777/ModularForm/GL2/Q/holomorphic/13/4/1/?group=0 <http://127.0.0.1:37777/ModularForm/GL2/Q/holomorphic/13/4/1/?group=0>) you do not give sage/magma code to get the coefficient field
True, I thought if would be a bit too much on this page, no?
> in most places you do not give magma code; I presume this is due to difficulty in matching the forms (and characters). In principle this should be doable, but I don't think it is a priority right now, I suggest removing the magma option for the moment rather than putting in partial support.
That’s all due to the possibly different ordering in Magma. The characters are easily identified and I think it is worth having the partial support because a magma user can easily get the same space of newforms in magma this way and match the forms if needed.
> The download option is currently for sage; it should be labelled as such (as on the elliptic curve and genus 2 curve pages, for example). Here it would not be hard to also give a magma option (but again I don't think this is a priority, getting the data cleaned up is more important)

Magma code requires a bit thinking for nontrivial characters as the output will give all galois conjugates if I see it correctly (i.e. also of the conjugate characters) and I or someone else has to check which one is which (this probably goes along with having records for the Galois conjugates in the lmfdb as well instead of just representatives if we want).

AndrewVSutherland:
@sehlen, @JohnCremona, @jwj61 Regarding the layout style for "show commands" (in sage/magma/...) There are currently 6 html pages with "show commands" in them.  They all put it at the top (so modular forms should definitely do the same).  Otherwise there is a lot of variation, and I suggest we pick one format and make it the standard.  Here is what we currently have:

- Dirichlet characters (e.g. http://www.lmfdb.org/Character/Dirichlet/13/2) has "show commands" in the center and code in fixed-width font on blue panel.
- elliptic curve over Q (e.g. http://www.lmfdb.org/EllipticCurve/Q/15/a/1) has "show commands" in the center and code in proportional font on blue panel
- elliptic curve over Q isogeny class (e.g. http://www.lmfdb.org/EllipticCurve/Q/15/a/) has "show commands" on the right and code in fixed-width font with no panel
- elliptic curve over number field (e.g. http://www.lmfdb.org/EllipticCurve/2.2.209.1/50.3/g/2) has "show commands" on the right and code in proportional font on blue panel
- genus 2 curve over Q (e.g.http://lmfdb.org/Genus2Curve/Q/169/a/169/1) has "show commands" in the center and code in proportional font on blue panel
- number fields (e.g. http://lmfdb.org/NumberField/2.2.13.1) has "show commands" on the right and code in proportional font on blue panel

I think showing code in fixed width font on a blue panel (e.g. as with Dirichlet characters) looks the best.  I'm agnostic as to whether "show commands" should be on the right or in the center (currently we center it in 3 places and right justify it in 3 places), lets just pick one and go with it.

Opinions?  Once we agree I'll open a new issue to change the six cases above to conform to whatever we choose as the standard style.

jwj61:
Wow, that is a lot of variation.

I agree with uniformizing it, and fix-width font certainly makes sense, and the blue panel is most common (and fine with me).

My opinion: "show commands" on the right looks better than centered.  Arguably it's position should be controlled by a css class.

There is also variation of "sage" vs "sagemath" and "pari/gp" vs "pari".  I don't know which is better for sage; for pari I think pari/gp is better than pari since these are gp commands.

AndrewVSutherland:
OK, so I would propose that we adopt the style on http://www.lmfdb.org/Character/Dirichlet/13/2, except replace "pari" with "pari/gp" as @jwj61 suggests (and use "sage" rather than "sagemath").  So this means fixed-width font on blue panel with "show commands" right-justified at the top.

Eventually we should certainly put this into css somewhere, but since most pages don't yet have this it's probably easier for the moment to just change the pages we have (and then anyone cloning any of them will at least do the same thing).  If there are no objections I will go ahead and merge this pull request and then submit a new pull request that makes the style consistent everywhere (7 places).

@JohnCremona do you have any comments/opinions before I proceed?

JohnCremona:
Yes: the official name for the software we still call Sage usually is SageMath.  So we should call it that.

From http://pari.math.u-bordeaux.fr/ it would appear that PARI/GP is the full name, PARI refers to the library, gp to the interactive software and GP to the language of gp!  I think that means that we should use 'gp' since we are giving commands to b used in a gp session.
If in doubt we should ask (respectively) William Stein and Bill Allombert.

I agree that css is the proper way to make sure that these are in the same place on each page, but...

AndrewVSutherland:
@JohnCremona OK, sage --> sagemath and pari --> pari/gp.  I'm going to merge this PR and then submit a new PR that makes everything conform to the same standard and then you and/or @jwj61 can review it.

PR #1439, opened 2016-05-22
Precomputing extra data for elliptic curves/Q, and using it

JohnCremona:
I think this is ready for review now.  Some of the code here in the revised import script is for the one-off creation of a new collections called curves2 with the same data as in curves but with additional fields for each curve to avoid computation on the fly.  This has been run so at Warwick both collections exist.

The live code gets its data from curves 2 if it exists otherwise from curves.  This is a temporary measure and once the new collection is in the cloud too that test can be removed (we will eventually rename curves to something else and rename curves2 to curves).

Independently of that, if the extra data cannot be read from the database (whatever the collection was called) it is still recomputed on the fly.  I put that in for testing and while I had not yet converted all the data.  It is harmless, but also could be removed later.

The new collection is fully indexed.  I have a list of the indices created which I will put in the correct place in the inventory, as well as updating the inventory, when this is finished.  Anyone is welcome to see it.

Finally: even with the new data we still create the Sage elliptic curve just to create the plot.  I see no way round this and do not think it costs a lot.  One plan was to make each curve's home page load before the plot was ready, as we do for L-function plots, but that would take me some time to work out how to do and I don't want to delay the rest of this any more (I did most of it a week ago).  We could also adapt the plot function to simply take the 5-tuple of coefficients as input.  I have not seriously considered precomputing and storing all the plots, but that is certainly an option.

### DIFF
--- a/lmfdb/ecnf/hmf_check_find.py
+++ b/lmfdb/ecnf/hmf_check_find.py
@@ -16,9 +16,9 @@ from lmfdb.base import getDBConnection
 from sage.rings.all import ZZ, QQ
 from sage.databases.cremona import cremona_to_lmfdb
 
-from lmfdb.website import DEFAULT_DB_PORT as dbport
-from pymongo.mongo_client import MongoClient
-C= MongoClient(port=dbport)
+from lmfdb.base import getDBConnection
+print "getting connection"
+C= getDBConnection()
 C['admin'].authenticate('lmfdb', 'lmfdb')
 
 print "authenticating on the elliptic_curves database"
--- a/lmfdb/ecnf/import_ecnf_data.py
+++ b/lmfdb/ecnf/import_ecnf_data.py
@@ -72,9 +72,9 @@ from sage.rings.all import ZZ, QQ
 from sage.databases.cremona import cremona_to_lmfdb
 from lmfdb.ecnf.ecnf_stats import field_data
 
-from lmfdb.website import DEFAULT_DB_PORT as dbport
-from pymongo.mongo_client import MongoClient
-C= MongoClient(port=dbport)
+from lmfdb.base import getDBConnection
+print "getting connection"
+C= getDBConnection()
 
 print "authenticating on the elliptic_curves database"
 import yaml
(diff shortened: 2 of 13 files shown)


JohnCremona:
One more thing: I had to slightly refactor the ec import script to properly use the getDBCollection() function, and while I was at it I made the same change in a few other places so I would not forget.  That is all in the 4th commit.

AndrewVSutherland:
Reviewing now.

AndrewVSutherland:
So I poked around a bit and did not see any obvious problems, I think it makes sense to merge and push to beta (but do more testing there before pushing to prod).

AndrewVSutherland:
@JohnCremona  I see there are conflicts that need to be resolved (no doubt caused by the "show commands" change, I was hoping they were far enough apart for git to be able to handle).

Do you want to take a look at these?

JohnCremona:
@AndrewVSutherland OK I'll look at it -- not your job!

JohnCremona:
OK I fixed that -- just one small thing.  In the section Modular Form (in the page of a single curve) I had added the label of the modular form in the title.  I am not sure that this is a good idea (comments? include the label but not in the title of the section? keep it as it was?) but that was on a line adjacent to a code snippet, hence the conflict.

AndrewVSutherland:
@JohnCremona I trust it is safe to copy curves2 to the cloud now?

JohnCremona:
Yes.  I think we can merge master to beta now (it is 74 commits ahead)?  And then to prod, possibly after the curves2 transfer (though it should work OK without)?  Or are there other reasons for delaying either of these merges?

AndrewVSutherland:
curves2 transfered and master pushed to beta.  Let's test and if all goes well, plan to push to prod later today.

JohnCremona:
On 24 May 2016 at 20:07, Andrew Sutherland <notifications@github.com> wrote:

> curves2 transfered and master pushed to beta. Let's test and if all goes
> well, plan to push to prod later today.
>

Great.   Is renaming a collection harmless?  I was thing of renaming curves
to old-curves to make sure that the new code on bet is happy (as it should
be).  But I don't want to trigger a lot of useless replication frenzy.

John

AndrewVSutherland:
This should be harmless (the rename operation will be replicated but the 
data won't be copied).  In any case its not that much data anyway (about 
3GB).

edgarcosta:
renaming a collection should be instantaneous on all the servers, as the servers replicate the command log, and not the data.

JohnCremona:
@edgarcosta Thanks for confirming that.,

PR #1441, opened 2016-05-22
Changes to allow us to use an index on ramified primes

jwj61:
Set the "all" toggle for ramiified primes when searching for number fields, and search.

This should work with the warwick database -- you need to move the numberfield database to the cloud before moving to prod because it involves a new "field", ramps_all, in the database, and two new indexes (ramps_all, and (degree, ramps_all)).

As a side note, the data has a few more number fields.  A bigger upload I mentioned elsewhere is on whole while I revise the preparation/upload process to make use of insert_many.

### DIFF
--- a/lmfdb/number_fields/number_field.py
+++ b/lmfdb/number_fields/number_field.py
@@ -455,9 +455,10 @@ def number_field_search(**args):
         parse_primes(info,query,'ur_primes',name='Unramified primes',qfield='ramps',mode='complement',to_string=True)
         if 'ram_quantifier' in info and str(info['ram_quantifier']) == 'some':
             mode = 'append'
+            parse_primes(info,query,'ram_primes','ramified primes','ramps',mode,to_string=True)
         else:
-            mode = 'exact'
-        parse_primes(info,query,'ram_primes','ramified primes','ramps',mode,to_string=True)
+            mode = 'liststring'
+            parse_primes(info,query,'ram_primes','ramified primes','ramps_all',mode,to_string=True)
     except ValueError:
         return search_input_error(info, bread)
     count = parse_count(info)
--- a/lmfdb/number_fields/number_field.py
+++ b/lmfdb/number_fields/number_field.py
@@ -458,7 +458,7 @@ def number_field_search(**args):
             parse_primes(info,query,'ram_primes','ramified primes','ramps',mode,to_string=True)
         else:
             mode = 'liststring'
-            parse_primes(info,query,'ram_primes','ramified primes','ramps_all',mode,to_string=True)
+            parse_primes(info,query,'ram_primes','ramified primes','ramps_all',mode)
     except ValueError:
         return search_input_error(info, bread)
     count = parse_count(info)
--- a/lmfdb/search_parsing.py
+++ b/lmfdb/search_parsing.py
@@ -253,6 +253,9 @@ def parse_primes(inp, query, qfield, mode=None, to_string=False):
             primes = [str(p) for p in primes]
         if mode == 'complement':
             query[qfield] = {"$nin": primes}
+        elif mode == 'liststring':
+            sprimes = sorted(primes)
+            query[qfield] = ",".join(sprimes)
         elif mode == 'exact':
             query[qfield] = sorted(primes)
         elif mode == "append":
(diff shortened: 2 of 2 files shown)


JohnCremona:
I tested this and it works fine.  I cannot say that I have worked through th logic of the code, so it would help if you could give some sample use cases.

Can I do the following (which I actually wanted to do):   find all S3 extensions of Q unramified outside {2,3,23}?  After putting S3 into the Galois group box and 2,3,23 into the ramified primes box, setting the switch to "all" gives no hits -- though there are certainly such fields in the database ramified at precisely these 3 primes, e.g. 6.0.120891312.1.  Putting "some" yields 248 fields which appear to be ramified at these 3 primes as well as some others.

I think this shows that the search page options, even now, are hard to use!  The "unramified primes" box is clear, I think:   do not list any fields which are ramified at an primes in this list.  For the "ramified primes" box the user might want one of at least three things:  to see only fields which are ramified at exactly the primes given;  fields whose ramified primes are contained in the primes given (that is what I wanted);  fields whose ramified primes include all the primes given.  Those could be labelled "contained in", "exactly", "include".

Is that possible?

AndrewVSutherland:
@JohnCremona Does it make sense to also merge this and push to beta so we can test the performance impact?  Or do you want to wait for John to response to your question about "subset/equal/superset" (which I agree is a good idea the word "some" is ambiguous, but this is not something that changed with this PR, it has been that way for a long time).

jwj61:
There was a bug in the searching, which is why it missed the search for S3 fields ramified at {2,3,23}.

The feature request of having unramified outside a set should be a separate PR.

JohnCremona:
I agree on making more options a separate PR.   I am still getting no results for S3 fields with "all" 2, 3, 23 ramified.  I am running the last commit (2657b5d).  Am I doing something stupid?

jwj61:
I am not sure.  When I do the same, I see 20 of the 44 matching fields (if I don't specify degree), and 20 of 22 fields if I specify degree=3 as well.

jwj61:
Maybe delete your test branch and recreate it.

JohnCremona:
I am all confused now and must go to a seminar.....for a start I thought that group S3 implied degree 6!

AndrewVSutherland:
I just pulled the new PR and tested it locally, and I get the same result as @jwj61, I think this is OK now and will go ahead and merge. (@JohnCremona Galois group S3 implies the normal closure as degree 6)

AndrewVSutherland:
@jwj61 (cc @edgardcosta) Is it safe for me to copy the numberfields database to the cloud now?  Are you done adding fields for the moment?

I think I would actually be happier if we moved the data before merging this PR so that an inadvertant push to prod won't break anything.

edgarcosta:
On the cloud side, if you follow the same procedure as before, everything
should go smooth.

AndrewVSutherland:
Yup, just giving you a heads up.  I wrote an update-numberfields.sh 
script (simplified version of the one I used for modular forms) that I 
am testing on ms1 now.  Even though it does not matter in this case, I 
decided to do it at the collection level rather than the database level 
for consistency (we'll need to do this for elliptic_curves soon too).

jwj61:
I was away from my computer.  I am not currently touching the number field database.

AndrewVSutherland:
@jwj61 I will begin the copy shortly, I'm guessing it will take 15-20 
minutes.  I'll let you know when it is done.

AndrewVSutherland:
@jwj61 I'm done reading the numberfields database on lmfdb.warwick.ac.uk, its safe for you to make updates again.  The new data is still in the process of loading to the cloud.

JohnCremona:
In my rush to re-test before my seminar I had entered C3 instead of S3, which certainly does not interact well with degree 6....  sorry about that, I am happy with what I see now.
I understand that we a re waiting for some data to move before merging this?

AndrewVSutherland:
The new numberfields data is now on the cloud.  Let me just sanity check a few things and then I think we are good to go.

AndrewVSutherland:
@JohnCremona OK everything looks good on www.lmfdb.org, and the new ramps_all field is now present (and indexed) in numberfields.fields I'm going to go ahead and merge.

JohnCremona:
+1

PR #1445, opened 2016-05-24
uniformize "show commands using" appearance and implementation

AndrewVSutherland:
Addresses the non-uniformity of "Show commands" code that was implemented in eight similar but all slightly different ways: see https://github.com/LMFDB/lmfdb/pull/1433#issuecomment-220806447 for a list of six example pages that you can compare side-by-side, and there are two new modular forms pages that use show commands as of PR #1433:

http://127.0.0.1:37777/ModularForm/GL2/Q/holomorphic/3/6/1/
http://127.0.0.1:37777/ModularForm/GL2/Q/holomorphic/3/6/1/a/

(note PR #1433 was never pushed to beta, I wanted to get this PR in first).

This was achieved by moving almost all the html code into homepage.html, so that it is implemented in just one place.  The macro place_code defined in homepage.html can then be used to place code snippets where ever they are needed.

I also fixed a few bugs I noticed along the way (e.g. the display code for Galois orbits of Dirichlet characters was broken), and cleaned up some inconsistent formatting in some of the code snippets.

Aside from improving the appearance (and getting rid of a few hundred lines of duplicated code), there should be no functional changes.  One thing worth noting is that the existing implementation of code snippets in the elliptic_curves and ecnf pages uses a lazy evaluation approach (it doesn't actually create code snippets until they are needed), with this PR this is no longer the case.  This has no impact on performance -- the code snippets are generally not much longer than the lazy macros that server as stand-ins (page sizes changed by less than 1 percent in the cases I measured), and in either case the time to display them is completely negligible compared to the work mathjax is doing to render the rest of the page.

I ran the full test suite on atkin and (after fixing test_root.py) everything passed except for test_acknowledgements which fails when attempting to connect to hobbes.la.asu.edu (known temporary problem).

### DIFF
--- a/lmfdb/WebCharacter.py
+++ b/lmfdb/WebCharacter.py
@@ -298,7 +298,7 @@ class WebDirichlet(WebCharObject):
         if self.modlabel:
             self.modulus = m = int(self.modlabel)
             self.H = H = DirichletGroup_conrey(m)
-        self.credit = 'Sage'
+        self.credit = 'SageMath'
         self.codelangs = ('pari', 'sage')
 
     def _char_desc(self, c, mod=None, prim=None):
@@ -454,7 +454,7 @@ class WebHecke(WebCharObject):
         #self.number = lmfdb_label2hecke(self.numlabel)
         # make this canonical
         self.modlabel = self.ideal2label(self._modulus)
-        self.credit = "Pari, Sage"
+        self.credit = "Pari, SageMath"
         self.codelangs = ('pari', 'sage')
         self.parity = None
         logger.debug('###### WebHeckeComputed')
@@ -599,7 +599,7 @@ class WebHecke(WebCharObject):
 class WebCharFamily(WebCharObject):
     """ compute first groups """
     _keys = [ 'title', 'credit', 'codelangs', 'type', 'nf', 'nflabel',
-            'nfpol', 'codeinit', 'headers', 'contents' ]   
+            'nfpol', 'code', 'headers', 'contents' ]   
     headers = [ 'modulus', 'order', 'structure', 'first characters' ]
 
     def __init__(self, **args):
@@ -886,7 +886,7 @@ class WebDirichletGroup(WebCharGroup, WebDirichlet):
     @property
     def codeinit(self):
         return {
-                'sage': 'H = DirichletGroup_conrey(%i)\n'%(self.modulus),
+                'sage': 'H = DirichletGroup_conrey(%i)'%(self.modulus),
                 'pari': 'g = idealstar(,%i,2)'%(self.modulus)
                 }
 
@@ -914,7 +914,7 @@ class WebSmallDirichletGroup(WebDirichletGroup):
         if self.modlabel:
             self.modulus = m = int(self.modlabel)
             self.H = Integers(m).unit_group()
-        self.credit = 'Sage'
+        self.credit = 'SageMath'
         self.codelangs = ('pari', 'sage')
 
     @property
@@ -1019,12 +1019,11 @@ class WebSmallDirichletCharacter(WebChar, WebDirichlet):
             return { 'sage': 'kronecker_character(%i)'%m }
         return None
 
-
     @property
     def codegaloisorbit(self):
         return { 'sage': 'chi_sage.galois_orbit()',
                  'pari': [
-                     '[mod,num,order] = [%i,%i,%i]'%(self.modulus,self.num,self.order),
+                     '[mod,num,order] = [%i,%i,%i]'%(self.modulus,self.number,self.order),
                      '[Mod(num,mod)^k | k<-[1..order-1], gcd(k,order)==1]',
                      #'order = charorder(g,chi)',
                      #'[ chi*k % order | k <-[1..order-1], gcd(k,order)==1 ]'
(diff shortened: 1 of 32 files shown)


JohnCremona:
Testing now.

JohnCremona:
Thanks a lot for doing this -- completing what I started (in a much smaller way) when I started using macros to insert code snippets set elsewhere than in the template, though I was doing it piecemeal and you have done it much more systematically.

Small point:  the lines 
      info['code'] = dict([(k[4:],info[k]) for k in info if k[0:4] == "code"])
        info['code']['show'] = { lang:'' for lang in info['codelangs'] } # use default show names
appear several times in  the file characters/main.py.   Since the first is a bit technical is might be safer to make it a macro (i.e. separate python function) defined in one place as a transformation to be done on info.

Still trying out the various page types...

JohnCremona:
Looks excellent.  I must add code snippets to ecnf isogeny class pages.  Also:

*Every page which has such code snippets available should provide a way to download a file containing all of them for the page*.  I did this for elliptic curves over Q and believe that it is useful to have everywhere.  It is possible to cut-and-paste the individual snippets but that can get quite tedious.  Another issue for this?  (Certainly I am not suggesting delaying this one.)

AndrewVSutherland:
@JohnCremona That is a good idea, I actually didn't even notice the E/Q pages had this until I was working on this PR; I agree it is a useful feature, especially since the download feature removes all the prompt, which can be annoying when cutting and pasting (Sage ignores them, but not Magma).

I suggest creating a new issue for this. I'd be happy to work on it (since this if all fresh in my head), but probably not for a few days (other things I want to get done, including reviewing your outstanding PR).

AndrewVSutherland:
Regarding https://github.com/LMFDB/lmfdb/pull/1445#issuecomment-221211871, this is really just a legacy of the way code snippets were implemented in the character pages (every code snippet is put in an attribute whose name starts with "code" rather than in a dictionary names code, which is what is done everywhere else).  I'd be happy to make this into a separate function, but I'd be even happier to get rid of it entirely by putting all the code attributes into a single dictionary.

The only reason I did not do this in this PR is that I wanted to keep the code changes as minimal as I could (I already had to touch 31 files!), and I thought it safer to get something out there that worked across the board first, and then I can go back and do some clean up to get rid of this (and a few other bits of legacy code).

AndrewVSutherland:
@JohnCremona One other thing I wanted to mention.  With the new layout, one thing that occurred to me is that it might be better to have the "Show commands" text in a box on the right sidebar (which is also where the "download code" options are for E/Q).  This would have the virtue of getting rid of some blank vertical space at the top of the center of the page (which is really "prime real estate" from a U/I perspective, a shame to waste it).  I didn't want to do this in this PR (again trying to keep the changes minimal), but it would be logical to make such a change at the same time as implementing a "download all code" option (so maybe something to include in the new issue if you agree).

pascalmolin:
This is definitely a very nice PR, thank you Drew.

Concerning the code snippets for Dirichlet characters: they appear as
methods instead of dict entries just because I designed this scheme before
John started his yaml version.
I had two things in mind to prefer methods instead of dict entries:
- class heritage to handle both Dirichlet and Hecke characters, although
this is not an issue because we could do this with a dict, I guess
- letting the class being as lazy as possible, so that one can call a
character without computing all the invariants (a problem is that
characters are called not only for their web page)
Anyway, I prefer defining these codes inside the object class instead of
static yaml because some arguments can be computed.

But now there is a unified method I will use it, I will just think a bit
more about it.


2016-05-24 11:38 GMT+02:00 Andrew Sutherland <notifications@github.com>:

AndrewVSutherland:
@pascalmolin I agree that lazy evaluation at the class/method level is a 
good idea for supporting objects that are created internally but not 
necessarily rendered on web pages.  If you look at the EC and ECNF code, 
you will see that they also use lazy evaluation, but they put their code 
snippets in a "code" dictionary rather than having separate attributes 
prefixed by "code" for each one -- this is the only change I might 
suggest for the characters code.

The EC and ECNF code also use a .yaml file (with place holders for 
values that get filled in dynamically), but I think I agree with you 
that this isn't necessarily always the best choice (I'm not saying it's 
a bad choice either).  If you look at the code_snippets function in 
web_g2c.py, it does not use a .yaml file (which it easily could) but I 
think it is still very easy to read, understand, and maintain -- I'm not 
sure there would be any benefit to using yaml there.

BTW, this PR also fixes a problem with the Galois orbit code snippet in 
the existing implementation (which I expect went unnoticed until now).

pascalmolin:
Sorry if my answer seemed reluctant, I replied in a hurry: I completely
agree with you, and I saw that the code snippets for genus 2 curves are
done exactly the way I should reproduce for Dirichlet characters.


2016-05-24 12:33 GMT+02:00 Andrew Sutherland <notifications@github.com>:

JohnCremona:
I have just caught up with this thread.  I am not at all wedded to the yaml solution, though it was a marked improvement to having all snippets for all languages hard-coded into the html template!  @jwj61 and I talked about moving the snippets into the database itself since then, retrieving them might be faster than reading from a file.
I agree with not doing more in this PR.  The extra coding to put (up to) 3 Download Code option on each page was extremely simple!  See the last 13 lines of elliptic_curve.py and the 30 preceding lines.  Should be easy to extend to the other pages which have code.

AndrewVSutherland:
@JohnCremona Great, do you want to go ahead and merge?  It might be good to clear this one quickly just because it touches so many files (if we find there is anything we want to change once we push it to beta and more people have a look, I'll be happy to make changes in another PR).

JohnCremona:
Done.  Is it about time we updated beta?

AndrewVSutherland:
Yes, please do (and I just merged the half_integral PR).

PR #1450, opened 2016-05-24
When calling polredabs on a polynomial provided by a user, call polredbest first

jwj61:
This speeds up the computation of polredabs when the coeffiicents of the given polynomial are large.  When coefficients are small, this is slower, but things are already fast in that case.  So this makes fast cases slightly slower, and slow cases.

For example, in gp,

x^5 - 331137220*x^4 + 37922047405356360*x^3 - 1127174691845938128093840*x^2 + 52208293424667465123438066822480*x - 16507431553557006099641796204889368224

takes 1/50 the time with this change.

### DIFF
--- a/lmfdb/WebNumberField.py
+++ b/lmfdb/WebNumberField.py
@@ -178,7 +178,7 @@ class WebNumberField:
         pol = PolynomialRing(QQ, 'x')(str(pol))
         pol *= pol.denominator()
         R = pol.parent()
-        pol = R(pari(pol).polredabs())
+        pol = R(pari(pol).polredbest().polredabs())
         return cls.from_coeffs([int(c) for c in pol.coefficients(sparse=False)])
 
     # If we already have the database entry


AndrewVSutherland:
One wonders why polredabs does not do this automatically, but I'm glad to see you put this in!

AurelPage:
I have asked Karim why polredabs does not do that. Here is (part of) his answer:
"It is a very bad idea in this form: useless overhead, potentially a factor 2 for an improvement that will only be sensible in trivial cases. In this particular example, the gain exclusively comes from the fact that polredabs has to completely factor the discriminant of the field, but does it the wrong way (in this example). One should use a partial factorisation + nfcertify strategy and only factor the discriminant if it fails."

AurelPage:
Last commit in development version of pari/gp:

commit dd2a64ac084fde50251b8003e881ac71bfd213f8
Author: Karim Belabas <Karim.Belabas@math.u-bordeaux1.fr>
Date:   Thu May 26 15:34:45 2016 +0200

    improve polredabs()
    
    76- polredabs([T,listP]) no longer returns 0 if the attached order cannot
       be proven to be maximal: it computes the expected canonical polynomial
       in all cases, which can be very slow. Always use polredbest() if you
       don't require a canonical output.
    77- polredabs(T) now internally uses the polredabs([T,listP]) strategy,
       making it much faster in favourable cases, while still always returning
       a canonical defining polynomial.
    
    E.g. t=x^5-331137220*x^4+37922047405356360*x^3-1127174691845938128093840*x^2+52208293424667465123438066822480*x-16507431553557006099641796204889368224
    polredabs(t) is now 100 times faster.

AndrewVSutherland:
@AurelPage Thanks for the update, this is good news.

Issue #1434, opened 2016-05-21
Random modular form


sehlen:
Other pages have a link that leads to a random object, modular forms don't but this is a nice feature and they should have it.

(The pull request that will fix this issue is coming very soon ;-)

AndrewVSutherland:
You can now access random modular forms via http://www.lmfdb.org/ModularForm/GL2/Q/holomorphic/random.  Once issue #1248 has been addressed we can link this to a button on the main page for classical modular forms (I'll leave the issue open until that happens).

AndrewVSutherland:
Addressed by #2717.  In addition to a random form link, you can also choose random newforms that match specified search criteria (e.g. a random newform with trivial character, or a random CM newform) using the "random form" button on http://www.lmfdb.org/ModularForm/GL2/Q/holomorphic/

Issue #1442, opened 2016-05-23
"No newforms" on Galois conjugate pages


jvoight:
Y'all'll probably fix this in v.1.1, but right now on lmfdb.org, if you go to
   http://lmfdb.org/ModularForm/GL2/Q/holomorphic/32/2/29/
you'll see a page that says it has no newforms in it, but it is conjugate to a space that does have newforms.  This is confusing.

AndrewVSutherland:
@jvoight Thanks for letting us know!.  I'm sure you are just trying to be diplomatic, but this isn't "confusing", it is a mathematical error in the data related to the open critical issue https://github.com/LMFDB/lmfdb/issues/1248 (and this form is within the extent we claim to have, see https://github.com/LMFDB/lmfdb/issues/873).  We are working on getting this (and all other errors in the data) fixed as quickly as possible (the fix will not wait for version 1.1, we will update the data on lmfdb.org whenever we have a new batch of verified data, we have already updated it once to fix a bunch of errors/omissions for the forms with trivial character, we are now working on the forms with non-trivial character).

So great observation (and please be sure to tell us about anything else you spot), but I don't think we should open a new issue for this.  I'll add a note in the two related issues.

jvoight:
OK, thanks for sorting this as appropriate--I wasn't sure if it was included in the sweeping net of other issues.  Is someone keeping a track in the test files of any one of these that is reported, to make sure in the corrected version that they are correct?

AndrewVSutherland:
I have a test script that runs through every space in the database that 
checks basic consistency and that all the pages load without errors, I 
am going to update it to check that the Galois conjugates listed for 
each Hecke orbit all give consistent information (i.e. display a page 
with the correct dimension or say they are conjugate to one that does).

This test script is rerun every time there is a code or data change.

PR #1453, opened 2016-05-24
Random modular form and some bugfixes

sehlen:
This resolves issue https://github.com/LMFDB/lmfdb/issues/1434.
Moreover, it includes a few bugfixes in the backend and, more importantly, corrects a mistake that showed "no newforms" on pages that are Galois conjugates of nonzero spaces in the db and in particular   nonzero (https://github.com/LMFDB/lmfdb/issues/1442#issue-156183423).


AndrewVSutherland:
@sehlen, After testing this for a while, I think it would be better if we held off making the random modular form visible until we fix the data issues, but I do not want to hold up this PR, since it fixes some other bugs as well.  Could you just remove the "Go to a random modular form from the database." line from the html template but leave everything else as is (so we can still test random modular forms if we wish using the URL http://127.0.0.1:37777/ModularForm/GL2/Q/holomorphic/random, which is useful!).

The problem is that in our current state, it is just too easy to click the random modular form button and have things break, so we don't want to push this to www.lmfdb.org.  But I do want to be able to merge all the other corrections you made (and tomorrow I do plan to push master to prod, it has a number of performance improvements to elliptic curves and number fields that we want to make available).

AndrewVSutherland:
This is superseded by https://github.com/LMFDB/lmfdb/pull/1465.  Eventually a new pull request should be submitted to add back in the random modular form capability, but this should be done after https://github.com/LMFDB/lmfdb/pull/1465 has been tested and merged and issue #1248 has been addressed.

PR #1455, opened 2016-05-25
Factor reindexing out of rewrite_collection.

jwj61:
Default behavior is the same, but now you can separate copying and reindexing.  Why?

Suppose you want to add many new records to an existing database.  You can make a copy using rewrite_collection and not build new indexes, add the entries, then create the indexes in the new database.


AndrewVSutherland:
This is a good idea!  But it looks to me like you edited an old version of the file.  Can you pull upstream master and reimplement the new functionality (or I can do it if you prefer).

jwj61:
I think I got the unintended changes.

AndrewVSutherland:
Just tested it and it looks good, thanks.  Merging.

Issue #1333, opened 2016-05-13
Change ideal labels (for conductors and levels) for imaginary quadratic fields


JohnCremona:
At present imaginary quadratic fields have a different way of labelling their ideals, which form part of the labels for elliptic curves, isogeny classes, and (in future) Bianchi modular forms.  See
http://beta.lmfdb.org/knowledge/show/ec.conductor_label

These labels have the form [N,c,d] where N, c, d are positive integers;  and the brackets and commas make for terrible URLS (try http://beta.lmfdb.org/EllipticCurve/2.0.7.1/%5B44%2C8%2C2%5D/a/3)

I want to change [N,c,d] to N.c.d so that the previous URL would become

http://beta.lmfdb.org/EllipticCurve/2.0.7.1/44.8.2/a/3

This is still different from the ideal labels for totally real fields where that ideal 44.8.2 would be 44.i with i an (arbitrary) index , but not so bad.

I want to do this before adding Bianchi modular forms to the LMFDB.  Right now nothing links to or from these Bianchi elliptic curve labels.  I'll just need to update various scripts (including external ones).  I assume there will be no objections!

AndrewVSutherland:
Definitely no objections, this is a good idea.

AndrewVSutherland:
Once this is done we can easily address #1147.

JohnCremona:
This was done ages ago.  Ideal labels used for conductors over IQFs do now all have the form N.c.d where N is the norm.  So I think this can be closed.

Issue #1267, opened 2016-05-09
Precompute much more data for elliptic curves over number fields


JohnCremona:
See #1199 .  Currently quite a lot of data displayed for al elliptic curve over number fields is not stored in the database, but computed on the fly.

Here's what happens when we make an ecnf web page:

the corresponding Sage elliptic curve object is created
its conductor is computed and factored (without the need to factor it we could of course construct the conductor from the db)
the discriminant ideal is computed and factored. This involves working out whether we have a global minimal model, of if not then at which primes it is non-minimal
the regulator of the generators is computed. (We have gens for rather few of the curves in the db and I need to be reminded to store the regulator of any points which we do have.
the local data data at all primes of bad reduction is computed from the Sage curve

My plan: add code to the ecnf upload scripts to add data fields into the database at upload time, so that the WebEllipticCurve objects (class ECNF) do not need to create Sage elliptic curves.

JohnCremona:
Now that #1318 is done (under review at least) I'll start on this one.  After #1318 I will do things a little differently which will simplify the coding and the transition.
 - 0. Work out exactly which additional data fields to precompute and store.
 - 1. Add to the current upload script (which already does some of the precomputations) to do everything and store it.  
 - 2. Run this on all my input data files (~300,000 curves) while uploading them to a brand new collection, say nfcurves2.
 - 3. Add all the indices needed to nfcurves2.
 - 4. Change the code in the ecnf module to read from nfcurves2 instead of nfcurves and use the precomputed data to avoid all computation on the fly (except the plot).
 - 5. Test this new setup thoroughly.

After that two things have to happen ideally simultaneously: (a) rename nfcurves2 to nfcurves (and nfcurves to nfcurves_old say); and (b) change the name nfcurves2 to nfcurves in the code (which should be in exactly one place).  These need to be done with care since (a) has to happen in the cloud (after copying nfcurves2 there) and in Warwick at the same time as the code changes get pushed.

I'm sure we can work out how to do the last step.  I would prefer not to do what I have done over Q which is to have code which works both with the old smaller data and the new larger data for each curve, since that was rather tedious.  but not impossible.

AndrewVSutherland:
[...]

PR #1458, opened 2016-05-25
minor changes to ecnf import file (new IQF labels), with no effect

JohnCremona:
A trivial step towards #1333.  The only file chaned is the import script for ecnf, and all but one change is in a docstring showing sample input (for which I used IQF examples since at the time that is all I had), and the remaining one is in a function I don't actually use.
Since the conductor  labels are only used as strings in the website code nothing will happen until I re-upload the curves with their new labels, and that will be done as part of #1267


AndrewVSutherland:
Looks fine to me.

PR #1461, opened 2016-05-25
Add the option to search for number fields unramified outside a set

jwj61:
This is a feature requested in the discussion of another pull request, and is (to me) the most natural way to specify ramification.

Comments, particularly on wording and placement of the ramified primes entry area are welcome.  They are on the main number field search/browse page and the search results page.

### DIFF
--- a/lmfdb/number_fields/number_field.py
+++ b/lmfdb/number_fields/number_field.py
@@ -453,9 +453,13 @@ def number_field_search(**args):
         parse_ints(info,query,'class_number')
         parse_bracketed_posints(info,query,'class_group',split=False,check_divisibility='increasing')
         parse_primes(info,query,'ur_primes',name='Unramified primes',qfield='ramps',mode='complement',to_string=True)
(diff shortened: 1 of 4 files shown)


AndrewVSutherland:
One thought on alternate wording: "subset of" "superset of" and "equal to"

And I would make "equal to" the default.  It is about 20x faster and encourages users to at least think about whether they can give this level of precision.  It's also what I would expect to happen if no options were given.

JohnCremona:
Thanks for doing this!  (it's a feature I myself wanted.)  Finding a wording which is both concise and clear is not easy.  @AndrewVSutherland 's suggestion is most precise though less concise, and I think I would want to try out both versions on the page before deciding.  I am fine with "equal to" being the default (though a pedant might then argue that when no primes are give we would be asking for no ramified primes by default...).
Grammatically these options apply to the "*set of* ramified primes", while the page says just "ramified primes", and perhaps that is why none of these wordings feels perfect.

jwj61:
Wording now adjusted to set of ramified primes, selection options suggested by Drew, and exactly the default.

PR #1463, opened 2016-05-26
simplify (and possibly speed up) random ec/Q

JohnCremona:
A small change which should make showing a random elliptic curve oever Q faster.

### DIFF
--- a/lmfdb/elliptic_curves/elliptic_curve.py
+++ b/lmfdb/elliptic_curves/elliptic_curve.py
@@ -99,12 +99,8 @@ def rational_elliptic_curves(err_args=None):
 
 @ec_page.route("/random")
 def random_curve():
-    label = random_object_from_collection( db_ec() )['label']
-    # This version leaves the word 'random' in the URL:
-    # return render_curve_webpage_by_label(label)
-    # This version uses the curve's own URL:
-    return redirect(url_for(".by_ec_label", label=label), 301)
-
+    label = random_object_from_collection( db_ec() )['lmfdb_label']
+    return render_curve_webpage_by_label(label)
 
 @ec_page.route("/curve_of_the_day")
 def todays_curve():
--- a/lmfdb/elliptic_curves/elliptic_curve.py
+++ b/lmfdb/elliptic_curves/elliptic_curve.py
@@ -100,7 +100,8 @@ def rational_elliptic_curves(err_args=None):
 @ec_page.route("/random")
 def random_curve():
     label = random_object_from_collection( db_ec() )['lmfdb_label']
-    return render_curve_webpage_by_label(label)
+    cond, iso, num = split_lmfdb_label(label)
+    return redirect(url_for(".by_triple_label", conductor=cond, iso_label=iso, number=num))
 
 @ec_page.route("/curve_of_the_day")
 def todays_curve():


AndrewVSutherland:
I found one small issue, if you click the random button it takes you to the page

http://127.0.0.1:37777/EllipticCurve/Q/random

It does not replace "random" with the url corresponding to the label of the curve.  This didn't immediately bother me (but it is a change), but then I noticed that if you click on the last bread crumb, instead of taking you to the page you are on, it takes you to a new random page (which it definitely should not do).

davidfarmer:
There should be a test that the URL of a random object does
not contain the string "random"!

Initially "random" was part of the URL, and it took a bit of discussion
to decide that we wanted the URL of the object we are looking at.

JohnCremona:
This is strange.  I did not intend that to happen, and when I run the code it does not happen!  I get a genuine curve URL with proper breadcrumbs.  **But** however many times I run this I always get http://localhost:37777/EllipticCurve/Q/33165/m/1
which is seriously strange.

AndrewVSutherland:
@JohnCremona So try this on atkin.  In one terminal do "sage start-lmfdb.py" and in another terminal do

curl -silent "http://127.0.0.1:37777/EllipticCurve/Q/random"

and you will get a full page back (not a redirect), at least that is what I am seeing when I checkout the branch JohnCremona/ec-random.  But if you instead to

curl -silent "http://www.lmfdb.org/EllipticCurve/Q/random"

You will get a redirect (as expected)

AndrewVSutherland:
One thing I noticed on the console is the message

Falling back to constructing E

JohnCremona:
@AndrewVSutherland You are right and it is a bug (two bugs in line 224 of web_ec.py) which in effect causes the efficiency saving of reading extra data from the curves2 db to be lost!  I am making an urgent patch for that.  (Not based on this one!)

JohnCremona:
@AndrewVSutherland I am running this on atkin using the system installed Sage -- are you testing there too?  If not we should perhaps check that we have the same versions of pymongo, flask, etc, since I am not seeing what you are.  (Look at atkin:37002 if you like).
More seriously in my logs I am seeing the statement printed "Falling back to constructing E" which is a left-over from earlier work (testing the new curves2 collection) and it should not be there.

jwj61:
I am not testing this (and am about to be away from a computer for the rest of the day), but the code diffs show a change from using redirect, which would change the url, to just render_curve_... which I would expect to leave the url as /random.  Maybe look to see what that line says in the code you are actually running.

AndrewVSutherland:
@JohnCremona I was initially testing on my system and then switched to Atkin and tested there.

Should I continue looking at this or are you going to update it to reflect #1466 (I'm running to lunch now, back in an hour or so).

JohnCremona:
Thanks for merging #1466 and sorry for the bugs.  I'll update this one since I don't like the mysterious behaviour.

JohnCremona:
OK, so the weirdness went away, and I have put back the redirect so there is no 'random' in the url.  There are still fewer redirects than before this patch.  It used to go from random_curve() --> by_ec_label(), which is more general than we need here since it parses Cremona and LMFDB labels for a curve or a class and just redirected in this situation to  by_triple_label().  Now we know we have an LMFDB label for a curve so can go directly to by_triple_label().
All in all I doubt this will have a noticeable effect but it is slightly cleaner so may as well go in.

AndrewVSutherland:
OK, I will review it shortly.

AndrewVSutherland:
Looks good.

Issue #1454, opened 2016-05-25
Classical modular form page won't load


AndrewVSutherland:
The page http://www.lmfdb.org/ModularForm/GL2/Q/holomorphic/23/6/2/ times out and produces a 502 error (also on beta).

sehlen:
This is caused by a large record (the file in gridfs is 50MB).
Issues like this are all over the place but are solved by creating a smaller record which I'll do systematically when updating all of the data.

However, what is strange is that when I just created a much smaller record for 23.6.2.a, it still times out on beta and loads but takes quite a while if I run it locally. Could this have to do with mongo being busy replicating my update of the dimension table?
I just stopped the update (all records that we actually do have in the db are changed anyway, the dimension table has records for high level spaces, which is ok I guess but it just seems to takes a long time to update all the records. I guess I might have to drop the indices - was that one of the things you learned from the other larger updates you and others have been working on? I haven't followed the whole discussion and maybe you can just give me a quick advice here. 

(What I want to do (as described in my other post) is to add a field to all records that are missing it, namely the space_orbit_label, which should by default be level.weight.i, where i is the smallest number appearing in the list of numbers of characters in the same Galois orbit. This had been added at some point and is a good field for indexing (I want to index it as being "unique" in fact to avoid duplicates which we have right now and having it will make it easy to remove the duplicates as the character_galois_orbit field (a list) apparently can appear with different orderings in the db!)

sehlen:
But there are certainly also some other issues here because for this particular form (the number field has degree 90 over Q), the function determining if the q-expansion should be hidden or not takes 10s ;-)
So we have to save that in the db! I will include this into the list of things I need to prepare before we recompute/fix the data.

AndrewVSutherland:
@sehlen -- The advice I gave to others was to rewrite all the records to a new collection (with no indexes on it) and to then create the indexes, and I provided a utility function to do this (see https://github.com/LMFDB/lmfdb/blob/master/data_mgt/utilities/rewrite.py).  When this has completed you can rename the collections to replace the old one with the new one (this takes no time).

[...]

PR #1465, opened 2016-05-26
Mf fixes

sehlen:
This is an alternative to https://github.com/LMFDB/lmfdb/pull/1453
which does not show the random modular form *for now* but includes the bugfixes (+1 more bugfix that does not affect the website at all).


AndrewVSutherland:
Thanks @sehlen, reviewing now.

AndrewVSutherland:
So if I go to the problematic page (noted in https://github.com/LMFDB/lmfdb/issues/1442#issue-156183423)

http://127.0.0.1:37777/ModularForm/GL2/Q/holomorphic/32/2/29

it tells me that it is conjugate to S_2(chi_32(21,*)) (and does not say the space is empty, which is good).  If I then go to

http://127.0.0.1:37777/ModularForm/GL2/Q/holomorphic/32/2/21

I get a "Problem with Hecke orbits in the database!" error message, which is maybe what I should get if there is indeed a problem in the database.  But what is confusing is that if I do the same thing on beta.lmfdb.org (so same data but different software), it shows me a *different* conjugate, i.e. if you go to

http://beta.lmfdb.org/ModularForm/GL2/Q/holomorphic/32/2/29/

it says it is conjugate to S_2(chi_32(5,*)) (note 5 not 21), and also that there are no newforms in this space (which is fixed in this PR).  But the form S_2(chi_32(5,*)) is there and displays ok:

http://beta.lmfdb.org/ModularForm/GL2/Q/holomorphic/32/2/5/

So my question is this: Is this what you expect to see, and is it correct that the Galois conjugate listed for S_2(chi_32(29,*)) should be S_2(chi_32(21,*)) (as it is now), rather than S_2(chi_32(5,*))?

This surprises me just because I would have thought it should depend only on the data, not the code.

sehlen:
That's confusing, I get redirected to character 5 and can't see a problem.

It somehow almost has to do with a problem in the dimension table (this is one of the cass where the dimension table contains a duplicate of the galois orbit of the character which I started working on two days ago and then made a mistake and got stuck for a while - I'll go back it it now and will fix all records today!) and the process not being deterministic - I haven't looked at this code in quite a while but that's what I'm going to do now to find out how this could possibly happen.

sehlen:
To see what I mean, go to /ModularForm/GL2/Q/holomorphic/32/2/ and observe that the orbit [5, 13, 21, 29] appears twice on this page (due to someone trying to insert a record for 21 apparently although it is not clear to me at all how the different ordering of the orbit was determined when that was done ;-)

sehlen:
OK, so the code that determined the representative was a bit complicated.
I rewrote it and it should be faster and given that the data is consistent it should also be deterministic for sure.

I also updated the dimension_table data completely - I rewrote it to a new collection and (re)-created indexes, renamed the old one to dimension_table_old and renamed the new one to dimension_table, so it is in use now on beta and there shouldn't be any duplicated records anymore.

In particular, after deleting the incomplete record for 32/2/21, you can navigate to all pages in this galois orbit and they all refer to 5, which loads just fine.

Note that the code works with the old dimension_table but might still in theory redirect you to the space 21 as the old dimension_table contains such a record. So, bringing this to prod after some testing should improve things for sure but it would be much better if dimension_table was copied as well.

sehlen:
Let me add that I started working on documentation along with this update.
You can learn about the dimension_table at https://github.com/sehlen/lmfdb-inventory/blob/master/db-modularforms2.md

sehlen:
@fredstro Maybe you want to check that everything I did make sense as well.
Note that I used the convention for the space_orbit_label as it used to be to not cause confusion for now.

sehlen:
Hold on, I just saw a problem.

AndrewVSutherland:
OK, I'll hold on, let me know when it is ready to review.

And I can easily copy the new dimension_table over to prod when we are ready, it isn't very big.

sehlen:
Well, the problem was a rather small one in a sense but I got sidetracked, mostly by really annoying bugs in sage.
Anyway, I decided to update this PR with everything I've done otherwise which prepares for much larger updates once I settled some more issues on the computing/verifying data side of things. 

First things first, the issue I was holding this one back is that I updated the code to follow the convention that the smallest (Conrey) number in a character Galois orbit ist always the representative for the orbit as it should have been all the time (or at least for quite a while). But the data in the DB is sometimes just not consistent with this convention. So I "fixed" this for now in the following way:

If a user clicks on a space that's not supposed to be there but is currently there, then I show the space but if the representative of a Galois orbit is not available, the page for the representative does not link to any other record in the Galois orbit, even though it might be in the DB as of now. The reason is that I don't want to introduce code that does something only for the case that data is not as its supposed to be and more importantly probably will not give a result before leading to a timeout in many cases anyway or link to wrong data.

Example (where it would be reasonable to have a link but the data issues will be resolved shortly I hope and then everything will be more consistent):

Go to ModularForm/GL2/Q/holomorphic/13/7/?group=1 and click on 13/7/5. The space is not in the DB and we give a polite error message. If you click on 13/7/8, you get to that space.
Usually, clicking on 13/7/8 would result in a message that the pace is conjugate to 13/7/5 with a link and 13/7/5 would load. This (correct) behaviour can e.g. be exhibited for the Galois orbit 2,6,7,11 on the same page.

The example given above in https://github.com/LMFDB/lmfdb/pull/1465#issuecomment-221914136 is resolved by the cleaned up dimension_table as indicated earlier. 

A few remarks about all the other changes I pulled in: first off, we should carefully test that everything works on beta and run the test script as soon as everything is on prod to make sure these larger changes really work as intended and are completely "backwards compatible".

The main point of the changes is to introduce a way to store coefficients of modular forms in a different way than what we do now. The way I implemented now is to store them as polynomials, which loads much much faster and used less space and is a bit more convenient to use than matrices (as discussed in another thread on some mailing list ;-) But the code is compatible with having different data formats coexist and thus is compatible with the current data and future updates to the data.
A change like this *has to be done* in order to make pages like the one mentioned in https://github.com/LMFDB/lmfdb/issues/1454 to load. With the current data and the current flaws/bugs in sage regarding relative number fields, a page like that will just never load. (That's not entirely true. Stripping off coefficients from the record will make it work as well. But then you will also never be able to download the sage code, for instance. Not even 10 coefficients.)
It's not perfect yet but having it tested with the existing data certainly will help.

A few other changes include bugfixes to the sage download code option and the possibility to save the coefficient complexity and some other properties (norm and trace of first a(n) with n>1 s.t. a(n)!=0) in the db, which is much faster and the only practicable way for most records in fact. Most of the data is unchanged so it does not make use of this new code but I'm preparing the data update in the background and it's good to have the code ready for this.

All tests pass and emf_test_pages passes as well for me.
@AndrewVSutherland Maybe you want to update the emf_test_pages.py to include some of the issues mentioned here in some way?

sehlen:
One more remark: After the last change that I added to the branch, one very visible improvement is the loading speed of the page /ModularForm/GL2/Q/holomorphic/83/12/1/b/ (which now has the additional properties stored in the record) vs. how it loads on the current beta, which should hold the same data.

AndrewVSutherland:
Thanks for the updates @sehlen I will test this thoroughly later this morning and update emf_test_pages.py as you suggest.

AndrewVSutherland:
@sehlen http://127.0.0.1:37777/ModularForm/GL2/Q/holomorphic/13/2/4/a/ raises a ValueError saying "Wrong format in database!" (the user just sees a server error), even though the page http://beta.lmfdb.org/ModularForm/GL2/Q/holomorphic/13/2/4/a/ appears to work.  Is this a bug or just the new code noticing a problem that the old code ignored?

AndrewVSutherland:
@sehlen If you go the page http://127.0.0.1:37777/ModularForm/GL2/Q/holomorphic/7/3/6/a/ and click the download button you get a server error caused by "AttributeError: 'RationalField_with_category' object has no attribute 'relative_degree'" (the coefficient field is Q) but this does not happen on http://beta.lmfdb.org/ModularForm/GL2/Q/holomorphic/7/3/6/a/ where the download works correctly

AndrewVSutherland:
This is not an issue with this PR, but as I was extending the test ranges I noticed that
http://127.0.0.1:37777/ModularForm/GL2/Q/holomorphic/16/2/13/ gives a "Problem with Hecke orbits in the database error".  The same thing happens on http://beta.lmfdb.org/ModularForm/GL2/Q/holomorphic/16/2/13/ and on www.lmfdb.org.

sehlen:
@AndrewVSutherland Thanks a lot for testing all of this!

The problem you mention in https://github.com/LMFDB/lmfdb/pull/1465#issuecomment-222332467 is weird because I can't reproduce it right now. Can you give me some more detailed logs?

sehlen:
The problem noticed in https://github.com/LMFDB/lmfdb/pull/1465#issuecomment-222332883 is fixed with the latest commit that I added to this PR. Thanks a lot! We should add some tests for downloads in various cases, I guess.

AndrewVSutherland:
@sehlen I can confirm that your latest set of commits fixes http://127.0.0.1:37777/ModularForm/GL2/Q/holomorphic/7/3/6/a/, thanks!  Any luck with https://github.com/LMFDB/lmfdb/pulls#issuecomment-222332467?

sehlen:
@AndrewVSutherland 
I guess you mean https://github.com/LMFDB/lmfdb/pull/1465#issuecomment-222332467.
I couldn't reproduce the error whatsoever. If you set the loglevel of emf_logger to 0 or so, could you send me the logs that you get?

AndrewVSutherland:
@sehlen I just tested it on Atkin and it works there, so it must be an issue with my test setup, sorry for the confusion.

sehlen:
@AndrewVSutherland  Great! No worries - testing is very important. Although I would like to understand how that could happen...

AndrewVSutherland:
@sehlen It could be because I do not have a few of the additional sage packages the LMFDB uses (notably: git+git://github.com/jwbober/conrey-dirichlet-characters.git@master#egg=dirichlet_conrey) installed on my dev system at home (I haven't been able to get Sage 7.1 to rebuild properly with all the  packages listed in requirements.txt).  I'll be careful to double check on Atkin before reporting further errors.

sehlen:
@AndrewVSutherland OK it almost has to be the conrey dirichlet characters although I would expect that you should see more errors.
When a form with non-trivial character is loaded from the db, a conrey dirichlet character is initialized and so errors are caught (and maybe raised again which we might not want to do, not sure) and give the general log message that you saw, I think.

AndrewVSutherland:
Page http://127.0.0.1:37777/ModularForm/GL2/Q/holomorphic/11/10/4/a/ (running on Atkin now) produces a server error caused by "AttributeError: 'unicode' object has no attribute 'absolute_degree'"

The pages http://127.0.0.1:37777/ModularForm/GL2/Q/holomorphic/13/10/4/a/ and http://127.0.0.1:37777/ModularForm/GL2/Q/holomorphic/23/6/2/a/ produce server errors caused by "TypeError: 'NoneType' object is not iterable"

The page  http://127.0.0.1:37777/ModularForm/GL2/Q/holomorphic/31/30/1/a/ produces a server error caused by "TypeError: unsupported operand type(s) for -: 'Undefined' and 'int'"

There are data problems (hecke orbit space dimensions not summing to the correct value) for the spaces 11.4.3, 11.6.3, 16.2.13, 16.2.13, 18.2.7, 20.2.3, 21.2.4, 21.2.5, 22.2.3, 32.4.21, 36.5.5.  This isn't an issue for this PR (they already cause errors on www.lmfdb.org), just letting you know.

sehlen:
For /ModularForm/GL2/Q/holomorphic/11/10/4/a: This space shouldn't be there (I'm not fixing this now as a more systematic update will run soon, I hope). Go to /ModularForm/GL2/Q/holomorphic/11/10/4/ and observe that it links to /ModularForm/GL2/Q/holomorphic/11/10/3.

The second one  http://127.0.0.1:37777/ModularForm/GL2/Q/holomorphic/13/10/4/a/ seems like a bug, I will check tomorrow.

The last one is a data issue and has maybe been observed before, not sure.
Observe that /ModularForm/GL2/Q/holomorphic/31/30/1/ tells you that "Unfortunately, we do not have this space in the database, yet.". But I will fix this in the code nevertheless so that the user does not see a server error.

AndrewVSutherland:
@sehlen OK, thanks, just let me know when you make a new commit and I will retest.  I have updated emf_test_pages.py to check all the spaces it can find in the database within our stated extent (any character) -- it hasn't finished yet but I'll let you know first thing tomorrow if it finds any more (I'll also merge the new emf_test_pages.py into the lmfdb master branch).

sehlen:
All right - all problems should be solved with a combination of a code update and a small data update in the two cases where the "TypeError: 'NoneType' object is not iterable" error occurred.

Note that these forms do have "new" records with the properties we show when hiding the q-expansion stored in the db and they will be different on prod. As far as I can see, everything should work on prod but I have to admit, I replaced some data without making a backup to check with.

But since the modular_forms2 db certainly still contains many records that are identical to what we have on prod, this should be well tested already.

When you feel this is tested enough, it should go to beta, and after a little testing to prod, I think.
The dimension table can go to prod right away, I think.
Maybe you can make a backup so that we could switch back very fast (I made one on the warwick mongodb: dimension_table_old).

AndrewVSutherland:
@sehlen Thanks for the update, I'll test it now.  In terms of the data, will any of the changes you made cause problems for the software currently running on prod?  If not I might be inclined to copy the 11 collections we copied before over to prod -- and yes I always save one backup that we can recover from very quickly (just a rename) if anything goes wrong.

Another question, which collections on prod are actually used by the software, is it just the 11 collections: 

"dimension_table" "dimension_table.chunks" "dimension_table.files" "webmodformspace" "webmodformspace.chunks" "webmodformspace.files" "webnewforms" "webnewforms.chunks" "webnewforms.files" "webeigenvalues.chunks" "webeigenvalues.files"

or are there others?  It would be nice to remove anything we don't need on prod (it would free up space to keep backup collections around).

AndrewVSutherland:
@sehlen It looks good, I think we are ready to merge.

sehlen:
@AndrewVSutherland Regarding the data:
1) The collections you list should indeed be the full list of collections that are currently used.
We might offer the modular symbols stored in Modular_symbols and Newform_factors for download at some point. But otherwise, I don't think anything else is used. @fredstro could you please confirm?
If we delete these from prod, ithey would still be backed-up by some backup running in Warwick as well as the replication to beta, right? 

2) The data right now should at least not be worse than what we have on prod because but I've been playing around with it a little bit - so there are some new records and some old ones. When I make the larger update, I guess it's best to make a copy of the collections. But it should indeed be safe to copy it over to prod, although the difference is small.

sehlen:
@AndrewVSutherland Probably it's indeed a good idea to copy the data.
Take the loading time of the following page (where the record has the new fields on beta running the code that uses it):

http://www.lmfdb.org/ModularForm/GL2/Q/holomorphic/56/10/27/
(right now I get a timeout)
vs
http://beta.lmfdb.org/ModularForm/GL2/Q/holomorphic/56/10/27/

However, this has not been updated systematically, yet.

fredstro:
regarding the data. the only thing I'm not sure if (and I don't have access
to my computer now) is the dimension table. I'm not sure if this is the
correct dimension table. it seems that the one mentioned is using gridfs
and I think the one we use now only uses mongo records. Can you check?
@sehlen

sehlen:
Thanks for noticing, @fredstro. It is the correct dimension table (pretty
sure, I recreated it a few days ago and it is in use) but the .files and
.chunks don't exist in fact ;-)

AndrewVSutherland:
@sehlen  Great, I can trim the list down to 9 collections.  I will update these 9 on prod later today.

AndrewVSutherland:
On 2016-05-29 10:08, Stephan Ehlen wrote:
> @AndrewVSutherland Regarding the data:
> 1) The collections you list should indeed be the full list of
> collections that are currently used.
> We might offer the modular symbols stored in Modular_symbols and
> Newform_factors for download at some point. But otherwise, I don't
> think anything else is used. @fredstro could you please confirm?
> If we delete these from prod, ithey would still be backed-up by some
> backup running in Warwick as well as the replication to beta, right?

Correct, beta (by which I mean the mongo db running on 
lmfdb.warwick.ac.uk and used by beta.lmfdb.org) is backed up regularly, 
and it is not synched to prod.  Deleting collections from prod has no 
impact on beta.  The beta mongo db is also replicated in the cloud (but 
note that deletions are replicated automatically as well, so this does 
not protect you against accidentally deleting something you need).

But I might suggest that if you know you can get rid of any of the 
collections on beta it would be worth doing so (they will still exist on 
backup).

For example, are the collections WebNewForms.files and 
WebNewForms.chunks relevant (there is no corresponding WebNewForms).

> 
> 2) The data right now should at least not be worse than what we have
> on prod because but I've been playing around with it a little bit - so
> there are some new records and some old ones. When I make the larger
> update, I guess it's best to make a copy of the collections. But it
> should indeed be safe to copy it over to prod, although the difference
> is small.

I think I might be happier having the data we tested on prod, so that it 
matches what we tested the new software against (we can always revert to 
the old data if need be, but reverting to the old software would 
actually be a pain with our current setup).

When do you expect to have the larger update done?

AndrewVSutherland:
@sehlen @fredstro @edgarcosta 

I removed all but the following 12 collections from the production mongo db and everything still seems to work:

dimension_table
webmodformspace webmodformspace.chunks webmodformspace.files
webnewforms webnewforms.chunks webnewforms.files
webeigenvalues.chunks webeigenvalues.files
webchar webchar.chunks webchar.files

I'm hesitant to remove the last three because they are referenced in the code (in web_character.py).

sehlen:
I haven't looked at this in a while but I believe that the webchar collections are in fact used but not necessary (i.e. the website also runs without it)
In theory they could be used more if we would insert records for all characters for which we have any modular forms spaces in the dab. The code will work just fine without these collections but maybe a bit slower, in case a character has a record in the db - I haven't really tested it.

AndrewVSutherland:
Thanks.  I'll leave them in for the moment (the don't take up much space 
in any case).

fredstro:
@sehlen at some point we should also see if we can use the webchars class
from the dirichlet character pages instead of having our own (or make them
use ours...)

PR #1466, opened 2016-05-26
fix ec sign(disc)

JohnCremona:
A bug in PR #1439 meant that even when we have extra ec data as in the curves2 collection we were in effect ignoring it.  (Cautionary tale: I had a block of a dozen or more lines all within a single try/except block and something triggered the exception which was not one of the things I was planning for).  Then when I fixed that *and ran the tests* another small bug was discovered (curves with negative discriminant omiited the minus sign in the disc's factorization).


AndrewVSutherland:
Reviewing now.

AndrewVSutherland:
OK it looks good, I don't see any "falling back" errors and all tests pass, so I'm going to merge and push to beta.

PR #1471, opened 2016-05-28
General code cleanup + performance improvements

AndrewVSutherland:
I went through the code cleaning up a number of legacy issues and left over detritus from the original "clone=and-conquer" implementation that simply adapted the elliptic curves python code and html templates.  I also made a number of performance improvements and added many new tests.  There are a lot of code changes, but most do not impact the functionality in any way (or do so invisibly).  Here are a few functional changes that are visible:

- If you go to http://www.lmfdb.org/Genus2Curve/Q/169/ and click "search again" you will get a server error.  This is fixed in http://127.0.0.1:37777/Genus2Curve/Q/169/.
- If you go to http://www.lmfdb.org/Genus2Curve/Q/random you will be redirected to a random curve page with a "dotted" curve label in the URL, but if you go to http://127.0.0.1:37777/Genus2Curve/Q/random you will see a "slashed" curve label in the URL (as desired).
- The bread crumbs on curve pages have been made more uniform and now reflect all four components of the label.  Clicking on any intermediate bread crumb takes you to an appropriate page (see http://127.0.0.1:37777/Genus2Curve/Q/15360/f/983040/, for example), and when that page is a search page (as opposed to an isogeny class home page), refining the search works as it should.  If you change the search parameters in a way that contradicts the bread crumbs it will take you to a general search page.

The main performance improvements involve avoiding some unnecessary redirects, avoiding querying the isogeny_classes collection for data that is now stored in the curves collection (and has been for a while), and using the projection parameter in queries (the second optional parameter to .find(), see https://docs.mongodb.com/manual/reference/method/db.collection.find/) to restrict the amount of data that needs to transferred from the mongo db to the web server.  This significantly speeds up operations that involve large amounts of data, such as the download button.
If you go to http://www.lmfdb.org/Genus2Curve/Q/?cond=, for example and click the "sage" button to download all 66158 curves, it will take 5-10 seconds, but if you go to http://127.0.0.1:37777/Genus2Curve/Q/?cond= and do the same thing it will take only a few seconds (depending on your internet connection) even though the later may need to transfer data much further than the cloud server.

Analogous versions of two of the bugs fixed in this PR exist in the elliptic curves code, I am about to open an issue for this.

The performance improvements gained by using the projection parameter in large .find() calls would also help in many other places (even more than it does in genus2_curves).  Notable examples include the search and download functions in elliptic_curves and ecnf (see lines 243 and 563 of LMFDB/lmfdb/elliptic_curves/elliptic_curve.py, and lines 407 and 588 of lmfdb/ecnf/main.py,  @JohnCremona), and the search function in numberfields (see line 477 of LMFDB/lmfdb/number_fields/number_field.py, @jwj61 ).  There is no need to force mongo db to send the full contents of every record that matches a query to the LMFDB application when only a few attributes are needed (those displayed in the columns of the search results, of the data included in the download file).  In fact, for queries whose purpose is just to retrieve a list of labels, if the projection parameter is set to {'_id':int(0),'label':int(1)} and there is an index on 'label', mongo db can process the query using only the index, it doesn't need to look at the collection at all.

### DIFF
--- a/lmfdb/genus2_curves/genus2_curve.py
+++ b/lmfdb/genus2_curves/genus2_curve.py
@@ -1,24 +1,21 @@
 # -*- coding: utf-8 -*-
+# -*- coding: utf-8 -*-
 import StringIO
-import ast
+from ast import literal_eval
 import re
 import time
-import pymongo
 from pymongo import ASCENDING, DESCENDING
 from operator import mul
-import lmfdb.base
 from lmfdb.base import app
-from flask import Flask, flash, session, g, render_template, url_for, request, redirect, make_response, send_file
+from flask import flash, render_template, url_for, request, redirect, make_response, send_file
 from markupsafe import Markup
-import tempfile
-import os
 
-from lmfdb.utils import ajax_more, image_src, web_latex, to_dict, comma, random_object_from_collection
+from lmfdb.utils import to_dict, comma, random_object_from_collection
 from lmfdb.search_parsing import parse_bool, parse_ints, parse_signed_ints, parse_bracketed_posints, parse_count, parse_start
 from lmfdb.number_fields.number_field import make_disc_key
 from lmfdb.genus2_curves import g2c_page, g2c_logger
-from lmfdb.genus2_curves.isog_class import G2Cisog_class, url_for_label, isog_url_for_label, st_group_name, st_group_href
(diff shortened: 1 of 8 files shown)


jwj61:
I typed "hello" in the box for conductor.  Before, the bad input was caught, now it crashes.

AndrewVSutherland:
@jwj61 Fixed in the most recent commit.

PR #1477, opened 2016-05-29
Optimize random_object_from_collection

AndrewVSutherland:
This pull request implements a faster way to obtain random objects from a collection.

The file rewrite.py contains a new utility function create_random_object_index, which, given a database to which the caller has write access and a collection name "coll" creates a new collection "coll.rand" with the same object ids as coll and a single indexed attribute "num" that runs from 1 to coll.count()

The function random_object_from_collection has been modified to look for the colleciton coll.rand, and if it exists use an indexed lookup on num to more efficiently retrieve random objects.  From measurements taken directly on the mongo db and in sage/python, it is 20-30 times faster for large collections than using the .aggregate method.

Random object indexes have been added to the larger collections (e.g. curves2, nfcurves, fields, etc.. on both lmfdb.warwick.ac.uk and in the cloud).  They will not be used until this PR is merged and pushed (and if they are not present, there is no harm, the random object index will be used if it is there and otherwise random_object_from_collection falls back to using .aggregate).

There is also a new function random_value_from_collection in utils.py that takes a collection name and an attribute and returns the value of that attribute in a randomly chosen object (it will return None if the value is not present in the chosen object).  This is slightly more efficient and useful, for example, when one just wants the label of a random object.  Currently the function random_value_from_collection is not used anywhere, but most of the places where random_object_from_collection is currently used could use this function instead.

### DIFF
--- a/data_mgt/utilities/rewrite.py
+++ b/data_mgt/utilities/rewrite.py
@@ -1,6 +1,5 @@
-from pymongo import WriteConcern
-from time import time
-#
+import time
+
 # rewrite_collection(db,incoll,outcoll,func)
 #
 # function to pipe an existing collection to a new collection through a user specified function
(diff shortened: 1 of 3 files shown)


AndrewVSutherland:
To test this PR, all you need to do is fetch some random elliptic curves or number fields (the random object indexes for these have been created on lmfdb.warwick.ac.uk).  To verify that it works as it should in the absence of a random object index, fetch some random local fields (for example).  You can use beta.lmfdb.org/api to see which collections have random object indexes on them (you will see an additional .rand collection).

I should note that the random object index is not automatically updated when new objects are added, it needs to be rebuilt -- this can be done automatically when the new data is moved to the cloud server, and on demand if desired (but there is really no harm in letting the random object index get slightly out of date).

JohnCremona:
Thanks for doing this.  My idea (never implemented) instead of the random_value_from_collection function was to have an additional "projection" parameter to random_object... like the one in find(), default None to return the whole thing but able to return one *or more* data fields from the random object found.

Issue #1343, opened 2016-05-13
Integral lattice and Gram matrix knowls


brianconrey:
I feel like the knowls for "Integral lattice" and "Gram matrix" could use a little amplification.
Suppose I think of lattices as collections of classical lattice points in
R^n. It's not clear to me from the knowls what the bilinear form is.
If it is supposed to be defined in terms of the Gram matrix then
the Gram matrix should be defined without referring to the bilinear
form; otherwise the definition is circular.  Perhaps the Gram matrix could be defined in terms of the inner products of the elements of a basis; then the bilinear form in terms of the Gram matrix.

annahaensch:
I just saw this, and I'm working on it today.

annahaensch:
I edited the knowls for integral lattice and Gram matrix in the beta version.  Please tell me if this sufficiently addresses your concerns.

sanni85:
Also this issue can be closed now.

AndrewVSutherland:
Agreed.

Issue #1342, opened 2016-05-13
Learn more about lattices


brianconrey:
On the page http://www.lmfdb.org/Lattice/
it would be nice to include a "History of Lattices"
or some at least some explanation of the role of
lattices in the grand scheme of things. Maybe 
how they relate to modular forms and to number fields
and and perhaps some simple examples that 
illustrate these connections.

davidfarmer:
Anna Haensch would be the right person to do this, and also issue #1343

annahaensch:
I just saw this issue, and I am working on that today. I'll comment again when it's done.

annahaensch:
I've built up a knowl called lattice.history giving a brief history and highlighting some of the specific connections between lattices and LMFDB objects as well as some relevant lattice topics of interest.  I'll continue to work on it.  But there is something there now. @sanni85 will add a link to the knowl on the lattice homepage.

sanni85:
Now that the pull request #1485 is merged this issue can be closed after the lattice database is updated on the cloud: the record for the Leech lattice has been changed and the link to it  in lattice.history is broken on the main website.

AndrewVSutherland:
I updated the database and the link now works, closing.

PR #1485, opened 2016-06-01
Lattice13

sanni85:
This pull request has the following effect:

1. an history link is added in the learn more box addressing #1342  (also #1343 is solved now)  http://localhost:37777/Lattice/

2. the search by label or name is more robust: white spaces are handled better and capitalization too: try to search leech+spaces or a label with spaces in between like 2.  1.2.     1.1
the result is given to the url with the label, so also a test was fixed


sanni85:
By the way I have deleted the record of the Leech lattice from the database and made a new one since the class number and genus representatives have been computed (the label contains the class number and it was set to 0 since it was unknown, now it is 24).

PR #1487, opened 2016-06-01
Add label to nf pages

jwj61:
As discussed at AIM, home pages should have the item's label at the top of the properties box.   The label also should be somewhere in the main body of the page, with a knowl for the label structure.  This adds these things to number field pages.  It then removes "learnmoreabout"s on the labels.

### DIFF
--- a/lmfdb/number_fields/number_field.py
+++ b/lmfdb/number_fields/number_field.py
@@ -104,8 +104,7 @@ def NF_redirect():
 @nf_page.route("/HowComputed")
 def how_computed_page():
     info = {}
-    info['learnmore'] = [('Global number field labels', url_for(".render_labels_page")), 
-        ('Galois group labels', url_for(".render_groups_page")), 
+    info['learnmore'] = [('Galois group labels', url_for(".render_groups_page")), 
         (Completename, url_for(".render_discriminants_page")) ]
     t = 'How Number Field Data was Computed'
     bread = [('Global Number Fields', url_for(".number_field_render_webpage")), ('Galois group labels', ' ')]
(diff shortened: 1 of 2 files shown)


AndrewVSutherland:
I wasn't at the AIM workshop so perhaps this has already been debated, but for what it's worth, here are my four cents:

(1) I agree the label should be at the top of the properties box, so +1 for making this change.

(2) I'm not sure the label needs to appear both in the heading of the page and in the main text in the body, but if is going to be in the body it seems strange not to have it at the top, as it is in the properties box.  But of course then it is right next to the header, which makes it seem redundant.

(3) The label now appears in three places on the page -- it seems like the only reason for putting it in the main body is so that you have a place to hang the knowl.

(4) I see no harm in leaving the labels knowl in the learnmoreabout box (regardless of whether the label is in the main body or not).  If I'm a first time user I think I am more likely to look and click there than I am to click on a labels knowl that is embedded in a non-obvious place in the middle of the page.

AndrewVSutherland:
One other thing I noticed while reviewing this (obviously these don't need to be fixed in this PR unless you want to).  On the home page for a number field the captions for "degree", "discriminant", "signature", "ramified primes", "rank", "torsion generator", "regulator" should all have knowl links (I think there are knowls for all these things, they just need to be linked to the captions).  The knowl link on "label" would look less out of place if all these other things had knowl links as well.

Regarding my comments above, I'm happy to merge this if you (and others) think this is the right way to go, I just wanted to give my feedback.  The only thing I feel slightly strongly about is that I don't think there is any harm in have an entry for number fields label in the learnmoreabout box and there is possibly some harm in removing it, so I would suggest putting it back.

jwj61:
I will get back to this with updates and comments after fixing bugs.

jwj61:
The goal was to bring the page more in line with other sections and what people generally wanted.  Most pages have learnmoreabout labels, so I restored those.  The label at the top of properties is standard, so I left that change.

Different pages do different things when it comes to putting the label in the body of the page.  Personally, I don't like it there, so I removed it again.  A majority of the people at AIM thought it should be there, but they were split between it being first or after the individual invariants.  So, there is currently no convention on that.

I could make invariants like degree knowls, but the status quo was an attempt to be consistent with other areas, where Invariants is the knowl which has an explanation and knowls for the individual pieces.  When there is a consensus as to what we are doing on that score, I will make the page conform.

AndrewVSutherland:
@jwj61 regarding knowls on the invariants, my impression was that the cases where the knowl is on "Invariants" it is typically because the invariants are being identified by mathematical symbols (e.g. N for conductor and \Delta for discriminant) rather than words.  But for things captioned by a mathematical term (e.g. "conductor") I think it makes sense to put a knowl on it.  But I agree that we currently do not have a consistent convention.

JohnCremona:
For what it's worth, that was the reasoning behind pages such as http://www.lmfdb.org/EllipticCurve/Q/54600/v/1 in the invariant section.  There was some discussion of this point at a workshop (in a distant galaxy long long ago).

PR #1498, opened 2016-06-02
fixed ecnf browse templates to sort by discriminant, Issue 1484

### LINKED ISSUE #1484: ECNF: sort summary pages
Once on http://lmfdb.org/EllipticCurve/browse/ if you click on any of the options, for example
http://lmfdb.org/EllipticCurve/browse/2/%282%2C0%29/
you get a list of results, indexed by number fields. Should the number fields be ordered also by discriminant?
This is done for Hilbert modular forms e.g. http://lmfdb.org/ModularForm/GL2/TotallyReal/browse/2/

JohnCremona:
This fixed #1484 .  I tturned out to be very simple since I already had a correctly orderd list of field labels, so just ad to edit two templates to use it instead of the random order of a dictionary's keys.

To test this click on both a degree and a signature from the page EllipticCurve/browse/ since these use different templates and both have been changed.  A good example is signature (0,1).

### DIFF
--- a/lmfdb/ecnf/templates/by_degree.html
+++ b/lmfdb/ecnf/templates/by_degree.html
@@ -26,8 +26,10 @@
  </h3>
 <table>
 <tr><th> Field <th> Number of curves <th> Number of isogeny classes <th> Maximum conductor norm </tr>
-{% for F, Fdat in sdat.iteritems() %}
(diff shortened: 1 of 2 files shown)


AndrewVSutherland:
Much better :)

PR #1504, opened 2016-06-02
Add label to properties box of artin rep, galois groups, local fields

jwj61:
Can check on the home pages of each type of object.  Also added learnmore for artin rep labels, and removed colons from Galois groups properties boxes.

### DIFF
--- a/lmfdb/artin_representations/main.py
+++ b/lmfdb/artin_representations/main.py
@@ -184,7 +184,8 @@ def render_artin_representation_webpage(label):
         processed_root_number = "not computed"
     else:
         processed_root_number = str(the_rep.sign())
-    properties = [("Dimension", str(the_rep.dimension())),
+    properties = [("Label", label),
+                  ("Dimension", str(the_rep.dimension())),
                   ("Group", the_rep.group()),
                   #("Conductor", str(the_rep.conductor())),
                   ("Conductor", "$" + the_rep.factored_conductor_latex() + "$"),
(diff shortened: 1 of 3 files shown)


AndrewVSutherland:
@jwj61 Looks good.  I notice that on the Artin reps home page you have knowls on all the search parameters except ramified and unramified primes -- maybe worth adding these? (unlike the "Invariants" situation, on the search page there is nowhere else to find the definition of these terms)

JohnCremona:
In the example (which I got to using the random function!) http://localhost:37777/ArtinRepresentation/2.2e4_5_131.4t3.2c1  in the "Roots of defining polynomial" there is redundant discussion of an extension of Q_{37} with generator a having polynomial x-1.  It would be nice to special-case this.  (Not necessarily in this PR.)

I agree with @AndrewVSutherland about needing knowls for ramification.  In the knowls artin.conductor and artin.lfunction there is a reference to primes being ramified and these could then cross-reference the new knowl.  (The second of those has many terms which could be links to other knowls, such as the Inertia group, and more).

jwj61:
I am a little concerned with the de facto change in merging criteria.  It used to be pretty much: merge if it improves the site.  I can understand that people see new things when reviewing a pull request, but I would put those in the issue tracker.

This pull request is about putting object labels in the properties box.  It was not merged, and the comments are unrelated to that.  It seems merging this is now contingent on making other changes, or at least debating whether those changes should be made.  Neither of those things should be necessary when making a change.

The suggested changes are now in other pull requests.

JohnCremona:
@jwj61 Please don't overreact -- when I review something and spot something which might make it *even better* it seems most natural to make that comment on the PR itself, but I do not thereby mean that the thing I suggest is a prerequisite to merging.  The person who made the PR might if convenient add a new commit to the same PR, or make it a new issue,or just disagree.  Fine.

To show I mean it I am merging this one now, and then will review #1520 (thanks!) for which the code looks good.

AndrewVSutherland:
@jwj61 I was only waiting to merge in case you wanted to respond to my question/suggestion, I certainly did not mean to insist that you make a change and I apologize if I gave that impression.  I'm happy to merge it as is, and I see that John Cremona has already done so.

PR #1512, opened 2016-06-03
Ec fix download all code

AndrewVSutherland:


JohnCremona:
OK -- just eliminating spurious newlines I think?
One day it might be nice to remove those unicode srting 'u' characters, which happen because the dict being printed comes from pymongo.  I was going to try something like x.str().replace("u'","'") but now you have fixed the test I don't have to.

sehlen:
The ‘u’ indicates that it is a unicode string.
If you don’t want unicode strings, you can call .encode() on them, which will return them encoded in the default encoding.
(I don’t know if unicode might happen to be the default encoding, probably yes, in which case one has to find out how to pass encodings to the .encode() method which I was too lazy to do ;-)

Example:
sage: s=u’a'
sage: s
u’a’
sage: s.encode()
‘a’

AndrewVSutherland:
@JohnCremona, right except it was putting a newline after every character :) (this is because I changed how multi-line code snippets are handled to deal with inconsistencies between how you implemented it (using an array of lines) and how others did it (embedding carriage returns).).

JohnCremona:
Thanks @sehlen -- I had tried adding .encode() already but it made no difference, and I just tried again both with defaults and with 'utf-8' and 'ascii' but again no difference.  I think the trouble is that the code does 
response = make_response('\n\n'.join(str(d) for d in data_list))
where data_list is a list of dicts (as returned by pymongo) and each dictionary item has a unicode string for the key as well as (in some cases) the value.  You can see the result at http://www.lmfdb.org/EllipticCurve/Q/download_all/11.a

OK with help from Stackexchange I find that this works:

import json
    response = make_response('\n\n'.join(json.dumps(d) for d in data_list))

so I hope you'll tolerate a newPR (which will revert that test...)

PR #1520, opened 2016-06-06
In showing roots of an Artin representation, modify output depending on

jwj61:
whether roots are in Q_p or an extension.

This implements a change suggested by @JohnCremona in a different pull request

Example with roots in Q_p: ArtinRepresentation/3.2e2_523e2.6t8.1c1

Example with roots in an extension of Q_p:  ArtinRepresentation/5.2e4_12119e3.6t14.1c1


JohnCremona:
Looks good.

Issue #122, opened 2013-09-17
never do from something import *


haraldschilly:
```
…/lmfdb/characters/test_characters.py:15: SyntaxWarning: import * only allowed at module level
  def test_Heckemethods(self):
```

This is just one part, also on module level we should never just import *. That's a very good practice and allows for static code checking like `pyFlakes` to work better!

(if you do not have `pyFlakes`, install it and run it on your files …)

edgarcosta:
@AndrewVSutherland what about making running your code through pyflakes a standard thing?

AndrewVSutherland:
@edgarcosta Yes, this should be in the developers guide, and this is something I would ask reviewers to start checking on pull requests, as I have done.  pyflakes is very lightweight (do sudo apt get pyflakes to install) and in addition to catching unneeded imports, it finds a lot of bugs (I found several in the siegel modular forms and lfunctions code using it); usually these are caused by spelling errors in variable names, which without a tool like pyflakes are very difficult to catch in python code if they are hiding in a line of code your tests do not hit.

edgarcosta:
no need for apt-get or sudo, pyflakes is a pip package `pip install --upgrade pyflakes`, suffices.

JohnCremona:
!
$ pip install --upgrade pyflakes
The program 'pip' is currently not installed. You can install it by typing:
sudo apt-get install python-pip

I normally use the python and pip in Sage, i.e. "sage -pip ..." and did install pyflakes that way but then did not know how to use it.  So in fact using apt-get to install pyflakes was useful.  (This is on lehner.)

PR #1541, opened 2016-06-12
Modify higher genus w/automorphisms pages to align more closely with other future work.

jenpaulhus:
These are modified versions of the group actions pages based on conversations with @jvoight to try to align these pages more with future pages relating to Belyi maps.  The data is basically the same, but centered around conjugacy classes of elements in a group (so called passports), instead of the individual elements of the group.

The first commit is two new files.
(1) /higher_genus_w_automorphisms/templates/hgcwa-show-family.html
will give pages like: /HigherGenus/C/aut/3.8-3.0.2-2-4-4
where given a group and signature, all possible conjugacy classes of the group which satisfy the signature restrictions are listed.

(2) /higher_genus_w_automorphisms/templates/hgcwa-show-passport.html
will give pages like: /HigherGenus/C/aut/3.8-3.0.2-2-4-4.1
where given a specific set of conjugacy classes from the pages in (1) will  give all generating vectors (up to conjugation by elements in G) and list properties like whether the family is hyperelliptic. 

The second commit is modifications to old files to support the changes above.

The download functionality is not coded yet.

Any and all feedback welcome.

### DIFF
--- a/lmfdb/higher_genus_w_automorphisms/main.py
+++ b/lmfdb/higher_genus_w_automorphisms/main.py
@@ -19,7 +19,19 @@ from lmfdb.higher_genus_w_automorphisms import higher_genus_w_automorphisms_page
 from lmfdb.genus2_curves.data import group_dict
 
 
-HGCwA_credit = 'J. Paulhus'
+
+
+# Determining what kind of label
+family_label_regex = re.compile(r'(\d+)\.(\d+-\d+)\.(\d+)\.(\d+-)')
+passport_label_regex = re.compile(r'((\d+)\.(\d+-\d+)\.(\d+)\.(\d+.*))\.(\d+)')
+cc_label_regex = re.compile(r'((\d+)\.(\d+-\d+)\.(\d+)\.(\d+.*))\.(\d+)')
+
+def label_is_one_family(lab):
+    return family_label_regex.match(lab)
+
+def label_is_one_passport(lab):
+    return passport_label_regex.match(lab)
+
 
 
 def get_bread(breads=[]):
@@ -29,7 +41,7 @@ def get_bread(breads=[]):
     return bc
 
 
-
+#pretty printing functions 
 def groupid_to_meaningful(groupid):
     if groupid[0] < 120:
         return group_dict[str(groupid).replace(" ", "")]
@@ -51,7 +63,24 @@ def group_display_shortC(C):
         return group_display_short(nt[0], nt[1], C)
     return gds
 
-    
+        
+def sign_display(L):
+    sizeL = len(L)                
+    signL = "[ " + str(L[0]) + "; "
+    for i in range(1,sizeL-1):
+        signL= signL + str(L[i]) + ", "                    
+     
+    signL=signL + str(L[sizeL-1]) + " ]"                  
+    return signL                
+
+def cc_display(L):
+    sizeL = len(L)
+    stg = str(L[0])+ ", "
+    for i in range(1,sizeL-1):
+        stg =stg + str(L[i])+", "
+    stg=stg+ str(L[sizeL-1])
+    return stg
+
     
 @higher_genus_w_automorphisms_page.route("/")
 def index():
@@ -66,15 +95,29 @@ def index():
                 ('Source of the data', url_for(".how_computed_page")),
                 ('Labeling convention', url_for(".labels_page"))]
     
-    return render_template("hgcwa-index.html", title="Higher Genus Curves with Automorphisms", bread=bread, credit=HGCwA_credit, info=info, learnmore=learnmore)
+    return render_template("hgcwa-index.html", title="Higher Genus Curves with Automorphisms", bread=bread, info=info, learnmore=learnmore)
 
 
 
 
 @higher_genus_w_automorphisms_page.route("/<label>")
 def by_label(label):
-    return render_hgcwa_webpage({'label': label})
 
+    if label_is_one_passport(label):
+        return render_passport({'passport_label': label})
+    elif label_is_one_family(label):
+        return render_family({'label': label})    
+    else:
+        info = {}
+        bread = get_bread([("Search error", url_for('.search'))])
+        info['err'] = "Higher Genus Curve with Automorphism " + label + " was not found in the database."
+        info['label'] = label
+        return search_input_error(info, bread)
+    
+
+@higher_genus_w_automorphisms_page.route("/<passport_label>")
+def by_passport_label(label):
+    return render_passport({'passport_label': label})
 
 
 @higher_genus_w_automorphisms_page.route("/search", methods=["GET", "POST"])
@@ -95,7 +138,7 @@ def higher_genus_w_automorphisms_search(**args):
     C = base.getDBConnection()
     query = {}
     if 'jump_to' in info:
-        return render_hgcwa_webpage({'label': info['jump_to']})
+        return render_family({'label': info['jump_to']})
 
     try:
         parse_list(info,query,'group', name='Group')
@@ -112,7 +155,7 @@ def higher_genus_w_automorphisms_search(**args):
     count = parse_count(info)
     start = parse_start(info)
     
-    res = C.curve_automorphisms.families.find(query).sort([(
+    res = C.curve_automorphisms.passports.find(query).sort([(
          'g', pymongo.ASCENDING), ('dim', pymongo.ASCENDING)])
     nres = res.count()
     res = res.skip(start).limit(count)
@@ -136,24 +179,24 @@ def higher_genus_w_automorphisms_search(**args):
         else:
             info['report'] = 'displaying all %s matches' % nres
 
-    return render_template("hgcwa-search.html", info=info, title="Higher Genus Curves with Automorphisms Search Result", bread=bread, credit=HGCwA_credit)
+    return render_template("hgcwa-search.html", info=info, title="Higher Genus Curves with Automorphisms Search Result", bread=bread)
 
 
 
 
 
-def render_hgcwa_webpage(args):
-    data = None
+def render_family(args):
     info = {}
     if 'label' in args:
         label = clean_input(args['label'])
         C = base.getDBConnection()
-        data = C.curve_automorphisms.families.find_one({'label': label})
-        if data is None:
+        dataz = C.curve_automorphisms.passports.find({'label': label})
+        if dataz.count() is 0:
             bread = get_bread([("Search error", url_for('.search'))])
             info['err'] = "Higher Genus Curve with Automorphism " + label + " was not found in the database."
             info['label'] = label
             return search_input_error(info, bread)
+        data=dataz[0]
         g = data['genus']
         GG = data['group']
         gn = GG[0]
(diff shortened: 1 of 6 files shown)


AndrewVSutherland:
This isn't specific to this pull request, but we are trying to make sure we have complete inventory information for all the databases in the LMFDB.  This is needed so we know exactly which collections need to be preserved as we migrate to the new storage engine (undocumented collections may get left behind).

@jenpaulhus, can you please fill in https://github.com/LMFDB/lmfdb-inventory/blob/master/db-curve_automorphisms.md with a description of the two collections in the curve_automorphisms database?  You can check out some of the other inventory entries (e.g. https://github.com/LMFDB/lmfdb-inventory/blob/master/db-sato_tate_groups.md) for an example of what to do.

jenpaulhus:
@AndrewVSutherland Did a rudimentary update to the file https://github.com/LMFDB/lmfdb-inventory/blob/master/db-curve_automorphisms.md .  I will add more detail later once the collection passports is in its final form.

AndrewVSutherland:
@jenpaulhus 

I just did a quick review and noticed a bunch of things (all but one of which are very minor and easy to address):

### Issues specific to this PR

* On http://127.0.0.1:37777/HigherGenus/C/aut/3.8-3.0.2-2-4-4 the word passports should be a knowl (you don't have to create the knowl right now, but you should put in the knowl link in so that you can then easily create it on beta.lmfdb.org).  I would also put a know on genus (yes, anyone looking at the page probably knows what this is, but there is no harm in putting one in).

* The breadcrumbs on the passport pages have the wrong URL.  If you click on the second to last bread crumb on the page http://127.0.0.1:37777/HigherGenus/C/aut/3.8-3.0.2-2-4-4.1, it should take you to the page you came from http://127.0.0.1:37777/HigherGenus/C/aut/3.8-3.0.2-2-4-4, but it keeps you on the passport page (the problem is in line 360 of main.py, you need to put a url into the bread crumb, putting in ' ' keeps you where you are and is only appropriate for the last bread crumb).

* I think the name of the automorphism group (e.g. **D_8**) should appear in the properties box.  I don't know that the small group id needs to be there, but if it there you should se square brackets **[8,3]** rather than round brackets, as you do elsewhere (and is also done in the genus 2 pages).  In particular, you cannot type **(8,3)** into your search forms, you need to type **[8,3]**

* It's not clear to me why http://127.0.0.1:37777/HigherGenus/C/aut/3.16-11.0.2-2-2-4 is in the related objects list on http://127.0.0.1:37777/HigherGenus/C/aut/3.8-3.0.2-2-4-4.1,, perhaps it should be a related object on http://127.0.0.1:37777/HigherGenus/C/aut/3.8-3.0.2-2-4-4 instead?  I would have expected to see  http://127.0.0.1:37777/HigherGenus/C/aut/3.8-3.0.2-2-4-4 as  a related object on http://127.0.0.1:37777/HigherGenus/C/aut/3.8-3.0.2-2-4-4.1 (yes it is also a bread crumb, but that's OK, the same thing happens with elliptic curves and their isogeny classes, for example).

* I would put spaces between the cycles in your permutation cycle notation make it more readable (e.g. in **(1,3) (2,4) (5,7) (6,8)** rather than **(1,3)(2,4)(5,7)(6,8)**

* The search by label function on http://127.0.0.1:37777/HigherGenus/C/aut/  should accept passport labels as well as family labels (just like elliptic curve label searches accept both curve labels and isogeny class labels)

### Issues not specific to this PR:

* On http://127.0.0.1:37777/HigherGenus/C/aut/ I think it might be clearer to use the phrase "families of higher genus curves" in the title and in the the browse and find headings.  In the search results you refer to families of curves (which I think makes sense), not automorphisms
* On http://127.0.0.1:37777/HigherGenus/C/aut/ the word curve should be plural in "Find specific automorphisms of higher genus curve".
* On http://127.0.0.1:37777/HigherGenus/C/aut/?Submit=Search the headings Passport Genus and Signature should have knowls.
* If I type a signature exactly as it appears on the screen (e.g. **[0;2,3,3,3,6]**, with a semicolon), I get an error -- you should change the search processing to simply replace semicolons with commas before parsing the signature so that both **[0;2,3,3,3,6]** and **[0,2,3,3,3,6]** work.

### DB issue that  will be easier to address now (or soon) than it will be later:

You are storing vectors of integers in the database in several places (e.g. signatures and small group ids).  As we have discovered elsewhere, this causes problems when you want to search on them efficiently.   Especially for the signatures, it would be better to encode them as strings, e.g. **'[0,2,3,3,3,6]'** (no spaces) that you then convert to arrays of integers when you read them from the database (use ast.literal_eval to do this, not eval).  This may not be an issue while the database is as small as it is, but it will start to become noticeable if it ever gets larger in the future, and it will be more painful to change in the future than it would be now.  For small group ids, in genus 2 curves I encode these as strings (e.g. **'[8,3]'**), and I note that the group_dict in genus2_curves/data.py that you are using does the same thing.  Alternatively, you could use the small groups collection in the sato_tate_groups database which gives LMFDB style labels (e.g. **'8.3'**) to each of the small groups.  It does not currently have all the groups in data.py stored in it, but I could easily add them -- just let me know if you would like me to do this.  There is some argument for going this route, since I expect that eventually we will want small groups to have LMFDB pages (if for no reason other than so that we can link to them rather than expecting people to go look them up in the small groups atlas).

I expect that was probably more feedback than you were looking for :), but I hope it is helpful.

jenpaulhus:
Thanks @AndrewVSutherland  for all the feedback. I greatly appreciate it. I’ve got most of the changes made, including converting the lists of integers to strings in the data.  That has caused some issue with the searches, though.  Once I get that figured out, I’ll make another commit.

In regard to your comments about a dictionary of group names, I think it might be nice to have a central place for this in lmfdb. I used the list from genus 2 simply because I found it first.   But I also added a few groups which are larger than order 120 (like PSL(2,7)) directly into my code.  This is probably less than ideal, and why I think one central place for a conversion from the SmallGroup databases to “nice” names would be good.  I certainly am happy to help with this, but I’m not sure what form it should take or where it should be located.

AndrewVSutherland:
@jenpaulhus I added the groups of order <= 255 to sato_tate_groups.small_groups (you can view the collection at http://lmfdb.org/api/sato_tate_groups/small_groups).  Each group is identified by a string label of the form "a.b" where [a,b] is the GAP id (so a is the group order).

For the moment, you can import sato_tate_groups.main and use the function sg_pretty to convert a small group label into a latex string for display.  I will create a small_groups.py module in lmfdb.lmfdb to provide generic access to small groups data (and eventually move the collection out of the sato_tate_groups database into its own database), but it will be easier to make this change (and update your import) *after* this PR is merged.

Let me know if you need groups larger than 255, it's easy to add more.

jenpaulhus:
Another series of updates.  I’ve modified the relevant files to accommodate the following changes: 

1. Added a bunch of knowl links on pages to genus, passport, signature on different pages.

2. Fixed breadcrumbs on pages like http://127.0.0.1:37777/HigherGenus/C/aut/3.8-3.0.2-2-4-4.1

3. Put “pretty name” of groups in Properties box, like [http://127.0.0.1:37777/HigherGenus/C/aut/3.16-11.0.2-2-2-4](http://127.0.0.1:37777/HigherGenus/C/aut/3.168-42.0.2-3-7)  and pulled that information from sato_tate database.

4. Added spaces between cycles in permutations.

5. The “search by label” on http://127.0.0.1:37777/HigherGenus/C/aut/ now accepts passport labels too. Also, the signature box will accept values like “[0;2,3,7]”.

6. On http://127.0.0.1:37777/HigherGenus/C/aut/ I pluralized “curve” in “Find specific autos of…”  and added the phrase “families of higher genus curves” to the title, but I kept the words “with automorphisms” too since any family with no automorphisms will not appear ever in this database.  


7. Changed from list of integers [1,2,3,4] to strings “[1,2,3,4]”  in the database for fields: _group, con, full_auto, signature,_ and _signH_.  This caused an issue with the search parser functions I found in lmfdb/search_parsing.py.  They appear to remove the “[“ and “]” from strings representing lists, but I want to keep those to query the database.   So I wrote another function “ parse_bracketed_posints2” which does the same thing as “parse_bracketed_posints” except keeps the brackets in the string.  The function is currently sitting in my code lmfdb/higher_genus_w_automorphisms/main.py 

8.  Added the family as a “Friend”  for each individual passport.  

9. 
> It's not clear to me why http://127.0.0.1:37777/HigherGenus/C/aut/3.16-11.0.2-2-2-4 is in the related objects list on http://127.0.0.1:37777/HigherGenus/C/aut/3.8-3.0.2-2-4-4.1

This is “working as intended”.  This “friend” corresponds to the statement on the page that “The full automorphism group for this family is C_2 \times D_4 with signature [0;2,2,2,4]”, meaning the generic element of the subspace of the moduli space M_g defined by the generating vectors in this passport  3.8-2.0.1-2-2-4-4.1  actually has automorphism group [16,11] with signature [0;2,2,2,4].

jenpaulhus:
@AndrewVSutherland For now groups up to 255 will be sufficient. Starting at genus 7 and up to genus 15 eventually, I will need groups up to order 504 plus (1092,25) which is PSL(2,13).  I will, however, not need any groups of order 256 until genus 17. Thanks for maintaining that database of small groups.

jenpaulhus:
Forgot to update the tests.  They should be fixed on this latest commit.

AndrewVSutherland:
@jenpaulhus Nice work!  These changes all look good, but I did notice two things:

* It looks like you changed the database format of the signature and group attributes in the curve_automorphisms.passports collection to strings, but you did not do this in the curve_automorphisms.families collection, was that intentional?  It might be confusing to have different formats in the two collections.
* The usage "from xxxx import *" is deprecated (see https://github.com/LMFDB/lmfdb/issues/122).  Can you please amend the imports in lines 14 and 18 of main.py accordingly?  Thanks!

jenpaulhus:
I got rid of the import * notation (so much for me being lazy!).  

The curve_automorphisms.families collection is what is currently on beta, but the new pages are all from curve_automorphisms.passports. It was easier to start from scratch with the data. So the families collection can be removed once these pages are up and running.

AndrewVSutherland:
@jenpaulhus Looks good, I will go ahead and merge this now.  Can you update https://github.com/LMFDB/lmfdb-inventory/blob/master/db-curve_automorphisms.md to reflect the fact that now only the passports collection is relevant?  Once this new code has been pushed to beta and prod I will remove the families collection.

AndrewVSutherland:
@jenpaulhus the families collection has been removed from beta and prod.  I also added indexes on the new passports collection for all the fields you search on (genus group, dim, hyperelliptic, signature, label, passport_label) and updated https://github.com/LMFDB/lmfdb-inventory/blob/master/db-curve_automorphisms.md.

I was going to add an index to support the sort you do on search results (lines 221-223 of main.py, but I noticed that you are sorting on 'g', 'dim', 'cc' but none of the records in the passports collection has an attribute named 'g'; did you perhaps mean to use 'g0' instead?

jenpaulhus:
@AndrewVSutherland  Oops! That should be 'genus' not 'g'.  I'll fix that for my next round of corrections.

PR #1543, opened 2016-06-13
removed references to now-redundant curves2 collection

JohnCremona:
This code change is **necessary** for the website to work with no curves2 collection, only the curve collection.  It will make beta work again, and is **necessary** that it is pushed to prod before the production database is updated.

Apologies for this not being as smooth as it should have been, owing to a typo introduced a few months ago which means that (without this change) it does not work to have curves2 existing but not curves.

### DIFF
--- a/lmfdb/elliptic_curves/import_ec_data.py
+++ b/lmfdb/elliptic_curves/import_ec_data.py
@@ -89,10 +89,8 @@ pw_dict = yaml.load(open(os.path.join(os.getcwd(), os.extsep, os.extsep, os.exts
 username = pw_dict['data']['username']
 password = pw_dict['data']['password']
 C['elliptic_curves'].authenticate(username, password)
-print "setting curves and curves2"
+print "setting curves"
 curves = C.elliptic_curves.curves
-curves2 = C.elliptic_curves.curves2
-
 
 def parse_tgens(s):
     r"""
@@ -601,8 +599,10 @@ def make_extra_data(label,number,ainvs,gens):
     return data
 
 def add_extra_data(N1,N2,store=False):
-    """
-    Add these fields to curves in the db with conductors from N1 to N2:
+    """Add these fields to curves in the db with conductors from N1 to
+    N2: NB This referes to a new collection 'curves2' which was
+    created temporarily when upgrading the data stored, and no longer
+    exists.
 
    - 'xainvs': (string) '[a1,a2,a3,a4,a6]' (will replace 'ainvs' in due course)
    - 'equation': (string)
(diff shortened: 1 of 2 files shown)


AndrewVSutherland:
checking now

AndrewVSutherland:
@JohnCremona As soon as this is live on beta (6 minutes or so) and we confirm it is working as expected, we should push it to prod, then update the production database with the curves2 --> curves switch, correct?

AndrewVSutherland:
This PR is now live on beta and EllipticCurve/Q is working again.

AndrewVSutherland:
@JohnCremona This PR is now live on prod and I made the database change on prod (curves --> old_curves, curves2 --> curves, curves2.rand --> curves.rand).  Everything looks OK but please take a look and let me know if anything seems amiss.

JohnCremona:
Thanks!  Will check all is OK when back from lunch.

JohnCremona:
All looks good.  The old_curves collection can be deleted.  If I had been more fail-safe I would have done similar chenges instead of using the dropTarget=True option which meant no going back.  (But even so it was only that typo which stopped things working momentarily).

Who actually (a) can and (b) knows how to do things like copying from warwick db to cloud db and renaming-typoe operations on the cloud db?  I have a feeling that only two people even know what all these versions of the db are called.  Is it documented somewhere?

edgarcosta:
(a) Who can do it? Short answer: anyone that can login on the machine ms.lmfdb.xyz (or any other cloud machine), in another words, the people that sent me their credentials.
Long answer:
- Sarunas Burdulis	
- Edgar Costa
- John Cremona
- David Farmer
- Andrew Sutherland
- John Voight
By changing a rule on the firewall, I can allow anyone that has access to lmfdb.warwick.ac.uk to do this. However, I don't think we want that.

(b) Drew has been the doing all the pushes, but yeah it is just me an him.

(c) Documentation: https://github.com/edgarcosta/lmfdb-gce/blob/master/setup/mongodb.md
We should move it at some point part of that repo to be under LMFDB

JohnCremona:
Thanks.  I though that the answer to (a) was as you wrote, though it does not work for me:
$ ssh ms.lmfdb.xyz
Permission denied (publickey).
For (b) it makes sense to have one person as lead on this so we don't create conflicts, but it also makes sense for others to know what to do.  (c) is helpful, though it will need several readings!

thanks

JohnCremona:
PS I will send you a new public key file

edgarcosta:
I emailed you about your username.

Regarding (c), please feel free to email me all the questions that you might have, and I will try to address them, and reword the description to make it a bit more clear.

AndrewVSutherland:
@JohnCremona I will create and document a general purpose script that automates what I have been doing to copy individual collections over on demand.  It probably does make sense to have one person managing this at any one point in time because we don't really have a good tool like git to coordinate data migration activity.

Issue #433, opened 2015-10-27
Add all automorphic friends of genus 2 curves


jvoight:
We did this in Dublin, but we didn't have access to all of the classical modular forms we needed.

jvoight:
For example, http://www.lmfdb.org/Genus2Curve/Q/169/a/169/1 splits into elliptic curves over QQ(sqrt(13)) (I think--certainly over the sextic totally real subfield of QQ(zeta_13)), so it should at least link to the HMF and probably also the elliptic curve.

jvoight:
Also, http://www.lmfdb.org/Genus2Curve/Q/676/b/ is the square of an elliptic curve, but is missing the corresponding modular form for that elliptic curve.

AndrewVSutherland:
In addition to missing automorphic friends, there are also some inconsistencies in the current data between what is listed in the curve pages, isogeny class pages, and L-function pages (see the following message from Dan Yasaki, for example).  I am working on fixing all the inconsistencies now and should have a fix ready later today.

> http://www.lmfdb.org/Genus2Curve/Q/1088/a/1088/1
> is a curve that is friends with two elliptic curves over a real
> quadratic field.  The isgoeny class
> http://www.lmfdb.org/Genus2Curve/Q/1088/a/
> is friends with no elliptic curves (or ec isogeny classes).

> Contrast that with
> http://www.lmfdb.org/Genus2Curve/Q/1696/b/434176/1
> which is a curve that is friends with two elliptic curves over an
> imaginary quadratic field.  The isogeny class
> http://www.lmfdb.org/Genus2Curve/Q/1696/b/
> is friends with the corresponding two elliptic curve isogeny classes.

> The curve
> http://www.lmfdb.org/Genus2Curve/Q/12500/a/12500/1
> is not friends with a HMF, but its isogeny class
> http://www.lmfdb.org/Genus2Curve/Q/12500/a/
> is friends with a HMF.

Doing a systematic match on the hashes of all relevant objects currently in the lmfdb (products of classical modular forms, products of elliptic curves over Q, Hilbert modular forms over real quadratic fields, and elliptic curves over real and imaginary quadratic fields) would not be difficult and is something I think should be fixed in Release 1.1.

JohnCremona:
[...]

PR #1546, opened 2016-06-13
Genus2 cleanup

AndrewVSutherland:
This PR cleans up a bunch of minor issues in the genus2 code/data and is in preparation for uploading a much larger set of genus2 curves with 5-smooth discriminants planned for later this month.

There are a lot of code changes (but the total amount of code is now significantly smaller), but the only visible changes should be:

* fixed a few small cosmetic/wording issues (e.g. tildeS_4 ->\tilde{S}_4)
* removed number of curves from isogeny classes properties box (this only counts elements of the isogeny class that happen to be Jacobians of curves that happen to currently be in the database, not very meaningful and not invariant)
* related objects are now consistent between curves and their isogeny classes and include some objects that were previously missing (this addresses part of #433).

Non-visible code changes include:

* isogeny_classes.py is now gone (this functionality is now included in web_g2c.py and much simpler), and all references to the isogeny_classes collection have been removed.  Once this PR is pushed to beta and prod the isogeny_classes collection can be completely removed (the fields root_number and bad_lfactors were the only ones not already present in curves and these have been moved there).
* modified versions of several array fields have been added to the genus2_curves collection that encode these arrays as strings for faster lookup (the documentation in LMFDB/inventory has been modified to reflect this -- this includes the fields: torsion_subgroup, g2_inv, igusa_inv, igusa_clebsch_inv, and eqn).
* a new collection Lfunctions.instances has been added for listing automorphic friends (related objects with the same L-function).  These records will eventually all move to the instances collection in the Lfunctions database (at which point I will change the code and this collection can be removed), but it makes sense to wait until #433 has been addressed before doing this).


AndrewVSutherland:
I meant to add that I did verify that all tests in test_genus2_curves.py pass (and I added a few new ones), and I also ran a test that loads every single genus 2 curve and isogeny class home page to make sure nothing got broken as a result of all these changes (see genus2_test_pages.py, it takes about 10 hours to run so I don't recommend rerunning it unless you are really paranoid).

AndrewVSutherland:
The relevant data changes have also been made on the production database (they are backward/forward compatible so both the new and old code work on the new data -- I only added new fields, I will remove the old fields once this PR has been pushed to both beta and prod).

JohnCremona:
checking and testing now

JohnCremona:
In db-elliptic_curves.md I'm not sure why the utf-8 encoding line has to be in there 3 times -- wasn't 2 enough?  (see the 4th quote at https://en.wikiquote.org/wiki/The_Hunting_of_the_Snark)

JohnCremona:
In line 185 of  lmfdb/genus2_curves/genus2_curve.py I don't like the "assert" statement, though testing for a None result is good.

What is the final web page section "Remainder of the endomorphism lattice by field" about?  e.g. at http://localhost:37777/Genus2Curve/Q/336/a/172032/1

I like the simplified functions like g2c_db_curves() without the global variable.  I think I started that with elliptic curves, and have no idea why I thought it worthwhile.

JohnCremona:
(continuing)
http://localhost:37777/Genus2Curve/Q/336/a/  has 2 elliptic curves as related objects, though they are not mentioned on the page itself -- more importantly, the links do not work!  The url_for function should be used, rather than whatever the function lfunction_friend_from_url is currently doing.

This last point is the only one which I think must be fixed before merging -- and I know that it is not an issue claused by the PR, I just think it better to fix now if possible.

AndrewVSutherland:
Agreed, I will fix this (and respond to your other points) later today (I'm currently attending Rubin 60).  Thanks for the feedback!

AndrewVSutherland:
@JohnCremona I have addressed all the issues you identified (I also incorporated the change you made to web_ec.py in PR #1543, since I could not test without it; this should have no impact when merging).

BTW, there was no problem in lfunction_friend_from_url, which was just reading URLs from the database just like the Lfunctions code does (and will be reading them from the same collection once the records in the temporary genus2_curves.Lfunctions.instance collection move to Lfunctions.instances where they belong).  These URLs can/should be changed when you change the label/url format for ecnf curves.

The Lfunctions code uses URLs rather than object labels and calls to url_for(...) for a good reason -- you don't want code that could potentially need to display a link to any object in the database to have to know the names of all the appropriate entry points to pass to url_for(...), and the names of the arguments those methods are expecting.   URLs for an object's home page in the LMFDB should be as immutable as the object's label, since third parties may very well use these links externally.

Of course the code that creates the Lfunctions.instances record in the first place should use "url_for(...)" to construct the URL it places there.

JohnCremona:
OK, I checked that the elliptic curve links now work, and I see that the final section is omitted when it has not content.  Good!
As for the URL manipulation, so be it.

PR #1559, opened 2016-06-15
Fix for #1555

### LINKED ISSUE #1555: Bad bread on Siegel modular forms pages
If you go to http://www.lmfdb.org/ModularForm/GSp/Q/Sp4Z_j/ and click on any entry in the table with rows indexed by k and columns indexed by j you will be taken to the page:

http://www.lmfdb.org/ModularForm/GSp/Q/Sp4Z_j/j/k/

which displays information about the space M_{k,j}.  If you then click on the last bread crumb (labelled M_{k,j} it will try to take you to the page

http://www.lmfdb.org/ModularForm/GSp/Q/Sp4Z_j/k/j/

for the space M_{j,k} (note that j and k are now reversed).  This is confusing in general, and when only one of the spaces M_{j,k} and M_{k,j} exists it causes internal server errors, for example, go to the page

http://www.lmfdb.org/ModularForm/GSp/Q/Sp4Z_j/0/10/

and click on the last breadcrumb to get an internal server error.

AndrewVSutherland:
In addition to fixing #1555, this fix swaps the order of j and k in the URLs previously written as

http://www.lmfdb.org/ModularForm/GSp/Q/Sp4Z_j/j/k/

to

localhost:37777/ModularForm/GSp/Q/Sp4Z_j/k/j/

which is now consistent with the displayed M_{k,j}.  This inconsistency is likely what caused the bug in the first place.  This does mean that external URLs linking to these pages could be broken.  If this is a concern, please say so and I will change it back to its previous inconsistent state (but still fix #1555).

### DIFF
--- a/lmfdb/siegel_modular_forms/siegel_modular_form.py
+++ b/lmfdb/siegel_modular_forms/siegel_modular_form.py
@@ -48,13 +48,13 @@ def rescan_collection():
     global COLNS
     COLNS = colns
 
-@app.route('/ModularForm/GSp/Q/Sp4Z_j/<j>/<k>')
-@app.route('/ModularForm/GSp/Q/Sp4Z_j/<j>/<k>/')
+@app.route('/ModularForm/GSp/Q/Sp4Z_j/<k>/<j>')
+@app.route('/ModularForm/GSp/Q/Sp4Z_j/<k>/<j>/')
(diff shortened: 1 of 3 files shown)


JohnCremona:
@nilsskoruppa Can perhaps comment on the last part.  Otherwise I tested this and am happy with it.

Issue #1561, opened 2016-06-16
ecnf twist link broken


JohnCremona:
Dan Yasaki reported this.  Go to any home page of an elliptic curve over a number field, for example http://beta.lmfdb.org/EllipticCurve/4.4.4913.1/169.6/a/2 and click on the Twists link under Related Objects.  It fails since when we do a search for curves with a fixed j-invariant we need to specify the field.
I will fix this.

AndrewVSutherland:
Fixed by #1562

PR #1562, opened 2016-06-16
fix for #1561: ecnf Twists link

JohnCremona:
Two bugs in one line, both caused by the search-parsing code (this used to work, honest!):  a parameter name is 'field' and not 'field_label', and if you pass something like jinv=L as part of a **kwargs where L is a list, then it arrives as multiple arguements jinv=L[0], jinv-L[1], ... which fails.  This was fixed by passing in the j-invariant as a string representing a polynomial in the variable, which the parser converts to a coefficient list before looking in the database.


AndrewVSutherland:
Looks good.

Issue #1548, opened 2016-06-13
Elliptic curves could/should show Atkin-Lehner eigenvalues (local root numbers)


JohnCremona:
For example http://localhost:37777/EllipticCurve/Q/90/c/2 could show the local root numbers in the Local Data section.  It is +1 or -1 respectively for split and non-split multiplcative reduction, which can be read off instantly (but would still be worth showing since it is hard to remember which is which), and requires more work in the additive case.  Because of the latter it should be stored in the database.  If this had been suggested to me before we made the switch to having local doata stored it would have been trivial to add it.   As it is, it will require adding to the 'local_data' component to the stored data for every curve.

JohnCremona:
See #1574.  All the precomputation has been done and stored in a new collection.  When that has been merged and the collections renamed, this can be closed.

AndrewVSutherland:
As discussed on #1574, the collections have now all been updated.  You should still change the code to actually use the data at some point (not urgent).  Feel free to close or leave this issue open as a reminder as you see fit.

JohnCremona:
Ah, now I see where the confusion is.  The code already uses the new data when it is there!  That code changes was already in #1563 .  So I'll close this.

Issue #534, opened 2015-10-29
Knowl opened from a table header appears centered


sibilant:
From the Dirichlet character main page, you can search for characters with specified modulus, conductor and order. The result is a displayed table of characters with header titles such as "character", "conductor", etc. If you click on a header to open the knowl, the knowl text appears centered (presumably because the table headers are). 

Example: 
http://www.lmfdb.org/Character/Dirichlet/?modulus=1&conductor=1&order=1
Click on "character" in the table header for a particularly striking example.

alinabucur:
This was a problem with a lot of pages, Anna Haensch and I fixed a bunch in Corvallis. It is basically an issue with redefining table styles in various homepage templates.

AndrewVSutherland:
After the recent round of cleanup to the Dirichlet character pages this problem no longer occurs.

PR #1563, opened 2016-06-16
added display of local root numbers to elliptic curves /Q

JohnCremona:
See Issue #1548.  This shows local root numbers for elliptic curves over Q (and adds a knowl saying what they are).
Since these are not currently stored in the database for primes of additive reduction, they are currently computed on the fly if (and only if) needed.  I will also add the additional data to the database, but that is a separate operation, and this will work both with and without that being done.

A similar thing will be done for elliptic curves over number fields, which are having a data update anyway.


AndrewVSutherland:
This is not specific to this PR (which I will now merge), but I notice that the knowls on the column headings of the local data table all get center justified when they are displayed, which seems very strange.

JohnCremona:
Yes, I noticed that.  Someone with more css knowledge than me might know how to fix that -- I don't!

davidfarmer:
The CSS problem is issue #534

PR #1567, opened 2016-06-17
Add st links

AndrewVSutherland:
This PR adds links to appropriate Sato-Tate group pages to L-function pages, e.g., compare

http://www.lmfdb.org/L/EllipticCurve/Q/11.a/
http://localhost:37777/L/EllipticCurve/Q/11.a/

http://www.lmfdb.org/L/Genus2Curve/Q/336/a/
http://localhost:37777/L/Genus2Curve/Q/336/a/

and also to endomorphism lattice pages, e.g., compare (after scrolling down)

http://www.lmfdb.org/Genus2Curve/Q/4608/c/27648/1
http://localhost:37777/Genus2Curve/Q/4608/c/27648/1

Additionally, it standardizes Sato-Tate group page link formatting (anyone who wants to display a Sato-Tate link should use lmfdb.sato_tate_groups.st_link_by_name) and always redirects to page with the full Sato-Tate group label in the URL.

All tests in test.sh pass.

### DIFF
--- a/lmfdb/ecnf/WebEllipticCurve.py
+++ b/lmfdb/ecnf/WebEllipticCurve.py
@@ -6,6 +6,7 @@ from sage.all import ZZ, var, PolynomialRing, QQ, GCD, RR, rainbow, implicit_plo
 from lmfdb.base import app, getDBConnection
 from lmfdb.utils import image_src, web_latex, web_latex_ideal_fact, encode_plot
 from lmfdb.WebNumberField import WebNumberField
+from lmfdb.sato_tate_groups.main import st_link_by_name
 from kraus import (non_minimal_primes, is_global_minimal_model, has_global_minimal_model, minimal_discriminant_ideal)
 
 ecnf = None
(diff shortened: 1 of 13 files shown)


JohnCremona:
Sorry, just noticed this.  Will test.

JohnCremona:
Just one thing: in the last of the examples above, the SU(2) is not a link, it seems.  Is that because it is ST^0 and not ST?  If there is a reason, please merge this.

AndrewVSutherland:
@JohnCremona This is intentional, SU(2) is used to identify component of the Sato-Tate group (and really should be written SU(2)_2, I will change this in a later PR).  We don't currently have separate home pages for identify components; there is a corresponding connected Sato-Tate group E_1 (which is of course a subgroup of J(E_4)), but it would be confusing to list this instead of SU(2).

I guess the key point is that identify components of Sato-Tate groups and Sato-Tate groups live in different categories; the latter comes equipped with a map on Frobenius elements and a (possibly trivial) Galois action on its components, while the former is just a connected compact Lie group.  Going from E_1 to SU(2) amounts to applying a forgetful functor to E_1.

JohnCremona:
Thanks for the detailed explanation!  It confirms that what I know about S-T groups could be written on the back of a 2x2 matrix.

PR #1574, opened 2016-06-21
Switch to new ec/Q collection

JohnCremona:
There is a new version of the collection curves in database elliptic_curves, called new_curves.  Two differences: (1) local data now includes root numbers, which are already displayed but now require no computation on the fly; (2) updated minimal quadratic twist information, following a well-defined definition (documented elsewhere) rather than the current Sage version which, when there is more than one twist of the same minimal conductor picks one essentially at random, and also, when the current curve is already of minimal conductor, use itself.  The new scheme sort first by conductor, then by abs value of discriminant, and finally (there can be at most 2 remaining candidates, which are -1-twists of each other) picks the one with positive c6.

Most of the new code here is in the import script which was run once-only to create the new database using the rewrite_collection function.  Otherwise all it does is make sure that the collection new_curves is used when present, but it should work fine when not present (please check that part of the code to make sure).  When this is on beta (resp. prod) we can do the usual: rename 'curves' to 'old_curves' and 'new_curves' to 'curves'; then drop 'old_curves'.

### DIFF
--- a/lmfdb/elliptic_curves/import_ec_data.py
+++ b/lmfdb/elliptic_curves/import_ec_data.py
@@ -559,6 +559,42 @@ def add_numerical_iso_codes(N1,N2):
         data['iso_nlabel'] = numerical_iso_label(C['lmfdb_iso'])
         curves.update_one({'_id': C['_id']}, {"$set": data}, upsert=True)
 
+def xainvs(E):
+    return ''.join(['[',','.join([str(a) for a in E2.ainvs()]),']'])
+
+def min_quad_twist(E):
+    r""" Refined version of Sage function (as of version 7.2).  Sage picks
+at random if several twists have the same conductor, but we want to
+sort by (conductor, abs(discriminant), label).
+    """
+    E1, D1 = E.minimal_quadratic_twist()
+    # E1 is minimised at primes >3 but we want to choose
+    # systematically from the 8 twists by <-1,2,3>:
+    tw = [-1,2,-2,3,-3,6,-6]
+    Elist = [E1] + [E1.quadratic_twist(t).minimal_model() for t in tw]
+    Elist=list(Set(Elist))
+    # We sort these first by conductor, then by abs(disc), then by
+    # LMFDB label.  For the latter we need to look them all up in the
+    # database.
+    ainvs = [xainvs(E2) for E2 in Elist]
+    labels = [curves.find_one({'xainvs':ai},fields={'lmfdb_label':True})['lmfdb_label'] for ai in ainvs]
+    for Ei,lab in zip(Elist,labels):
+        Ei.lab = lab
+    sort_key = lambda e: (e.conductor(), e.discriminant().abs(), e.lab)
+    Elist.sort(key=sort_key)
+    # we report when the tie-break on label is used
+    keys = [sort_key(E2) for E2 in Elist]
+    if keys[0][:2]==keys[1][:2]:
+        print(labels)
+        D = Elist[0].is_quadratic_twist(Elist[1])
+        N = E.conductor()
+        print("N={}. D={}.  {} ~ {}".format(N,D,Elist[0].ainvs(),Elist[1].ainvs()))
+    Et = Elist[0]
+    D = E.is_quadratic_twist(Et) # 1 or square-free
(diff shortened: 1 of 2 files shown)


AndrewVSutherland:
I'll make you a deal :), I will quickly review and merge this if you do 
the same for #1567 (which is really quite minor and has been sitting 
there for 4 days).

JohnCremona:
It's a deal -- see #1567 comment.

AndrewVSutherland:
:) You beat me to it.

AndrewVSutherland:
@JohnCremona It looks fine to me (I didn't test the import script, I'll leave that to you), just one question, what is your intended use of the automatic "new_curves" check? I ask because normally when I am testing a change to the data I create a new_curves file and modify my local branch to read from new_curves, but I do not necessarily want the beta.lmfdb.org to immediately start using data in the new_curves file, since I am still testing it (I will then move curves to old_curves and new_curves to curves once I am happy that new_curves works and is backward compatible).

JohnCremona:
@AndrewVSutherland I did do my own testing of the new data vs. the old, so by making this PR I am asking for the process of switching over to the new collection to start.  So I don't really understand the question -- apart from the non-live code, the **only** code change is to read by default from new_curves.  The purpose of making the db_ec() function try both collection names is precisely so that this code will work through the collection renaming and that propagating through to the production database.  Does that make sense?

AndrewVSutherland:
@JohnCremona So here is the problem I see.  The process for moving a collection (or group of collections) from beta (Warwick) to prod (the cloud) works by first copying all the data over into new collections named "new_*" on the assumption that the software will not look at them while the copying is under way (not all the collections will get there at the same time, index creation may not have happened yet, etc...).

Only once all the data has been copied over and indexed, the existing collections (e.g. "curves") are renamed to old_* and the new_* collections are renamed *.  This last step is essentially instantaneous (a few milliseconds even when there are several collections involved).

Once we are happy that everything is working as expected, the old_* collections get dropped.  The key point is that on prod there is never a new_* in existence other than during the copying operation, and during this time we don't want the software to look at it.

Now I could switch to using a different prefix (e.g. "temp_*") if you have a good reason for wanting to use new_curves, but that would be a reason different from the propagation issue, which is why I asked.

JohnCremona:
Understood.  Don't merge this until after the collections have migrated, and then I may/will change the PR not to mention new_curves.  The website should work whether the collection has the new data or not (since it works right now when there are no root numbers stored).

AndrewVSutherland:
@JohnCremona In terms of the collection migration, here is what I think you want me to do, but please clarify if I am wrong.

(1) on warwick: curves -> old_curves, new_curves -> curves
(2) copy curves collection from warwick to cloud
(3) wait for you to decide if you want to change this PR or not (I predict you will decide that you do)
(4) merge, push to beta, and then push to prod (after confirming all is well on beta)

In step (2) the current curves collection in the cloud will be moved to old_curves.  As noted above, during the copying process (which will take an hour or so) there will be a temporarily be a collection called "new_curves" in the cloud, but the fact that this is the same you used for the "new_curves" collection currently in the elliptic_curves database in warwick is purely coincidental (and if you decide you do not want to change this PR, I will change the script to use the prefix "xyz_" instead (a nonce might be better, but that has downsides as well).  This won't be a problem because the software currently running on prod does not look at new_curves.

JohnCremona:
That sounds perfect, thanks.

AndrewVSutherland:
@JohnCremona Steps (1) and (2) are now complete.  As you can see at http://www.lmfdb.org/api/elliptic_curves/curves, the 'rootno' attribute is now  there.  The old collection is at http://www.lmfdb.org/api/elliptic_curves/old_curves.  It actually only took 10 minutes (wired tiger is fast).

I am now on step (3) (waiting for you to decide to update this PR :)).

JohnCremona:
No merging is needed.  The existing code does the right thing with the enhanced database.  I can later simplify the code by assuming that the rootno data is already there, but that's not urgent.  When I do I will make a new PR, so after this comment I'll close this one.  Thanks!

I think what I really needed instead of this PR was to issue a **database** pull request!  In effect that is what you have acted on.

PR #1578, opened 2016-06-22
fixed download and added test

AndrewVSutherland:


AndrewVSutherland:
One line fix, also added test to check for this bug recurring in the future.

PR #1583, opened 2016-06-23
sort databases/collections/fields in api html interface

AndrewVSutherland:
Small change to the api web pages so that the databases and collection lists on http://127.0.0.1:37779/api/ or now sorted (as opposed to http://www.lmfdb.org/api/).

On the html pages for individual collections and documents, document attributes are now sorted by key name.  This only affects the html display, the yaml and json formats already get sorted this way.


AndrewVSutherland:
As this PR has received no comments and has no impact on the main website (it is just a cosmetic change to the /api html pages to make it easier to find things), I am going to go ahead merge.  If anyone objects, please say so and I will revert.

PR #1595, opened 2016-06-27
meta description is now hardcoded in base.py

### LINKED ISSUE #1594: The meta description is a knowl
***1 Upvote*** Hello,

This line:
`
<meta name="description" content="Welcome to the LMFDB, the database of L-functions, modular forms, and related objects. These pages are intended to be a modern handbook including tables, formulas, links, and references for L-functions and their underlying objects.">
`
common to all the webpages, comes from this knowl:

http://beta.lmfdb.org/knowledge/show/intro.description

see: https://github.com/LMFDB/lmfdb/blob/master/lmfdb/base.py#L203

Shouldn't this be hardcoded?

Best,
Edgar

edgarcosta:
Fixes #1594

### DIFF
--- a/lmfdb/base.py
+++ b/lmfdb/base.py
@@ -33,8 +33,9 @@ def get_logfocus():
 # getDBConnection() and should always be obtained from that)
 _C = None
 
+#FIXME perhaps the globals dbport and DEFAULT_DB_PORT should be removed?
+#we could had it a default argument to makeDBConnection
 DEFAULT_DB_PORT = 37010
-#FIXME perhaps the global dbport removed
 dbport = DEFAULT_DB_PORT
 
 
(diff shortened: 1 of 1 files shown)


edgarcosta:
If this ends up getting merged to prod, we should then consider deleting the knowl: 
http://beta.lmfdb.org/knowledge/show/intro.description

AndrewVSutherland:
Once this is pushed to beta and prod I will delete the info.description knowl.  FYI, I edited http://beta.lmfdb.org/knowledge/show/intro to make the first para match the (now hard-coded) meta text (of course the intro text can still be edited by anyone with access to the knowl database, but I think that is OK (it only appears on the page http://www.lmfdb.org/intro, as opposed to the meta tag which gets inserted into every page we serve).

PR #1598, opened 2016-06-27
Index.html cleanup

AndrewVSutherland:
This PR removes the obsolete lmfdb/templates/index.html which references info.description (see #1595) but is no longer used (the page route "/" is handled in line 61 of pages.py and uses index_boxes.yaml).

I also modified about.html to use meta_description rather than the (empty) description.


JohnCremona:
checking...

PR #1606, opened 2016-06-27
Fix for #1006

### LINKED ISSUE #1006: Siegel modular forms half integral weight page
The URL is

 http://beta.lmfdb.org/ModularForm/GSp/Q/Gamma0_4_half/

The page does not indicate that the degree is 2.

Also, it does not point to a knowl for half-integral weight Siegel modular forms; in fact, there may not be such a knowl.

AndrewVSutherland:
This PR addresses #1006 by adding a degree attribute to families of Siegel modular forms that is then displayed whenever a dimension table is shown.  The word degree has a knowl link in it (the knowl can be created after this PR is merged), it refers to to the integer g for which the group G appearing in M_{k,j}(G) is a group of 2g x 2g matrices (currently g is 2, 3, or 4).

This meant touching a lot of files, but the changes are all very minor.

### DIFF
--- a/lmfdb/siegel_modular_forms/collection.py
+++ b/lmfdb/siegel_modular_forms/collection.py
@@ -38,6 +38,7 @@ class Collection (SageObject):
         if not self.__latex_name:
             self.__latex_name =  Latex(self.name())
         self.__description = doc.get('description')
+        self.__degree = doc.get('degree')
         self.__dimension = doc.get('dimension')
         if self.__dimension:
             a,b,c = self.__dimension.rpartition('.')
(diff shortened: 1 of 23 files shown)


AndrewVSutherland:
If no one has any comments, I am going to go ahead and merge this -- there are some other SMF issues I'd like to address and I don't want to add those changes to this PR or create a merge conflict.  The visible change here is minor and uncontroversial (it just adds the degree to the dimension table heading).

Once I have finished addressing all the SMF issues that can easily be addressed by someone who is not an expert on Siegel modular forms (i.e., me) I will invite Jerry Shurman to take another look and hopefully provide content for some of the knowls I have created.

JohnCremona:
Please go ahead.  I cannot test right now but the code changes are much less than the number of files touched might suggest (which is not to say that they are not worth making, every one!)
In fact I'll merge it myself, so I can share any blame ;)

Issue #1609, opened 2016-06-28
Remove upload from website.py


AndrewVSutherland:
If you go to http://www.lmfdb.org/upload/viewAll you can see the contents of the upload database, which we would like to remove from the cloud mongo db (there is no reason for it to be there).  Moreover, if you go to http://www.lmfdb.org/users and log in, you will then see upload files on the side bar and can attempt to upload a file.  It will fail because of the access controls we now have in place.

This is also true in beta, i.e. you can try to upload a file but it will fail.  Until we decide that we want to allow file uploads and make a decision as to whether this functionality should be available on www.lmfdb.org or only on beta.lmfdb.org, I think we should remove it from the sidebar in homepage.html and not load it in website.py.

JohnCremona:
I agree.  I think this idea that we should make it easy for anyone to upload data and have something sensible happen to it was never a good one.  It is fine to delete any such data with my name on it, which I once uploaded at a workshop to try out the facility when it had just been set up (several years ago).

AndrewVSutherland:
Addressed by #1610 and now on beta and prod.

AndrewVSutherland:
The upload database has been removed from the cloud (still present on Warwick).

PR #1610, opened 2016-06-28
Remove upload from sidebar

AndrewVSutherland:
Addresses issue #1609

All tests now pass (I commented out Box 6 sage link test which I could not get to work -- I did change the link to http://www.sagemath.org and the link is correct).

### DIFF
--- a/lmfdb/index_boxes.yaml
+++ b/lmfdb/index_boxes.yaml
@@ -46,7 +46,7 @@ links:
 ---
 title: Code and Open Software
 image: compute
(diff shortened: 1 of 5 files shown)


JohnCremona:
Did you try using https?  The sage servers have switched over, I think.

AndrewVSutherland:
@JohnCremona Yes I did, and it also does not work.

JohnCremona:
OK just checking.  I'll look at this one this evening unless someone else gets there first.

PR #1613, opened 2016-06-28
Fix regex for checking artin rep labels.

jwj61:
The previous regex failed in two ways.  The "e" is optional in the conductor part (was manditory in the old regex), and T-numbers can have more than one digit in the Galois part.

To test, enter the following label into the corresponding search box: 2.2e2_3_11e2.12t12.2c1 .  The parts that didn't work before are the second 12 in 12t12 and the _3_ in the conductor part (2e2_3_11e2 means 2^2*3*11^2, and there is no exponent given for 3 since it is 1).

### DIFF
--- a/lmfdb/artin_representations/main.py
+++ b/lmfdb/artin_representations/main.py
@@ -44,7 +44,7 @@ def make_cond_key(D):
 
 def parse_artin_label(label):
     label = clean_input(label)
-    if re.compile(r'^\d+\.\d+(e\d+)?(_\d+(e\d+)?)*\.\d+t\d+\.\d+c\d+$').match(label):
+    if re.compile(r'^\d+\.\d+(e\d+)?(_\d+(e\d+)?)*\.\d+(t\d+)?\.\d+c\d+$').match(label):
         return label
     else:
         raise ValueError("Error parsing input %s.  It is not in a valid form for an Artin representation label, such as 9.2e12_587e3.10t32.1c1"% label)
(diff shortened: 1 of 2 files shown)


AndrewVSutherland:
The label you give as an example works with the new regex, but neither the new nor the old regex accepts the label 18.2e39_5e8_71e12_26449e12.60.1c1: 

http://localhost:37777/ArtinRepresentation/?natural=18.2e39_5e8_71e12_26449e12.60.1c1&search=Go
http://lmfdb.org/ArtinRepresentation/?natural=18.2e39_5e8_71e12_26449e12.60.1c1&search=Go

But this is the label of an Artin representation in the database:

http://lmfdb.org/ArtinRepresentation/18.2e39_5e8_71e12_26449e12.60.1c1

Also, I suggest widening the input box for labels on the search screen so you can see what you are typing with long labels such as this one.

jwj61:
Good points, will do.

jwj61:
In addition to the adjustments in the code, I updated the knowls for the artin labels and the smallest containing permutation representation to include this case, when the degree of that representation is too big to give a t-number.

AndrewVSutherland:
Looks good, merging.

But I think there is a typo in the label knowl, it says "we give the degree d and t-number j in the form ntj; otherwise we use just the degree d", but shouldn't "ntj" be $d$t$j$?

jwj61:
Fixed n to d, and switch to html-code from math mode for the label part since you type it.

Issue #1599, opened 2016-06-27
HMF search for degree=2 and dimension=2 times out on prod (but not on beta)


AndrewVSutherland:
The search http://www.lmfdb.org/ModularForm/GL2/TotallyReal/?field_label=&deg=2&disc=&weight=&level_norm=&dimension=2&count=100 times out after 30 seconds with a server error.  But the same search on beta http://beta.lmfdb.org/ModularForm/GL2/TotallyReal/?field_label=&deg=2&disc=&weight=&level_norm=&dimension=2&count=100 works fine, and on prod searching for just dimension=2 or just degree=2 works fine.

http://www.lmfdb.org/ModularForm/GL2/TotallyReal/?field_label=&deg=&disc=&weight=&level_norm=&dimension=2&count=100
http://www.lmfdb.org/ModularForm/GL2/TotallyReal/?field_label=&deg=2&disc=&weight=&level_norm=&dimension=&count=100

Reported by Marc Masdeu at Warwick (respond via e-mail when fixed).

AndrewVSutherland:
I am adding an index to better support this kind of query; HMF searches may be unavailable for the next few minutes while the index builds.

AndrewVSutherland:
@edgarcosta I have done some testing and I believe this is may be a performance issue caused by the switch to wired tiger zlib.  I just recopied all the HMF data over from Warwick so the hmfs.forms collection in the two dbs are identical, and they have the exact same indexes, in the same order.

The issue is that when performing a query that cannot be resolved entirely by index lookup, some subset of the records in the collection will need to be scanned, and in the case of the forms collection, the average size of these records is quite large (about 200kb) and my guess is that with zlib each of these is getting decompressed (even though only one or two attributes needs to be examined) and this takes a lot of time.

Here is a timing comparison between a query run on atkin and a query run on www.cental0.lmfdb.xyz that I believe reflects what mongo db is doing internally to resolve this query (it searches the list of records with dimension 2 to figure out which also have degree 2):

On Atkin:

time A=[(r['label'],r['deg']) for r in db.forms.find({'dimension':int(2)},{'label':True,'deg':True})]
CPU times: user 607 ms, sys: 48 ms, total: 655 ms

On www.central0.lmfdb.xyz:

time A=[(r['label'],r['deg']) for r in db.forms.find({'dimension':int(2)},{'label':True,'deg':True})]
CPU times: user 464 ms, sys: 44 ms, total: 508 ms
[...]

PR #1615, opened 2016-06-29
Fix for #1599

AndrewVSutherland:
This pull request addresses issue #1599 by creating a new collection hmf.forms.search that is a copy of hmf.forms without the eigenvalue data.  The total size of this collection is less than 1/400 the size of hmf.forms (about 90MB on mmap, about 14MB on wired tiger), so the space increase is completely negligible (this collection is now present both on warwick and in the cloud).

All queries to hmf.forms in hilbert_modular_forms.py that do not require eigenvalue data are now redirected to hmfs.forms.search and are very fast.  On beta.lmfdb.xyz, they are faster than they were (by a factor of up to 10 or 20), and once this PR is pushed to prod, searches on www.lmfdb.org will be as much as 1000 times faster -- e.g. searching for weight 2 and dimension 1, which currently takes more than 10 minutes (as noted on #1599, iit keeps running even after the web page times out) should take less than half a second.

I would be grateful if someone would please review this PR and merge it quickly; until it is merged www.lmfdb.org can easily be brought down by a half dozen ill-chosen queries on hmfs.

### DIFF
--- a/lmfdb/hilbert_modular_forms/hilbert_modular_form.py
+++ b/lmfdb/hilbert_modular_forms/hilbert_modular_form.py
@@ -1,4 +1,5 @@
 # -*- coding: utf-8 -*-
+# -*- coding: utf-8 -*-
 
 import re
 import pymongo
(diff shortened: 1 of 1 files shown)


JohnCremona:
Looking now

Issue #1663, opened 2016-07-07
Higher Genus Curve download not working


jenpaulhus:
The pages from #1657 are returning an error in beta.
http://beta.lmfdb.org/HigherGenus/C/aut/3.14-2.0.2-7-14/download/magma

When I checked locally, it looks like the issue is that "make_response" is not being imported from flask which is called on line 497 of main.py.

However, when I look at the files I committed (which include the import), they are a bit different than the ones that were merged
https://github.com/LMFDB/lmfdb/pull/1657/commits/1c1ff28b47fa2e2520b0694e28a8f437035e095f
vs
https://github.com/LMFDB/lmfdb/commit/c2d10f19d6d652279f2099a3be6f8bc5bec5959a

Somehow along the way a few imported commands got lost.  Is there a way to fix this directly or should I just make a new pull request with that command added back in?

AndrewVSutherland2:
@jenpaulhus -- let me figure out what went wrong here before you make any
further changes.

On Thu, Jul 7, 2016 at 8:23 AM, Jen Paulhus, Ph.D. <notifications@github.com

AndrewVSutherland:
@jenpaulhus OK, the make_response import inadvertently got removed in PR #1592 (which was applied to a version of main.py that was not using make_response).  This wasn't flagged as a merge conflict by git because your most recent PR didn't change this line.  I have corrected it and pushed the change to beta, once I confirm the fix works I will close this issue.

pyflakes did not report any other missing imports, so this should be the only issue.  But it did notice that line line 542 of main.py the variable num_entries is assigned but never used -- you might want to take a look at this.

AndrewVSutherland:
OK, the downloads now work on beta.  Closing.

PR #1657, opened 2016-07-05
Add download functionality to higher genus curves w/automorphisms

jenpaulhus:
This pull request adds download options to both the family pages and passport pages off of http://127.0.0.1:37777/HigherGenus/C/aut.  Code for both Magma and GAP can now be downloaded.  For examples see:
http://127.0.0.1:37777/HigherGenus/C/aut/3.96-64.0.2-3-8.1
http://127.0.0.1:37777/HigherGenus/C/aut/3.8-2.0.2-2-4-4
and click on the "Download [language] code"

A couple of other small changes:
* changed the word "passport" to "refined passport" in a number of places
* fixed issue of searching over 'g' instead of 'genus' as referenced in #1541. 
* removed "Completeness of the data" from the "Learn more about"  here: http://127.0.0.1:37777/HigherGenus/C/aut/ since it now shows up at the top of the page.

A couple of questions:
* An old file lmfdb/higher_genus_w_automorphisms/template/hgcwa-show-curve.html is no longer needed.  I'm not sure how to request a delete of a file on github.
* Once I write a few remaining knowls, these pages are probably ready to go to "beta", i.e. have an official link on the sidebar for beta.  Do I just make a pull request with the new proposed link on the sidebar, or is there another way to request this?

PR #1678, opened 2016-07-08
Higher Genus curves w/autos: a series of small fixes

jenpaulhus:
Suggestions from #1657:
1. Changed the link from /HigherGenus/C/aut to HigherGenus/C/Aut.  But, could we put a short term redirect from the lower case version to the upper case version? I have given a few people the link with "aut" and would like it to redirect instead of returning an error.

2. Separated breadcrumbs and added links for each one, see for example:
http://127.0.0.1:37777/HigherGenus/C/Aut/5.48-49.0.2-6-6
http://127.0.0.1:37777/HigherGenus/C/Aut/5.48-49.0.2-6-6.5

3. Added random refined passport option at http://127.0.0.1:37777/HigherGenus/C/Aut/

4. Uniformized error handling

5. Added Next/Previous at the top of searches, see:
http://127.0.0.1:37777/HigherGenus/C/Aut/?genus=5

6. Added more tests to test_hgcwa.py

From #1663 deleted "num_entries" variable which was assigned and never used.


AndrewVSutherland:
@jenpaulhus These look like good changes, just a few small issues:

* You need to enclose the latex breadcrumb for the group in $'s or \( \);
* The random link should redirect so that the user sees the correct URL for the passport page in their URL bar (e.g. if they want to bookmark the page, they don't want to bookmark /random);
* Now that the error handling is uniform, I think you can remove the "Error" heading and info.err display from hwcga-search.html (just display "Either modify your search above or go back").
* I think you can remove hwcga-error.html, which does not appear to be used any longer.

Regarding redirecting .../aut to .../Aut, it would be a slight pain to put in redirect, but what you could do instead (in the short term) is to have both lines

app.register_blueprint(higher_genus_w_automorphisms_page, url_prefix="/HigherGenus/C/aut")
app.register_blueprint(higher_genus_w_automorphisms_page, url_prefix="/HigherGenus/C/Aut")

in your __init__.py (all the internal links will still use /Aut, and anyone poking around will see this).  

When we go to beta you should remove the "/aut" line (the link will then be accessible on the sidebar anyway, so it shouldn't be a problem for people to find it).

jenpaulhus:
@AndrewVSutherland Made the small changes you suggested above.  

1. Breadcrumb groups are now showing as latex.
2. Random curves are given their proper passport label.
3. I deleted the Error piece of the hgcwa-search.html file. (In a minute I will make one last commit because I just remembered that I need to  delete commented out code).
4. I removed /template/hgcwa-error.html file.
5. I added the line app.register_blueprint(higher_genus_w_automorphisms_page, url_prefix="/HigherGenus/C/aut")  back into the init.py file (although I put it second because if it was first, the urls on the rest of the pages all defaulted to the "aut" instead of "Aut".)

AndrewVSutherland:
@jenpaulhus It looks like there is a typo in hgcwa-search.html causing a server error on http://127.0.0.1:37777/HigherGenus/C/Aut/?genus=&group=&signature=&dim=&inc_hyper=include&count=20&Submit=Search, maybe an extra {% endif %}.

jenpaulhus:
@AndrewVSutherland Oops!  Should be fixed now.

AndrewVSutherland:
Looks good!

Issue #1692, opened 2016-07-09
Server error on degree 16 Number field L-functions


AndrewVSutherland:
In the course of investigating #1687 I stumbled across the problems on the following 4 L-function pages:

http://www.lmfdb.org/L/NumberField/16.2.14383898168677573334952601242898232687.4/
http://www.lmfdb.org/L/NumberField/16.4.1055734705924451249419325715250393166653097913001.1/
http://www.lmfdb.org/L/NumberField/16.4.35504497706629289335330118915361516626281.21/
http://www.lmfdb.org/L/NumberField/16.0.1599229661124692101046354619853228081.5/

In each case the traceback is the same:

Traceback (most recent call last):
  File "/home/drew/sage-6.5-x86_64-Linux/local/lib/python2.7/site-packages/Flask-0.10.1-py2.7.egg/flask/app.py", line 1817, in wsgi_app
    response = self.full_dispatch_request()
  File "/home/drew/sage-6.5-x86_64-Linux/local/lib/python2.7/site-packages/Flask-0.10.1-py2.7.egg/flask/app.py", line 1477, in full_dispatch_request
    rv = self.handle_user_exception(e)
  File "/home/drew/sage-6.5-x86_64-Linux/local/lib/python2.7/site-packages/Flask-0.10.1-py2.7.egg/flask/app.py", line 1381, in handle_user_exception
    reraise(exc_type, exc_value, tb)
  File "/home/drew/sage-6.5-x86_64-Linux/local/lib/python2.7/site-packages/Flask-0.10.1-py2.7.egg/flask/app.py", line 1475, in full_dispatch_request
    rv = self.dispatch_request()
  File "/home/drew/sage-6.5-x86_64-Linux/local/lib/python2.7/site-packages/Flask-0.10.1-py2.7.egg/flask/app.py", line 1461, in dispatch_request
    return self.view_functions[rule.endpoint](**req.view_args)
  File "/home/drew/Dropbox/github/lmfdb/lmfdb/lfunctions/main.py", line 344, in l_function_nf_page
    return render_single_Lfunction(DedekindZeta, args, request)
  File "/home/drew/Dropbox/github/lmfdb/lmfdb/lfunctions/main.py", line 397, in render_single_Lfunction
    L = Lclass(**args)
  File "/home/drew/Dropbox/github/lmfdb/lmfdb/lfunctions/Lfunction.py", line 1094, in __init__
[...]

Issue #1691, opened 2016-07-09
Server error when computing L-function zeros of number fields


AndrewVSutherland:
The page www.lmfdb.org/L/Zeros/NumberField/8.0.2178981.1/ causes a server error (as does the page www.lmfdb.org/L/NumberField/8.0.2178981.1/, but in this case the error is not reported to the user, it just fails to compute the Z-plot and logs the exception in flasklog).  Similar errors happen on many other pages, including:

www.lmfdb.org/L/Zeros/NumberField/8.2.32941720000.1/
www.lmfdb.org/L/Zeros/NumberField/10.0.1185762483.1/
www.lmfdb.org/L/Zeros/NumberField/10.0.52089208083.1/
www.lmfdb.org/L/Zeros/NumberField/8.0.3211264.1/

This is possibly related to #1687, in that both involve std::bad_alloc exceptions inside sage.libs.lcalc.lcalc_Lfunction, but in this case the error is handled more gracefully (it logs the error in flasklog and does not crash the LMFDB server).

AndrewVSutherland:
This problem is more pervasive than #1687, affecting about 1/3 of all number fields (37 out of 100 randomly chosen fields), including:

http://www.lmfdb.org/L/Zeros/NumberField/2.2.2105897.1/
http://www.lmfdb.org/L/Zeros/NumberField/7.3.48090396.1/
http://www.lmfdb.org/L/Zeros/NumberField/7.1.132713424.1/
http://www.lmfdb.org/L/Zeros/NumberField/7.1.136749644.1/
http://www.lmfdb.org/L/Zeros/NumberField/5.5.30500193199204.1/
http://www.lmfdb.org/L/Zeros/NumberField/9.1.35044743937.1/
http://www.lmfdb.org/L/Zeros/NumberField/2.2.277288.1/
http://www.lmfdb.org/L/Zeros/NumberField/9.3.821824919051661504.33/
http://www.lmfdb.org/L/Zeros/NumberField/5.1.1420419586896.1/
http://www.lmfdb.org/L/Zeros/NumberField/5.5.3004489415647729.1/
http://www.lmfdb.org/L/Zeros/NumberField/7.3.280169328600000.1/
http://www.lmfdb.org/L/Zeros/NumberField/7.1.145983303.1/
http://www.lmfdb.org/L/Zeros/NumberField/9.1.29054491825.1/
http://www.lmfdb.org/L/Zeros/NumberField/7.1.97523156.1/
[...]

Issue #1687, opened 2016-07-09
Many number field L-function pages crash the LMFDB


AndrewVSutherland:
The page http://www.lmfdb.org/L/NumberField/14.14.28152039412241052225421312.1/ causes an unhandled SIGABRT that crashes the web server (which then restarts).  Because it is an unhandled error, nothing appears in flasklog or accesslog.  A (long) traceback taken on Atkin appears below.

Repeatedly accessing this page can easily make the webserver unavailable (because it has to keep restarting and the restarts take time), and it seems likely that a robot accessing this page repeatedly may be what caused last night's outage (because of the crash the URL requester gets an immediate error message allowing them to try again, causing another crash, etc...)

>  
terminate called after throwing an instance of 'std::bad_alloc'
  what():  std::bad_alloc
------------------------------------------------------------------------
/usr/local/sage/sage-1/local/lib/python2.7/site-packages/cysignals/signals.so(+0x4965)[0x7fa351664965]
/usr/local/sage/sage-1/local/lib/python2.7/site-packages/cysignals/signals.so(+0x49b5)[0x7fa3516649b5]
/usr/local/sage/sage-1/local/lib/python2.7/site-packages/cysignals/signals.so(+0x6c81)[0x7fa351666c81]
/lib/x86_64-linux-gnu/libpthread.so.0(+0x10330)[0x7fa357720330]
/lib/x86_64-linux-gnu/libc.so.6(gsignal+0x37)[0x7fa35737ec37]
/lib/x86_64-linux-gnu/libc.so.6(abort+0x148)[0x7fa357382028]
/usr/lib/x86_64-linux-gnu/libstdc++.so.6(_ZN9__gnu_cxx27__verbose_terminate_handlerEv+0x155)[0x7fa347b28535]
/usr/lib/x86_64-linux-gnu/libstdc++.so.6(+0x5e6d6)[0x7fa347b266d6]
/usr/lib/x86_64-linux-gnu/libstdc++.so.6(+0x5e703)[0x7fa347b26703]
/usr/lib/x86_64-linux-gnu/libstdc++.so.6(+0x5e922)[0x7fa347b26922]
/usr/lib/x86_64-linux-gnu/libstdc++.so.6(_Znwm+0x7d)[0x7fa347b26e0d]
/usr/lib/x86_64-linux-gnu/libstdc++.so.6(_Znam+0x9)[0x7fa347b26ea9]
[...]

PR #1693, opened 2016-07-09
Fix for #1692

AndrewVSutherland:
This PR addresses #1692 by modifying the code for creating Dedekind zeta functions to handle situations in which the class number and/or regulator are not known.  In this case it no longer attempts to create a sageLfunction for computing values/zeros and instead sets it to None.  The /Zeros page simply returns "not available" in this case, and the /Plot page will return a 404 (getLfunctionPlot already handles situations where the plot cannot be computed and did not need to be modified).


AndrewVSutherland:
This PR is very small and safe and filtering out these fairly benign problems will make it easier to work on the more difficult Dedekind zeta function issues caused by lcalc crashes (#1687 and #1691), so I am going to merge this now.

Issue #12, opened 2013-01-29
broken L/NumberField/14.14.28152039412241052225421312.1/


haraldschilly:
the page http://www.lmfdb.org/L/NumberField/14.14.28152039412241052225421312.1/ *maybe* worked before and is broken now

haraldschilly:
see issue #430

jvoight:
This number field doesn't even exist any more, so that might be why it doesn't have an L-function!  http://beta.lmfdb.org/NumberField/14.14.28152039412241052225421312.1/

jwj61:
Interesting.  I don't recall deleting any fields, but since it is not there, the issue is closed.

jwj61:
On a closer look, this field does exist.  JV didn't see it because of the trailing slash in the url.  This should be readily fixed once the L-function database extends far enough since it is an abelian field, so it will just need to reference some Dirichlet l-functions.

jvoight:
My mistake!  I just copied the link from the top (deleting the L).  (It looks like the L-function page redirects; should we do the same for number fields?)

jwj61:
Probably, but I think this is a general issue.  I tried the analogue with a specific elliptic curve page, and adding a slash put me on the search results page.

JohnCremona:
Why is this critical?

davidfarmer:
I marked it as critical because it is a link that shows
a bad error.

My most recent pull request hides the links to Dedekind
zeta functions that are too big to computer on the fly, so
I removed the critical label.

The proper fix will come when we have those L-functions
in a database.

jwj61:
I am not sure this deserves a production problem label since there is no place in the lmfdb which links to this page.  Some day, that L-function may come back (which involves the addition of more L-function data, and building the mechanism for products of L-functions).  This is more like a feature request, or really a byproduct of two or three planned features.

AndrewVSutherland:
Agreed.

AndrewVSutherland:
This issue is fixed by PR #1701, which by preventing the computation of special values for number field L-functions of degree > 4 allows this page to load correctly.

AndrewVSutherland:
Closed via #1701.

Issue #1700, opened 2016-07-10
Store Dedekind zeta L-function data in the database


AndrewVSutherland:
As discussed in #1687 and #1691, due to myriad problems with computing special values, zeros, and Z-plots of L-functions on the fly using lcalc, we would like to have the L-function data for Dedekind zeta functions stored in the database (as we now do for Dirichlet L-functions).

davidfarmer:
Duplicate of #420

I also want us to (re)visit the issue of storing the Artin L-functions
and using the factorization of the Dedekind zeta function.

AndrewVSutherland:
@davidfarmer I think you meant #430 not #420.  I note that #430 is tagged with a 1.2 milestone rather than 2.0, which I like.   I agree that it we should store Artin L-functions in the database.  I'm not 100% convinced about the idea of using the factorization, I would want to be sure it can be done quickly and safely (preferably without relying on Sage).

Given that there is now more conversation about this here rather than at #430, does it make sense to close #430 as duplicate or do you want to close this one?

AndrewVSutherland:
@davidfarmer BTW, I am now 100% in agreement with your [comment](https://github.com/LMFDB/lmfdb/issues/500#issuecomment-231255861) about pre-computing and storing tensor product L-functions on #500; we shouldn't be trying to compute L-functions on the fly unless it is truly trivial.  Aside from the resources it uses, there are just too many ways for things to go wrong.

AndrewVSutherland:
This has been done; there are now more than 7 million precomputed L-function of Dirichlet characters, including all of modulus up to 3600.  L-function links only appear on Dirichlet character pages when the L-function has been precomputed (none are computed on the fly).

jwj61:
What about other Dedekind zeta functions?  This relies on getting Artin L-functions in the database, and then having the L-function code take products when the L-function is imprimitive.

AndrewVSutherland:
Apologies @jwj61 I was too hasty!

PR #1701, opened 2016-07-10
Fix for #1687 and #1691

AndrewVSutherland:
This is a one-line change to address issues #1687 and #1691 by disabling the use of sageLfunction for Dedekind Zeta functions of number fields of degree > 4.  In particular, all the URLs listed [here](https://github.com/LMFDB/lmfdb/issues/1687#issuecomment-231550693), which crash the LMFD server on www.lmfdb.org load without problems with this PR (special values are simply listed as "not computed" and the zeros and Z-plot are shown as "not available").  Compare:

http://127.0.0.1:37777/L/NumberField/5.5.2337227518904161.1/
http://127.0.0.1:37777/L/NumberField/10.0.162778775259375.1/

and

http://www.lmfdb.org/L/NumberField/5.5.2337227518904161.1/
http://www.lmfdb.org/L/NumberField/10.0.162778775259375.1/

for example.  The L/Zeros/... listed [here](https://github.com/LMFDB/lmfdb/issues/1691#issuecomment-231556070) now simply return "not available" rather than a server error (except for the two cases of degree 2 which are unaffected by this change, they result in timeouts).

Merging this change will allow #1687 and #1691 to be closed (the first is tagged as critical because it is related to the www.lmfdb.org outage caused (we believe) by a robot hitting a long list of URLs that crash sage).  In the longer term, we can again make special value, zeros, and Z-plots available for Dedekind Zeta functions once they are stored in the database (this is issue #1700).

### DIFF
--- a/lmfdb/lfunctions/Lfunction.py
+++ b/lmfdb/lfunctions/Lfunction.py
@@ -833,8 +833,7 @@ class Lfunction_Dirichlet(Lfunction):
             if self.selfdual:
                 self.coefficient_type = 2
                 for n in range(0, self.numcoeff - 1):
-                    self.dirichlet_coefficients[n] = int(
-                        round(real(self.dirichlet_coefficients[n])))
+                    self.dirichlet_coefficients[n] = int(round(real(self.dirichlet_coefficients[n])))
             else:
                 self.coefficient_type = 3
 
@@ -1151,7 +1150,8 @@ class DedekindZeta(Lfunction):   # added by DK
         self.citation = ''
         
         # If we know the residue create a Sage L-function we can call to compute values
-        if self.res:
+        # But only when the degree is at most 4, due to bugs in the Sage lcalc library (see issues #1687 and #1691)
+        if self.res and self.degree <= 4:
             generateSageLfunction(self)
         else:
             self.sageLfunction = None


AndrewVSutherland:
I should note that there are still a few Dedekind zeta function pages of degree <= 4 that cause problems, e.g. http://www.lmfdb.org/L/Zeros/NumberField/4.2.238186250894592000.1/, but these are rare (only one showed up in a random sample of 100 quartic fields).  Given that most of the degree 4 pages work and do show special values and (eventually) zeros and Z-plots (e.g. http://www.lmfdb.org/L/NumberField/4.0.888276.1/), I'm inclined to leave them as is.

AndrewVSutherland:
This also fixes https://github.com/LMFDB/lmfdb/issues/12, which has the distinction of being the oldest open issue for the LMFDB project.  The broken page

http://www.lmfdb.org/L/NumberField/14.14.28152039412241052225421312.1/

now works at

http://127.0.0.1:37777/L/NumberField/14.14.28152039412241052225421312.1/

AndrewVSutherland:
I'm going to go ahead and merge this (easy enough to undo if anyone disagrees with this change).  There are more changes that need to be made to the L-functions code that I don't want to include in this PR but would like to build on top of it (including tests that won't work without this PR).

PR #1702, opened 2016-07-10
remove WebDirichletCharacter from Lfunction.py

AndrewVSutherland:
This PR removes the use of WebDirichletCharacter from Lfunction.py; it is unnecessary now that L-function data for Dirichlet L-functions is stored in the database.  This fixes a bug that could cause the LMFDB application to crash on URLs like www.lmfdb.org/L/Character/Dirichlet/1000000000/3 due to memory allocation failures (the old code was creating the WebDirichletCharacter even before checking the database for the L-function data).  It also eliminates some computations that were being overwritten by the L-function data from the database.

It additionally fixes a small typo - the functional equations for Dirichlet characters that are odd (so not self-dual) now have \overline{\chi} on the RHS as they should.


AndrewVSutherland:
Pages to compare:

www.lmfdb.org/L/Character/Dirichlet/3/2/
http://127.0.0.1:37777/L/Character/Dirichlet/3/2/  (should be identical)

www.lmfdb.org/L/Character/Dirichlet/5/2/
http://127.0.0.1:37777/L/Character/Dirichlet/5/2/  (should differ only in F.E. display bar over chi)

www.lmfdb.org/L/Character/Dirichlet/1000000000/3/ (timeout or server error, do not repeat)
http://127.0.0.1:37777/L/Character/Dirichlet/1000000000/3/  (message saying not found)

pascalmolin:
2016-07-10 23:28 GMT+02:00 Andrew Sutherland <notifications@github.com>:

> It additionally fixes a small typo - the functional equations for
> Dirichlet characters that are odd (so not self-dual) now have
> \overline{\chi} on the RHS as they should.
>
No, the functional equation always relates chi to chi^{-1}=overline(chi),
so the criterion to print this overline is that the Dirichlet character
should have order > 2
(equivalently the conrey index has order > 2).

AndrewVSutherland:
@pascalmolin, sorry, I misspoke, I should have said "not self-dual", which is what the code is checking (not the parity).  I believe self-dual is set to false in the L-function data if and only if the order is > 2.

pascalmolin:
@AndrewVSutherland <https://github.com/AndrewVSutherland> I just took a
quick look to check this, and I do not see where self.selfdual is defined
(it used to be on line 831 with the correct definition).
But since the display is correct it must come from somewhere (db?)

otherwise, self.selfdual = (self.characternumber ** 2 %
self.charactermodulus == 1)

2016-07-11 0:05 GMT+02:00 Andrew Sutherland <notifications@github.com>:

pascalmolin:
@AndrewVSutherland <https://github.com/AndrewVSutherland> sorry, now I've
seen the makeLfromdb !

2016-07-11 0:43 GMT+02:00 Pascal Molin <molin@math.univ-paris-diderot.fr>:

AndrewVSutherland:
@pascalmolin Right selfdual (and just about everything) now gets set in makeLfromdata, which uses lfunc_data read from the mongo db.  Before this PR the code in DedekindZeta was trying to set overline{chi} correctly (see lines 843-846 in https://github.com/LMFDB/lmfdb/blob/master/lmfdb/lfunctions/Lfunction.py), but then incorrectly overwriting it in line 852 (and not looking at selfdual at all when setting texnamecompleted1ms_arithmetic).  This is what I fixed, so it now does what it was doing in lines 843-846 to both texnamecompled1ms and texnamecompled1ms_arithmetic.

JohnCremona:
When I go to http://127.0.0.1:37777/L/Character/Dirichlet/1000000000/3/ I get a server error:
KeyError: 'No L-function data for the Dirichlet character $\\chi_{1000000000}(3,\\cdot)$ found in the database.'

BTW there was already a test_lfunctions.py but not in the lfunctions directory, the one above.  So there might be some duplication.

AndrewVSutherland:
@JohnCremona That's probably because you are running in debug mode, try it without --debug

AndrewVSutherland:
@JohnCremona I can look at combining the test scripts, but I'd rather wait and do that in a separate PR.

AndrewVSutherland:
@JohnCremona I went ahead and consolidated the test scripts.

JohnCremona:
Sorry for delay.  Will get back to this shortly.

JohnCremona:
I have not checked all the new character code in detail but trust that the two of you have worked on it and know what you are doing.  Test run fine.  I am puzzled by the different error behaviour with / without running in beta mode -- that means that the error-catching behaviour will be different on prod and beta, doesn't it?
I'll be happier when I know the reason for that, and merge this.

AndrewVSutherland:
@JohnCremona See https://github.com/LMFDB/lmfdb/blob/master/lmfdb/lfunctions/main.py#L381, the function render_lfunction_exception is designed to catch errors only when current_app.debug is not set -- this is actually a useful feature.  It does not impact behavior on beta.lmfdb.org, compare for example:

http://beta.lmfdb.org/L/Character/Dirichlet/10/2/
http://127.0.0.1:37777/L/Character/Dirichlet/10/2/ (running with --debug)

Running locally in a terminal with --debug set is not the same thing as beta (e.g. beta does not display tracebacks in the browser).

JohnCremona:
Thanks -- great explanation, and I learnt something useful!

Issue #1184, opened 2016-05-05
Sato-Tate groups for Dirichlet L-functions


davidfarmer:
The Sato-Tate groups of Dirichlet L-functions are listed in the form "mu(N)",
which is supposed to represent the subgroup the unit circle consisting of the Nth roots of 1.

These should be replaced by an official name, and linked to the newly created Sato-Tate material.

JohnCremona:
Agreed.  Drew?  But is it only the L-function which has a ST group, or does the character itself?
The same applied for elliptic curves/Q: the ST group is shown on the L-function page but not as a link to the ST group page.
For ec/nf we now show ST groups but not yet for their L-functions.

AndrewVSutherland:
ST group links now appear on Dirichlet L-function pages and as friends of Dirichlet characters (PR #1658 and PR #1703).

PR #1703, opened 2016-07-11
Restrict number of components for ST groups to 10^20

AndrewVSutherland:
Restricts the maximum size of ST groups mu(n) to 10^20 to avoid attempting to factor a large number.  Also adds ST groups as friends on Dirichlet character pages (addressing #1184).

### DIFF
--- a/lmfdb/WebCharacter.py
+++ b/lmfdb/WebCharacter.py
@@ -846,6 +846,8 @@ class WebChar(WebCharObject):
             url = url_character(type=self.type, umber_field=self.nflabel, modulus=self.modlabel, number=self.numlabel)
             if lmfdb.lfunctions.LfunctionDatabase.getInstanceLdata(url[1:]):
                 f.append( ('L-function', '/L'+ url) )
+        if self.type == 'Dirichlet':
+            f.append( ('Sato-Tate group', '/SatoTateGroup/0.1.%d'%self.order) )
         if len(self.vflabel)>0:
             f.append( ("Value Field", '/NumberField/' + self.vflabel) )
         return f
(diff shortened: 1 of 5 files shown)


AndrewVSutherland:
added some more tests and fixed a few more large modulus issues.  all tests now pass (except one known classical mf issue)

AndrewVSutherland:
In addition to addressing Sato-Tate groups mu(n) with n up to 10^20, this PR ensures that Dirichlet characters of modulus up to 10^20 now work correctly (as claimed on www.lmfdb.org/Character/Dirichlet/).  Previously characters with modulus > 2^64 did not work, compare

http://www.lmfdb.org/Character/Dirichlet/99999999999999999999/
http://www.lmfdb.org/Character/Dirichlet/99999999999999999999/2

with

http://127.0.0.1:37777/Character/Dirichlet/99999999999999999999/
http://127.0.0.1:37777/Character/Dirichlet/99999999999999999999/2

also handles characters with order > 2^64, such as

http://127.0.0.1:37779/Character/Dirichlet/99999999999999999931/2

JohnCremona:
Quick question -- is there a reason to write Xrange instead of using srange or xsrange?  The ones with s in them are Sage functions and yield Sage integers.

AndrewVSutherland:
No, I will switch to xsrange and remove Xrange.

JohnCremona:
Good!  Just running tests, the pages and code look fine.

JohnCremona:
Running all tests revealed a couple of things -- not relevant for this patch: some extraneous print output in some ecnf tests, and a failed test /home/jec/lmfdb/lmfdb/modular_forms/elliptic_modular_forms/test_emf.py
No reason not to merge this one though.

AndrewVSutherland:
@JohnCremona the failure in test_emf.py is due to bad data in modularforms2 (see #1248).

PR #1708, opened 2016-07-11
Template collision check

AndrewVSutherland:
This PR adds test_template_namespace.py that searches the LMFDB directory tree for template files with the same name (flask puts these all into the same namespace, regardless of which directory they are in, and this can cause unexpected behavior when collisions occur).

In the process of testing this I found two duplicate template files.

* Two essentially identical copies of lf-list.html appear in lmfdb/zeros/templates and lmfdb/lfunction_db/templates; this template is only used in lfunction_db, so I deleted the copy in lmfdb/zeros/templates.
* Two slightly different copies of maas_form_picard.html appear in the modular_forms/... directory tree.  This template is referenced in only one place "lmfdb/modular_forms/maass_forms/picard/views/mwfp_main.py" and the reference is in an unreachable line of code in "render_picard_maass_forms".  I kept the version in the corresponding templates subdirectory (which also appeared to be newer) and deleted the other one.

To test, copy any template .html file into another templates directory and from the the top lmfdb directory run ./test.sh lmfdb/test_template_namespace.py -- it should find the duplicate you created (don't forget to delete it when you are done).


JohnCremona:
I think I would have chosen a different name for the new test class than LfunctionTest ;) but it does work as advertised!

AndrewVSutherland:
Oops, test class name is now something more sensible.

JohnCremona:
Thanks!

PR #1713, opened 2016-07-12
pyflakes cleanup

edgarcosta:
Hello,

I ran pyflakes on all the folders from r to z in lmfdb/ and addressed all of its warnings for relevant code.

Best,
Edgar

### DIFF
--- a/lmfdb/rep_galois_modl/__init__.py
+++ b/lmfdb/rep_galois_modl/__init__.py
@@ -12,5 +12,6 @@ def body_class():
     return {'body_class': 'rep_galois_modl'}
 
 import main
+assert main #silence pyflakes
 
 app.register_blueprint(rep_galois_modl_page, url_prefix="/Representation/Galois/ModL")
--- a/lmfdb/rep_galois_modl/main.py
+++ b/lmfdb/rep_galois_modl/main.py
@@ -14,7 +14,9 @@ from sage.all import ZZ, conway_polynomial
 from lmfdb.rep_galois_modl import rep_galois_modl_page #, rep_galois_modl_logger
 from lmfdb.rep_galois_modl.rep_galois_modl_stats import get_stats
 from lmfdb.search_parsing import parse_ints, parse_list, parse_count, parse_start
-from lmfdb.lattices.main import vect_to_sym, vect_to_matrix
+
+#should these functions be defined in lattices or somewhere else?
+from lmfdb.lattices import vect_to_sym, vect_to_matrix
 
 from markupsafe import Markup
 
--- a/lmfdb/rep_galois_modl/main.py
+++ b/lmfdb/rep_galois_modl/main.py
@@ -3,25 +3,22 @@ import pymongo
 ASC = pymongo.ASCENDING
 LIST_RE = re.compile(r'^(\d+|(\d+-\d+))(,(\d+|(\d+-\d+)))*$')
 
-import flask
-from flask import render_template, render_template_string, request, abort, Blueprint, url_for, make_response, Flask, session, g, redirect, make_response, flash,  send_file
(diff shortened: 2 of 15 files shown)


JohnCremona:
What are the assertions after some import statements for?

edgarcosta:
It is a way to avoid the pyflakes warning:
`blah imported but unused`

the asserts should only be used in the __init__.py files.

JohnCremona:
OK thanks.  I ought to start using myflakes myself -- what do you have to do?

edgarcosta:
I just do:
```pyflakes filename```
for example"
```pyflakes web_ec.py```

but sometimes, specially with the `__init__.py` pyflakes has no way to know
that the file imported but unused, will be used later on, when someone
imports the module itself.

**Comment by JohnCremona** [comment]
When I ran the tests on this I got an import error:
/home/jec/data/zeros/
Failure: ImportError (No module named lattices.main) ... ERROR

**Comment by edgarcosta** [comment]
@sanni85, please see `lmfdb/rep_galois_modl/main.py`, where I added:
```
#should these functions be defined in lattices or somewhere else?
from lmfdb.lattices import vect_to_sym, vect_to_matrix
```

**Comment by AndrewVSutherland** [comment]
In my experience the only error you should see in pyflakes is a single unused import in __init__.py that does need to be there.  But you do need to use caution when doing pyflakes cleanup, especially when it warns about unused variables -- you can get rid of the warning by deleting the assignment of the unused variable, but it is possible that the variable should be getting used, or that the variable name is misspelled and the assignment should be to a slightly different variable name.  It is worth spending some time looking at the code to try and figure these out, and in cases where you really cannot be sure but suspect that some version of the offending line of code does need to be there, it is probably better to leave it in but put a # FIXME THIS VARIABLE IS NEVER USED comment right above it in the hopes that someone who knows the code better can fix it later.

@JohnCremona install pyflakes using sudo apt-get and then just run pyflakes filename (you can use wildcards).

**Comment by edgarcosta** [comment]
@JohnCremona, still running the test.sh, I will let you know once it is done

**Comment by edgarcosta** [comment]
@AndrewVSutherland I will add those FIXMEs

**Comment by AndrewVSutherland** [comment]
@edgarcosta, @JohnCremona I would be happy to review this PR as well - it might be worth another pair of eyes.

**Comment by edgarcosta** [comment]
@AndrewVSutherland, what about adding the 
# FIXME THIS VARIABLE IS NEVER USED
and commenting out the unused variable?

Otherwise, we will never get rid of the warnings, and unless the variable is global, if the code was broken, we won't fix it by leaving it there.

**Comment by AndrewVSutherland** [comment]
@edgarcosta Sure.

**Comment by JohnCremona** [comment]
The test pass for me (with your latest commit), apart from the irrelevant one in test_emf.py so I am happy to germe this now, or was there more testing needed?

**Comment by edgarcosta** [comment]
I don't think so. I'm quite sure I didn't break anything that wasn't already broken. 
Mine test.sh is still running...

**Comment by AndrewVSutherland** [comment]
Tests all pass for me (except test_emf.py) and I took a look through the code and it looks fine (I agree that it certainly does not break anything that is not already broken).  I will go ahead and merge.

@edgarcosta You might want to switch to using atkin for tests, it only takes 3 minutes to run test.sh there.

**Comment by JohnCremona** [comment]
Running test.sh takes <3 minutes for me.  What's causing the delay for you?

**Comment by edgarcosta** [comment]
Done. I was running it on my laptop, connected to warwick, not ideal, given that I'm in the other side of the world...
```
Ran 260 tests in 4603.798s

FAILED (SKIP=7, errors=1, failures=1)
```

JohnCremona:
!!!

PR #1720, opened 2016-07-12
tidied up elliptic_curves with pyflakes

JohnCremona:
As well as cleaning everything up, I removed a long obsolete import script.


AndrewVSutherland:
It makes me very happy to see a factorization that was not being used (jfac) get deleted (of course the j-invariant is still getting factored, but least it now only gets factored once :))

JohnCremona:
@AndrewVSutherland That was not a factorization really, just the construction of a Sage Factorization object out of a list or (prime, exponent) pairs, simply for easy of latexification.

PR #1721, opened 2016-07-12
updated user-manager script to use authentication

JohnCremona:
This updates the script (in the top directory) user-manager.py, whose sole purpose is to change a user's login password, so that it uses the authentication mechanism to have write acceess to the users database.  It can only be used when the appropriate authentication passwords are available.

### DIFF
--- a/user-manager.py
+++ b/user-manager.py
@@ -17,7 +17,7 @@ def changepassword(username, newpassword):
     from lmfdb.users import pwdmanager
     pwdmanager.change_password(username, newpassword)
 
-if len(sys.argv) < 2:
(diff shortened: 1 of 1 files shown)


edgarcosta:
I might be missing something, but I think you want
```
if len(sys.argv) != 4:
```

JohnCremona:
Fair enough - I changed it from <1 to <2 so that if you put nothing on the command line you got to see the usage text, but of course the length should exactly 4.  I'll change that.

AndrewVSutherland:
Looks good (I just used it to change the password of the user I e-mailed you about).

Issue #1570, opened 2016-06-20
robots.txt broken


haraldschilly:
Hi, I just saw a notification, that the robots.txt file is broken. There is a mechanism, that disables indexing the page, if it is hosted somewhere else -- as a result, only the main page should be indexed and not the other ones. The transfer of the lmfdb page must have broken this heuristic, such that http://lmfdb.org/robots.txt is wrong.

Heuristic is here:

https://github.com/LMFDB/lmfdb/blob/master/lmfdb/website.py#L89

haraldschilly:
ok, no, I think it's all fine. This circles back to the question, if there should be www. or not.

i.e. http://www.lmfdb.org/robots.txt is correct!

hence I'm closing it, but since many got this as an email, it's maybe some food for thought about this. (i.e. should be add an automatic redirection to include www. in the url)

edgarcosta:
> should be add an automatic redirection to include www. in the url

where should we do that? code? or gce?

haraldschilly:
if at all, that should be in the code. we also don't have to forget to declare the preferred way in the webmaster tools...

but for now, I wouldn't do anything.

PR #1729, opened 2016-07-13
redirects lmfdb.org to www.lmfdb.org

edgarcosta:
Hello,

I added a hook to redirect requests from "lmfdb.org" to "www.lmfdb.org".

There is really way for me to test this until it gets to prod, thus please double check.

I'm using the code 301, which means, permanent redirect, see: https://en.wikipedia.org/wiki/HTTP_301 .

Best,
Edgar

### DIFF
--- a/lmfdb/website.py
+++ b/lmfdb/website.py
@@ -56,6 +56,16 @@ import sys
 #import base
 from base import app, render_template, request, DEFAULT_DB_PORT, set_logfocus, _init
 
+@app.before_request
+def redirect_nonwww():
+    """Redirect lmfdb.org requests to www.lmfdb.org"""
+    from urlparse import urlparse, urlunparse
+    urlparts = urlparse(request.url)
+    if urlparts.netloc == 'lmfdb.org':
+        urlparts._replace(netloc='www.lmfdb.org')
+        from flask import redirect
+        return redirect(urlunparse(urlparts), code=301)
+
 @app.errorhandler(404)
 def not_found_404(error):
     return render_template("404.html"), 404


JohnCremona:
I don't understand this at all (not knowing much about web servers).  Surely by the time any of our code gets run the request has already reached us via the URL?  So how can this make a difference?

Right now if I put "lmfdb.org" in my browser's URL bar it goes to http://lmfdb.org which does appear to be our prod server.

edgarcosta:
if you type `http://lmfdb.org/whatever` in the URL bar, before anything else, the code will tell your browser that the content  `http://lmfdb.org/whatever` has been permanently moved to  `http://www.lmfdb.org/whatever`.

For example, if you type http://ams.org you get redirected to www.ams.org

JohnCremona:
Sorry to be so dim about this -- but if I start up a browser and type something into its URL bar, from where does it get its instructions aboutt his redirect *before* going to the website?  Is that not handled by some DNS lookup?

PS Very interesting PR number, this one!

edgarcosta:
Yes, we could do all this on the DNS.
In this case, is the website itself that is telling you that we moved the webpage to somewhere else. So yes, you type `lmfdb.org/whatever`, asks the DNS where is lmfdb.org, talks to our servers that it wants `whatever`, and our servers immediately reply and say, you should ask `www.lmfdb.org` for that, your browsers does that, and only afterwards our servers will handle the `whatever` request.

JohnCremona:
Great explanation, thanks.

For the review itself, I think you are saying that we have to wait until this gets pushed to prod before we can see if it works as intended.  I checked out https://docs.python.org/2/library/urlparse.html and there is one thing I noticed: the 'netloc' field might end with ";port", which your code would not catch.  Perhaps that is not an issue here since we only want to redirect the literal URL "lmfdb.org" and not something like "lmfdb.org:12345".

edgarcosta:
the goal of all this is to address #1570 and have consistent urls when people copy and paste them

edgarcosta:
Yes, I don't care about those situations.

JohnCremona:
OK -- @haraldschilly does the proposed change look good to you?

AndrewVSutherland:
Given that this looks completely safe and that the only way to test it is on prod, I suggest we merge this.

JohnCremona:
Agreed.

AndrewVSutherland:
It works :) But now it appears we need to change the address stackdriver is using for health checks.

edgarcosta:
Does it? 
For me it doesn't work. I'm investigating.

edgarcosta:
It has something to do with these two functions:

```
@app.before_request
def get_menu_cookie():
    g.show_menu = request.cookies.get('showmenu') != "False"

@app.after_request
def set_menu_cookie(response):
    response.set_cookie("showmenu", str(g.show_menu))
    return response
````

AndrewVSutherland:
If I type lmfdb.org into my url bar (in either Chrome or Firefox) I get redirected to www.lmfdb.org.

edgarcosta:
I have fixed it a couple of minutes ago.

Issue #1209, opened 2016-05-06
Losing connection to DB


edgarcosta:
Hello,

At the moment the client quits whenever it loses the connection to the DB.
In my opinion it should not.

Best,
Edgar

JohnCremona:
I take you do not just mean that this is an inconvenience for developers, but actually a problem for the live website?

jvoight:
Give us a minimal nonworking example or what you'd like to have changed and what the implications would be.  Also, an indication of how serious this is.

edgarcosta:
If the client loses the connection to the mongodb server you get an "Internal Server Error" from gunicorn.

Minimal example, http://lmfdb.org/acknowledgment

jvoight:
Yes, that is pretty unhelpful for it act in this way.  But when the client loses connection to the mongodb server, isn't that a bigger problem anyway?  I mean, there isn't really any point in displaying pages not requiring mongodb.  (I realize you're trying to give an example of the problem; I'm trying to determine what needs to be fixed here, and how critical it is.)  

How often is this happening in the cloud?  

Shouldn't the client try first to restore the mongodb connection, and when it fails try a different connection, and then when that fails display a more helpful error/kill itself?

edgarcosta:
Further, this needs to be fixed at some point, but not by May 10.

It is also related to #1165

edgarcosta:
Yes, it's way bigger problem to lose the connection to mongodb. However informing the user that something went wrong is much better than just displaying "Internal Server Error" in big letters.

Not critical at all. 

It hasn't yet happened in the cloud, except when the client by itself was closing the connection, see #1201 

Yes, if we loose the connection we will keep retrying forever and ever. Basically the client dies immediately, but we launch another one right away.

edgarcosta:
More on this, from Sarunas, I bolded the key part:

> Organizational matters aside (not less important, I just don't feel that
> I have been with the project long enough), I'd like to chime in on
> today's interruption. It was about MongoDB underperforming while under
> index creation task, and the Sage clients not handling a slow-responding
> DB backend gracefully. MongoDB replica members, m0 and m1, didn't crash,
> but apparently became “disk-bound” (judging by Munin monitoring data
> https://math.dartmouth.edu/munin/LMFDB/comparison-day.html). I
> understand you have been running a respectably-sized MongoDB for quite a
> while — so is this a known condition? If so, shouldn't index creation be
> done more cautiously, perhaps with some advance warning/consensus,
[...]

PR #1730, opened 2016-07-13
Fix for #1209

AndrewVSutherland:
This PR addresses #1209 by allowing the LMFDB application to stay up when mongo db goes down.  If/when this happens, the user will see the main page (with sidebar and title) with a message indicating that the LMFDB is unavailable and the option to visit an alternate site (e.g. lmfdb.warwick.ac.uk) or to contact us (the contact page does not require mongo db).

While mongo db is down, clicking on any sidebar links (or directly trying to visit any URL) will take the use to the same front page (with a 503 status returned to the browser).  If/when the mongo db becomes available again, everything will start working again, no restart is required.

To test this, its best to run warwick.sh in a terminal where you can easily stop and start it.  Try starting the lmfdb without the tunnel to the mongo db at warwick in place, then run warwick.sh and see that everything works, then stop warwick.sh in the middle of something and see what happens.

### DIFF
--- a/lmfdb/base.py
+++ b/lmfdb/base.py
@@ -10,9 +10,9 @@
 import os
 import logging
 from time import sleep
-from flask import Flask, g, url_for, abort, render_template, request
+from flask import Flask, g, url_for, abort
 import pymongo
-from pymongo import MongoClient, MongoReplicaSetClient, Connection
+from pymongo import MongoClient, MongoReplicaSetClient
 from pymongo.cursor import Cursor
 from pymongo.errors import AutoReconnect
 from os.path import dirname, join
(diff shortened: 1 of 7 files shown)


edgarcosta:
Looks good to me.

ps: this line https://github.com/LMFDB/lmfdb/blob/master/lmfdb/pages.py#L30 will never be executed, as the getDBconnection will abort first.

AndrewVSutherland:
@edgarcosta Good catch I'll change Line 30 and add another commit to this PR momentarily.

AndrewVSutherland:
@edgarcosta Actually I think it makes sense to live line 30 as is (BTW it is changed to raise a 503 not a 501 in this PR).  I can imagine a scenario in which the LMFDB is connected to a mongo db but the userdb isn't there for some reason, and there is no harm to leaving the code as is it in any case.

edgarcosta:
Yes, please leave it there.

AndrewVSutherland:
I added one small change to 503.html.

Given that Edgar has reviewed this and the changes only affect the behavior of the system in situations where the current code would either crash the LMFDB or throw an internal server error anyway, I am going to merge this so we can test it further and review #1732.

PR #1732, opened 2016-07-13
some improvements on handling mongoclient.config

edgarcosta:
The goal of this PR, is to be able to set timeout times for the connection to the MongoDB on the  mongoclient.config file.
See the default config file.
Further, now you can specify the filename of your config as an argv, if you are not running tests.

### DIFF
--- a/default_mongoclient.config
+++ b/default_mongoclient.config
@@ -4,8 +4,8 @@
 # To change the defaults, edit a copy of the file below, renamed mongoclient.config
 #
 # settings without values are parsed as None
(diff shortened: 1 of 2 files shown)


AndrewVSutherland:
I like these changes.  But I would suggest that we review and merge #1730 first.  In theory there should be no conflicts between #1730 and #1732 (they both touch website.py but in different places), but if there is a problem it will be simpler to tweak #1732 than #1730 (fewer files/changes).

edgarcosta:
I agree with that.

Issue #1733, opened 2016-07-13
Server error on many GL3 and GL4 Maass form lcalc files


AndrewVSutherland:
Several examples from flasklog

http://www.lmfdb.org/L/ModularForm/GL3/Q/Maass/1/1/16.40312_0.171121/-0.4216864/?download=lcalcfile
http://www.lmfdb.org/L/ModularForm/GL3/Q/Maass/4/1/9.632444_1.374060/0.15012282/?download=lcalcfile
http://www.lmfdb.org/L/ModularForm/GL3/Q/Maass/4/1/8.954662_2.936591/0.36025530/?download=lcalcfile
http://www.lmfdb.org/L/ModularForm/GL4/Q/Maass/1/1/16.89972_2.272587_-6.03583/0.55659019/?download=lcalcfile
http://www.lmfdb.org/L/ModularForm/GL3/Q/Maass/1/1/-16.4031_-0.17112/-0.4216864/?download=lcalcfile
http://www.lmfdb.org/L/ModularForm/GL3/Q/Maass/4/1/-9.63244_-1.37406/0.15012282/?download=lcalcfile
http://www.lmfdb.org/L/ModularForm/GL3/Q/Maass/4/1/-8.95466_-2.93659/0.36025530/?download=lcalcfile
http://www.lmfdb.org/L/ModularForm/GL4/Q/Maass/1/1/-16.8997_-2.27258_6.035835/0.55659019/?download=lcalcfile
http://www.lmfdb.org/L/ModularForm/GL3/Q/Maass/1/1/16.40312_0.171121/-0.4216864/?download=lcalcfile

Failures occur in parse_complex_number (in LfunctionLcalc.py)

JohnCremona:
I do not know why we have to write such a function in the first place, since Sage is good at construction numbers from a variety of different strings.

sage: x,y = CC('3+4*i'); (x,y)
(3.00000000000000, 4.00000000000000)

That parse function does not actually specify what forms of in put are expected and catered for.

AndrewVSutherland:
@JohnCremona Do you want to take a stab at removing it?  It not I can look at it.

JohnCremona:
In one of the above the string being parsed is u'(1.0)' and the code I pasted above handles that fine.

I will therefore make a patch for this myself.

JohnCremona:
Testing my patch now.

JohnCremona:
[...]

PR #1736, opened 2016-07-13
fix for complex number parsing Issue 1733

JohnCremona:
I changed the parsing function, and added tests to show that the problem pages at #1733 are fine now.


AndrewVSutherland:
Looks good, thanks for the quick fix!

PR #1739, opened 2016-07-13
Hypergm cleanup

AndrewVSutherland:


jwj61:
Why is it important that files which are essentially in development have their imports stripped?  It is hidden on the production site, and it seems like it will just create additional hassle for when we return to working on this.

AndrewVSutherland:
Sorry @jwj61, I've been trying to get pyflakes to run cleanly on the entire codebase in an effort to address #122 (it has also turned up quite a few subtle bugs).

I'm happy to revert this PR if you like.  It just would be very nice to by able to run pyflakes on the entire directory tree and have it not report any issues.  It gives a very quick and easy way to spot problems.

JohnCremona:
It does get to be a bit of an obesssion, especially when applied to files like import scripts (which I did).  The advantage of having a totally clean codebase is that then important things will not be hidden behind the ones we know are ok to ignore.
I'm a little afraid of what I'll find when I resurrect some of my older branches and try to rebase them, but it can't be that bad,

Issue #1662, opened 2016-07-07
Timeout attempting to load degree 16 number field page


AndrewVSutherland:
If you go to http://www.lmfdb.org/NumberField/?degree=16&discriminant=13307764731675384304522756096 you will see 4 number fields listed.  The last 3 links work fine, but the first times out, apparently while in pari_integeral_basis.  From the flasklog:

> Exception on /NumberField/16.0.13307764731675384304522756096.1 [GET]
Traceback (most recent call last):
  File "/home/sage/sage-7.1/local/lib/python2.7/site-packages/Flask-0.10.1-py2.7.egg/flask/app.py", line 1817, in wsgi_app
    response = self.full_dispatch_request()
  File "/home/sage/sage-7.1/local/lib/python2.7/site-packages/Flask-0.10.1-py2.7.egg/flask/app.py", line 1477, in full_dispatch_request
    rv = self.handle_user_exception(e)
  File "/home/sage/sage-7.1/local/lib/python2.7/site-packages/Flask-0.10.1-py2.7.egg/flask/app.py", line 1381, in handle_user_exception
    reraise(exc_type, exc_value, tb)
  File "/home/sage/sage-7.1/local/lib/python2.7/site-packages/Flask-0.10.1-py2.7.egg/flask/app.py", line 1475, in full_dispatch_request
    rv = self.dispatch_request()
  File "/home/sage/sage-7.1/local/lib/python2.7/site-packages/Flask-0.10.1-py2.7.egg/flask/app.py", line 1461, in dispatch_request
    return self.view_functions[rule.endpoint](**req.view_args)
  File "/home/lmfdb/lmfdb-git-prod/lmfdb/number_fields/number_field.py", line 411, in by_label
    return render_field_webpage({'label': nf_string_to_label(label)})
  File "/home/lmfdb/lmfdb-git-prod/lmfdb/number_fields/number_field.py", line 302, in render_field_webpage
    data['frob_data'], data['seeram'] = frobs(nf.K())
  File "/home/lmfdb/lmfdb-git-prod/lmfdb/number_fields/number_field.py", line 548, in frobs
    frob_at_p = residue_field_degrees_function(K)
  File "/home/lmfdb/lmfdb-git-prod/lmfdb/number_fields/number_field.py", line 532, in residue_field_degrees_function
    k1 = pari(K)
[...]

PR #1744, opened 2016-07-13
Avoid sage nf

jwj61:
This fixes issue #1662 .  The issue lists the field which used to timeout.

The problem was that several parts of displaying a number field created a sage number field.  If the polynomial discriminant is too large, this can take too long.  For number field pages, we avoid this by creating a pari number field instead (feeding it our precomputed integral basis lets it run quickly).

This also makes pyflakes run cleanly on number_field.py.  Hopefully it doesn't break other parts of the web site.  This pull request changes other files which were importing functions from number_field.py when they probably shouldn't have (the functions are defined elsewhere and used to be imported into number_field.py).

### DIFF
--- a/lmfdb/WebNumberField.py
+++ b/lmfdb/WebNumberField.py
@@ -343,10 +343,21 @@ class WebNumberField:
         t = self.galois_t()
         return modules2string(n, t, ugm)
 
+    # Sage version of K -- should be avoided since it can be slow
+    # in extreme cases
     def K(self):
         if not self.haskey('K'):
             self._data['K'] = NumberField(self.poly(), self.gen_name)
         return self._data['K']
+
+    # pari version of K
+    def gpK(self):
+        if not self.haskey('gpK'):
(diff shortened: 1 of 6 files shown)


AndrewVSutherland:
I'm getting the error

AttributeError: 'NumberField_absolute_with_category' object has no attribute 'gpK'

at line 533 of number_field.py when I run test.sh lmfdb/lfunctions/test_lfunctions.py

AndrewVSutherland:
But it definitely fixes #1662, very snappy!

AndrewVSutherland:
@jwj61 You can reproduce the problem in https://github.com/LMFDB/lmfdb/pull/1744#issuecomment-232481952 by going to

http://127.0.0.1:37777/L/ArtinRepresentation/2.23.3t2.1c1/

jwj61:
Good catch!  That is now fixed as well.

AndrewVSutherland:
Looks good, merging.

PR #1747, opened 2016-07-13
Fix part of issue 1661, integers being too long

jwj61:
Test with second field reported in that issue.  As suspected, the two problems are unrelated.


AndrewVSutherland:
Looks good

PR #1751, opened 2016-07-14
change a to alpha in a place which was forgotten

### LINKED ISSUE #1748: alpha / a mixup in q-expansions
As reported via the bug tracker:

"Not at all a serious problem, but a curious issue:

When looking at Fourier expansions of modular forms with coefficients in a number field, the number field is given as Q(alpha), and the initial q-expansion involves alpha, but for some reason when you click on "show more coefficients", the alpha's become a's."

For example

http://www.lmfdb.org/ModularForm/GL2/Q/holomorphic/23/2/1/a/

JohnCremona:
This fixes #1748: compare http://www.lmfdb.org/ModularForm/GL2/Q/holomorphic/23/2/1/a/ (after clicking the "show more coefficients" button.


AndrewVSutherland:
Checking now.

PR #1758, opened 2016-07-14
Higher Genus  tiny changes and connect to genus2/Q pages

jenpaulhus:
1. Fixed a typo for calling one KNOWL on passport pages like
http://127.0.0.1:37777/HigherGenus/C/Aut/2.5-1.0.5-5-5.2
The "refined passport" KNOWL had the wrong name in the file /lmfdb/higher_genus_w_automorphisms/template/hgcwa-show-passport.html

This fix broke a test too, so small change to one test as well.

2. On the index page  http://127.0.0.1:37777/HigherGenus/C/Aut/ the highest genus listed for the browse options is now a variable which is found by query to the database to find the highest genus available (before it was just a static number).

3.  Added connections to the genus 2/Q data.  For example, there is now a friend here:
http://127.0.0.1:37777/HigherGenus/C/Aut/2.12-4.0.2-2-2-3 
which takes you to the search in the genus2/Q data for automorphism group over \bar{Q} of D_6.

### DIFF
--- a/lmfdb/higher_genus_w_automorphisms/main.py
+++ b/lmfdb/higher_genus_w_automorphisms/main.py
@@ -157,7 +157,10 @@ def index():
     bread = get_bread()
     if request.args:
         return higher_genus_w_automorphisms_search(**request.args)
-    genus_list = range(2,6)
+
+    C = base.getDBConnection()
+    genus_max = C.curve_automorphisms.passports.find().sort('genus', pymongo.DESCENDING).limit(1)[0]['genus']  + 1
+    genus_list = range(2,genus_max)
     info = {'count': 20,'genus_list': genus_list}
     
 
@@ -322,8 +325,13 @@ def render_family(args):
             
         info.update({'passport': Lall})
 
-            
-        friends = [ ]
+
+        g2List = ['[2,1]','[4,2]','[8,3]','[10,2]','[12,4]','[24,8]','[48,29]']
+        if g  == 2 and data['group'] in g2List:
+            g2url = "/Genus2Curve/Q/?geom_aut_grp_id=" + data['group']
+            friends = [("Family Over $\Q$", g2url ) ]
+        else:    
+            friends = [ ]
         
 
         br_g, br_gp, br_sign = split_family_label(label)
--- a/lmfdb/higher_genus_w_automorphisms/main.py
+++ b/lmfdb/higher_genus_w_automorphisms/main.py
@@ -89,6 +89,11 @@ def signature_to_list(L):
     return L
 
 
+def sort_sign(L):
+    L1 = L[1:]
+    L1.sort()
+    return [L[0]] +L1
+
 def label_to_breadcrumbs(L):
     newsig = '['
     for i in range(0,len(L)):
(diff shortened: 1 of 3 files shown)


jenpaulhus:
Added one more change in last commit.  Signature values can now be entered in the search in any order (except for the first value which must be the quotient genus).

So in the past if you entered [0;4,4,2,2,2] as a search signature you would get no hits.  Now this will sort in non-decreasing order first before querying the database.

AndrewVSutherland:
Looking at this now.

AndrewVSutherland:
@jenpaulhus I like all these changes, especially linking to the genus 2 curves (connecting different parts of the LMFDB is always a good thing!).  It also made me realize that I should change the names of the dihedral groups to match the convention used in small_groups (which both your code and the Sato-Tate groups code use), so that the dihedral group of order 12 id denoted D_6.  I will do that shortly.

One suggestion: in the friend link "Family Over Q" I might put "Genus 2 curves over Q", to make it clearer where the link is taking you (and note that there is no reason to capitalize "over").

While reviewing this I noticed what I think might be a data issue.  If you got to the page

http://www.lmfdb.org/HigherGenus/C/Aut/?start=20&paging=0&genus=2&group=%5B6%2C1%5D&signature=&dim=&inc_hyper=include&count=

there are 2 passports listed, but they appear to be identical (they both take you to the same page http://www.lmfdb.org/HigherGenus/C/Aut/2.6-1.0.2-2-3-3.1).

jenpaulhus:
The issue mentioned above isn't a data issue but an issue of what I return from database queries.  I'm not sure how to fix it in a "nice" way.

Here's what's happening.  Each entry in this database represents one action, specifically one generating vector.  So this example you mention above is one where there are two distinct actions in the same passport (http://www.lmfdb.org/HigherGenus/C/Aut/2.6-1.0.2-2-3-3.1  has two generating vectors listed) and both of these entries in the database have the same entry in the "passport_label" field.  

So when I query the database for say genus 2 curves with group (6,1) and signature [0,2,2,3,3],  the "find" command returns 2 entries, but with exactly the same value in the 'passport_label' field, and this value is what I use to generate the url to go to that passport page from the search result page.  

Now what I would like to do to fix this is say "Run the query but really just return one entry for each "passport_label" which shows up in the search."   I had some hope for the "distinct" command in mongodb, but that just gives me the distinct 'passport_label' field entries, and not the whole entry corresponding to each of these.  Is there a way to do this in mongodb? To ask it to return one document corresponding to each of the distinct 'passport_label' fields which show up in any given query?

AndrewVSutherland:
@jenpaulhus So I think you could the same thing that is done to search for isogeny classes of elliptic curves: you only want records where total_label ends in ".1".  But rather than searching for this (which could be done but it would be slow) it would be better to add a new integer attribute 'num' which is equal to this last entry and distinguishes vectors that have the same passport label.

jenpaulhus:
@AndrewVSutherland  Great.  That was easy to fix since I already stored that integer in the database separate from the 'total_label'.  This last commit fixes the wording for the friend link to genus 2 curves and makes it so that repeated passport labels are not returned in searches.  

See for example
http://127.0.0.1:37777/HigherGenus/C/Aut/?start=20&paging=0&genus=2&group=%5B6%2C1%5D&signature=&dim=&inc_hyper=include&count=
vs
http://www.lmfdb.org/HigherGenus/C/Aut/?start=20&paging=0&genus=2&group=%5B6%2C1%5D&signature=&dim=&inc_hyper=include&count=

or

http://127.0.0.1:37777/HigherGenus/C/Aut/?start=0&paging=0&genus=3&group=%5B8%2C3%5D&signature=&dim=&inc_hyper=include&count=20
vs
http://www.lmfdb.org/HigherGenus/C/Aut/?start=0&paging=0&genus=3&group=%5B8%2C3%5D&signature=&dim=&inc_hyper=include&count=20

AndrewVSutherland:
Looks good.

PR #1759, opened 2016-07-14
remove sage_eval from p2sage function

JohnCremona:
All test pass so I think I caught all the types which p2sage() was called on; the only input not converted is '??' which seems to cause no problems to the caller (and would hve been unchanged anyway).

All tests in test.sh pass, so let's hope that we have enough tests.

### DIFF
--- a/lmfdb/lfunctions/Lfunctionutilities.py
+++ b/lmfdb/lfunctions/Lfunctionutilities.py
@@ -13,22 +13,29 @@ from lmfdb.base import getDBConnection
 ###############################################################
 
 def p2sage(s):
-    """ THIS FUNCTION IS DEPRECATED AND SHOULD NOT BE USED"""
-    from sage.all import sage_eval
-    # I really don't like the use of sage_eval here. It may be ok as long
-    # as we are only calling this function on trusted input from the database,
-    # but someone is going to forget that someday... --JWB
-    # This function should be removed and all calls to it replaced with something appropriate (e.g. PolynomialRing(QQ)(s) or something similar) -- AWS
-
-    x = PolynomialRing(RationalField(),"x").gen()
-    a = PolynomialRing(RationalField(),"a").gen()
-    try:
-        z = sage_eval(str(s), locals={'x' : x, 'a' : a})
-    except:
-        z = s
+    """Convert s to something sensible in Sage.  Can handle objects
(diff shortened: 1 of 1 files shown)


AndrewVSutherland:
Checking now.

AndrewVSutherland:
I tested this on every L-function listed on http://www.lmfdb.org/L/degree3/ and http://www.lmfdb.org/L/degree4/, and about half of those listed on http://www.lmfdb.org/L/degree2/, including checking Zeros, Plot, and lcalcfile and everything seemed to work (except in one case that is not related to your changes, the page http://www.lmfdb.org/L/ModularForm/GL3/Q/Maass/4/1/8.239796_2.641226/1.066042%20+/ does not show zeros (or say they are unavailable), even though http://www.lmfdb.org/L/Zeros/ModularForm/GL3/Q/Maass/4/1/8.239796_2.641226/1.066042%20+/
works fine -- I will look into this and raise a new issue if need be).

AndrewVSutherland:
@JohnCremona I made one minor change after merging your PR to use logger.error rather than print to record cases where p2sage fails to convert something -- this will make it easier to catch them in flasklog (I periodically search for errors).

One question in the line:
        Qa = PolynomialRing(RationalField(),"a"); a=Qa.gen()
you create the python variable a, which never gets used (as noted by pyflakes).  Is there a reason for this or is it just force of habit?

JohnCremona:
No, the variable a was not needed.  I added in its definition when getting passed the (obviously) failing attempt at CC('a^4'), but that gives a NameError even if a is defined, so I should have deleted it.

Although this is certainly better now than it was (since we should never use eval() and that is what sage_eval() eventually calls), I found myself first thinking that the caller of the p2sage function should themselves specify the parent structure (i.e. type, essentially), of the expected return value, only to realize that if they know that then they can do the conversion themselves, making this function redundant.

In fact there are only 7 (if I counted right) calls to p2sage from anywhere, and it might have been simpler to just put in an appropriate special-purpose conversion in each of those places.  For the next iteration.

AndrewVSutherland:
@JohnCremona Agreed, that was why I had commented it as "DEPRECATED", I think it would be better to remove all calls to p2sage, but what you did is a great first step and addresses the most important issue (not using eval).

PR #1760, opened 2016-07-14
Fixes #1752

### LINKED ISSUE #1752: many api json downloads fail with internal server errors
The following pages all give internal server errors; the problem appears to occur on collections whose documents contain attributes that are object ids (other than _id, which is handled separately), which causes errors like:

TypeError: ObjectId('4f7052dc5009fb18ef000811') is not JSON serializable

The following pages all have this problem:

http://www.lmfdb.org/api/MaassWaveForms/Coefficients?_format=json
http://www.lmfdb.org/api/MaassWaveForms/Coefficients.chunks?_format=json
http://www.lmfdb.org/api/MaassWaveForms/Coefficients.files?_format=json
http://www.lmfdb.org/api/MaassWaveForms/FS?_format=json
http://www.lmfdb.org/api/MaassWaveForms/Table.chunks?_format=json
http://www.lmfdb.org/api/MaassWaveForms/maassform_plots?_format=json
http://www.lmfdb.org/api/SL2Zsubgroups/groups?_format=json
http://www.lmfdb.org/api/modularforms2/webchar.chunks?_format=json
http://www.lmfdb.org/api/modularforms2/webeigenvalues.chunks?_format=json
http://www.lmfdb.org/api/modularforms2/webmodformspace.chunks?_format=json
http://www.lmfdb.org/api/modularforms2/webnewforms.chunks?_format=json
http://www.lmfdb.org/api/siegel_modular_forms/samples?_format=json

AndrewVSutherland:
This PR changes api.py in two ways:

- All object id's in documents are now encoded as "ObjectId('hex-string')", not just _id
- Uses ISO-8859-1 encoding and json_util.default to ensure all bson objects can be encoded as json objects in a reasonable way (including binary data, date-times, etc...)

All the problematic URLs listed in #1752 now work.

The resulting json output is slightly different (but has the same data content), compare:

http://www.lmfdb.org/api/genus2_curves/curves?_offset=0&_format=json
http://127.0.0.1:37777/api/genus2_curves/curves?_offset=0&_format=json

If we are concerned but 100% exact backward compatibility, I could put the existing call to flask.jsonify (which fails on the problem URLs) in a try/except and only use the new method when necessary.

My view is that this isn't critical and I would rather have the json pages all use the same format.

### DIFF
--- a/lmfdb/api/api.py
+++ b/lmfdb/api/api.py
@@ -4,13 +4,15 @@ import pymongo
 import urllib2
 ASC = pymongo.ASCENDING
 DESC = pymongo.DESCENDING
-import flask
 import yaml
+import json
+import flask
 import lmfdb.base as base
 from lmfdb.utils import flash_error
 from datetime import datetime
-from flask import render_template, request, url_for
+from flask import render_template, request, url_for, app, current_app
 from lmfdb.api import api_page, api_logger
+from bson import json_util
 from bson.objectid import ObjectId
 
 # caches the database information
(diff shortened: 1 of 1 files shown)


JohnCremona:
Looks fine to me.  If there is anyone out there who actually uses the json output they could test of the new output is readable, but I will assume so and certainly the server errors are fixed.

PR #1787, opened 2016-07-16
Sage version check at startup

AndrewVSutherland:
prints/logs a warning if starting the LMFDB with a Sage version older than LMFDB_SAGE_VERSION (currently set to 7.1)

### DIFF
--- a/lmfdb/website.py
+++ b/lmfdb/website.py
@@ -89,7 +89,7 @@ from flask import g, render_template, request, make_response, redirect, url_for,
 import sage
 
 DEFAULT_DB_PORT = 37010
-LMFDB_SAGE_VERSION = 7.1
+LMFDB_SAGE_VERSION = '7.1'
 
 @app.before_request
 def redirect_nonwww():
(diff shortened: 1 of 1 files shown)


JohnCremona:
Thanks for implementing this.  However, converting to float will not work in two ways: the minor version might be >9 (e.g. version 7.10), not likely perhaps but has happened;  and if I were running a development version I might find (as I just did):
sage: sage.version.version
'7.3.beta8'

I think it would work to do a lexicographical compare on 
sage: [int(c) for c in sage.version.version.split(".")[:2]]

By the way, the Sage version is obtained and used elsewhere already: in base.py (where instead of the --better-- sage.version.version we have sage.env.SAGE_VERSION which has to be imported) for display in the footer of every page.  And in lmfdb/modular_forms/elliptic_modular_forms/__init__.py we already extract the major and minor versions in code similar to the above .

Sorry, it would have been quicker for me to have amended the PR than write all this!

JohnCremona:
There's also Fredrik's install-deps.sh script, which I am not sure anyone actually uses...

JohnCremona:
Shall I make an updated version of the patch and let you test it?  I don't have an old enough version of Sage around to test on anyway ;)

AndrewVSutherland:
@JohnCremona I can amend my PR as you suggest, let me test and then I'll commit.

AndrewVSutherland:
@JohnCremona Done - it warns on Sage 6.5 but not on Sage 7.1 or 7.2.

JohnCremona:
That was quick!

JohnCremona:
BTW, it works fine with version 7.3.beta8!

Issue #1697, opened 2016-07-10
Bad labels for many degree 4 number fields on beta


AndrewVSutherland:
If you go to http://beta.lmfdb.org/NumberField/?discriminant=1944468 (for example) you will see 6 results

http://beta.lmfdb.org/NumberField/4.4.1944468.1
http://beta.lmfdb.org/NumberField/4.1944468.0.1
http://beta.lmfdb.org/NumberField/4.1944468.0.2
http://beta.lmfdb.org/NumberField/4.1944468.0.3
http://beta.lmfdb.org/NumberField/4.1944468.0.4
http://beta.lmfdb.org/NumberField/4.1944468.0.5

Only the first of these has a label of the form d.r.D.i where the second part r is the number of real embeddings (see http://beta.lmfdb.org/NumberField/FieldLabels) and the third part D is the absolute discriminant (it looks like r and D may have been inadvertently swapped).

This appears to affect more than a million recently added quartic fields (there are 2109331 degree 4 fields on beta, but only 618536 on prod, the problems all appear in the fields that are new on beta).

jwj61:
A byproduct of making a new import script.  Luckily, it appears it won't be too hard to fix.

jwj61:
The recently imported fields now have their labels fixed as can be seen from the search mentioned above.

This might be a good time to copy fields to the cloud -- it has many more fields than the cloud version, both a big import in degree 4, and some higher degree fields from Kluners-Malle.

I should do something about updating the random field database, but I forget what I am supposed to do.

AndrewVSutherland:
Use create_random_object_index in lmfdb/data_mgt/utilities/rewrite.py.  It will automatically drop the existing fields.rand collection when you run it.  You want to do

create_random_object_index (db, "fields")

where db is a pointer to the numberfields database on which you have authenticated as editor.

jwj61:
Done, and I will make a note of what to do.

AndrewVSutherland:
@jwj61 Do you want me to update the fields collection on the cloud now?

jwj61:
Yes, please.

AndrewVSutherland:
[...]

PR #1788, opened 2016-07-16
fix empty nf search label bug

JohnCremona:
A tiny fix this:  go to http://www.lmfdb.org/NumberField/ and click on the Go button without entering anything, and you get a server error, because of a small error which this fixes.  Now you just get the same page back again which seems OK to me.

### DIFF
--- a/lmfdb/number_fields/number_field.py
+++ b/lmfdb/number_fields/number_field.py
@@ -450,7 +450,7 @@ def number_field_search(**args):
         query = {'label_orig': info['natural']}
         try:
             parse_nf_string(info,query,'natural',name="Label",qfield='label')
-            return redirect(url_for(".by_label", label= clean_input(query['label'])))
+            return redirect(url_for(".by_label", label= clean_input(query['label_orig'])))
         except ValueError:
             query['err'] = info['err']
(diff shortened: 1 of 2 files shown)


AndrewVSutherland:
@JohnCremona This works, but I notice that test_search_poly_mean2parser now fails.

JohnCremona:
Sorry, I clearly did not test thoroughly.  Let me do that and fix it while you glance at PR#1772.

jwj61:
Do you want me to fix this?

JohnCremona:
Thanks @jwj61 but I just did -- a redirect is needed since (in effect) we have removed spaces from the URL.

AndrewVSutherland:
@JohnCremona I expect this is now fine and can be merged, but the numberfields database appears to be busy at the moment so I can't test (I assume @jwj61 is working on #1697, which may involve some updating a lot of records), http://beta.lmfdb.org/NumberField/ times out).

jwj61:
Yes, although it is using rewrite which only does read on the fields database.  In any case, that is finished.

AndrewVSutherland:
Looks good now.

PR #1789, opened 2016-07-16
fix gp code for ecnf

AurelPage:
The number field was missing in the arguments of the initialization of elliptic curves over number fields.

Example:
/EllipticCurve/3.3.148.1/356.1/a/1
(click on show commands for pari/gp)


JohnCremona:
Thanks Aurel -- as reported by me on pari-dev!

Issue #1358, opened 2016-05-14
Some ECNF pages produce 502 Server Error messages


AndrewVSutherland:
The page

 http://www.lmfdb.org/EllipticCurve/6.6.371293.1/79.1/c/4

(and several others) produces a 502 server error message.

edgarcosta:
```
Exception on /EllipticCurve/6.6.371293.1/79.1/c/4 [GET]
Traceback (most recent call last):
  File "/home/sage/sage-7.1/local/lib/python2.7/site-packages/Flask-0.10.1-py2.7.egg/flask/app.py", line 1817, in wsgi_app
    response = self.full_dispatch_request()
  File "/home/sage/sage-7.1/local/lib/python2.7/site-packages/Flask-0.10.1-py2.7.egg/flask/app.py", line 1477, in full_dispatch_request
    rv = self.handle_user_exception(e)
  File "/home/sage/sage-7.1/local/lib/python2.7/site-packages/Flask-0.10.1-py2.7.egg/flask/app.py", line 1381, in handle_user_exception
    reraise(exc_type, exc_value, tb)
  File "/home/sage/sage-7.1/local/lib/python2.7/site-packages/Flask-0.10.1-py2.7.egg/flask/app.py", line 1475, in full_dispatch_request
    rv = self.dispatch_request()
  File "/home/sage/sage-7.1/local/lib/python2.7/site-packages/Flask-0.10.1-py2.7.egg/flask/app.py", line 1461, in dispatch_request
    return self.view_functions[rule.endpoint](**req.view_args)
  File "/home/lmfdb/lmfdb-git-prod/lmfdb/ecnf/main.py", line 313, in show_ecnf
    ec = ECNF.by_label(label)
  File "/home/lmfdb/lmfdb-git-prod/lmfdb/ecnf/WebEllipticCurve.py", line 125, in by_label
    return ECNF(data)
  File "/home/lmfdb/lmfdb-git-prod/lmfdb/ecnf/WebEllipticCurve.py", line 116, in __init__
    self.make_E()
  File "/home/lmfdb/lmfdb-git-prod/lmfdb/ecnf/WebEllipticCurve.py", line 138, in make_E
    N = E.conductor()
  File "/home/sage/sage-7.1/local/lib/python2.7/site-packages/sage/schemes/elliptic_curves/ell_number_field.py", line 1600, in conductor
    for d in self.local_data()],\
[...]

PR #1792, opened 2016-07-16
Avoid server errors on EC download queries

AndrewVSutherland:
Recently we have seeing a lot of errors in the flasklog of the form:

Failed URL: http://www.lmfdb.org/EllipticCurve/2.2.229.1/?download=1&query={'field_label': '2.2.229.1'}
Failed URL: http://www.lmfdb.org/EllipticCurve/Q/21780/?download=1&query={'conductor': 21780}
Failed URL: http://www.lmfdb.org/EllipticCurve/2.2.17.1/136.1/?download=1&query={'field_label': '2.2.17.1', 'conductor_label': '136.1', 'conductor_norm': 136}
Failed URL: http://www.lmfdb.org/EllipticCurve/Q/14/?download=1&query={'conductor': 14}
...
I'm not sure where these are coming from (they don't look like a robot), but they cause a server error because the "Submit" parameter is not set.  This PR makes a one line change in two places to verify that all the arguments the download_search function requires are actually present before attempting the download (otherwise it just treats the request as it usually would).

### DIFF
--- a/lmfdb/ecnf/main.py
+++ b/lmfdb/ecnf/main.py
@@ -344,8 +344,8 @@ def show_ecnf(nf, conductor_label, class_label, number):
 
 
 def elliptic_curve_search(info):
-    
-    if 'download' in info and info['download'] != 0:
+
+    if info.get('download') == '1' and info.get('Submit') and info.get('query'):
         return download_search(info)
 
     if not 'query' in info:
--- a/lmfdb/elliptic_curves/elliptic_curve.py
+++ b/lmfdb/elliptic_curves/elliptic_curve.py
@@ -159,7 +159,7 @@ def elliptic_curve_jump_error(label, args, wellformed_label=False, cremona_label
 
 def elliptic_curve_search(info):
 
-    if 'download' in info and info['download'] != '0':
+    if info.get('download') == '1' and info.get('Submit') and info.get('query'):
         return download_search(info)
 
     if 'SearchAgain' in info:


AndrewVSutherland:
BTW, I ran test.sh and all tests pass.

JohnCremona:
Will look at this tomorrow

JohnCremona:
I don't understand this at all, and certainly see no way to test that the change has any effect!
Looking at the code /Q the only way I can see that the function elliptic_curve_search() ever gets called with download!=0 is when the "Download all search results" button is clocked  on the search results page, and then Submit is also set (to 'gp' or something else).

AndrewVSutherland:
@JohnCremona My guess is that someone is manually typing in URLs are running a script (I do this a lot myself).  You can easily test that the change has an effect by doing so yourself, e.g., compare:

http://www.lmfdb.org/EllipticCurve/Q/?jinv=1728&download=1
http://127.0.0.1:37777/EllipticCurve/Q/?jinv=1728&download=1

The goal of this PR is simply a one-line change that avoids calling the download_search function in situations where we know it is going to fail.  I think you can agree that it does no harm -- in the case of requests triggered by pressing one of the download buttons, the code will do exactly the same thing it does now, and if the download argument is not present it will also do the exact same thing it does now.

The benefit of the PR is that (1) it will stop the errors from showing up in flasklog, and (2) it will make it easier for whoever is doing this to figure out what they should be doing.

JohnCremona:
Good points, and I'll now do that test and will probably merge this -- though I don't much like the principle of catching errors cause by people making up URLs since there is no way we can catch all such!

AndrewVSutherland:
@JohnCremona I agree that we should feel no obligation to make malformed URLs work.   I only did this because it was easy and I could see it being caused by a well-meaning user (and it will make the errors stop showing up in the flasklogs so that I can find actual bugs more easily).

JohnCremona:
FIne --as you see I merged it.
I can see errors in the flasklog on beta -- the most recent ones are below -- but am not sure which of these are already known about:

lmfdb@lmfdb:~/lmfdb-git-beta$ grep error flasklog
[2016-07-16 22:08:30 UTC] 500 error on URL http://beta.lmfdb.org/api/MaassWaveForms/FS/4f5d5aea88aece444000002c
[2016-07-17 02:33:18 UTC] 500 error on URL http://beta.lmfdb.org/NumberField/6.6.371293.1
[2016-07-17 03:52:01 UTC] 500 error on URL http://beta.lmfdb.org/ArtinRepresentation/?natural=2.3e3_11e2.12t12.2c1&search=Go
[2016-07-17 04:15:53 UTC] 500 error on URL http://beta.lmfdb.org/ModularForm/GL2/Q/holomorphic/20/9/11/
[2016-07-17 04:41:51 UTC] 500 error on URL http://beta.lmfdb.org/ArtinRepresentation/?natural=2.3e3_11e2.12t12.2c1&search=Go
[2016-07-17 12:30:04 UTC] 500 error on URL http://beta.lmfdb.org/EllipticCurve/6.6.371293.1/53.2/a/2
[2016-07-17 12:31:21 UTC] 500 error on URL http://beta.lmfdb.org/EllipticCurve/6.6.371293.1/79.1/c/3
[2016-07-17 14:13:26 UTC] 500 error on URL http://beta.lmfdb.org/EllipticCurve/6.6.371293.1/79.1/c/3
[2016-07-17 14:14:34 UTC] 500 error on URL http://beta.lmfdb.org/ModularForm/GL2/Q/holomorphic/20/9/11/

AndrewVSutherland:
@JohnCremona These are all known problems, many of which have recently been fixed.  With your most recent merge the only ones that should still be open are those related to modular forms.

AndrewVSutherland:
@JohnCremona And the three related to elliptic curves over number fields  (these are related to issue #1358 that you are looking at now).

PR #1793, opened 2016-07-16
Fix #1285

### LINKED ISSUE #1285: Quotation marks
The latex use of `` and " don't render correctly. For example see
http://www.lmfdb.org/L/history just below the header Riemann's Memoir.

AndrewVSutherland:
This PR addresses #1285 adds one additional re.sub to the knowl md_latex_accents preprocssing code to replace any occurence of ``something'' with &ldquo;something&rdquo; (i.e. enclose something in HTML left and right double quotes).  To see the difference, compare the first sentence following the heading "Reimann's Memoir" in

Before: http://www.lmfdb.org/L/history
After: http://127.0.0.1:37777/L/history

(you may have to zoom in to see that they really are left and right double quotes).

Per the discussion in #1285, this only matches the "official" latex syntax, i.e. two left single quotes followed by something followed by two single quotes.


JohnCremona:
Result looks good, though I'm not a good enough reader of regular expressions to understand the one you are matching with except that it starts with `` and ends with '' !

Issue #112, opened 2013-09-16
Degree 1 L function browsing plot


pdehaye:
The plot is broken at the moment on beta.lmfdb.org
If you slide ot the range 45-50, for instance, you get something in the range 15-20 (it appears)

jwbober:
This is still broken 2 years later.

JohnCremona:
Example of a URL please?

jwbober:
http://beta.lmfdb.org/L/degree1/

davidfarmer:
still broken on chrome, works on firefox

edgarcosta:
Broken on Chrome and Safari. Works on Firefox.
<img width="2560" alt="screen shot 2016-04-07 at 20 16 30" src="https://cloud.githubusercontent.com/assets/322822/14370637/21d29bce-fcfe-11e5-9b09-8b5fa341d38f.png">

AndrewVSutherland:
Confirmed fix now on http://www.lmfdb.org/L/degree1/ works on chrome/safari/firefox,

PR #1801, opened 2016-07-17
pyflakes changes for assorted import scripts

jwj61:


AndrewVSutherland:
Thank you for doing this!

jwj61:
I have webnumberfield done, but accidentally did it on top of the current pull request.

AndrewVSutherland2:
I am working my way through the modular forms code (lot's of import *'s to
deal with -- I'm testing the changes by checking every modular form page
and making sure all the ones that used to load correctly (many on the
production system do not) still do).  I should finish today, which will let
us close #112.

PR #1802, opened 2016-07-17
Fix for bug in Artin representations caused by recent removal of construction of sage number fields.

jwj61:
The recent change to avoid constructing sage number fields overlooked the case where an artin representation needs a field which is not in the database.  Then we need to construct it, which leads to other changes to accomodate both cases.

Artin representation 2.3e3_11e2.12t12.2c1 is an example which currently crashes beta and prod.  To see other cases which involves this code, try NumberField/x%5E4-2 (i.e., x^4-2) and see that Frobenius splittings at the bottom are still ok, and the Artin representations linked to at the bottom still work.

### DIFF
--- a/lmfdb/math_classes.py
+++ b/lmfdb/math_classes.py
@@ -770,8 +770,14 @@ class NumberFieldGaloisGroup(object):
         try:
             return self._residue_field_degrees(p)
         except AttributeError:
-            from number_fields.number_field import residue_field_degrees_function
-            fn_with_pari_output = residue_field_degrees_function(self.wnf())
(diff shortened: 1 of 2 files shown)


AndrewVSutherland:
Looking at this now.

AndrewVSutherland:
Looks good; all tests pass (except for known MF problems) and the page http://127.0.0.1:37777/ArtinRepresentation/2.3e3_11e2.12t12.2c1 works, whereas http://www.lmfdb.org/ArtinRepresentation/2.3e3_11e2.12t12.2c1 is currently broken.

I clicked around on the number field page you indicated and didn't notice any obvious problems.

Mergin.

Issue #1803, opened 2016-07-17
Server errors on many Maass form pages


AndrewVSutherland:
Many Maass form pages are throwing internal server errors that show up in the flasklogs of the www.lmfdb.org server.  This is apparently due to some sort of data inconsistency -- the specified ids are object ids of records in MaassWaveForms/Coefficients (e.g. see http://www.lmfdb.org/api/MaassWaveForms/Coefficients/4f55571188aece241f00000d for the first example listed below), but the ObjectId listed in the coeff_id attribute of that record is not found.

There are at least 49 affected pages:

http://www.lmfdb.org/ModularForm/GL2/Q/Maass/4f55571188aece241f00000d
http://www.lmfdb.org/ModularForm/GL2/Q/Maass/4f5692ca88aece2afe000002
http://www.lmfdb.org/ModularForm/GL2/Q/Maass/4f5693de88aece2afe00000e
http://www.lmfdb.org/ModularForm/GL2/Q/Maass/4f56952f88aece2afe00001e
http://www.lmfdb.org/ModularForm/GL2/Q/Maass/4f5697fc88aece2afe00003e
http://www.lmfdb.org/ModularForm/GL2/Q/Maass/4f56984e88aece2afe000044
http://www.lmfdb.org/ModularForm/GL2/Q/Maass/4f56990188aece2afe00004d
http://www.lmfdb.org/ModularForm/GL2/Q/Maass/4f56990388aece2afe00004e
http://www.lmfdb.org/ModularForm/GL2/Q/Maass/4f56990888aece2afe00004f
http://www.lmfdb.org/ModularForm/GL2/Q/Maass/4f56990b88aece2afe000050
http://www.lmfdb.org/ModularForm/GL2/Q/Maass/4f56991588aece2afe000051
http://www.lmfdb.org/ModularForm/GL2/Q/Maass/4f56991f88aece2afe000053
http://www.lmfdb.org/ModularForm/GL2/Q/Maass/4f56992188aece2afe000054
http://www.lmfdb.org/ModularForm/GL2/Q/Maass/4f56998f88aece2afe00005e
http://www.lmfdb.org/ModularForm/GL2/Q/Maass/4f56999c88aece2afe000060
http://www.lmfdb.org/ModularForm/GL2/Q/Maass/4f5699a288aece2afe000061
http://www.lmfdb.org/ModularForm/GL2/Q/Maass/4f5699a588aece2afe000062
[...]

PR #1807, opened 2016-07-17
Interim partial fix for #1803

AndrewVSutherland:
This PR is an initial attempt to address part of #1803 by displaying a meaningful error message to the user and giving them an opportunity to report the problem.  It also logs an error in the flasklog.

I added the template problem.html (which is essentially a copy of LfunctionSingle.html used by render_lfunction_exception in similar situations) which can be used to report problems where the application actually has some clue as to what the problem is (as opposed to a generic 404 or 500 errors).

To see it in action compare

http://www.lmfdb.org/ModularForm/GL2/Q/Maass/4f55571188aece241f00000d
http://127.0.0.1:37777/ModularForm/GL2/Q/Maass/4f55571188aece241f00000d

All modular forms tests that used to pass still pass.


JohnCremona:
Looks good.  The new problem.html template is something we should think of using in other palces perhaps.

PR #1808, opened 2016-07-17
Cleanup emf imports

AndrewVSutherland:
This PR cleans up pyflakes issues in the lmfdb/modular_forms/elliptic_modular subtree of the LMFDB codebase, including 8 occurrences of import * (see #122).  pyflakes now runs cleanly on all .py files in this branch.  It also includes a cleanup of mwf_classes.py that was necessary because it was indirectly accessing symbols imported via import *'s in the elliptic_modular_forms code (there are actually nested import *'s in the modular forms code).

All tests hit by test.sh run cleanly except for the known failure in test_galois_conjugate that was already present.  I also ran emf_test_pages.py which attempts to load every classical modular form in the database within our declared extent ranges, and no new errors appeared (as before, there are still 128 bad pages, these are listed in emf_test_pages.log which I generated both before and after this PR with identical results).

Once this PR is merged there should be only 15 import *'s left to fix, all in the lmfdb/modular_forms/mass_forms subtree.  I will fix these next (but this PR needs to be merged first due to dependencies).


AndrewVSutherland:
I should add that I also fixed a few obvious bugs that pyflakes found, see extract_data_from_jump_to in  lmfdb/modular_forms/elliptic_modular_forms/backend/emf_utils.py, for example.

JohnCremona:
I'll try to go through this today.

AndrewVSutherland:
@JohnCremona Thanks, I don't want to tackle lmfd/modular_forms/maas_forms (which is all the remains between us and closing #112) until this one is merged.

JohnCremona:
Just one small thing:

l.43 of modular_forms/elliptic_modular_forms/backend/__init__.py: after deleting {2} I think that the {3} should be changed to {2}.  This would cause an error if the functino was actually used, so it certainly isn't used in any test!

[I approve of just commenting out assigments to variables whcih are never used rather than deleting them.]

JohnCremona:
@AndrewVSutherland By #112 you meant #122?

AndrewVSutherland:
@JohnCremona Good catch, fixed.  And yes I did mean #122 (comment edited).

Issue #1236, opened 2016-05-08
Store counts/statistics in the database and put in place a procedure to update them


jvoight:
I think we could easily put in place procedures to have counts automatically updated every night, but we don't necessarily need this in place for the launch.  (So we'll want a simple script that can be run when someone updates the data, but is always run at night.)

AndrewVSutherland:
The first step is to have a generic mechanism for computing and storing counts; I am going to work on this and implement it in genus 2 curves.

Once this is in place, I don't think we need to do it nightly, at least not on the production server, we only need to do it when the database changes (which is infrequent).

AndrewVSutherland:
I'll also note that, aside from a few issues in ecnf and number fields, this issue is the main thing that is keeping me from closing #1247.

JohnCremona:
Quite a few collections / sections do this, with some version of an update_stats() script in the scripts directory which can be easily run after any data updates.  Quite possibly, all places where we display counts have such scripts in place.  If that were true then this could be closed.

edgarcosta:
> Quite possibly, all places where we display counts have such scripts in place.

I think that is the case, and I think we can close this. 

ps: If this is not the case, someone will tumble up on this issue again, and raise a new issue or email us.

PR #1814, opened 2016-07-17
Store statistics in the database

AndrewVSutherland:
This is a first step toward addressing #1236.  It consists of two parts:

- A new function update_attribute_stats has been added to lmfdb/data_mgt/utilities/rewrite.py. It takes a db, a collection name (say coll), and a list of attribute names.  For each attribute in the list, it creates a record in the collection coll.stats (which will be created if not present) that contains statistics for the attribute, including a sorted list of (value,count) pairs that counts how many objects in coll have attr set to value.  This is intended for attributes whose values lie in a reasonably small set (e.g. torsion structures, rank, etc...) on which one might want to report statistics (but even an attribute like "base_field" that might take hundreds of values should still be fine, it will work up to the pymongo bson limit of 16MB).  You can view an example at http://beta.lmfdb.org/api/genus2_curves/curves.stats/.

- A new function attribute_value_counts has been added to lmfdb/utils.py that take a collection name and an attribute name.  It returns a sorted list of (value,count) pairs for the specified attribute as above.  If the collection coll.stats is present and contains a record for the specified attribute **and** the number of objects in coll is the same as it was when that record was created, it will return the list stored in coll.stats, otherwise it will compute it on the fly (if the statistics are present but out of date a warning will be written to flasklog suggesting that the statistics should be updated).

You can see this in action on http://127.0.0.1:37777/Genus2Curve/Q/stats, the genus 2 stats page has been updated to use attribute_value_counts when it computes statistics, which will read the stats from the database because they are present in genus2_curves.curves.stats.  You can compare to http://www.lmfdb.org/Genus2Curve/Q/stats to see that the statistics page looks the same.

This shaves about one second off the load time of for the genus 2 curves statistics page; the improvement should be greater for other statistics pages for larger collections if they are changed to use the new functions introduced in this PR.

### DIFF
--- a/data_mgt/utilities/rewrite.py
+++ b/data_mgt/utilities/rewrite.py
@@ -100,7 +100,7 @@ def update_attribute_stats(db, coll, attributes):
     for attr in attributes:
         db[statscoll].delete_one({'_id':attr})
     total = db[coll].count()
-    counts = [[value,db[coll].find({attr:value}).count()] for value in sorted(db[coll].distinct(attr))]
-    min, max = counts[0][0], counts[-1][0]
-    db[statscoll].insert_one({'_id':attr, 'total':total, 'counts':counts, 'min':min, 'max':max})
-
+    for attr in attributes:
+        counts = [[value,db[coll].find({attr:value}).count()] for value in sorted(db[coll].distinct(attr))]
+        min, max = counts[0][0], counts[-1][0]
+        db[statscoll].insert_one({'_id':attr, 'total':total, 'counts':counts, 'min':min, 'max':max})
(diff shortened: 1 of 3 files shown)


JohnCremona:
This look like a really useful thing to have -- thanks!  I'm sure you have tested it well with genus 2 curves, and of course there may be issues which arise when more of use try it out, but that's not a reason to delay merging this.  I would prefer to merge sooner, and then try it out on (for example) the elliptic_curves/curves collection, rather than try it out when it is still a separate branch.

One small thing (literally 2): in utils.py you twice have "if m != m" when it should be "if m != n" I think!

AndrewVSutherland:
@JohnCremona Good catch!  Fixed.

AndrewVSutherland:
@JohnCremona If you hit any issues with elliptic_curves/curves or think of additional functionality you would like the .stats collection to support, please let me know -- I initially planned to do more (e.g. storing counts from arbitrary queries) but decided to keep it minimal and then add functionality once it became clearer what was needed/useful.

PR #1816, opened 2016-07-18
Fix #1815

### LINKED ISSUE #1815: Server error on number field discriminant search
On http://www.lmfdb.org/NumberField, type in any comma-delimited list of integers (e.g. "-3,8") into the discriminant box and press search to get a server error.  Comma delimited lists work fine in other fields (e.g. degree)

The problem is caused by a bug in parse_signed_ints in search_parsing.py, which calls parse_range3 which attempts to call an undefined function parse_discs (this function does not appear in any of the LMFDB code, nor is it defined in sage) -- I discovered this while doing a pyflakes cleanup of the files in lmfdb/lmfdb.

AndrewVSutherland:
Fixes #1815 and does a pyflakes cleanup of the top level directory lmfdb/lmfdb (there is still one pyflakes complaint that is not a problem).  All tests pass (except the usual test_galois_conjugate failure).

To test the fix to #1815, compare

http://www.lmfdb.org/NumberField/?discriminant=-3%2C8
http://127.0.0.1:37777/NumberField/?discriminant=-3%2C8

Then go to http://127.0.0.1:37777/NumberField/ and try various inputs for the discriminant field, e.g. "-8,-3,5,28", or "-20..10" or "-30..10,20..40"


JohnCremona:
Some comments / queries:

In Webcharacter.py some lines have been deleted, perhaps because pyflakes said that things were defined but not used, which look gangerous to me: (a) how can we be sure thet the attribute _order2 is never accessed, or no-one will ever want to access it? (b) in the line above "_ = self.valuefield" there was a comment suggesting that this line was tere for a reason, perhaps there is a side-effect? 

In lmfdb/__init.py__ there's no lewline at the end of the line inserted

Why has the method poles() of class ArtinRepresentation been removed?

Now I'll test that #1815 is fixed.

AndrewVSutherland:
@JohnCremona The variable _order2 is not accessed anywhere in the LMFDB (as you can confirm by doing grep -r "_order2" . from the top lmfdb directory), and the fact that the variable name begins with an underscore suggests that it was never intended to be visible outside of the class.  As you can see from looking at the code in valuefield, its only side effect is to compute _order2, it does not do anything else other than return the string vf.  Calling valuefield every time vflabel is called for the sole purpose of recomputing _order2 (which is a one line computation) is confusing and unnecessary.

There were two identical copies of the method poles() in ArtinRepresentation, I only deleted the second one.  The first one is still there at line 399.

I added the newline to the end of __init__.py, thanks for catching that!

AndrewVSutherland:
@JohnCremona I meant to add, thank you for looking at the changes so carefully, and I appreciate the questions!  This PR and #1808 both make a lot of small changes to a bunch of files (thereby fixing many unnoticed bugs as well as cleaning up imports).  They definitely deserve to have a second pair of eyes look through them carefully.

JohnCremona:
Good, thanks for the full answers -- I was sure there would be good answers, but as you say, a second pair of eyes is a good thing.
About the lack of a final newline, it was github which showed me that in the way it displayed the diffs.

PR #1818, opened 2016-07-18
small tweak to make pyflakes happy

AndrewVSutherland:


JohnCremona:
:) clearly pyflakes is always right: code is simpler and a few bytes shorter now!

Issue #1812, opened 2016-07-17
Maass forms with 500 blank Fourier coefficients


AndrewVSutherland:
If you go to http://www.lmfdb.org/ModularForm/GL2/Q/Maass/4cb8503a58bca91458000008 you will see a table of Fourier coefficients c(n) for n from 0 to 499, all of which are blank (for what a typical page looks like, click the next or prev buttons).

AndrewVSutherland:
Two more pages with the same problem:

http://www.lmfdb.org/ModularForm/GL2/Q/Maass/4cb8503a58bca91458000009
http://www.lmfdb.org/ModularForm/GL2/Q/Maass/4cb8503a58bca9145800000a

The associated L-function pages appear to work, they show zeros and plot as unavailable.

AndrewVSutherland:
The same problem affects all the pages listed below.  The problem appears to be caused by bad data in Coefficients.files/Coefficients.chunks.  In each case the value of 'Numc' in the FS record for the form is at least 2000, as is the value of the 'numc' record in the corresponding Coefficients.files record, but the 'length' of the Coefficient data is just 29 (certainly not long enough to hold 2000 coefficients).

http://www.lmfdb.org/ModularForm/GL2/Q/Maass/4cb8503a58bca91458000008
http://www.lmfdb.org/ModularForm/GL2/Q/Maass/4cb8503a58bca91458000009
http://www.lmfdb.org/ModularForm/GL2/Q/Maass/4cb8503a58bca9145800000a
http://www.lmfdb.org/ModularForm/GL2/Q/Maass/4f4d22a788aece4908000061
http://www.lmfdb.org/ModularForm/GL2/Q/Maass/4f4d22a788aece4908000062
http://www.lmfdb.org/ModularForm/GL2/Q/Maass/4f4d22a788aece4908000063
http://www.lmfdb.org/ModularForm/GL2/Q/Maass/4f4d22a788aece4908000064
http://www.lmfdb.org/ModularForm/GL2/Q/Maass/4f4d22a788aece4908000065
http://www.lmfdb.org/ModularForm/GL2/Q/Maass/4f4d22a888aece4908000066
http://www.lmfdb.org/ModularForm/GL2/Q/Maass/4f4d22a888aece4908000067
http://www.lmfdb.org/ModularForm/GL2/Q/Maass/4f4d22a888aece4908000068
[...]

PR #1819, opened 2016-07-18
Cleanup maass form imports, fixes #122

### LINKED ISSUE #122: never do from something import *

AndrewVSutherland:
This PR is a pyflakes cleanup of the modular_forms/maass_forms/* subtree of the LMFDB that removes the last remaining import *'s in the LMFDB codebase (thereby allowing us close #122).

It uncovered many bugs due to references to misspelled or never defined variables; these all appear to be in code branches that are not (typically) executed or in functions that are not called anywhere in the LMFDB (a disturbing number of them were in maass_forms_db.py which is used and is the source of a few open production issues).

Where it was possible to discern the intent, I fixed these bugs, and where it was not I commented out the broken code and added a FIXME comment.  I also commented out some code that requires SQL imports and a SQL database to run (this code has not been touched in 4 years and appears not to be used).

All tests runs successful (except the known test_galois_conjugates errors) and browsing through the pages under  http://127.0.0.1:37777/ModularForm/GL2/Q/Maass/ did not reveal any changes -- everything that works on www.lmfdb.org appears to still work, and everything that is broken on www.lmfdb.org is still broken (but hopefully will now be easier to fix).

Pages I checked include:
http://127.0.0.1:37777/ModularForm/GL2/Q/Maass/BrowseGraph/1/10/0/10/
http://127.0.0.1:37777/ModularForm/GL2/Q/Maass/4cb8503a58bca91458000032
http://127.0.0.1:37777/L/ModularForm/GL2/Q/Maass/4cb8503a58bca91458000032/
http://127.0.0.1:37777/L/ModularForm/GL2/Q/Maass/4cb8503a58bca91458000032/?download=lcalcfile
http://127.0.0.1:37777/ModularForm/GL2/Q/Maass/1/
http://127.0.0.1:37777/ModularForm/GL2/Q/Maass/4cb8503a58bca91458000031
http://127.0.0.1:37777/ModularForm/GL2/Q/Maass/BrowseGraph/100/250/0/2/
http://127.0.0.1:37777/ModularForm/GL2/Q/Maass/BrowseGraph/100/1000/0/1/
http://127.0.0.1:37777/ModularForm/GL2/Q/Maass/53234e85acf756120945d7d0
http://127.0.0.1:37777/L/ModularForm/GL2/Q/Maass/53234e85acf756120945d7d0/
http://127.0.0.1:37777/L/ModularForm/GL2/Q/Maass/53234e85acf756120945d7d0/?download=lcalcfile
http://127.0.0.1:37777/ModularForm/GL2/Q/Maass/997/

I also checked that prev/next and download links (both of them) on Maass form pages still work.

### DIFF
--- a/lmfdb/modular_forms/backend/__init__.py
+++ b/lmfdb/modular_forms/backend/__init__.py
@@ -1,4 +1,6 @@
-# from mf_utils import ModularFormDisplay
-from mf_classes import MFDataTable
+# -*- coding: utf-8 -*-
+
 import mf_classes
+assert mf_classes
 import mf_utils
+assert mf_utils
--- a/lmfdb/modular_forms/backend/mf_classes.py
+++ b/lmfdb/modular_forms/backend/mf_classes.py
@@ -1,3 +1,5 @@
+# -*- coding: utf-8 -*-
+
 from lmfdb.base import getDBConnection
 from lmfdb.modular_forms import mf_logger
 
@@ -97,7 +99,7 @@ class MFDisplay(object):
         self.table = {}
 
     def connect(self):
-        self.db = base.getDBConnection()[dbname]
+        self.db = getDBConnection()[self._dbname]
 
     def set_table(self):
         raise NotImplementedError("Needs to be overwritten in subclasses!")
--- a/lmfdb/modular_forms/maass_forms/__init__.py
+++ b/lmfdb/modular_forms/maass_forms/__init__.py
@@ -1,4 +1,4 @@
-# /modular_forms/maass_forms/__init__.py
+# -*- coding: utf-8 -*-
 
 from lmfdb.utils import make_logger
 from flask import Blueprint
@@ -8,5 +8,6 @@ maassf = Blueprint(MAASSF, __name__, template_folder="views/templates")
 maassf_logger = make_logger(maassf)
 
 import maass_waveforms
-from maass_waveforms import *
+assert maass_waveforms
 import picard
+assert picard
--- a/lmfdb/modular_forms/maass_forms/maass_waveforms/__init__.py
+++ b/lmfdb/modular_forms/maass_forms/maass_waveforms/__init__.py
@@ -1,3 +1,5 @@
+# -*- coding: utf-8 -*-
+
 from lmfdb.utils import make_logger
 from flask import Blueprint
 
@@ -6,6 +8,6 @@ mwf = Blueprint(MWF, __name__, template_folder="views/templates", static_folder=
 mwf_logger = make_logger(mwf)
 
 import backend
+assert backend
 import views
-from backend import *
-from views import *
+assert views
--- a/lmfdb/modular_forms/maass_forms/maass_waveforms/backend/__init__.py
+++ b/lmfdb/modular_forms/maass_forms/maass_waveforms/backend/__init__.py
@@ -1,3 +1,8 @@
-from web_maassforms import WebMaassFormSpace, WebMaassForm
-from mwf_utils import *
-from mwf_classes import *
+# -*- coding: utf-8 -*-
+
+import web_maassforms
+assert web_maassforms
+import mwf_utils
+assert mwf_utils
+import mwf_classes
+assert mwf_classes
--- a/lmfdb/modular_forms/maass_forms/maass_waveforms/backend/maass_forms_db.py
+++ b/lmfdb/modular_forms/maass_forms/maass_waveforms/backend/maass_forms_db.py
@@ -284,7 +284,7 @@ class MaassDB(object):
             date_begun = ff['date']
             now = bson.datetime.datetime.utcnow()
             elapsed = now - date_begun
-            # FIXME: date_begnus is never used
+            # FIXME: date_beguns is never used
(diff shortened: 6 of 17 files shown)


JohnCremona:
I can look at this (again) but not until tomorrow.  It would be nice if one of the authors of the code could be pinged about it!

BTW, in using pyflakes on my own code I came across several instances of something which can lead to bugs which are very hard to track down otherwise, though are sometimes harmless:  the variables used in a list comprehension are *not* local to that as one might think!  So it you have defined (say) "E = EllipticCurve(...)" and need to use the value of E later, you really should not put, a few lines later, "jlist = [E.j_invariant() for E in Elist]".  But I think that the people reading this conversation already know that...

AndrewVSutherland:
@JohnCremona thanks, tomorrow is fine.

Regarding pyflakes, yes it can definitely reveal some very subtle bugs, which is why I think our efforts to do a pyflakes cleanup of the entire codebase are well worth while (and we are not all that far from being done, in the process of cleaning up imports we have also cleaned up almost everything flagged by pyflakes).

The list comprehension variable scoping you mention is a particularly nasty "feature" of python that I have always viewed as a bug -- the python implementation team apparently agrees, since they actually fixed this in python 3.  But of course we are still on python 2.x and likely will be for a long time due to dependencies in sage.  All the more reason to run pyflakes on everything.

Actually, once we have cleaned up all the pyflakes issues, I would like to propose that we add pyflakes to requirements.txt and modify test.sh to include the line "find | grep "\.py$" | xargs pyflakes" to automatically run pyflakes on all directories being searched for test scripts.  This will keep new problems (including import *'s) from creeping into the code.

JohnCremona:
I didn't know that about python 3 -- good!

I agree with adding the running of pyflakes as standard to test.sh.  One thing which would make this work slightly better would be to move all code which is not used in running the website (import scripts mainly, but there are some others) to a new place: we already have lmfdb/import_scripts (which has some but not all impost scripts) but I was thinking of a separate top-level directory alongside perfmon and data_mgt.

AndrewVSutherland:
@JohnCremona I like the idea of pulling code that is not part of the website out of the lmfdb directory tree a lot (I would still suggest people don't use import * and do use pyflakes -- bugs in import scripts are not a good thing either -- but I would not be so worried about eval and sage_eval).

edgarcosta:
+1 for this:
> One thing which would make this work slightly better would be to move all code which is not used in running the website (import scripts mainly, but there are some others) to a new place

If we clearly separate the what is used for the website and what is not used in the website, then we can slowly raise the standards  and easily "enforce" the new standards for all the code that is used for the website.
For example, enforcing no sage_eval or eval, etc...

JohnCremona:
We should probably open a new issue for the topic of the last few comments -- put all code not used by the live database into a separate subdirectory (say "local") of the top-level, starting by moveing lmfdb/import_scripts to local/impost_scripts.

Before we start doingt his, I think we should acknoeledge the fact that there ahve been a vast number of mostly minor code changes happening, which is very likely to have the side-effect that anyone out there with branches they were working on will now find themselves having to merge with the current master and dealing with some conflicts.  Of course this is inevitable, but from a human point of view it is likely to result in delays in people working on useful things, at least until they know that this process has been completed!

JohnCremona:
About 80% or even 90% of the changes here are instantly mergable, but for the remaining 10-20% it would be preferable for someone who has written or worked on this code before to look at the changes, especially as there are only 2 small tests in lmfdb/modular_forms/maass_forms/, so saying that all tests pass is not saying so much.
Who should asked to take a look?  Or shall we merge this now anyway -- I saw nothing obviously wrong?

AndrewVSutherland:
@JohnCremona Most of the code was originally written by @lemurell but has not been modified by him in a long time, so I'm not sure how familiar he is with the code.  I agree with you about the test cases (in fact I will raise an issue about this), which is why I also did a fair amount of manually browsing to test (e.g. you might also want to click through the list of URLs I put in the initial description of the PR to).

If @lemurell wants (and has time) to look through this PR that would be great!  But I don't think we should wait to merge.  There are at least 4 significant outstanding issues in the Maass forms code that need to be fixed for Release 1.1 (#1803, #1812, #1820, #1821), and we can't really start to work on them until after this PR is merged (otherwise we are going to have all sorts of merge conflicts to deal with).

I realize there is some risk, but I don't think anything significant is going to break (and I will take responsibility for fixing it if it does).  There are already a lot of bugs in the Maass forms code (a few of which are fixed by this PR), and while  I can't guarantee this PR won't introduce any new problems, on balance I think it will fix more than it breaks and more importantly, it puts us in a much better position to fix things going forward (the current import *'s in the code hide a lot of issues).

In other words, I think it does meet our standard for merging by making things "less bad".

JohnCremona:
OK!

AndrewVSutherland:
Per https://github.com/LMFDB/lmfdb/pull/1819#issuecomment-233564612, I just opened #1837.

PR #1825, opened 2016-07-18
pyflakes cleanup of hilbert_modular_forms

AndrewVSutherland:
This is a pyflakes cleanup of the hilbert_modular_forms code.  I fixed what I believe is a bug in web_HMF.py (see what is now line 144) -- I presume this code branch never gets hit on the website.

All tests pass and most of the other changes are straight-forward, but @JohnCremona and/or @AurelPage might want to sanity check.

### DIFF
--- a/lmfdb/hilbert_modular_forms/__init__.py
+++ b/lmfdb/hilbert_modular_forms/__init__.py
@@ -12,5 +12,6 @@ def body_class():
     return {'body_class': 'hmf'}
 
 import hilbert_modular_form
+assert hilbert_modular_form
 
 app.register_blueprint(hmf_page, url_prefix="/ModularForm/GL2/TotallyReal")
--- a/lmfdb/hilbert_modular_forms/check_conjugates.py
+++ b/lmfdb/hilbert_modular_forms/check_conjugates.py
@@ -12,13 +12,14 @@ Initial version (University of Warwick 2015) Aurel Page
 
 """
 
+import os
 import sys
 sys.path.append("../..");
-import pymongo
-from lmfdb.website import DEFAULT_DB_PORT as dbport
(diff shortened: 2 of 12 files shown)


AndrewVSutherland:
I noticed that some of the older recompute/import scripts are importing dbport from lmfdb.website, but there is no global variable dbport in lmfdb.website; I changed them to instead do

from lmfdb.website import DEFAULT_DB_PORT as dbport

as is done in the newer recompute_Galois_squarefull_forms.py (for example).

AndrewVSutherland:
I could argue that it would better not to fiddle with dbport at all in the import scripts, just call lmfdb.base.getDBConnection(), which will use the default or whatever you have set up in your mongoclient configuration.  But I'll leave them be for the moment.

AndrewVSutherland:
@JohnCremona: As motivation to review and merge this (when you have time, no rush), I note that after all the cleanup I did today, once this PR and #1819 are merged, pyflakes should run cleanly on the entire LMFDB code base.

JohnCremona:
OK< I will loook at this.  You are right that some of these scripts have not been used for a while and still do things the old way.   Here I have to say that it used to be the case that the db port number (37010 by default) were hard-coded into the code in dozens of separate places, and I was the one who set it in one place to be imported, so that I could experiment with my own db copies; but I did not at that time change anyone else's import scripts.

JohnCremona:
I made some additional changes in a new commit currently waiting as a PR to @AndrewVSutherland 's pyflakes-hmf branch.  If that is merged tehre then I'll merge the updated PR here.

AndrewVSutherland:
@JohnCremona Your additional changes all look good and have been merged into this PR, thanks!

PR #1833, opened 2016-07-19
Handle non-existent elliptic curve isogeny class requests

AndrewVSutherland:
If you attempt to access a URL for an elliptic curve that is not in the database, e.g.

http://www.lmfdb.org/EllipticCurve/2.2.5.1/31.1/b/1

you get an error message explaining that no elliptic curve with this label is in the database.  But if you instead try to access the corresponding isogeny class, i.e.

http://www.lmfdb.org/EllipticCurve/2.2.5.1/31.1/b/

you get an internal server error.  This tiny PR fixes this by displaying an error message when the isogeny class is not found, similar to what is done when the elliptic curve is not found, see:

http://127.0.0.1:37777/EllipticCurve/2.2.5.1/31.1/b/


JohnCremona:
Thanks!

Issue #1832, opened 2016-07-18
remove calls to eval from all lmfdb website code


AndrewVSutherland:
This is a followup to the recently closed issue #1539, which removed all calls to sage_eval.  There are only a few calls to eval left, and most appear in import or verification scripts that are not part of the website.  Running following command in the top level lmfdb directory will list all calls to a naked "eval" function or method (this excludes calls to ast.literal_eval, which are safe):

```
 find | grep "\.py$" | xargs grep  "[^_a-zA-Z]eval[\s]*(" -c | grep -v ":0"

./lmfdb/lfunctions/Lfunction.py:2
./lmfdb/modular_forms/elliptic_modular_forms/backend/plot_dom.py:1
./lmfdb/elliptic_curves/import_ec_lfunction_data.py:1
./lmfdb/hilbert_modular_forms/recompute_AL.py:19
./lmfdb/hilbert_modular_forms/recompute_Galois_squarefull_forms.py:3
./lmfdb/hilbert_modular_forms/recompute_AL_modp.py:29
./lmfdb/hilbert_modular_forms/q_expansion.py:16
./lmfdb/hilbert_modular_forms/import_hmf_data.py:29
./lmfdb/hilbert_modular_forms/import_hmf_data_new_Galois_squarefull_forms.py:27
./lmfdb/ecnf/hmf_check_find.py:16
./lmfdb/transitive_group.py:1
./lmfdb/databases/standard_types.py:2
```

The filenames beginning with "import" and "recompute" are not used in the website code, nor is ecnf/hmf_check_find.py or hilbert_modular_Forms/q_expansion.py (and most of the evals therein are calls to magma.eval).  That leaves only

```
./lmfdb/lfunctions/Lfunction.py:2
./lmfdb/modular_forms/elliptic_modular_forms/backend/plot_dom.py:1
./lmfdb/transitive_group.py:1
./lmfdb/databases/standard_types.py:2
```

A total of 6 occurrences.

JohnCremona:
Your regexp is not perfect!  In *my* file hmf_check_find.py, every instance of "eval" is in "mag.eval" and these are all in a function which is sending magma commands to an inferior process.

Anyway, I strongly agree with the proposal to remove all genuine eval() calls from all "live" code.

AndrewVSutherland:
[...]

PR #1835, opened 2016-07-19
fix eval in Lfunction.py

AndrewVSutherland:
Addresses #1832 for Lfunction.py.

### DIFF
--- a/lmfdb/lfunctions/Lfunction.py
+++ b/lmfdb/lfunctions/Lfunction.py
@@ -530,7 +530,6 @@ class Lfunction_HMF(Lfunction):
         validate_required_args ('Unable to construct Hilbert modular form L-function.', args, 'label')
 
         self._Ltype = "hilbertmodularform"
-
         logger.debug(str(args))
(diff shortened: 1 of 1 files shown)


JohnCremona:
Good.  By the way, I think it is terrible for a function in Lfunction.py to have to worry about extracting the Hecke field from a Hilbert newform by constructing polynomial rings etc.  The class WebHMF (which exists, I started working on it but did not finish) should do all this low-level interface which required knowledge of the exact format of an HMF in the database.  In thisparticular block of code, the L-function wnats to know the sign of the functional equation, and this should be possible via something as simple as setting F = WebHMF(...) and then self.sign = F.sign().

AndrewVSutherland:
@JohnCremona This and many other "terrible" things in Lfunction.py will all go away once all the L-functions are stored in the database (as Dirichlet, elliptic curve, and genus 2 curve L-functions now are).  Until this happens the L-functions code is going to remain a rat's nest.

In my view this should be one of the main goals of Release 1.2.

PR #1838, opened 2016-07-19
fix eval in plot_dom.py

AndrewVSutherland:
Simple fix, only Gamma0, Gamma1, and Gamma are ever used.


JohnCremona:
Looks good, butt will do a test anyway.

JohnCremona:
OK -- when I ran all tests I had a failure from the pari website from test_homepage.py but that is clearly nothing to do with this.

PR #1844, opened 2016-07-19
add pyflakes to requirements.txt and test.sh

AndrewVSutherland:
This PR adds a pyflakes check to test.sh that is run on all the python files in the specified subtree (or the entire codebase if the "coverage" option is used or no directory is specified -- this only takes 2-3 seconds).

To see it in action, from the top level lmfdb directory (where test.sh lives) try

./test.sh lmfdb
./test.sh lmfdb/elliptic_curves (this will only check *.py files in elliptic curves)

This PR also adds pyflakes to requirements.txt where it will get pip installed along with everything else.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -4,6 +4,7 @@ flask-login==0.2.11
 flask-markdown
 pymongo
 pyyaml
+pyflakes
 unittest2
 coverage
 nose
--- a/test.sh
+++ b/test.sh
@@ -35,6 +35,10 @@ else
   WHAT="$@"
 fi
 
+echo "Running pyflakes..."
+find $WHAT | grep "\.py$" | xargs pyflakes
+echo "pyflakes check complete"
+
 ARGS='-v -s --testmatch="(?:^|\/)[Tt]est_"'
 
 SAGE_COMMAND=$SAGE
--- a/test.sh
+++ b/test.sh
@@ -5,11 +5,12 @@
 
 # Note: to run the tests, sage must be in your path.  If necessary do
 # export PATH=$PATH:/path/to/sage
-# To run it, first install or upgrade "nose", "unittest2" and "coverage".
-# e.g. $ pip install --user -U nose coverage unittest2
+# To run it, first install or upgrade "nose", "unittest2" "coverage" and "pyflakes".
+# e.g. $ pip install --user -U nose coverage unittest2 pyflakes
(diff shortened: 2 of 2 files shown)


AndrewVSutherland:
I should have added the most important test of all, which demonstrates why merging this PR is a good idea.

Pick the python source file of your choice in your local lmfdb developement setup and introduce a subtle bug by changing the name of a function you are calling (e.g. change "getDBConnection" to "getDBconnection", or the name of a variable that appears as an rvalue (e.g. change "data['level'] = level" to "data['level'] = lvl" or "dat['level'] = level") and then run test.sh and watch pyflakes catch your mistake.

AndrewVSutherland:
Actually, hold off merging this for the moment, I am going to modify test.sh to automatically install pyflakes if it is not present so that this change will have no impact on those who have not already installed pyflakes.

AndrewVSutherland:
In fact no harm is done if pyflakes is not present, it will simply print a message saying it is not installed and continue as usual, so I don't think it makes sense to have test.sh auto-install it -- instead I just updated the comments in test.sh to add it to the list of packages that need to be installed (along with "nose" "unittest2" and "coverage").

This is ready to merge.  In our current state it will have no impact on devs who have pyflakes installed, and for those that do not they will just see a message saying pyflakes is not installed when they run test.sh but everything else will work as usual.

JohnCremona:
Thanks -- trying out now.  I would have recommended not automatically installing pyflakes anyway -- we have devs working with a system-wide install where they would not have the authority to install additional packages into Sage's python anyway.

JohnCremona:
Great -- I tried one file, one directory, and the whole shebang, pleased to see that the time taken for pyflakes is negligible. 

 With any testing which produces more than a few lines of output though, if there were a line or two of pyflakes errors would they get noticed?  Perhaps we could stop the test right away if the pyflakes  fails -- or is that too draconian?

AndrewVSutherland:
@JohnCremona I think it makes sense to let it go ahead, but I can modify the script to also display a message at the end telling you how many pyflakes errors there were in case you missed them first time around (there is a brief pause while sage starts during which you would get a chance to see pyflakes errors, but I agree they might scroll by if you turn away from the screen for a moment).

JohnCremona:
OK fair enough -- merging now, it can always be improved.

PR #1846, opened 2016-07-19
Avoid server errors on number field download queries

AndrewVSutherland:
This is similar to (but even simpler than) #1792, which was about avoiding server errors on malformed elliptic curve download requests.

If you enter the URL
http://www.lmfdb.org/NumberField/?degree=13&download=1
you get a server error (several of these have shown up in the flasklog).

But the URL
http://www.lmfdb.org/NumberField/?degree=13&download=1&Submit=
works fine.

This PR changes line 597 of number_field.py to use info.get('Submit') rather than info['Submit'] so that you can leave off the Submit and it will still work, e.g.

http://127.0.0.1:37777/NumberField/?degree=13&download=1
http://127.0.0.1:37777/NumberField/?degree=13&download=1&Submit=

both do the same thing (generate a download in the default format).

### DIFF
--- a/lmfdb/number_fields/number_field.py
+++ b/lmfdb/number_fields/number_field.py
@@ -594,7 +594,7 @@ def frobs(nf):
 
 
 def download_search(info, res):
-    dltype = info['Submit']
+    dltype = info.get('Submit')
     delim = 'bracket'
     com = r'\\'  # single line comment start
     com1 = ''  # multiline comment start


JohnCremona:
I just worked out what the difference is -- this way with no 'Submit' key you get None instead of an error.  As far as I can see (not tested just looked at the code) with no 'Submit' field the filename will end in .gp but the contents will not be correct gp?

AndrewVSutherland:
@JohnCremona It will do exactly what it does now when the URL http://www.lmfdb.org/NumberField/?degree=13&download=1&Submit= is entered, which is to default to gp (see download_search in number_field.py).

I just changed the genus 2 curve download to also have a default, but I didn't use gp, I use a generic text-file format (the downloader will presumably be able to figure it out, and they may be downloading it for use by their own code anyway).

JohnCremona:
Well, same as gp except for a semicolon from lines 650-653 ;)

Issue #1533, opened 2016-06-08
Sever error on classical modular form pages


AndrewVSutherland:
If you go to the page

http://www.lmfdb.org/ModularForm/GL2/Q/holomorphic/20/9/

and click on the link http://www.lmfdb.org/ModularForm/GL2/Q/holomorphic/20/9/11 for the Galois orbit of the space with dimension 16 you will see a server error.  Traceback appears below:

> ERROR:lmfdb.base@2016-06-08 11:08:30,431: Exception on /ModularForm/GL2/Q/holomorphic/20/9/11/ [GET] 
> Traceback (most recent call last):
>   File "/home/drew/sage-6.5-x86_64-Linux/local/lib/python2.7/site-packages/Flask-0.10.1-py2.7.egg/flask/app.py", line 1817, in wsgi_app
>     response = self.full_dispatch_request()
>   File "/home/drew/sage-6.5-x86_64-Linux/local/lib/python2.7/site-packages/Flask-0.10.1-py2.7.egg/flask/app.py", line 1477, in full_dispatch_request
>     rv = self.handle_user_exception(e)
>   File "/home/drew/sage-6.5-x86_64-Linux/local/lib/python2.7/site-packages/Flask-0.10.1-py2.7.egg/flask/app.py", line 1381, in handle_user_exception
>     reraise(exc_type, exc_value, tb)
>   File "/home/drew/sage-6.5-x86_64-Linux/local/lib/python2.7/site-packages/Flask-0.10.1-py2.7.egg/flask/app.py", line 1475, in full_dispatch_request
>     rv = self.dispatch_request()
>   File "/home/drew/sage-6.5-x86_64-Linux/local/lib/python2.7/site-packages/Flask-0.10.1-py2.7.egg/flask/app.py", line 1461, in dispatch_request
>     return self.view_functions[rule.endpoint](**req.view_args)
>   File "/home/drew/Dropbox/github/lmfdb/lmfdb/modular_forms/elliptic_modular_forms/views/emf_main.py", line 132, in render_elliptic_modular_forms
>     return render_web_modform_space(*_info)
>   File "/home/drew/Dropbox/github/lmfdb/lmfdb/modular_forms/elliptic_modular_forms/views/emf_render_web_modform_space.py", line 78, in render_web_modform_space
>     return render_template("emf_web_modform_space.html", *_info)
[...]

PR #1849, opened 2016-07-19
add --dbmon flag to support mongo db call monitoring

AndrewVSutherland:
This PR adds a --dbmon=NAME command line option that you can specify when you start the LMFDB webserver to log (with timings) all calls to mongo db on the specified database.  To monitor everything, use --dbmon=* (or even just --dbmon=), and to monitor everything except the specified datbase, put a twiddle in front of it (e.g. --dbmon=~knowledge will not log knowl lookups).

This will be useful both for analyzing performance, and for debugging issues like #1533 and #1803, both of which are caused by problems with invalid or inconsistent data across multiple collections (and it is not always easy to tell from the code which collections are being accessed when).

To see it in action just add --dbmon=* to your usual lmfdb startup and visit some of your favorite pages

sage -python start-lmfdb.py --dbmon=*

To prevent being overwhelmed by messages about knowl lookups use

sage -python start-lmfdb.py --dbmon=~knowledge

### DIFF
--- a/lmfdb/base.py
+++ b/lmfdb/base.py
@@ -37,22 +37,23 @@ _mongo_user = None
 _mongo_pass = None
 _mongo_dbmon = None
 
-# simple event logger that logs all interaction with mongo db
-# to activate set the mongo_kwargs to include event_listeners=[MongoEventLogger()]
-# see website.py --dblistener option where this is set
+# simple event logger that logs commands sent to mongo db
+# it will be used only if --dbmon is specified when the LMFDB server is started
+# this implementation assumes pymongo is begin used synchronously (currently true in the LMFDB, even on the webserver we fork rather than thread)
+# it would need to be modified to support multi-threading (although it would still work when _mongo_dbmon=None)
 from pymongo import monitoring
 class MongoEventLogger(monitoring.CommandListener):
     def __init__(self):
         self._last_request = 0
     def started(self, event):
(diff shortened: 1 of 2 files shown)


AndrewVSutherland:
EDIT: I changed the --dbmon command line option to take an argument that specifies the name of the database to monitor (e.g. --dbmon=MaassWaveForms) and updated the description above accordingly.  

You can specify an empty name to monitor everything (e.g. --dbmon=), but knowls make it a bit overwhelming.

AndrewVSutherland:
One more tweak, you can now use --dbmon=* to monitor everything, and --dbmon=~foo to monitor all databases **except** foo (so --dbmon=~knowledge will avoid logging every knowl lookup).  OK, I will stop fiddling with it now.

edgarcosta:
This is great and really helpful!

This is exactly what I used back in May to analyze why our first version of lmfdb.org in the cloud was so slow
https://github.com/LMFDB/lmfdb/compare/master...edgarcosta:mongostuff

I should have PRed a clean version of it back then, but at the time I only was trying to make everything work.

JohnCremona:
I'll let Edgar finish reviewing this one so I don't have to go into mongo configuration myself.

edgarcosta:
It looks good to me. 

Regarding the coding style, I prefer: `(a, b)` vs `(a,b)`, and parenthesis around logical operators. 
:)

PR #1853, opened 2016-07-20
don't raise error in render_lfunction_exception in debug mode

AndrewVSutherland:
@JohnCremona I commented out the two lines in render_lfunction_exception that raise an error in debug mode -- anyone who wants to use them while testing can always uncomment them.

### DIFF
--- a/lmfdb/lfunctions/main.py
+++ b/lmfdb/lfunctions/main.py
@@ -379,8 +379,8 @@ def l_function_lcalc_page():
 ################################################################################
 
 def render_lfunction_exception(err):
-    if current_app.debug:
-        raise err
+    #if current_app.debug:
+    #    raise err
     if err.args:
         errmsg = "Unable to render L-function page due to the following problem(s):<br><ul>" + reduce(lambda x,y:x+y,["<li>"+msg+"</li>" for msg in err.args]) + "</ul>"
     else:


JohnCremona:
Thanks.

JohnCremona:
BTW, I think that raising exceptions / errors should be very rare in the live website code, since we never want any such error *ever* to work its way through to the top level without being caught in a try/except block.   I find 280 lines in lmfdb/* with "raise" in them but of course this still includes non-live code (e.g. import scripts), and it may well be that some of the errors raised are caught higher up the chain.  I am not sure how to test that apart from watching logs, which we already do (at least, one of us does!).

AndrewVSutherland:
@JohnCremona A lot of our "standard" code, e.g. all the search parsing functions, raise exceptions on errors with the expectation that the caller will catch them.  I think this is a reasonable coding model -- typically one is validating a bunch of user input at once and writing code to check for errors at each step would be tedious.  Another standard place where exceptions get raised are in constructors for classes, although here there is a wide range of implementations -- in some places we return None, in others we return a string with an error message, and in others we raise an exception.  I actually thinking raising an exception is a better choice than returning a string with an error message in it, because in the latter case the caller has to check the type of the return value and if they forget to do this (which happens all the time) an exception is likely to occur later when the code tries to use the object it thinks it created but when it does the error that gets logged to flasklog will be much less helpful than the original one would have been.

JohnCremona:
I do agree, really -- even though I am one of those who does things like return a string instead of a proper object to indicate a problem.  Raising and catching errors is better, but even after 9 years of using python it still does not come naturally to me in some circumstances!
Is there any systematic way of finding all places where functions raise errors, and testing to see if these are being caught by the caller?

AndrewVSutherland:
@JohnCremona BTW, this discussion motivated me to look more closely at how the genus 2 code handles error messages returned by the WebG2C class (which Dan Yasaki originally cloned from you WebEC class).  The code was checking for a returned error message string, but it was doing so using isinstance(...,str), which does not work because the error message is actually a **unicode** string.  I changed WebG2C to raise errors instead (and the caller now uses a try/except).

I will think about how to search the code systematically for similar errors (the comparable situation in elliptic_curve.py is in theory OK, because you check for a specific string, but I'll note that if you ever change the spelling of an error message in WebEC or add  a new one, things will break), and for functions calling code that may raise errors outside of a try/except.

JohnCremona:
I intend to do something similar... sometime!

PR #1855, opened 2016-07-20
Genus2 error handling

AndrewVSutherland:
Fix internal server error on pages like http://www.lmfdb.org/Genus2Curve/Q/123/a/ and http://www.lmfdb.org/Genus2Curve/Q/123/a/456/1 that refer to genus 2 isogeny classes or curves that do not exist (@JohnCremona - FYI, this was caused by code not properly checking the return type of a class constructor that returned an error message as a **unicode string** not a string -- another reason to actually raise errors in this situation, which the code now does).

All genus2 tests pass.  To see the difference compare

http://www.lmfdb.org/Genus2Curve/Q/123/a
http://127.0.0.1:37777/Genus2Curve/Q/123/a

http://www.lmfdb.org/Genus2Curve/Q/123/a/456/1
http://127.0.0.1:37777/Genus2Curve/Q/123/a/456/1

There is also a change to

http://www.lmfdb.org/Genus2Curve/Q/123/a/456/
http://127.0.0.1:37777/Genus2Curve/Q/123/a/456/

The old version looks reasonable at first glance, but in fact the title refers to an isogeny class that does not exist and the bread crumb for the isogeny class will throw a server error if you click on it; the new version returns a 404, which makes more sense in this situation (the URL includes an isogeny class that does not exist).

### DIFF
--- a/lmfdb/genus2_curves/main.py
+++ b/lmfdb/genus2_curves/main.py
@@ -6,7 +6,7 @@ import re
 import time
 from pymongo import ASCENDING, DESCENDING
 from operator import mul
-from flask import render_template, url_for, request, redirect, send_file
+from flask import render_template, url_for, request, redirect, send_file, abort
 from sage.all import ZZ
 
 from lmfdb.utils import to_dict, comma, random_value_from_collection, attribute_value_counts, flash_error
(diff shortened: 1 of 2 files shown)


JohnCremona:
Looks good, still testing.  [I saw that running ./test.sh showed a pyflakes warning 
     ./lmfdb/lfunctions/main.py:25: 'current_app' imported but unused  
but that is presumably from a different PR.]

I noticed that the errors actually raised in the contructors contain some useful (for debugging) test, which is not used here.  It should be possible to catch those messages and pass them through to flask.abort, according to http://stackoverflow.com/questions/21294889/how-to-get-access-to-error-message-from-abort-command-when-using-custom-error-ha  (see the simple solution right at the end) , but if you don't fancy adding that now I will not object.

AndrewVSutherland:
@JohnCremona Nice idea, I like the idea of adding detail to our 404 messages when available, but this will require tweaking the 404 error handler registered in lmfdb/website.py and the 404.html template.

Let's merge this PR and I'll submit a new PR that does that adds that functionality and then update lmfdb/genus2_curves/main.py to use it.

BTW, the current_app import is fixed on master, just not on this branch.

JohnCremona:
OK!

PR #1857, opened 2016-07-21
display detail on 404 errors

AndrewVSutherland:
As suggested in https://github.com/LMFDB/lmfdb/pull/1855#issuecomment-234063144, error detail on 404 errors is now displayed when specified (use flask.abort(404,errmsg)), and this is used in the genus 2 code.

old: http://www.lmfdb.org/Genus2Curve/Q/169/b/
new" http://127.0.0.1:37777/Genus2Curve/Q/169/b/

old: http://www.lmfdb.org/Genus2Curve/Q/169/a/169/2
new" http://127.0.0.1:37777/Genus2Curve/Q/169/a/169/2


AndrewVSutherland:
I'm going to go ahead and merge this (it's a small safe change that only affects bad URL requests in any case) -- I want to use it in a few other places.

JohnCremona:
Fine -- sorry, I did test this today and thought I had done it already.

PR #1860, opened 2016-07-21
Hide *.chunks, *.rand, *.stats, ..., collections by default

AndrewVSutherland:
This PR hides certain collections in the api interface by default, including those that end in ".rand", ".chunks", ".old", and ".new", and those that begin with "test".  You can still access the full list of uncensored collections via /api/all, and the /api/stats page still shows all uncensored collections.

To see the difference, compare:

http://beta.lmfdb.org/api
http://127.0.0.1:37777/api
http://127.0.0.1:37777/api/all

The ".old" and ".new" suffixes are recommended for use when updating a collection using rewrite.py -- naming the new collection blah.new will keep it from being visible while it is being written, and after renaming blah to blah.old and blah.new to blah, the new collection will be visible and the old one will not be visible (but still available in case you want to revert).  This convention is also used when copying collections from Warwick to the cloud.

### DIFF
--- a/lmfdb/api/api.py
+++ b/lmfdb/api/api.py
@@ -71,6 +71,7 @@ def get_database_info(show_hidden=False):
     for db in C.database_names():
         if not censored_db(db):
             info[db] = sorted([(c, C[db][c].count()) for c in C[db].collection_names() if not censored_collection(c) and (show_hidden or not hidden_collection(c))])
+            print db, info[db]
     return info
 
 @api_page.route("/")
--- a/lmfdb/api/api.py
+++ b/lmfdb/api/api.py
@@ -71,7 +71,6 @@ def get_database_info(show_hidden=False):
     for db in C.database_names():
         if not censored_db(db):
             info[db] = sorted([(c, C[db][c].count()) for c in C[db].collection_names() if not censored_collection(c) and (show_hidden or not hidden_collection(c))])
-            print db, info[db]
     return info
 
 @api_page.route("/")
--- a/lmfdb/api/api.py
+++ b/lmfdb/api/api.py
@@ -16,9 +16,6 @@ from lmfdb.api import api_page, api_logger
 from bson import json_util
 from bson.objectid import ObjectId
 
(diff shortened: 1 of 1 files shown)


JohnCremona:
I have spent a few minutes comparing and cannot yet see anything on beta which is not visible with this test branch.  I like the idea, though it does rather rely on people following the rules when creating new collections (sorry about nfcurves2 which I'll rename nfcurves.new soon!)

AndrewVSutherland:
@JohnCremona The .rand files were already hidden, but with the new version, under MaassWaveForms and modularforms2, for example, on http://127.0.0.1:37777/api you should not see any *.chunks files, but on http://beta.lmfdb.org/api and http://127.0.0.1:37777/api/all, you should.

And under genus2_curves, you should not see curves.stats on the new /api page, but you should see it on the old /api page, and on /api/all, moreover, you should also see curves.rand on the new /api/all.

JohnCremona:
On my  http://127.0.0.1:37777/api  I see several .chunks files (2 in MaassFOrms, the rest in modularforms2.  I also see the stats collection in genus2curves!  Very mysterious -- I double-checked that I am running your latest commit.

AndrewVSutherland:
I just did "git checkout 15c956d9fa5994adb9fccfb1c8a0d5585ca90364" and I don't see them.  This is very bizarre.

AndrewVSutherland:
@JohnCremona Does your local copy of lmfdb/api/api.py have the function "hidden_collection" defined in it (at line 62)?

JohnCremona:
It's OK, I now see what I should see.  My fault,  I rely on the webserver automatically restarting when the source code changes as it does when I checkout a new branch , which is normally all that is needed.  But for some changes -- and this is clearly one -- you do actually have to stop the process and start a new one.  I have some efficiencies in place so I can easily review pull requests without interrupting anything else I might be doing (step 1: make a short alias for 'git remote update drew'!) but sometimes a few extra keystrokes are needed.  Sorry!

PR #1867, opened 2016-07-22
Fix #1866

### LINKED ISSUE #1866: Proxy error for some higher genus w/autos download
(possibly related to #1079)

If I go here:
http://beta.lmfdb.org/HigherGenus/C/Aut/10.6-1.0.2-2-2-2-2-2-2-2-2-2.1
and click on the "Download Magma Code" or "Download Gap Code" buttons, I get a proxy error. I suspect the issue is that the download is creating a file with about 3280 generating vectors (and each vector creates about 10 lines of text in the download file).  I can modify the code to not allow downloads quite that big, but I wasn't sure what the cutoff should be, i.e. what "too many generating vectors" means.

A page like this:
http://beta.lmfdb.org/HigherGenus/C/Aut/10.10-1.0.2-2-2-2-2-2-5
downloads just fine (and that would be 1250 generating vectors between the two passports).

AndrewVSutherland:
This PR addresses #1866 with a little python optimization; the problematic download now takes less than a second on a fast internet connection (all the other downloads are also faster).

@jenpaulhus You should test a couple of different code downloads just to make sure everything still works as it should.  I also removed the custom version of parse_bracketed_posints you created and added a "keepbrackets" option and created a parse_gap_id function in search_parsing.py as well (there are several other places in the LMFDB that take a gap id as input).

In order to make it easier for you to test this I am going to go ahead and merge it and push it to beta -- please let me know if you spot any problems.

### DIFF
--- a/lmfdb/higher_genus_w_automorphisms/main.py
+++ b/lmfdb/higher_genus_w_automorphisms/main.py
@@ -11,8 +11,7 @@ import os
 from lmfdb import base
 from flask import render_template, request, url_for, make_response, redirect
 from lmfdb.utils import to_dict, random_value_from_collection, flash_error
-from lmfdb.search_parsing import search_parser, parse_ints, parse_count, parse_start, clean_input
-BRACKETED_POSINT_RE = re.compile(r'^\[\]|\[\d+(,\d+)*\]$')
+from lmfdb.search_parsing import parse_ints, parse_count, parse_start, clean_input, parse_bracketed_posints, parse_gap_id
 
 from sage.all import Permutation
 from lmfdb.higher_genus_w_automorphisms import higher_genus_w_automorphisms_page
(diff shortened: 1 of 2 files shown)


jenpaulhus:
@AndrewVSutherland Thanks for making these changes. The downloads are substantially faster.  Just one small issue I found.  If you enter gibberish in the signature field, it returns a server error instead of a message about what should be in the signature field.  See for example:
http://beta.lmfdb.org/HigherGenus/C/Aut/?genus=&group=&signature=junk&dim=&inc_hyper=include&count=20&Submit=Search

The other search fields all work.  And other than that, everything seems to be working well.

AndrewVSutherland:
@jenpaulhus OK I just fixed this (it wasn't caused by any of the changes I made, it was an uncaught error in ast_literal_eval before the try/except parsing in your search function).

PR #1869, opened 2016-07-22
Move artin representation code

jwj61:
Some code for artin representations was under lmfdb/lmfdb.  This moves it into lmfdb/lmfdb/artin_representations and updates import statements.  Should be straightforward unless I used "mv" instead of "git mv" somewhere.

To test, you can use
 * http://127.0.0.1:37777/L/ArtinRepresentation/2.2e2_3_11e2.12t12.1c1/
 * http://127.0.0.1:37777/ArtinRepresentation/2.2e2_3_11e2.12t12.1c1
 * http://127.0.0.1:37777/NumberField/12.2.53246005216107264.1
which hit the three places artin code is used.


AndrewVSutherland:
@jwj61 I can take a look at this in a bit, I just need to get my testing config setup (I usually use atkin, but it still appears to be down).

jwj61:
No rush

AndrewVSutherland:
@jwj61 It looks like you need to fix a few imports of lmfdb.math_classes (the critical one is the one at line 494 of WebNumberField.py which is causing failures in test_lfunctions.py, the import at line 479 of WebNumberField.py is good).

```
./lmfdb/WebNumberField.py:494:
./lmfdb/import_scripts/artin_representations_Dokchitser/inject_conductor_len_artin_representation.py:8:
./lmfdb/import_scripts/artin_representations_Dokchitser/link_number_field_to_number_field_galois_group.py:16:
./lmfdb/import_scripts/artin_representations/extract_art.py:64:
./lmfdb/tensor_products/galois_reps.py:20:
./lmfdb/artin_representations/example_artin_representation.py:1:

```

jwj61:
Thanks for catching that.  It is now fixed.

AndrewVSutherland:
OK all tests pass (except for the usual problems with classical modular forms).  There are still one or two unused imports (e.g. in tensor_products) that should be fixed just to avoid possible confusion in the future, but I'll merge this and just fix those directly.

PR #1876, opened 2016-07-23
Ecnf extra data

AndrewVSutherland:


AndrewVSutherland:
Need to fix j-invariant lookup -- this PR will be replaced by a new PR from my repo that includes all these commits and a j-invariant fix.

PR #1879, opened 2016-07-23
Data format change for elliptic curves over number fields

AndrewVSutherland:
This PR is a replacement for #1876.  It includes three additional fixes:

- j-invariant lookups now work (as do twists links)
- urls for highlight curves with new labels have changed
- added redirect for URLs that use the old label format for elliptic curves over imaginary quadratic fields to avoid breaking links in the literature.


AndrewVSutherland:
One issue I noticed while testing this is that the very first time you do anything that involves an ideal over the field 6.6.371293.1 it is extremely slow (takes about 2 minutes just to do web_latex(N) where N is the conductor).  This has nothing to do with this PR and also affects the number field pages, e.g. try 127.0.0.1:/37777/NumberField/6.6.371293.1.

This appears to be a bug in Sage and/or Pari.  If you ran Sage and enter

` NumberField(QQ[x]([-1,-3,6,4,-5,-1,1]),'a').ideal([1,1])`

you can reproduce the problem.  If you ctrl-c you will see that it is stuck in bnfinit inside Pari.  If you then just enter the same command again it will complete instantly.

This is related to #1358 (but not the complete cause of it).

AndrewVSutherland:
I haven't found any more problems; I am going to go ahead and merge this and push it to beta where it will be easier to do further testing.

PR #1894, opened 2016-07-29
Higher Genus added some equations and cyclic trigonal

jenpaulhus:
Commits:

(1) I missed one case when I tested #1867 for the Magma downloads.  A tiny error with a missing '+'

(2) The data now includes a test of whether the curve(s) are cyclic trigonal.  This now shows up on refined passport pages of full automorphism like:
http://127.0.0.1:37777/HigherGenus/C/Aut/3.9-1.0.3-9-9.1
Also added a search Inc/Exc/Only for cyclic trigonal:
Also changed the download files to include this information.

(3) I am updating the data with known equations (particularly hyperelliptic curves) so I have added a spot for this on the refined passport pages under "Other Data", if the data includes it.  See for example:
http://127.0.0.1:37777/HigherGenus/C/Aut/5.48-30.0.3-4-4.1

(4) Added a search Inc/Exc/Only for whether the automorphism group is the full automorphism for that family. Again see:


AndrewVSutherland:
Looks good!

PR #1901, opened 2016-08-03
make Higher Genus w/autos data available on beta sidebar

jenpaulhus:
This pull is to request that the Higher Genus data go from alpha to beta testing.  It adds a link on the sidebar under "Higher Genus" 
http://127.0.0.1:37777/HigherGenus/C/Aut

Also, in higher_genus_w_automorphisms/_init_.py  I removed the second link to the old URL for this page (lower case "a" in Aut).

Most of my code is here: https://github.com/jenpaulhus/group-actions-RS
A preprint of what I've done so far is here: http://www.math.grinnell.edu/~paulhusj/Paulhus-lmfdb.pdf

### DIFF
--- a/lmfdb/higher_genus_w_automorphisms/__init__.py
+++ b/lmfdb/higher_genus_w_automorphisms/__init__.py
@@ -17,5 +17,5 @@ import main
 assert main # silence pyflakes
 
 app.register_blueprint(higher_genus_w_automorphisms_page, url_prefix="/HigherGenus/C/Aut")
-app.register_blueprint(higher_genus_w_automorphisms_page, url_prefix="/HigherGenus/C/aut")
+
 
--- a/lmfdb/sidebar.yaml
+++ b/lmfdb/sidebar.yaml
@@ -144,9 +144,10 @@
            status: future
          - title: "Higher genus:"
            part3:
-             - title: w/ automorphisms
+             - title: Families
                url_for:  "higher_genus_w_automorphisms.index"
-           colspan: onecolumn
+               status: beta
+               colspan: onecolumn
      - title: 2-d
        style: rotate
        status: future_rotated
--- a/lmfdb/sidebar.yaml
+++ b/lmfdb/sidebar.yaml
@@ -144,7 +144,7 @@
            status: future
          - title: "Higher genus:"
            part3:
-             - title: Group Actions
+             - title: Families
                url_for:  "higher_genus_w_automorphisms.index"
                status: beta
                colspan: onecolumn
--- a/lmfdb/sidebar.yaml
+++ b/lmfdb/sidebar.yaml
@@ -144,7 +144,7 @@
            status: future
          - title: "Higher genus:"
            part3:
-             - title: Families
+             - title: Group Actions
                url_for:  "higher_genus_w_automorphisms.index"
                status: beta
                colspan: onecolumn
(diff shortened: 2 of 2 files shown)


JohnCremona:
Jen, you need to put back the "status:" line changing "future" to "beta" as otherwise the new entry appears even when not running in beta mode.
I did not much like the abbreviation "w/" and tried changing to "with" .  In my browser it appears as "with" on one line and "automorphisms" on the next, which is OK for now.

I would be happy to merge this after making the one change to status -- then we can get more people to look at these pages!

AndrewVSutherland:
What if we just replaced "w/automorphisms" with "families"?  I realize this is less informative, but it would fit better and still be accurate.

On a separate note, I have reviewed much of the work Jen has done on this and I think it looks good and is ready for beta status.

JohnCremona:
I'm fine with families.

jenpaulhus:
I've now added the status: beta line to sidebar.yaml  and changed the link name from "w/ automorphisms" to "Families".  I wasn't a fan of "w/autos" either, but "families" does feel a little vague.  Some other possible ideas could be "automorphisms" only  or "group actions".

jenpaulhus:
Here it is with "Group Actions" instead. The sidebar renders a little strange locally for me so I can't tell if it will fit. It should, its exactly the same length as "/NumberFields" which fits.

davidfarmer:
"Families" may seem vague to the experts, but I like it because
it conveys the general idea, and those who care will understand
it well enough.   Titles don't have to be precise mathematical
statements.

AndrewVSutherland:
On my system "Group Actions" splits onto two lines; in view of @davidfarmer's comment I suggest  we just stick with "Families".

JohnCremona:
Agreed (it's on 2 lines for me too).

jenpaulhus:
Ok, sounds good. Do you need me to make the change back and push the branch again or are you easily able to merge from one commit back before I changed it to Group Actions?
4cf046bc0416bde5b5428e8ad803376f124668f7

JohnCremona:
I'm not sure how to do that on GitHub -- probably easier for you to revert the last change.

jenpaulhus:
Last commit should be back to "Families".

JohnCremona:
OK, looks good -- @AndrewVSutherland will you do the honours?

edgarcosta:
There are two issues with formatting.
"Higher Genus" is still under future, and we are also missing colspan above "Higher Genus", that messes up the css for :hover, over "Families":
screen shot 2016-08-04 at 14 12 20
screen shot 2016-08-04 at 14 11 47

Now going offtopic, why are we marking the previously visited links on the sidebar?
This is only noticeable on the ones that do not involve mathjax.
screen shot 2016-08-04 at 14 15 37



Cheers,
Edgar

AndrewVSutherland:
We want Higher Genus under future, otherwise it will show up on www.lmfdb.org with nothing below it.

edgarcosta:
Sounds good to me.

JohnCremona:
Surely having Higher Genus under beta would also work?

AndrewVSutherland:
Oops, I meant beta -- I just fixed this and the colspan issue Edgar noticed.

AndrewVSutherland:
@edgarcosta And I agree we shouldn't be marking previously visited links on the side bar, I'll see if I can tweak this.

AndrewVSutherland:
@edgarcosta The colors you are seeing are not previously visited links, there are links that have beta status, and I think it makes sense to keep them highlighted.

edgarcosta:
Ups... Sorry for the trouble

PR #1908, opened 2016-08-05
Show arithmetically equivalent field data in Galois group pages

jwj61:
This is a small feature addition, which is the first step towards showing arithmetically equivalent fields and other siblings on number field pages (which was demo-ed at AIM in May).  For that, we want to know when the siblings exist even where we haven't computed them (yet) so we can display a "data not computed message" in those cases.  Most of the information was there already in the transitive groups database.

This goes with new data in the transitivegroups.groups database to say how many arithmetically equivalent fields a number field will have for each Galois group.  (Note for degrees above 15, this does not seem to be easily available on the web, but was easy to compute with magma).

The only thing visible to the user now is on individual Galois group pages, depending on whether there would be 0, 1, or more arithmetically equivalent fields:

http://127.0.0.1:37777/GaloisGroup/16t98
http://127.0.0.1:37777/GaloisGroup/4t5
http://127.0.0.1:37777/GaloisGroup/7t5

The message is in the "Other representations" section since this is a special type of other representation.

### DIFF
--- a/lmfdb/galois_groups/main.py
+++ b/lmfdb/galois_groups/main.py
@@ -235,11 +235,11 @@ def render_group_webpage(args):
         ae = wgg.arith_equivalent()
         if ae>0:
             if ae>1:
-                data['arith_equiv'] = r'A number field with this Galois group has %d arithmetically equivalent fields'% ae
+                data['arith_equiv'] = r'A number field with this Galois group has %d arithmetically equivalent fields.'% ae
             else:
-                data['arith_equiv'] = r'A number field with this Galois group has exactly one arithmetically equivalent field'
+                data['arith_equiv'] = r'A number field with this Galois group has exactly one arithmetically equivalent field.'
         else:
-            data['arith_equiv'] = r'A number field with this Galois group has no arithmetically equivalent fields'
+            data['arith_equiv'] = r'A number field with this Galois group has no arithmetically equivalent fields.'
         if len(data['otherreps']) == 0:
             data['otherreps']="There is no other low degree representation."
         query={'galois': bson.SON([('n', n), ('t', t)])}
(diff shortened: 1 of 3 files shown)


AndrewVSutherland:
This is an awesome addition!  I do have two comments/suggestions:

- It would be good to put a knowl on "arithmetically equivalent" (which in addition to defining arithmetic equivalence would presumably also make clear that when you count arithmetically equivalent fields you are excluding the field itself).
- Probably not for this PR, but in the future it would be great to have arithmetically equivalent fields listed in the related objects of the home page of the fields involved (and of course they should all show up on the L-function page once we have them available for such fields (degree needs to be at least 7 if I recall correctly)).

jwj61:
You are right about the knowl -- I will do that.

The listing as related objects for the fields is already in that branch.  It is not (yet) there for the L-functions.  That will probably wait until we decide what to do with displaying dedekind zeta-functions which are not in the database since, yes, degree 7 is the first example.

AndrewVSutherland:
@jwj61 I'm confused -- if I go to http://127.0.0.1:37777/NumberField/7.3.2007889.1, it's Galois group 7T5 indicates that there should be an arithmetically equivalent field, and presumably http://127.0.0.1:37777/NumberField/7.3.2007889.2  is that field, but it is not listed as related object of http://127.0.0.1:37777/NumberField/7.3.2007889.1.

jwj61:
I see the confusion.  I have a branch which shows arithmetically equivalent fields and other siblings.  It is a different branch, and what was I was refering to as "that branch", but I can see how it would be hard for you to get all of that from the few words I wrote.

That other branch still needs messages about missing data programed in.  Just having the data in the database is connected to this pull request.  So, showing ae fields and other siblings is coming soon.

jwj61:
Made the words arithmetically equivalent a knowl.  The knowl now clarifies that a field is not arithmetically equivalent to itself

PR #1910, opened 2016-08-14
master -> beta

AndrewVSutherland:


jwj61:
Could you copy numberfields.fields (and numberfields.fields.rand) to the cloud.  It provides more examples or siblings, and my new favorites, the 16t98 fields.

AndrewVSutherland:
@jwj61 Done.  Can you take a look and let me know if everything appears to be OK (there was a slight hiccup when reloading the collection).

jwj61:
The number field database may be ok.  I am seeing some glitches, but I think that is probably that  transitivegroups.groups also needs to be updated.

AndrewVSutherland:
So should I copy over transitivegroups.groups now?

jwj61:
Yes, please.  (I am guessing that this is causing the difference between beta and the main site.)

AndrewVSutherland2:
@jwj61 Done.  Let me know if there are still any differences.

jwj61:
Looks good.

Issue #1915, opened 2016-08-30
good euler factors on isogeny class page of a genus 2 curve


edgarcosta:
Hello,

We display the good and bad euler factors on the L-function webpage:
e.g.: http://www.lmfdb.org/L/Genus2Curve/Q/284023/a/
However, the webpage for the isogeny class we only displays the bad factors:
e.g.: http://www.lmfdb.org/Genus2Curve/Q/284023/a/

Cheers,
Edgar

JohnCremona:
Also note that the web page of an elliptic curve isogeny class (over Q) had nothing about Euler factors at all, where it certailny could show something like the genus 2 isogeny class page does.

AndrewVSutherland:
Given that we already show the bad factors, it seems reasonable to show the good ones as well (which are in the Lfunctions database).  I can do this.

And I think it might be nice to do the same for elliptic curves (both over Q and number fields).

davidfarmer:
I think it would be odd to show the good local factors on the
page for an isogeny class, since it is the page for an
isogeny class, not the page for an L-function.

I might also think it is odd to show the bad factors, but I
am not an expert in that area and I would not disagree with
the experts deciding to include that information.
(And it would be fine with me to also include the
good factors, but I think that should be left to the experts
in charge of that page.)

AndrewVSutherland:
@davidfarmer I think someone interested in the arithmetic properties of a genus 2 curve might very well like to see the local factors at small good primes, which gives you the Frobenius charpolys -- this is analogous to showing the Frobenius cycle types at small unramifed primes on global number field pages (which we do).

AndrewVSutherland:
@davidfarmer Just to follow up on this, I spoke to Edgar and this request actually originated with a student (a non-expert user) who was interested in knowing the Frobenius charpolys at some small primes and did not know that these could be found at the L-function page of the curve -- I think it does make sense to list a few of these on the isogeny class page (and direct the user to the L-function page for more details).  This is done in PR #1916.

PR #1916, opened 2016-08-30
Addresses issue #1915

AndrewVSutherland:
This addresses issue 1915 by adding a list of local factors at good primes up to 30 to the isogeny class page for genus 2 curves.  To see the difference compare:

http://www.lmfdb.org/Genus2Curve/Q/565504/b/
http://127.0.0.1:37777/Genus2Curve/Q/565504/b/


JohnCremona:
Testing now.

PR #1917, opened 2016-08-31
fix 2 deprecation warnings from flask_markdown and flask_login

JohnCremona:
This fixes 2 issues identified in #1332 .

One is changing a function (called in lmfdb/users initialization) from flask-login since its name has changed from setup_app to init_app.

The other is to make a flask-markdown extension load cleanly.  For this one, I guess that the relevant lines 65-67 in lmfdb/knowledge/main.py can be deleted, since they seem to be about something which no longer exists (the URL http://wiki.l-functions.org seems to go back years to when this whole site was at l-functions.org).  But I left them in.

### DIFF
--- a/lmfdb/knowledge/main.py
+++ b/lmfdb/knowledge/main.py
@@ -63,7 +63,7 @@ class KnowlTagPatternWithTitle(markdown.inlinepatterns.Pattern):
         return "{{ KNOWL('%s') }}" % kid
 
 # Initialise the markdown converter, sending a wikilink [[topic]] to the L-functions wiki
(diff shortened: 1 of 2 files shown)


JohnCremona:
For the record, here are the versions of the various Flask modules I used when testing this:

Flask (0.10.1)
Flask-AutoIndex (0.6)
Flask-Babel (0.9)
Flask-Cache (0.13.1)
Flask-Login (0.2.11)
Flask-Markdown (0.3)
Flask-OldSessions (0.10)
Flask-OpenID (1.2.5)
Flask-Silk (0.2)

PR #1932, opened 2016-10-06
added list of citations

JohnCremona:
Using 2 html files created from the repository LMFDB/citations the URL /citation now gives a new short page linking to both the old version which now has URL /citation/citing and also to the new page at /citation/citations which is a (recently updated) list of over 100 citations.  The latter page also has links to (and back from) a separate bibtex file.

This is intended as a proof of concept.  No doubt people will have improvements to suggest.

### DIFF
--- a/lmfdb/pages.py
+++ b/lmfdb/pages.py
@@ -238,9 +238,26 @@ def management_board():
 
 @app.route("/citation")
 def citation():
-    t = "How to cite LMFDB"
+    t = "Citing the LMFDB"
     b = [(t, url_for("citation"))]
+    return render_template('citation.html', title=t, body_class='', bread=b)
+
+@app.route("/citation/citing")
+def citing():
+    t = "How to cite LMFDB"
+    b = [("Citing the LMFDB", url_for("citation")), (t, url_for("citing"))]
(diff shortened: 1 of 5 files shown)


AndrewVSutherland:
Looks good, the only minor thing I noticed is that the reference numbers (e.g. [17]) look like they are hyperlinks (i.e. they are green and highlight when you mouse over them) but they are not, which is slightly confusing.

JohnCremona:
True.  An example of the html is
[<a name="anni2014local">1</a>]
and that is there for the reverse link from the bibtex file.  That is put in by bibtex2html and I have not yet found a way to change it.  If bibtex2html cannot do it we could write a script to edit the file in the Makefile in the citations repository (which already runs both thml files through sed).

AndrewVSutherland:
@JohnCremona I don't have time to fiddle with this right now, but do you want me to go ahead and merge  the PR as is (I don't see any harm in this).  Then anyone who feels motivated to fix this small issue can do so.

edgarcosta:
I don't think this issue should hold up the merge. I would merge it and
create a new issue.

JohnCremona:
Thanks -- I though that David F would have a view?

Issue #1937, opened 2016-10-21
Sage 7.4


JohnCremona:
Since Sage 7.4 was just released I tried our test suite with it and found some issues which will need to be dealt with at some point:

 test_Lfunction_link (lmfdb.genus2_curves.test_genus2_curves.Genus2Test) ... ERROR
 test_Lec (lmfdb.lfunctions.test_lfunctions.LfunctionTest) ... ERROR
 test_Lgenus2 (lmfdb.lfunctions.test_lfunctions.LfunctionTest) ... ERROR
 Check that the links in Box 2 work. ... ERROR
 Check that the links in the /L/degree2/ table work. ... ERROR

The above all arise from

  File "/home/jec/lmfdb/lmfdb/genus2_curves/web_g2c.py", line 157, in list_to_factored_poly_otherorder
    this_t_number = this_gal.group()._pari_()[2]._sage_()
AttributeError: 'sage.libs.pari.gen.gen' object has no attribute '_sage_'

There is still this the old one

Testing that we link to Galois conjugate spaces ... FAIL

which has nothing to do with Sage verions.

AndrewVSutherland:
I can take a look at the problem in list_to_factored_poly_otherorder today -- something must have changed in the sage/pari interface.

JohnCremona:
Thanks but I fixed it already -- see the PR referenced here.  It was trivial.

AndrewVSutherland:
Nervermind then, thanks!

JohnCremona:
You can still help by testing that the patched code works fine with 7.3.  Note that the default sage on atkin & lehner is now 7.4, but 7.3 is still there:  /usr/local/sage/sage-2/sage is 7.3 and the test script can be made to use it via
SAGE=/usr/local/sage/sage-2/sage ./test.sh
which is what I already did.

AndrewVSutherland:
Fixed by PR #1938

JohnCremona:
Thanks.  It would be better if Sage added more member functoins so that one did not have to use the ._pari_() method and manually extract compnents and convert them, but that's not our job here.

PR #1938, opened 2016-10-21
tiny change to keep Sage-7.4 happy

JohnCremona:
This fixes Issue #1937 -- some tiny change between Sage 7.3 and 7.4 means that two underscores had to be deleted (in converting from a pari integer to a Sage integer).  I tested on both 7.3 and 7.4, the important thing to do before merging this is to check that all really is well with 7.3 since that's what all our sites currently run.


AndrewVSutherland:
Reviewing now

PR #1940, opened 2016-10-26
merge master to beta

JohnCremona:
I think it is time to do this -- main visible change will be the citation pages.  OK?


AndrewVSutherland:
Agreed, it's a bad idea to leave things unpushed for too long.

JohnCremona:
OK, I'll merge this one.  prod too?

AndrewVSutherland:
I suggest we let it run for a day on beta first, just in case.

edgarcosta:
> I suggest we let it run for a day on beta first, just in case.

I agree

Issue #1941, opened 2016-10-28
SSL module not available


JRSijsling:
Hello,

Recently I got an actual desktop for use in my office. I installed a Linux distribution on it (elementary 0.4, based on Ubuntu). One of the first things I did was to try and install Sage after installing all needed dependencies. To my amazement, that actually worked without any hitch whatsoever. The problems arose when I tried to go further so as to be able to work on the LMFDB.

Developing the LMFDB has a number of prerequisites. Among them is extending Sage by

sage -pip install --upgrade pip
sage -pip install --upgrade pymongo

This fails. I get

jrsijsling@JRS-ThinkCentre-M800z:~/Programs/sage-7.4$ ./sage -pip install --upgrade pip
pip is configured with locations that require TLS/SSL, however the ssl module in Python is not available.
Could not fetch URL https://pypi.python.org/simple/pip/: There was a problem confirming the ssl certificate: Can't connect to HTTPS URL because the SSL module is not available. - skipping
Requirement already up-to-date: pip in ./local/lib/python2.7/site-packages
jrsijsling@JRS-ThinkCentre-M800z:~/Programs/sage-7.4$ ./sage -pip install --upgrade pymongo
pip is configured with locations that require TLS/SSL, however the ssl module in Python is not available.
Collecting pymongo
  Could not fetch URL https://pypi.python.org/simple/pymongo/: There was a problem confirming the ssl certificate: Can't connect to HTTPS URL because the SSL module is not available. - skipping
  Could not find a version that satisfies the requirement pymongo (from versions: )
No matching distribution found for pymongo

I do have OpenSSL, and the notebook version of Sage works for me. And my version of Python 2.7 did in fact have the ssl module. So I do not know what causes this. The following attempts at a solution have failed:

* Fully upgrading Python (pip, pymongo, setuptools) and rebuilding Sage;
* Using Sage 7.3 instead;
* Running ./sage -i openssl and trying again;
* Using the Sage shell.

Running Python (2.7.10, which should have ssl installed internally) inside the Sage shell gives

(sage-sh) jrsijsling@JRS-ThinkCentre-M800z:sage-7.4$ python
Python 2.7.10 (default, Oct 28 2016, 09:09:23) 
[GCC 5.4.0 20160609] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> import ssl
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
[...]

PR #1942, opened 2016-10-31
Addition to GettingStarted.md on SSL dependency

JRSijsling:
On some systems the installation process in GettingStarted.md is not sufficient due to issues with the SSL module in Python. This addition describes what to do then.

This fixes Issue #1941 (already closed). See also https://groups.google.com/forum/#!topic/sage-support/c2fm64j55Jk, but the solution there is a suboptimal hybrid of the ones in the Sage installation guide to which this addition points.

### DIFF
--- a/GettingStarted.md
+++ b/GettingStarted.md
@@ -25,9 +25,18 @@ Installation
       sage -i database_gap
       sage -i pip
       sage -b
-      # in the `lmfdb/` directory:
+      # in the 'lmfdb/' directory:
       sage -pip install -r requirements.txt
    ```
+  * [SSL dependency] On some systems the final line above may fail with an
+    error message stating that the SSL module in Python was not found. If this
+    happens, then follow one of the three possible solutions in [Sage's
+    installation
+    guide](http://doc.sagemath.org/html/en/installation/source.html#building-the-notebook-with-ssl-support).
+    Depending on your chosen solution, Sage may thereafter claim upon shutdown
+    that the service_identity package is still missing; this can be installed
+    with
+    * `sage -pip install --upgrade service_identity`
   * [optional] Memcache.  *This step is not at all necessary and can
     safely be ignored!* Memcache speeds up recompilation of python
     modules during development.  Using it requires both installing the
--- a/GettingStarted.md
+++ b/GettingStarted.md
@@ -11,9 +11,17 @@ Installation
 
   and follow these instructions.
 
-* Make sure you have sage (>=7.0) installed and that
-  `sage` is available from the commandline.  In particular see
+* Make sure you have Sage (>=7.0) installed and that `sage` is available from
+  the commandline.  In particular see
   [Sage installation](http://doc.sagemath.org/html/en/installation/source.html).
+  Also check that your version of Sage has ssl available by checking that
+  `import ssl` works on its command line. If not, then the `pip install`
+  commands below will fail. To remedy this, either install SSL globally on
+  your system or have Sage build its own local version, as mentioned
+  [here](http://doc.sagemath.org/html/en/installation/source.html#notebook-additional-features)
+  and
+  [here](http://doc.sagemath.org/html/en/installation/source.html#building-the-notebook-with-ssl-support),
+  respectively.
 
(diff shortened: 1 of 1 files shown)


JohnCremona:
The problem I have with this is that I have no way to check if it works (without building a new Sage from scratch).  Perhaps someone should do that?

edgarcosta:
1- Jeroen, did you build sage from scratch or did you use pre-built version
of SAGE?


2- John, if we phrase it differently and put it as a possible solution
under troubleshooting, then we don't need to test that scenario.

JRSijsling:
Yes, I see your point! I can only vouch for the fact that it worked on my machine. So all that I can say is that there was at least one case where it was needed. It definitely belong to troubleshooting and I can move the remark there.

I encountered this when building from source. Tomorrow at the university I will download a binary and see if there the issue is also relevant.

edgarcosta:
Jeroen, then I would add that detail, that the issue showed up when you
built SAGE from source.
Further, I also believe that the crux of the problem is that you didn't
have the SSL development libraries and/or header files when you built SAGE
on your system aka "libssl-dev"

For example, these are "my" pre requisites for building SAGE in the cloud:
binutils gcc g++ gfortran make m4 perl tar git libssl-dev

For more details, see:
https://github.com/edgarcosta/lmfdb-gce/blob/master/server_scripts/prerequisites_sage.sh
https://github.com/edgarcosta/lmfdb-gce/blob/master/server_scripts/install_sage.sh

Cheers,
Edgar

JRSijsling:
That is exactly it Edgar. And I will tell you why this failed before. At

http://doc.sagemath.org/html/en/installation/source.html#notebook-additional-features

There is a line that says how to install the additional notebook features. I did once try to run that... but what happened is that

sudo apt-get install libssl libssl-dev

failed because on my machine the first of those is not available for some reason. And then the second was also aborted. I did not know about that feature of apt-get. When I just put the second package, then everything goes without a hitch.

So I know what to put as a canonical one-line solution. Confirm that it should go to the Troubleshooting folder and I will update the pull request.

JohnCremona:
I think we want to avoid adding to our Getting Started guide solutions to all possible Sage installation problems, since that will never end. We do list additional python modules which need to be installed (e.g. pymongo) and that already requires the user to have write access to the Sage build they are using.  But the ssl stuff is at a higher (or perhaps lower) level.  Sage itself only needs that for running its old (and becoming obsolete) notebook server;  and Sage was not able to ship ssl as standard because of licensing issues.  The Sage installation guide already has information about this:  simplest is to have the right ssl packages installed system-wide (which obviously depends on exactly what system you have);  next is to do the right thing when first configuring Sage before building (a step which is usually unnecessary); next is to post-install the ssl package within Sage and then force a rebuild of python and everything which depends on it (which is quite a lot).

For our own instructions it would (?) be sufficient to say to make sure that the Sage build you are using has ssl available (I think that testing that "import ssl" works at the Sage command line but I am not sure of that), with a cross-reference to the Sage manuals for what to do if not, and a comment to the effect that none of the pip install commands will work otherwise.

This problem may go away soon since I think that something changed concerning ssl licensing so that at some point (possibly already) Sage can ship its own ssl as standard, so be installed at build time if (and only if) there is no system-wide version.

JRSijsling:
That solution seems good. I remember that "import ssl" did signal the problem that I had. Shall I therefore add a line or two at "Make sure you have Sage [...]"?

edgarcosta:
> For our own instructions it would (?) be sufficient to say to make sure that the Sage build you are using has ssl available (I think that testing that "import ssl" works at the Sage command line but I am not sure of that), with a cross-reference to the Sage manuals for what to do if not, and a comment to the effect that none of the pip install commands will work otherwise

Yes! that sounds good to me.

JohnCremona:
Me too.  There is yet another post on sage-support today saying that after sage -i pip, actually using pip is impossible when openssh is not available.   So I think there will be a Sage solution before long; meanwhile let's do this to help lmfdb developers.

edgarcosta:
looks good to me.

Issue #1946, opened 2016-11-24
Bug in parsing user input


jwj61:
There seems to be a problem with parsing the input field for discriminants of number fields.  It shows up if the range spans 0 (e.g., -10-1 for -10 to 1).

AndrewVSutherland:
Fixed in #1947

JohnCremona:
Pushed to beta, can be closed in 11 mins if it works OK at beta.lmfdb.org (or do we keep issues open until pushed to master?)

jwj61:
I think it could have been closed once fixed.

PR #1947, opened 2016-11-24
Fix issue #1946

AndrewVSutherland:


JohnCremona:
I'll test this now.

JohnCremona:
I can confirm that entering -10-1 into the Discriminant range box at /NumberField/ does work (and entering -10..1 also works, converting into the first form).
Good that this is handled in search_parsing so it will apply in other contexts -- though I scould not think of another context where a range of integers spanning 0 would be valid!

PR #1952, opened 2016-12-13
Fixed bug in displaying p-adic regulators in rank 0 case.

rpollack9974:
For rank 0 curves, instead of displaying that the p-adic regulators were 1 it said something else.


rpollack9974:
Here's a URL which now works:

http://localhost:37777/EllipticCurve/Q/11/a/2

See the bottom of the page in the p-adic data section.

JohnCremona:
Looks good to me.

PR #1960, opened 2017-01-17
Toomanyzeros

edgarcosta:
Replaces the ugly message when someone wants to download too many zeros with a knowl.

I have also messed up the creation of the knowl. Could someone delete this knowl for me: http://www.lmfdb.org/knowledge/show/dq.zeros.toomany ?

### DIFF
--- a/lmfdb/zeros/zeta/zetazeros.py
+++ b/lmfdb/zeros/zeta/zetazeros.py
@@ -5,7 +5,7 @@ from flask import render_template, request, url_for
 ZetaZeros = flask.Blueprint("zeta zeros", __name__, template_folder="templates")
 logger = lmfdb.utils.make_logger(ZetaZeros)
 
-from platt_zeros import *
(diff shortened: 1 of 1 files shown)


jwj61:
Before I delete it I want to confirm, the knowl dq.zeros.toomany is not referenced anywhere, right?

edgarcosta:
Yes, it doesn't show up anywhere.

I was the one who created it today, and messed up the id.

jwj61:
Knowl deleted (on beta)

JohnCremona:
Thanks JJ.  I tried to do something cleaner about this download page a while back but got bogged down in details and never finished.  I cannot test right now but will later in the week if no-one else does.

jwj61:
I was testing this.  When I click on the link it provides, http://warwick.lmfdb.org/data/zeros/zeta/ I get that the address cannot be reached.  I realize this is inside the knowl, so can be changed independently of the pull request, but we should probably get it fixed before pulling.

edgarcosta:
fixed.

Issue #1950, opened 2016-12-04
Root number mismatch?


jwj61:
From the feedback page:

Dear LMFDB, I am confused about your entry http://www.lmfdb.org/L/ModularForm/GL2/Q/holomorphic/1/18/1/a/0/. It concerns the L-function of a holomorphic Hecke cusp form f of weight k=18 and level 1. You state that the root number is 1 and the central value L(1/2,f) is positive. However, as the level is 1, the root number should be i^k=-1, and so L(1/2,f) should vanish. Am I missing something? Please let me know by email: coolissues@freemail.hu Thank you in advance!

JohnCremona:
Is this the same as #1941?

davidfarmer:
It is the same as #1914

JohnCremona:
That's what I meant to type.  Then I suggest that we can close this one, provided that we add a note to #1914 including the email address of the person reporting this one.

davidfarmer:
I think this problem just comes from line 477 of lmfdb/lfunctions/Lfunction.py
which should say
     self.sign = (-1)**(self.weight/2)
instead of always equaling 1.

JohnCremona:
I agree and will approve a code change to that effect!  I think it should also work to remove the spacial case for level 1 since the general code a few lines below should handle this case, though that relies on self.AL[self.level] returning 1 which it might not do (I guess self.AL would be an empty list for level 1). 

I actually think that the WebNewForm class should have, and set, a sign field in its constructor.  But for now fixing this in Lfunction.py is simplest.

davidfarmer:
Has been fixed, and I replied to the person who brought it to our attention.

Issue #1914, opened 2016-08-18
Root number of classical modular form error


jwj61:
From the feedback page:

On the page http://www.lmfdb.org/L/ModularForm/GL2/Q/holomorphic/1/18/1/a/0/
there may be an error on this page (or maybe I'm just misunderstanding notation).  This form has weight $k \geq 2 \bmod 4$, so the sign in the functional equation is $(-1){k/2} = -1$, so $\epsilon$ should equal $-1$ and the central value $L(1/2, f)$ should be zero.

Please send feedback (like that it is being worked on, if it has been resolved, etc.) to martin.dickson34@gmail.com

jvoight:
Definitely an issue here.  Magma confirms:

> S := CuspidalSubspace(ModularForms(1, 18));
> f := Newforms(S)[1][1];
> Lf := LSeries(f);
> CentralValue(Lf);
0.000000000000000000000000000000
> LTaylor(Lf, 9, 1);
0.649604896356279588838547646141*$.1 + O($.1^2)

davidfarmer:
Has been fixed, and I replied to the person who brought it to our attention.

PR #1969, opened 2017-02-15
correct sign of level 1 modular form L-functions

davidfarmer:
This fixes issues #1950 and #1914 .

Once this fix is live, then I will email the people who noticed the problem and then
close the issues.

Once we read these L-functions from the database, all this code will be deleted.

### DIFF
--- a/lmfdb/lfunctions/Lfunction.py
+++ b/lmfdb/lfunctions/Lfunction.py
@@ -473,8 +473,8 @@ class Lfunction_EMF(Lfunction):
                 float(n ** self.automorphyexp))
 
         # Determining the sign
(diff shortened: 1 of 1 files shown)


JohnCremona:
Looks good to me, testing now.  I note that  when we have L-functions of forms of odd weight then this will need changing again (division by 2 of an odd python int will round down) but I guess that a lot of other things will too.

davidfarmer:
This is for level 1 only, so the weight has to be even.

And all this is just a mess that will be deleted once the
L-function data is take from the database.  That is really soon, right?

JohnCremona:
I have no idea who if anyone is precomputing L-function data for classival modular forms.  not me!

Issue #1970, opened 2017-02-15
Cache issues


edgarcosta:
Hello,

It looks like our generation of a random object is somewhat cached on the browser.
Is there a way for us to disable this for our users?

For example, if you open multiple new tabs with the link http://www.lmfdb.org/Genus2Curve/Q/random
then you always get the same curve.

I was able to recreate this throughout multiple browsers and in the respective private/incognito modes.
I was only able to avoid this issue by disabling the cache in Chrome/Firefox DevTools.

I also observed the same behavior in other classes: Modular Forms/Siegel, Elliptic curves/Numberfields, NF/Local. Artin representations, Galois Groups and Sato-Tate groups.

Cheers,
Edgar

AndrewVSutherland:
This issue is specific to the cloud, it does not happen on http://beta.lmfdb.org/Genus2Curve/Q/Random.

edgarcosta:
I agree.

But there is something underlying in all the classes that I mentioned.

For example:
http://www.lmfdb.org/EllipticCurve/Q/random
works fine.

AndrewVSutherland:
The only difference I can see is that in the elliptic curves case the function random_object_from_collection is used, but in the other cases the function random_value_from_collection is used (both are implemented in https://github.com/LMFDB/lmfdb/blob/master/lmfdb/utils.py).  But looking at the implementations of these functions I see no obvious reason why they would produce different caching behavior, they are basically the same except that the latter should in theory be slightly more efficient because it is reducing the amount of data being returned by mongo db.  Compare lines 41 and 56 to see the difference.

edgarcosta:
It is indeed a cloud issue, but nothing to do with the load balancer, as accessing directly the servers doesn't change anything: http://www-central0.lmfdb.xyz/Genus2Curve/Q/random

I will look into it.

edgarcosta:
I think I figured it out. 
All the problematic redirects use the code 301 to redirect to the object, instead of 302.

See http://flask.pocoo.org/docs/0.12/api/#flask.redirect and https://en.wikipedia.org/wiki/List_of_HTTP_status_codes#3xx_Redirection


```
$ for f in */*.py; do  egrep 'route\(".*/random"' -A10 -B1 $f -H | egrep 'redirect'  -B10 -A1  --color ; done
artin_representations/main.py-
artin_representations/main.py:@artin_representations_page.route("/random")
artin_representations/main.py-def random_representation():
[...]

PR #1973, opened 2017-02-17
301 -> 307

edgarcosta:
I tried to address #1970.

Went over the code searching for permanent redirects, i.e., code 301,  that perhaps should be temporary redirects.
On those, I switched the code from 301 to 307.


AndrewVSutherland:
Looks good.  Of course the real test will come when we push to the cloud.

PR #1979, opened 2017-02-28
update import script for ec lfunction data

JohnCremona:
The raw data format provided by Andy Booker changed between the first upload in March 2016 for conductors up to 380000 and the next batch for 380000-400000.  This patch updates the import script accordingly.


JohnCremona:
Oops! By mistake I added some innocuous script chages to a branch which already had some more substantial ones (displaying if Iwasawa invariants for elliptic curves) which I intended to be separate.  I will close this and reopen when I have fixed that.

Issue #1985, opened 2017-03-10
missing isogeny class data for elliptic curves over Q


JohnCremona:
The collection of elliptic curves over Q is supposed to include for each curve the "isogeny matrix" of degrees which are used to populate pages such as http://beta.lmfdb.org/EllipticCurve/Q/300027/l/.  However I just discovered that only half of the isogeny classes have this stored; this is invisible on the website since the code recomputes when it is missing.

I am currently filling in the missing data, which will speed up the display of isogeny class pages for conductors over 230000.  When that is done on the beta database, I will copy it to the cloud.  At the same time there will be some minor code changes, since when I adapted the code to not have to do any computation on the fly I did so systematically for individual curves but not for isogeny classes.

JohnCremona:
The newcurves collection (in the elliptic_curves databse) is a now copy of curves with the isogeny_matrix fields filled in for every curve (though it is the same for each curve in the class), together with a new field isogeny_degrees giving a sorted list of the distinct degrees of cyclic isogenies.

There will soon be a new branch which uses this, and no longer uses the EllipticCurve() constructor in displaying an isogeny class, saving time.

The new isogeny_degrees field will allow searching for elliptic curves with a given isogeny degree, as requested by Elkies.  These degrees should probably then be listed on an individual curve's web page.

JohnCremona:
The branch isogs at https://github.com/JohnCremona/lmfdb uses the newcurves collection and eliminates all use of the EllipticCurve constructor for isogeny classes over Q.  It also shows isogeny degrees on an elliptic curve's home page.

Next steps after testing that branch:

1. Merge that branch to master and beta, not prod;
2. Copy newcurves collection to cloud;
3. Merge branch into prod.

Then in due course the newcurves collection should be renamed curves and the code changed again to change the name newcurves --> curves (in one place).

edgarcosta:
Sounds good to me.

JohnCremona:
The new collection is in fact called curves.new (now).

JohnCremona:
There was a glitch in permuting the isogeny matrices when uploading from the source data (which uses Cremona labels) .  For an example, see class 14a.  I am fixing this; and the revised new collection will not have a new name, i.e. the previous collection curves will be renamed 'curves.old' while the new one will be just 'curves'.  Since the old code works fine with the new data there was no need for a new collection name.

JohnCremona:
This can be closed once #1986 is merged.  The new data is already on the beta database.

[...]

PR #1986, opened 2017-03-14
Isogeny classes for elliptic curves over Q

JohnCremona:
See issue #1985 

The missing data has been added by rewriting the collection 'curves' to 'newcurves'; at the same time, for each curve, a new data field called 'isogeny_degree' was added for every curve.

The code changes here 
1. change the collection name from curves to newcurves (temporary);
2. avoid all elliptic curve construction in creating an instance of class ECisog_class, which happens every time an isogeny class is displayed;
3. adjusted one test of an isogeny class page
4. Show on each e.c. home page a list of the isogeny degrees (if any)
4. fix an unrelated test in hmfs 
5. add an irrelevant file by mistake -- I will update this PR in a minute.

I have not yet implemented searching for curves by specifying a desired isogeny degrees; that will be done in a future PR.


JohnCremona:
NB It is safe to merge to master and beta but *not* to prod until the new collection has been copied to the cloud.

JohnCremona:
Please hold off testing on this: I am recreating the curves.new collection and will be updating the branch.

JohnCremona:
OK to review again.  Ignore commit 0fe0fe9, I started something new then commented it out.
The rewritten data is still in a collection called new since the old code should work fine on it.

There was a bug (the permutation used to convert between Cremona and LMFDB numbers was inverted in  the code which displays isogeny classes).  I also found some more data gliitches and fixed them, also providing a new script function which checked that every one of the 2.5 million curve recordsin the database has the correct fields and that these have the correct types.

JohnCremona:
Currently doing final testing after recreating the collection so that the isogeny matrices are stored in lmfdb order (as they used to be for N<230000) with no need to reorder on the fly, and so that the old code for isogeny classes displays properly with the new collection, for ease of transition.

Yet another new version of the isogs branch up soon.

JohnCremona:
OK, good for testing again.

AndrewVSutherland:
Looks good.

PR #1987, opened 2017-03-15
Show Galois representations for elliptic curves over number fields

JohnCremona:
A new collection nfcurves2 has been created, a copy of nfcurves but with information about images of mod-p Galois representations, as computed by @AndrewVSutherland .  This new code includes upload scripts for that and displays the new data on individual curve web pages.  (Later we can add additional search possibilities.)
In the CM case and for fields which intersect the p'th roots of unity nontrivially, we only list the finitely many primes p whose image is not "maximal", where the definition of maximal is essentially (as large as it can be given the field of definition and the endomorphism ring".  This should be defined by a suitable knowl.

This PR is for initial testing.  I am expecting to make adjustments before it is merged (which will also require the new collection to be copied to the cloud).  Here are some useful test cases for the new output block:

- random curve, no CM and no non-maximal primes:

http://localhost:37777/EllipticCurve/2.2.8.1/3346.4/g/1

- random curve, no CM but some (2) non-maximal primes:

http://localhost:37777/EllipticCurve/2.2.12.1/4356.3/j/8

- random curve, no CM but one non-maximal prime:

http://localhost:37777/EllipticCurve/2.2.5.1/1116.2/b/2

- CM curve, not rational CM, no non-maximal primes:

http://localhost:37777/EllipticCurve/2.0.11.1/6075.297.9/a/1

- CM curve, not rational CM, one non-maximal prime which is ramified:

http://localhost:37777/EllipticCurve/2.2.76.1/25.2/c/1
http://localhost:37777/EllipticCurve/2.0.11.1/6400.256.16/CMa/1


JohnCremona:
The new collection has been renamed curves.new (following standard practice) with associated code change.

JohnCremona:
The new collection has been renamed nfcurves.new to follow standard practice, and code updated.

AndrewVSutherland:
I think the pages look pretty good!  I notice that in page for CM curves that have at least one non-maximal image, like

    http://localhost:37777/EllipticCurve/2.2.237.1/16.1/a/3

you give a nice description of the maximal images (including at primes dividing the CM discriminant).  But at CM curves where all the images are maximal you do not -- is that intentional?

BTW, I wrote a "maximal images" knowl http://beta.lmfdb.org/knowledge/show/ec.maximal_galois_rep that you can link to (this means a small code change).

JohnCremona:
No, it's a bug -- I meant to add the text about images for all CM curves.  I will fix that and the knowls (and also rename the new collection again since it is unnecessary to use a new collection name as the revised collection will work fine wwth old code).

JohnCremona:
I also found a but that if a curve has no Galois image data yet it reports maximality at all primes since the code does not distinguish between non-existence of a field and existence with value [].  This occurs because I have uploaded curves since @AndrewVSutherland computed the data.  It will be fixed.

JohnCremona:
I have fixed the issues raised so far and invite a new check.  Here is a new checklist of curves to test (there are a few more possibilities than before).   One more thing to add, perhaps, is new knowls for Borel, split/non-split Cartan and their normalizers.  The existing knowls for these terms are in ec.q.* but could be copied to ec.* (and checked to make sure that they are correct when p=2!).

AndrewVSutherland:
I think it makes sense to have just one knowl for each of the various maximal subgroups of GL(2,Fp), which ultimately belongs in the "Subgroups of GL(2,Fp)" section of the LMFDB (I have all the data for p < 1000, I just need to get it loaded in to the LMFDB and write a search/browse interface -- on my list of things to do after genus 3 curves).

I am happy to either create such knowls now, in which case I think both the ec and ecnf code should be changed to point to them, or to leave them in ec.q.* for the moment and edit the content there as required (including correctly handling p=2) with the intention of moving them at a later date.

JohnCremona:
If you have time to create them now we can finish this off.  Tell me what they will be called and I can add links as you write.  Or if you don't have time just merge this as is (assuming that my template does now display what it should).

AndrewVSutherland:
OK, I'm working on it now.

JohnCremona:
That is great.  I will update the code in this PR to use these.

JohnCremona:
The knowls have been updated.  Don't delete the ec.q ones yet since I think there are links to them from elsewhere, e.g. from ec/Q pages.

AndrewVSutherland:
Looks good, I think this is ready to merge.

PR #1990, opened 2017-03-17
Iwasawa invariants for elliptic curves over Q

JohnCremona:
This displays information about Iwasawa lambda- and mu-invariants for elliptic curves over Q, where available.  So far the data exists on prod for conductors up to 90k and on beta up to 100k.  Joint work by me and Robert Pollack.

### DIFF
--- a/lmfdb/elliptic_curves/elliptic_curve.py
+++ b/lmfdb/elliptic_curves/elliptic_curve.py
@@ -439,6 +439,8 @@ def render_curve_webpage_by_label(label):
 
     if data.twoadic_label:
         credit = credit.replace(' and',',') + ' and Jeremy Rouse'
+    if data.data['iwdata']:
+        credit = credit.replace(' and',',') + ' and Robert Pollack'
     data.modform_display = url_for(".modular_form_display", label=lmfdb_label, number="")
 
     code = data.code()
--- a/lmfdb/elliptic_curves/templates/ec-curve.html
+++ b/lmfdb/elliptic_curves/templates/ec-curve.html
@@ -420,7 +420,9 @@ representation')}} is surjective for all primes \( p \).
 
 
 
-    <h2> p-adic data </h2>
+    <h2> $p$-adic data </h2>
+
+<h3>$p$-adic regulators</h3>
 
     {{ place_code('padicreg') }}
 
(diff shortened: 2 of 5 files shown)


AndrewVSutherland:
Reviewing now.

AndrewVSutherland:
I think it looks good.  One question: for some of the curves with rank 0 you say that the invariants are 0 for all p >= q for some prime q, but in other cases of rank 0 you don't and list values at all p <= 50 and primes of bad reduction.  Is there a reason for this discrepancy?

One comment: in the knowl for Iwasawa invariants I think it would be good to say a bit more about the ss reduction case where there are two p-adic L functions (corresponding to the two roots of x^2+p), maybe with a link to Pollack's Duke paper.

I'm happy to merge this, but I'll give you a chance to respond before I do.

JohnCremona:
1. Yes -- @rpollack9974  Pollack's code can sometimes show that all are 0 past some point, but not always.
2. Rob wrote the knowls so this should be done by him.
I'm not sure why putting @rpollack9974 did not do the usual magic so I'll try to get Rob listed as a member or whatever.

rpollack9974:
For (1), a rank 0 curve has non-zero Iwasawa invariants only for primes p such that either:
(a) L(E,1)/Omega_E has non-zero p-adic valuation;
(b) p is a split multiplicative prime
(c) a_p=1 (for p a good prime)

The first two conditions involve only finitely many primes while the last one is finite if E possesses a non-trivial isogeny of degree prime to the conductor of E.  

When (c) is known to be a finite set, then the code finds this prime q so that all invariants vanish after q.  Otherwise, it just computes for p < 50.  

Is this discussion worth including somewhere?

rpollack9974:
For (2), I'm happy to update the knowls.  But let me first ask, are there other knowls that link to papers?

AndrewVSutherland:
@rpollack9974 I do think this information is worth including; in general we want to keep knowls short and to the point, but information that clarifies the completeness/scope of the data is especially important (one can always look elsewhere for more details on the mathematical background, but not for information about exactly what has been computed).

There are many knowls that link to papers, see http://beta.lmfdb.org/knowledge/show/g2c.st_group, for an example.  You can create a link using \cite{MR:nnnnnn} or \cite{arXiv:nnnn.nnnn}.

rpollack9974:
Thanks!  Just finished updating the knowl.

PR #1996, opened 2017-03-21
Galois representations for elliptic curves over Q

JohnCremona:
@AndrewVSutherland has recomputed the mod-p Galois image data for elliptic curves over Q, changing the data for CM curves where the important distinction to make is between "maximal" and "non-maximal" image.  We have uploaded this, into a collection called (right now) curves.new.
The revised code is currently set to read from curves.new, but also works fine if the collection name is changed to curves in line 61 of web_ec.py.  To test this, please test as it is (which will show the new style) and also with that line changes back (to show that all is still well with the old collection).

Assuming that all looks good, I will revise the PR to read from curves, the this can be merged and pushed to beta and prod, where it will show the old style;  then we can rename curves.new --> curves and copy to the cloud, after which the beta and production websites will show the new data.  A final step will be to clean out the conditional code.

If this looks complicated it is because so in order to never break anything while making (fairly minor) change.

### DIFF
--- a/lmfdb/ecnf/main.py
+++ b/lmfdb/ecnf/main.py
@@ -114,7 +114,7 @@ def get_nf_info(lab):
     return label, pretty
 
 
-ecnf_credit = "John Cremona, Alyson Deines, Steve Donelly, Paul Gunnells, Warren Moore, Haluk Sengun, John Voight, Dan Yasaki"
+ecnf_credit = "John Cremona, Alyson Deines, Steve Donelly, Paul Gunnells, Warren Moore, Haluk Sengun, Andrew V Sutherland, John Voight, Dan Yasaki"
 
 
 def get_bread(*breads):
--- a/lmfdb/elliptic_curves/elliptic_curve.py
+++ b/lmfdb/elliptic_curves/elliptic_curve.py
@@ -95,6 +95,9 @@ def rational_elliptic_curves(err_args=None):
     credit = 'John Cremona and Andrew Sutherland'
     t = 'Elliptic curves over $\Q$'
     bread = [('Elliptic Curves', url_for("ecnf.index")), ('$\Q$', ' ')]
+    info['galois_data_type'] = 'old'
+    if 'non-maximal_primes' in db_ec().find_one():
+        info['galois_data_type'] = 'new'
     return render_template("ec-index.html", info=info, credit=credit, title=t, bread=bread, learnmore=learnmore_list_remove('Completeness'), **err_args)
 
 @ec_page.route("/random")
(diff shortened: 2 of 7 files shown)


AndrewVSutherland:
Hi John,

I'm tied up this afternoon hosting the BC-MIT number theory seminar (it 
should be good, Tim Browning and Roger Heath-Brown are both speaking), 
so it will likely be late tonight or tomorrow before I have a chance to 
look at this.

Drew

JohnCremona:
That's fine.  And this one takes precedence over #1990 which will probably need a manual merge as both change the same template,

AndrewVSutherland:
@JohnCremona The search for non-maximal primes doesn't behave as I would expect.  For example, there are curves with non-maximal images at 41 (e.g. http://127.0.0.1:37777/EllipticCurve/Q/15129/d/2) but searching on this gives no results.  And searching for curves with non-maximal image at 37 includes many CM curves whose image is maximal at 37 (e.g. http://127.0.0.1:37777/EllipticCurve/Q/27/a/1).

The style on the display pages looks good to me, but I think we will want to fix the browsing before this gets merged.

JohnCremona:
Thanks for spotting that.  I immediately know the problem which is that the ket used in the search has not been updated from non-surjective_primes to non-maximal_primes.  I'll fix it.

JohnCremona:
OK, I corrected that and updated the branch, **NB** I also switched the collection name in this branch from curves.new to curves since this is the one we need to merge before copying the revised data to the cloud, so this was necessary anyway.  In order to test the *new* data and behaviour you'll need to change that back to curves.new (in line 61 of web_ec.py).

AndrewVSutherland:
I think it looks good now!  It can even find my new favorite elliptic curve with maximal image at every prime except 193: http://127.0.0.1:37777/EllipticCurve/Q/335241/d/2.  I also note that it now correctly handles searches that have maximal image at primes greater than 37 (in the code we are currently running if you ask for curves with surjective image at 41 all the CM curves get included in the results, which is wrong even if you take the view that surjective=maximal for CM curves, there are 12 CM curves in the database that have non-maximal image at 41.

One small thing I noticed -- in the credits there needs to be a space after "John Cremona" in "John Cermonaand...." on search results pages where both our names are listed (and on a related topic, the credits for the ecnf curves should probably be updated now that they are showing Galois image data).

JohnCremona:
I added the missin space and added @AndrewVSutherland to the list of names for credit on ecnf pages.

AndrewVSutherland:
OK, I think this is ready to merge.  Are you happy for me to go ahead and do so?

Once I do, I think we then want to:

1) push to beta, make sure everything looks good, then push to prod
2) on beta move curves to curves.old and curves.new to curves, check that all is well
3) copy curves from beta to prod (on the cloud)

Sound good?

JohnCremona:
Sounds perfect to me.  Thanks.

PR #1998, opened 2017-03-23
Fixed list display in title

sanni85:
Display a list with dots only if there are 4 or more elements.

ArtinRepresentation/5.20627.6t16.1c1
ArtinRepresentation/2.31.3t2.1c1


JohnCremona:
I'll review this.

PR #2004, opened 2017-04-02
Determinants of artin reps

jwj61:
Fix rest of issue 1994, giving the determinant of an Artin rep as an Artin rep.

A decent test is

  http://beta.lmfdb.org/ArtinRepresentation/3.3e6_7e3.9t6.1c1

since it has to pick out the determinant from a Galois conjugate pair (change the c1 at the end to c2 to see the conjugate degree 3 representation, which has a conjugate determinant rep'n (so also the end switches from c1 to c2).

### DIFF
--- a/lmfdb/artin_representations/main.py
+++ b/lmfdb/artin_representations/main.py
@@ -173,14 +173,13 @@ def render_artin_representation_webpage(label):
         friends.append(("Artin Field", nf_url))
     cc = the_rep.central_character()
     if cc is not None:
-        if cc.modulus <= 100000: 
-            if the_rep.dimension()==1:
-                if cc.order == 2:
-                    cc_name = cc.symbol
-                else:
-                    cc_name = cc.texname
-                friends.append(("Dirichlet character "+cc_name, url_for("characters.render_Dirichletwebpage", modulus=cc.modulus, number=cc.number)))
-        if the_rep.dimension()>1:
+        if the_rep.dimension()==1:
+            if cc.order == 2:
+                cc_name = cc.symbol
+            else:
+                cc_name = cc.texname
+            friends.append(("Dirichlet character "+cc_name, url_for("characters.render_Dirichletwebpage", modulus=cc.modulus, number=cc.number)))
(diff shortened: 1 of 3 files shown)


pjbruin:
The example page looks good to me.  However, trying to display the following Artin representation gives an `IndexError`: http://localhost:37777/ArtinRepresentation/2.3_37.8t6.1c1

jwj61:
This should be fixed.  

The issue was that the minimal field containing character values of the original representation was bigger than the corresponding field for the determinant character.  This is related to the fact that we only store faithful Artin representations.

JohnCremona:
@jwj61 if @pjbruin gives you a posiive review feel free to merge.

pjbruin:
Thanks.  One more question: if a one-dimensional Artin representation has conductor <= 10^5, then the corresponding Dirichlet character is displayed both under "Basic invariants" and in the "Related objects" box (e.g. http://localhost:37777/ArtinRepresentation/1.2647.2t1.1c1), whereas if the conductor is > 10^5, then the corresponding Dirichlet character is displayed under "Basic invariants" but not in the "Related objects" box (e.g. http://localhost:37777/ArtinRepresentation/1.2e2_166603.2t1.1c1).  Wouldn't it be more consistent to always display it in both places?

It seems that checking for impractically large conductors is already done in the method `ArtinRepresentation.central_character()`. Would it make sense to simply remove the check `if cc.modulus <= 10000` in the function `render_artin_representation_webpage` in `artin_representations/main.py`?

jwj61:
Sure, I just changed that.

The rationale was that there are two on-the-fly computations here, identifying the character (which factors integers on the order of the modulus) and displaying the web page for the character.  It used to be that displaying a character could be slow for a large modulus.  So, we were avoiding a link to a page which wouldn't load.  Now that has been fixed (only some data is displayed for large moduli).  Using the two links you provide and following the friends links, you can see that there is a difference in between the Dirichelet character web pages.

pjbruin:
Perfect, thanks.

PR #2013, opened 2017-04-20
add search by isogeny degree

JohnCremona:
As requested by Elkies, this adds searching for elliptic curves over Q by isogeny degree.
I had added the additional data already (each curve now stores a list of the degrees of cyclic isogenies it admits, including 1) and also the template code but it was commented out.  This removes the commenting and adds the necessary item to the search query.
Check the main search page EllipticCurve/Q  -- you can put a single integer in the box, or a list or a range, as the standard parsing function is used.  The search results page has the same value preloaded as it should.  Searching for (for example) 3.5 will give all curves with either a 3-isogeny or a 5-isogeny; if you want those with both, use your own brain and search for 15!
Behind the scenes: I added a new index on the isogeny_degrees field, and I edited the knowl so that degrees are defined.

### DIFF
--- a/lmfdb/elliptic_curves/elliptic_curve.py
+++ b/lmfdb/elliptic_curves/elliptic_curve.py
@@ -235,6 +235,8 @@ def elliptic_curve_search(info):
             elif info['include_cm'] == 'only':
                 query['cm'] = {'$ne' : 0}
 
+        parse_ints(info,query,field='isodeg',qfield='isogeny_degrees')
+
         if info['galois_data_type'] == 'new':
             parse_primes(info, query, 'surj_primes', name='surjective primes',
                          qfield='non-maximal_primes', mode='complement')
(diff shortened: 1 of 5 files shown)


AndrewVSutherland:
Wonderful!  I will review this today.  How hard would it be to extend to elliptic curves over number fields (which is the case Elkies is most interested in)?

JohnCremona:
@AndrewVSutherland Not hard at all.   Just a one-off addition of the data field isogeny_degrees for each curve obtained from the isogeny_matrix (which is only there for curve with 'number' 1), then copying a few lines of code from elliptic_curves, and rememberiing to add the additional field when new curves are uploaded.  I'll get on to it now.

AndrewVSutherland:
I think it looks good!  Shall I merge or do you want me to wait for the number field update (which I understand will also require copying new data to the cloud)?

JohnCremona:
Please do this one now.  I have a separate branch for the number field version anyway -- and yes, that (but not this!)  will involve copying the nfcurves collection before the code running on prod is updated.

PR #2014, opened 2017-04-21
added search by isogeny degree for ecnf and added isogeny info on home pages

JohnCremona:
NB requires data now in beta db not on cloud


AndrewVSutherland:
A few comments:

!) One cosmetic suggestion: I would put the "isogeny degree" search box on the main search page in the second column under "torsion structure" -- it currently runs very far to the right and is not aligned with the items in the third column above it.

2) For CM curves with the CM defined over the ground field there are going to be cyclic isogenies of infinitely many degrees.  Currently this happens only in cases where the class number is 1, so all these isogenies are from the curve to itself (i.e. endomorphisms), but it is confusing to say "This curve has no rational isogenies." on pages like

    http://127.0.0.1:37777/EllipticCurve/2.0.11.1/961.703.1/CMa/1

where it is obvious from the Galois images listed immediately above that in fact the curve does admit rational isogenies (infinitely many in fact).  This will get even more confusing once we have curves with CM by an order of class number > 1 with the CM defined over the ground field.

Other than that I think it looks good!

Shall I go ahead and copy the new nfcurves collection to the cloud? (neither of my comments above are related to the data).

JohnCremona:
Excellent points.  For 1 I did a little space adjustment, but not enough.  On the search results page I split the boxes into 2 lines which looks better.  There will be more such boxes in future, I'm sure.
For 2. I'll do something simple now since it's only relevant for the 5 imaginary quadratic fields for which we have curves and as you say they all have class number 1.  It will get more complicated when we have curves defined over larger Hilbert class fields of IQFs but that will probably not be soon.

I think you might as well copy that data.  I discovered a little too late that when I did the same thing over Q I added 2 extra data fields for each curve: the number of curves in the class (trivial to compute as the length of the isogeny_degrees list, but having it stored would allow searches for "isogeny classes with more than x curves") and the maximum degree of a cyclic isogeny in the class (again easy to compute, as the largest entry in the isogeny_matrix, but again one might want to search on this, and also the isogeny_matrix is only stored for curve #1 in each class which unfortunately meant I had to put in an additional db search whenever displaying the home page of a curve which is not #1).   
That can be added later -- I don't have time now and do not want to delay this going live.  So if you start to copy the data, I will update the PR shortly.

JohnCremona:
OK I did those things.  For an example with rational CM which does have an isogeny see http://localhost:37779/EllipticCurve/2.0.7.1/16.10.1/CMa/1
In this case I decided that the wording "...consists of 2 curves linked by isogenies of degrees dividing 2." was not ideal, so   I put in a test so that when the max degree is a prime it says "...of degree..."
And then did that over Q too (e.g. http://localhost:37779/EllipticCurve/Q/1225/b/1)

AndrewVSutherland:
The data has been copied over, I'll review this shortly

AndrewVSutherland:
Looks good, merging.  BTW, I updated nfcurves.rand, which was out of date, on both beta and the cloud.

JohnCremona:
Thanks for doing the rand -- I was under the impression that the rewrite handled that but I don't know how it could.

AndrewVSutherland:
Rewriting a collection doesn't change the number of objects or their identifiers, so the rand index collection doesn't need to be updated.

It's only when you add or delete objects that updating the rand index collection becomes relevant.  If only a small number of records have been added, its not urgent to update it (the only downside is the new records will never get picked when a random object is selected).  But if a lot of records have been added (or if any have been deleted) it makes sense to update it using the create_random_object_index function in https://github.com/LMFDB/lmfdb/blob/master/data_mgt/utilities/rewrite.py.

While writing this I realized that the get_random_object functions don't handle failures that could be caused by deleting objects and not updating the rand index.  I will change the code in utils.py to catch any such errors and handle them (in which case there would be no urgent need to update the rand collection if only a few objects have been deleted).

JohnCremona:
Thanks for the explanation.  One thing I have been doing with the nfcurves collection as I slowly put more curve in is: if I only have a few curves for one field and am adding more (possibly to complete the set of curves for that field) I delete the ones in there already and upload the whole lot (for this one number field).  This is all for sextic fields only, and for ones where we do not yet have Galois image data.
I can change this policy if you think it better, when adding some more curves, but I have found that then I rather easily end up with duplicates (because of human error).
For example, I have just now finished finding curves over 6.6.1541581.1 where we already had 83 isogeny classes and I have the remaining 20 to add.

AndrewVSutherland:
I like your approach, incremental updates are always riskier.  I will go modify the get_random_object functions to handle any errors that might result from object ids changing (or objects being deleted) so you don't have to worry about updating the rand index when you make small changes.

Issue #962, opened 2016-03-31
smallest elliptic curve with given Mordell-Weil group


AurelPage:
From /roadmap: Add page with elliptic curves of smallest known conductor and Mordell-Weil group of the form F x Z^r for the 15 possible finite groups F and r = 0,1,2,...

JohnCremona:
One can already enter a specific rank and torsion structure into the search box: for example, r=2 and torsion [2,4]  gives 59 matches.  So what more could we do?  The most useful thing I can thonk of is to replace the free-format search box for torsion_structure with a drop-down menu of the 15 possibilities.

I am not strongly in favour of having special pages for lists of records, but am willing to be convinced (especially of someone else does the work).

JohnCremona:
[rank-torsion-data.txt](https://github.com/LMFDB/lmfdb/files/10353247/rank-torsion-data.txt)

Attached is a file of the smallest conductor for each (rank, trosion group) possibility for the current database, which has 47 different groups.  It would be possible for someone to add (or extend) bottom knowls for each curve in the list giving this information.  In cases where there is more than one candidate, the one listed is random.

PR #2023, opened 2017-04-23
ec/Q torsion structure search box now a drop-down list

JohnCremona:
See Issue #962 .
This changes the torsion structure search box for elliptic curves over Q to a drop-down list instead of free-format.
Doing the same for ecnf would be possible though it would mean being careful to update whenever a new torsion structure is present.  Right now there are 

sage: Set([tuple(e['torsion_structure']) for e in nfcurves.find()])
{(21,), (2, 18), (2, 8), (9,), (11,), (2, 12), (13,), (15,), (3, 3), (4, 4), (3,), (5,), (16,), (3, 6), (2, 2), (7,), (18,), (20,), (2, 16), (2, 6), (22,), (8,), (2, 10), (10,), (37,), (12,), (2, 14), (14,), (25,), (27,), (2,), (4,), (), (6,), (17,), (19,), (2, 4)}

### DIFF
--- a/lmfdb/elliptic_curves/templates/ec-curve.html
+++ b/lmfdb/elliptic_curves/templates/ec-curve.html
@@ -594,7 +594,7 @@ This curve has non-trivial cyclic isogenies of degree \(d\) for \(d=\)
 Its isogeny class <a href={{ data.class_url }}>{{data.lmfdb_iso}}</a>
 consists of {{data.class_size}} curves linked by isogenies of
 {% if data.one_deg %}degree{% else %}degrees dividing{% endif %}
-{{data.class_deg}}.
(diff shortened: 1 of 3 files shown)


AndrewVSutherland:
Looks good.

JohnCremona:
Thanks.
I was hoping for a cleverer way to list the options on the search results page than to have 15 separate if then else blocks.

JohnCremona:
Apologies @AndrewVSutherland after you reviewed this so promptly.  I'm about to make a new PR which deals with this in a very much better way: using a single 14-line python function I can reduce the template code to one line in each file.  And this can also be used (suitable modified) for ecnf.

PR #2024, opened 2017-04-24
refactor ec torsion structure search code to be much cleaner

JohnCremona:
An update to #2023.  Same functionality, better coded.


JohnCremona:
Hold off merging this since I am now adding similar functionality for ecnf.

JohnCremona:
OK, I added the same functionality for ecnf.  I found it not possible to quickly obtain the list of distinct values of the array field 'torsion_structure' so this is now hard-wired with the values now present.  This is not good -- I am still adding curves over totally real sextic fields, which include the unique examples of 37-torsion and 2x18-torsion as fairly recent additions!  So I would gratefully receive improvements.

JohnCremona:
Hold off again -- sorry.  Have been given a solution using mapreduce...

JohnCremona:
OK, I now have a solution using inline_map_reduce() with the results being cached so it only runs once for each thread / process.   I estimate about 4s for this (over a fairly slow connection).
If you think that's too slow I can put back the hard-wired list.

AndrewVSutherland:
@JohnCremona I just realized that mongodb's distinct method doesn't handle array fields the way we want (it looks for distinct values within the array), it update_attribute_stats won't work as written for the torsion_structure field in nfcurves (it would if it were a stored as a string like it is in genus2_curves).  I will modify update_attribute_stats so that it will handle array fields correctly (using inline_map_reduce as you did, rather than distinct).

AndrewVSutherland:
OK I modified update_attribute_stats and went ahead and created curves.stats and nfcurves.stats collections in elliptic_curves with torsion_structure data.  See

http://beta.lmfdb.org/api/elliptic_curves/curves.stats/
http://beta.lmfdb.org/api/elliptic_curves/nfcurves.stats/

All you need to do is change your code to grab the read from the torsion_structure object for these collections (and you can then remove your caching and map/reduce code).

If/when new data is added that contains new torsion structures just do update_attribute_stats(elliptic_curves,"nfcurves","torsion_structure") to update it.

JohnCremona:
Done.  Brilliant.  I really need to rewrite the whole of ecnf_stats but not now.  I also tested the update script as I had just added one more degree 6 field's data (but no more torsion structures!).

AndrewVSutherland:
Looks good. I will make sure the .stats file get copied to the cloud.

JohnCremona:
On 25 April 2017 at 18:31, Andrew Sutherland <notifications@github.com>
wrote:

> Looks good. I will make sure the .stats file get copied to the cloud.
>

I just updated it after adding another field's curves.  But this is
ongoing...

JohnCremona:
There's no hurry for enhancing the capabilities of the
update_attribute_stats() functoin but I have two suggestions about that
file:

@ move the instructions currently in comments before each function into
docstrings (between the def line and the code start) since then people can
see them by typing (e.g.) update_attribute_stats? at the command line.
 (i.e. allow python introspection).

@ (not sure about this one) Put a file __init__.py (which can be empty) in
the data_mgt directory, pr perhaps in the utilities subdirectory.  That
would all other scripts to do things like
from data_mgt.utilities.rewrite  import update_attribut_stats
instead of ( as I do now)
%runfile data_mgt/utilities/rewrite

Not urgent but might make these (even) easier to use.

John

AndrewVSutherland:
@JohnCremona Done -- see PR #2030.

PR #2030, opened 2017-04-28
Utility for collecting joint attribute statistics

AndrewVSutherland:
This PR adds a new "update_joint_attribute_stats" function to rewrite.py that allows the generation of statistics that count the number of distinct combinations of attribute values that occur in a specified collection for a specified list of attributes.

See the "two_selmer_rank:torsion_order" record in http://beta.lmfdb.org/api/genus2_curves/curves.stats/ for an example of its output.

I also cleaned up rewrite.py, added doc-strings, and added __init__.py files to data_mgt and utilities to make the functions in rewrite.py easier to use.

To test on your favorite collection, after authenticating as editor on the relevant database, you can do

> from data_mgt.utilities.rewrite  import update_joint_attribute_stats

The command used to create the "two_selmer_rank:torsion_order" attribute statstics was

> update_joint_attribute_stats(db,"curves",("two_selmer_rank","torsion_order"))

Please let me know if there is anything you think should be added or changed.  I'd like to make it as easy as possible for people to cache statistics in the database so that we can stop computing them on the fly.


JohnCremona:
Excellent, it looks as if this does everything I asked for and more so thanks.  Sorry but I'm unable to test for a few days.

JohnCremona:
Very good -- I just successfully used this to add a couple of joint attribute stats for elliptic_curves.curves.

Feature request:  get the stats over a subset of a collection as specified by a query.  Simple example:  to count isogeny classes of curves by rank, restrict to curves returned by {'number':1}.  I think this only means to change one line....

JohnCremona:
Thanks!  I made a feature request, went to a seminar, and on my return the feature was implemented.  And it works (see http://beta.lmfdb.org/api/elliptic_curves/curves.stats/).

This is definitely good for use now.

AndrewVSutherland:
I actually meant to add the same feature to joint attribute stats as well, but you merged before I got to it :)  I'll submit another PR for the joint attribute stats filter change shortly.

JohnCremona:
Thanks -- yes, that will be useful.   Before you do, we should think about this use case: we have elliptic curves over 396 number fields (plus Q) and it will be fine to have 396 separate stats entries, slightly less fine to run your command 396 times to get (for example) the rank distributions per field.  So yet another useful thing would be to collect stats for one attribute, or joint stats, for all distinct values of another, e.g. field_label, all in one command.  Not essential of course -- people can write their own loops!

JohnCremona:
Another comment http://johncremona.github.io/ecdata/table.htmland possible additional feature suggestion (though perhaps a closed PR is not the best place).  Take conductors of elliptic curves over Q.  I don't want to make a stats item for them since there are (nearly) 400,000 of them, but for the stats web page I do need to know the maximum.  I can do max(curves.distinct('conductor')) which takes 1.4s.  But for stats an option which I could use here, to make tables more like http://johncremona.github.io/ecdata/table.html
would be to collect stats for a derived quantity -- here, not the integer value of 'conductor' but rather that divided by 10000 and truncated.
It may be that there are few other palces where this would be used though.

AndrewVSutherland:
@JohnCremona PR 2033 addresses your comment https://github.com/LMFDB/lmfdb/pull/2030#issuecomment-298676362 above.  I'll need to think a bit about https://github.com/LMFDB/lmfdb/pull/2030#issuecomment-298723950 (I won't have time to do this today).

PR #2033, opened 2017-05-02
Adds more features to update_joint_attribute_stats

AndrewVSutherland:
This PR adds attributes prefix=None, filter=None, unflatten=False to the update_joint_attribute_stats function.  The functionality of prefix and filter is the same as for update_attribute_stats.  If the unflatten parameter is set to true, rather than create a single record with counts for all combinations of the specified attributes, it will split this information across several records, one for each distinct value of the first attribute.  

The key names for these records will be of the form "val1/attr2:...:attrn" (you can also include a prefix if you wish).  For example, I used

update_joint_attribute_stats(db, "curves", ["analytic_rank","torsion_order"],prefix="byrank",unflatten=True)

to create records 12-16 in http://beta.lmfdb.org/api/genus2_curves/curves.stats/

One further change: the behavior of both update_attribute_stats and update_joint_attribute_stats on empty collections (or using a filter that matches no records) has been changed -- a stats record with an empty counts list and a total of 0 will be created.

### DIFF
--- a/data_mgt/utilities/rewrite.py
+++ b/data_mgt/utilities/rewrite.py
@@ -193,13 +193,15 @@ def update_joint_attribute_stats(db, coll, attributes, prefix=None, filter=None,
         prefix: string used to prefix attribute name when constructing stats record identifier; this can be used to distinguish stats for the same attribute that were collected using different filters
         
         filter: pymongo filter that may be used to restrict stats to a subset of records
+        
+        unflatten: if true, rather than creating a single record with counts for each combination of values for attributes[0],...,attributes[-1],
+        a separate stats record will be created for each distinct value of attributes[0] with counts for combinations of values for attributes[1],...,attributes[-1].
 
     The joint statistics record contains a list of [jointvalue,count] where jointvalue is a colon-delimited string of attribute values and count is an integer,
     one for each distinct combination of values of the specified attributes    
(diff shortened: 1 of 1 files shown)


JohnCremona:
I successfully tried out the new functionality as can be seen at http://beta.lmfdb.org/api/elliptic_curves/curves.stats/ .

JohnCremona:
Sorry, I found a bug.  For example look at http://beta.lmfdb.org/api/elliptic_curves/curves.stats/ in the sha record (#3) in the counts field.  The first element is a string not an int, so the sorting is lexicographical and hence wrong, so that the max value given is 961 (from the end of the sorted list) instead of 5625.

In the actual database the 'sha' field is an int and it would be nice if the stats code preserved that.  Similarly for other fields such as 'rank'.

I also found that when processing the counts field for a joint attribute I was converting the list-of-pairs into a dict.  We could store this as a dict, but then the processing of it might need to sort the keys.

AndrewVSutherland:
@JohnCremona I can address this, but I won't have time to do it today -- I'll try to get to it tomorrow.

JohnCremona:
No hurry -- now I am aware of it I'll double check my use of 'max'.  I have just converted the page http://localhost:37779/EllipticCurve/Q/stats to get all its data from curves.stats!

PR #2034, opened 2017-05-03
refactor ec stats to use curves.stats collection

JohnCremona:
Using the new utilities in the rewrite module I have created new entries in the curves.stats collection which mean that the numbers appearing in EllipticCurve/Q/stats are now all precomputed with no quesries to the main curves collection being required.  This should make this page load faster, with no change in what is on it (these two are the things to test).

I also added a new function update_stats() to the ec import script file so that the required stats can be recreated and updated easily.

Later it will be possible to add additional information on the stats web page using this functionality, but for now I have made no changes to what is on the page, only changed how the numbers are obtained.

### DIFF
--- a/lmfdb/elliptic_curves/ec_stats.py
+++ b/lmfdb/elliptic_curves/ec_stats.py
@@ -1,8 +1,7 @@
 # -*- coding: utf-8 -*-
-from pymongo import DESCENDING
 from lmfdb.base import app
 from lmfdb.utils import comma, make_logger
-from lmfdb.elliptic_curves.web_ec import db_ec
+from lmfdb.elliptic_curves.web_ec import db_ecstats
 from flask import url_for
 
 def format_percentage(num, denom):
@@ -35,7 +34,7 @@ class ECstats(object):
 
     def __init__(self):
         logger.debug("Constructing an instance of ECstats")
-        self.ecdb = db_ec()
+        self.ecdbstats = db_ecstats()
         self._counts = {}
         self._stats = {}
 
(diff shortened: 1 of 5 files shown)


AndrewVSutherland:
I'm reviewing this now.  One question -- did you mean to include the changes to test_hmf.py and hmf_check_find.py in this commit?  They don't seem related to the ec_stats changes

AndrewVSutherland:
I think it looks good!  The performance improvement is dramatic, loading

http://www.lmfdb.org/EllipticCurve/Q/stats

typically takes 8-10 seconds, while loading

http://127.0.0.1:37777/EllipticCurve/Q/stats

takes 500-600 milliseconds (and this is through my ssh tunnel, it should be faster than this once it is in production).

I'm happy to merge, but I'll hold off until I hear back on the hmf question.

JohnCremona:
Apologies for sneaking in two unrelated changes.  The hmf_check_find one is completely harmless, I keep on tweaking the code there (I now have all elliptic curves for half the sextic fields).  The other one is a test of what happens when a rational HMF has no elliptic curve *yet*, which is a moveing goalpost; I just changed from a curve which we now have to one we still do not.  Perhaps that test should be removed now.
Neither affects production code so please go ahead.

JohnCremona:
Thanks -- could you now copy the collection curves.stats to the cloud so that when we push to prod it will work OK?

AndrewVSutherland:
Done.  See http://www.lmfdb.org/api/elliptic_curves/curves.stats/

AndrewVSutherland:
This PR is live on www.lmfdb.org.  Access times for http://www.lmfdb.org/EllipticCurve/Q/stats are now consistently under 200 milliseconds compared to 8-10 seconds previously (timings taken from MIT), so about a 50x improvement.

PR #2038, opened 2017-05-06
stats functions now attempt to preserve integers

AndrewVSutherland:
This PR addresses https://github.com/LMFDB/lmfdb/pull/2033#issuecomment-298866215.  If all the values of an attribute that is being analyzed are integers, counts will be stored and sorted as integers.  This also applies to "unflattened" joint stats.  See http://beta.lmfdb.org/api/genus2_curves/curves.stats/ for some examples.

### DIFF
--- a/data_mgt/utilities/rewrite.py
+++ b/data_mgt/utilities/rewrite.py
@@ -1,3 +1,4 @@
+# -*- coding: utf-8 -*-
 import time
 
 def rewrite_collection(db, incoll, outcoll, func, batchsize=1000, reindex=True, filter=None, projection=None):
--- a/data_mgt/utilities/rewrite.py
+++ b/data_mgt/utilities/rewrite.py
@@ -167,8 +167,11 @@ def update_attribute_stats(db, coll, attributes, prefix=None, filter=None):
     total = db[coll].find(filter).count()
     reducer = Code("""function(key,values){return Array.sum(values);}""")
     for attr in attributes:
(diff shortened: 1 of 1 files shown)


JohnCremona:
I will try this out on the elliptic curves collection, after taking a look to convince myself that the code will work without breaking if I recreate the curves.stats collection using this (which is of course better).
I see that this code is conservative in that it only converts to int (or float) if every instance can be so converted.  If we could be sure -- as surely we ought to be able to -- that for every object in a collection every value of each field had the same type, then this could be simplified using one function which converted any string to an int or a float if possible, in that order, otherwise leaving it alone.

AndrewVSutherland:
I think for utility code like this it is better to be conservative -- doing so really doesn't impact the running time.  I can imagine scenarios where the values are all strings but some of happen to look like numbers and converting some but not all of them could be confusing.

JohnCremona:
Good!  Before this when I ran my update_stats() function on nfcurves the torsion counts used strings and the max was therefore incorrectly shown as '9'.  Now the fields are converted to ints and the max is (correctly!) 37.

JohnCremona:
I'll merge this but to manage the changeover of data in stats collections I will change my code which read the stats data (in curves and nfcurves) to try/except ints and strings;  then use the new stats function to recreate the stats collections.  This should avoid a time interval in which things will not work (for elliptic curves).

PR #2041, opened 2017-05-07
last step in switching data types for ec stats

JohnCremona:
Apologies #2039 missed one thing which you can see at http://beta.lmfdb.org/EllipticCurve/Q/stats (no Sha stats!) but this PR fixes that and also removes the temporary code introduced in #2039.  (I have now recreated the curves.stats collection so it has ints not strings for torsion, rank *and* sha fields.)


JohnCremona:
Wait, I should leave the try/except code in until the new curves.stats collection is copied to the cloud....

JohnCremona:
OK now -- this will fix the blank Sha stats on beta and will also work on the old stats collection on prod.

AndrewVSutherland:
This PR and the new curves.stats file are now live on prod.

PR #2044, opened 2017-05-08
clean up transitional ec code for gal reps

JohnCremona:
When the galois image data for elliptic curves was changed over (from non-surjective to non-maximal) we put in temporary transitional code which worked on either the old data or the new.  For some time now both beta and production databases have had the new data so this code is no longer needed.  This PR simply deletes the redundant code.
To test either run the test suite or various random curves (both CM and not) in EllipticCurve/Q.


AndrewVSutherland:
Looks good.

JohnCremona:
Thanks.   I wish there was a simpler way of making this sort of change, where data format changes in an incompatible way, since it is very laborious to make the change-over, and prone to human error.  Though somehow it only seems to affect me...

jwj61:
If it makes you feel better, I have been through data format changes multiple times, and will soon do it with local fields.
\

JohnCremona:
Thanks @jwj61 it does.  And I recommend making functions like my check_database_consistency() in scripts/elliptic_curves/impost_ec_data and scripts/ecnf/import_ecnf_data.

PR #2046, opened 2017-05-11
EC/NF stats improvement

JohnCremona:
The nfcurves.stats collection has alot more in it now which means that the code for producing and displaying distributions of elliptic curves over number fields is much cleaner and faster.
Pages to check are:
EllipticCurve/
EllipticCurve/browse
EllipticCurve/browse/2 (and 3, 4, 5, 6)
EllipticCurve/browse/2/0 (and 2/2, 3/1, 3/3, 4/4, 5/5, 6/6)
Note that the URLs for the latter have changed, the old ones were 2/(0,1), 2/(2,0), etc;  the new URLs end in /d/r for degree d and r real embeddings, instead of /d/(r,s)/ where d=2+2s.  This was done since the parentheses and commas were mangled by browsers and looked ugly.

On the pages themselves there should be change of content whatsoever (comparing with the beta server, whose database has more curves in it then prod) with one exeption:  the numbers of real & imaginary quadratic fields really are 130 and 13, not 138 and 5!

With luck testing will be very simple and will reveal that the new version is a lot faster.
There's a new function update_stats() in scripts/ecnf/import_ecnf_data which recreates the necessary nfcurves.stats collection.  I checked that nothing was missing there by renaming it, running that function, and checking that all was well.


JohnCremona:
I'm about to add a new commit which makes the update_stats() function simpler and faster and eliminates some not-used things in nfcurves.stats.  (The new nfcurves.stats is already in place on beta).  No change at all to website code or function.

AndrewVSutherland:
Looks good, and much faster :)

AndrewVSutherland:
I copied the new nfcurves.stats to the cloud as well.

PR #2049, opened 2017-05-17
Update local fields code

jwj61:
This improves the local field code, and improves it in a couple of ways.  Because of changes to the way some of the data is stored (defining polynomial coefficients and inertia groups), this currently uses the collection newfields.  If/when it hits the big time, there will be a quick code change/pull request to switch the collection name back to fields.

There is a change to an Artin rep file, but it is just comments, and to an Artin rep import script, but it is just an import script.

The new local field database is bigger in 3 ways: more tame primes covered, new data for each field, and degrees for small primes now goes to 13.  The biggest addition are degree 12 for p=2 and 3.  When this gets to the main lmfdb, the data-extent knowl will be updated.

The new data consists of intermediate fields and a Galois Splitting Model (in many cases).  The latter is explained in the knowl for Invariants of a Galois closure).

When the inertia subgroup is not transitive, we now specify them by the Gap small groups library, and link into that data from the lmfdb.  This adds dynamic knowls for those groups.  Before going to the main server, that database needs to be expanded slightly (as discussed in lmfdb-inventory) to account for one group (so currently, some local field pages will crash -- I think they all have p=3 and n=12).

Under the hood, database connections was centralized to make it easier to switch the names.

Pages to possibly look at

http://127.0.0.1:37777/LocalNumberField/2.8.22.1 

This is a D_4 octic, so several intermediate fields, some with multiplicity, and the same page has an intransitive inertia subgroup, and a Galois splitting model

http://127.0.0.1:37777/LocalNumberField/2.8.0.1

This is the unramified octic field over Q_2 which famously does not have a Galois splitting model

http://127.0.0.1:37777/LocalNumberField/13.13.13.1

This is a field where we have not gotten around to computing the Galois splitting model.  Fields like this may be too hard, even if we get to it.

### DIFF
--- a/lmfdb/artin_representations/databases/Dokchitser_databases.py
+++ b/lmfdb/artin_representations/databases/Dokchitser_databases.py
@@ -1,5 +1,7 @@
 artin_location = ("artin", "representations")
 galois_group_location = ("artin", "field_data")
+#artin_location = ("artin", "representations_new")
+#galois_group_location = ("artin", "field_data_new")
 
 from type_generation import String, Array, Dict, Int, Anything
 
--- a/lmfdb/local_fields/main.py
+++ b/lmfdb/local_fields/main.py
@@ -12,7 +12,7 @@ from sage.all import PolynomialRing, QQ
 from lmfdb.local_fields import local_fields_page, logger
 from lmfdb.WebNumberField import string2list, do_mult
 
-from lmfdb.transitive_group import group_display_short, group_knowl_guts, group_display_knowl, group_display_inertia, small_group_knowl_guts
+from lmfdb.transitive_group import group_display_short, group_knowl_guts, group_display_knowl, group_display_inertia, small_group_knowl_guts, WebGaloisGroup
 
 LF_credit = 'J. Jones and D. Roberts'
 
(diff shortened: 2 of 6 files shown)


JohnCremona:
This looks fine to me --I tested it and looked at the pages you suggest.  I have not spent much time on the local fields pages though and perhaps someone else who has should comment.

2 questions: the first example http://127.0.0.1:37777/LocalNumberField/2.8.22.1 is Galois, but this is not made very clear except by the Automorphism group having order equal to the degree.  Why not say that it is Galois and show Gal() instead of Aut()?  And the polynomial given for the Galois closure at the bottom of the page as "Galois splitting model" is not the same as the defining polynomial of the field itself.  Since these are the same field, is there not a standard choice of polynomial?

Secondly, the  field 2.8.0.1 displays its Discriminant root field as Q_2(sqrt(*)).  Is that intentional?  I don't know what it means, and there is no knowl for "discriminant root field" though I can guess.

jwj61:
For the root disc. field, the knowl for invariants contains knowls for entries in that group.  At one time, that was the convention, but maybe we now do both.  Anyway, from the invariants knowl, you can get to "what is the root disc. field" and the conventions used for it.

This round of changes did not alter the original defining polynomials, and so the polynomial for the field and for the GSM are often different.  Some of the original polynomials were chosen to follow a convention which made their choice unique.  In the future (maybe this summer), many more of the fields will get such polynomials (namely Monge polynomials), which again won't serve as GSM's.

On Aut vs. Gal, yes, that sounds like a good idea, along with a statement for all fields as to whether or not they are Galois over Q, which would make it parallel to what is shown for global number fields.  So, I will work on that now.

jwj61:
The Aut/Gal change and putting the Discriminant root field knowl directly on the show field page are both done.

JohnCremona:
Thanks for the extra work -- I have not yet had time to look at this again, so if anyone else is listening and wants to join in the review, please don't wait for me.

JohnCremona:
Looks great now.

Is the notation Q_2(\sqrt{*}) at all standard? I haven't ever seen it.  In some contexts people write Q_4 for the unramified quadratic extension of Q_2, and in general Q_q where q is a power p^f of p for the unramified extension of degree f.  But I am not asking for this.

jwj61:
To be honest, I made it up when the local database was first born.  It actually abbreviates things even more, so would write this as just *.  For subfields of a V field, you would see the list *,3,3* (if the base field was Q_3).

PR #2055, opened 2017-05-25
New update_attribute_stats options

AndrewVSutherland:
This PR adds two optional parameters to update_attribute_stats.

The first is the boolean parameter _nocounts_.  If specified only the min and max of the specified attribute will be computed, not counts for each individual value.  For example:

> update_attribute_stats(db,"curves","cond",nocounts=True)

was used to create the stats record http://beta.lmfdb.org/api/genus2_curves/curves.stats/cond.

The second is the parameter _wrapper_ which is a string that specifies a Javascript wrapper function that can be used to compute a "derived" attribute.  This string must begin with "function f(rec){" and end with "};" and will be passed each record.  For example:

> update_attribute_stats(db,"curves","cond_ranges",wrapper=
> "function f(rec){return 100000*Math.ceil(rec.cond/100000);};")

was used to create the stats record http://beta.lmfdb.org/api/genus2_curves/curves.stats/cond_ranges.

Note that if _wrapper_ is specified, you should only specify one attribute, not a list of attributes.

### DIFF
--- a/data_mgt/utilities/rewrite.py
+++ b/data_mgt/utilities/rewrite.py
@@ -134,7 +134,7 @@ def create_random_object_index(db, coll, filter=None):
     rewrite_collection (db, coll, outcoll, add_counter, reindex=False, filter=filter, projection={'_id':True})
     db[outcoll].create_index('num')
 
-def update_attribute_stats(db, coll, attributes, prefix=None, filter=None):
+def update_attribute_stats(db, coll, attributes, prefix=None, filter=None, nocounts=False, wrapper=None):
     """
     
     Creates or updates statistic record in coll.stats for the specified attribute or list of attributes.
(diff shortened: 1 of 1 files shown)


JohnCremona:
This looks very useful -- I already have your curves conductors stat for ec/Q but did that manually, and have been waiting for this to do a similar for ranges.  Thanks!
For me it is a pity that we have to use javascript (mainly since I have never used it for anythng else).
I cannot test this myself yet, but feel free to merge.

JohnCremona:
Works perfectly (see http://beta.lmfdb.org/api/elliptic_curves/curves.stats/cond_ranges; I also recomputed http://beta.lmfdb.org/api/elliptic_curves/curves.stats/conductor).

I suggest that -- now that the available functionality in rewrote.py is quite extended, and potentially quite complicated -- the docstrings include a set of fully explicit examples.  This is most needed fo the latest additions because of the javascript, since (for me at least) I could not have got that right without following the example posted here, but some other examples would be helpful.
I'll merge this now anyway, as it is not urgent.  Thanks!

AndrewVSutherland:
@JohnCremona I added some examples to the doc strings, see PR #2057.

PR #2065, opened 2017-06-12
Fix #2060

### LINKED ISSUE #2060: typo in HMF L-function
Some dots appear incorrectly in the display of the sign of the functional equation.
Example:
/L/ModularForm/GL2/TotallyReal/2.2.104.1/holomorphic/2.2.104.1-5.2-c/0/0/
From looking at the code of the page it seems that a "&amp;middot;" became "&amp;amp;middot;", resulting in an incorrect display.

AurelPage:
This fixes #2060. It seems that the problem was that an html code (&amp;middot;) was used in math mode, and of course mathjax did not know how to interpret it. The page /L/ModularForm/GL2/TotallyReal/2.2.104.1/holomorphic/2.2.104.1-5.2-c/0/0/ now seems to display correctly, and I have not found any page that this PR breaks.


AurelPage:
Note: this PR passes the tests (by which I mean it fails the same ones as the master branch).

davidfarmer:
I will take care of this

davidfarmer:
Why does github show that every line in both files have changed?

AurelPage:
Because they actually changed : I realised that these two files had the DOS newline character at the end of each line, so John Cremona told me to dos2linux them (one reason to do this is that it confuses some text editors). In particular one invisible character changed at the end of each line of these two files, but they are functionally identical to the previous version. Only the second commit contains changes that affect the website.

PR #2066, opened 2017-06-12
Label check

fredstro:
After a request by Samuele I  added a function for finding the label of a newform in the database from a given level, weight, character, field and coefficients. 

I also added a test for this and fixed another test which failed due to a comment in the code containing the string "empty"...


davidfarmer:
Samuele will test this

sanni85:
Please merge this, it works fine for me.

PR #2071, opened 2017-06-12
Locsolv

nmascot:
New code to compute the finite list of places at which at which a genus 2 curve is not solvable


nmascot:
The code probably needs someone to review it and check that it is correct (also, it worked on the first time, which is rather suspicious :) )

AndrewVSutherland:
The code looks OK to me (although it could be quite slow at a large prime of bad reduction if we get particularly unlucky, as noted in your TDOD), but I don't think this a computation we should be doing on the fly, I'd rather store it in the database where it can be indexed (maybe I want a list of all the curves that are not locally solvable at 3, say).

If you run your code on all the curves in the database and send me a list, I'd be happy to cross check the results (e.g. using magma) and then add a new field to the curves collection that we can display on the genus 2 page (which was my original plan for addressing this issue).

nmascot:
Actually, I have just found a bug ! I am working on it; please do not merge for now.

PS: I completely agree that these data should be stored and not computed on the fly, but I am not sure I know how to change that.

AndrewVSutherland:
@nmascot If you use sage to run your code on all the genus 2 curves (which you can download by clicking the SageMath button on the bottom of http://www.lmfdb.org/Genus2Curve/Q/?count=50) and generate an output file (e.g. a text file in which each line contains the curve equation and a list of primes where the curve is not locally solvable, say with 0 denoting the prime at infinity), I can easily add it to the database.

There is no rush to merge this PR, it would be better to wait until you have run it against all the curves and we have double checked the results.  In fact the code doesn't even need to be in the LMFDB code base, it might make more sense to add this functionality to Sage or Pari, since it is generally useful and will work for any hyperelliptic curve over Q, not just genus 2 curves.

I'll be in Warwick on Wednesday if you want to work on this together then.

JohnCremona:
I agree 100% with @AndrewVSutherland about precomputing these -- even if the computation was instantaneous having the data in the db for statistics would be very interesting (for example seeing to what extent the local solubility proportion matches the distribution proved by me with Bhargave and Fisher.

nmascot:
I hve jut fixed the bug that I found yesterday (it was more a reasoning loophole, in fact), and now the results agree with what Magma says.

@AndrewVSutherland I'd be happy to work with you on adding these data to the LMFDB tomorrow, if you have time !

davidfarmer:
In the future (and maybe in this pull request), there should be one space on either side of =, ==,
and similar operations, and indentation should be 4 spaces for each level.

nmascot:
Here where each line is of the form
[Curve label, List of insoluble places]
where oo is represented as 0.
I have checked that Magma agrees exactly with these data, so we should include them in the database.
[InsolublePlaces.txt](https://github.com/LMFDB/lmfdb/files/1074240/InsolublePlaces.txt)

nmascot:
Once the data is in the database, I will remove my script form the LMFDB repository and modify the web_g2c.py so that it reads the data from the database.

JohnCremona:
Keep the script!  It can surely stay in scripts/genus2_curves

nmascot:
The data is now in the database, the webpage displays it, and the script is in scripts/genus2_curves.
I think that this is now ready to be merged (but please do test !).

AndrewVSutherland:
I have copied the new collection with the "non_solvable_places" attribute to the cloud, see http://www.lmfdb.org/api/genus2_curves/curves/.  I ran a quick check on the new genus 2 web pages and I think they look good!

However, I notice that your PR now includes two slightly different versions of the file g2LocSolv.py, one in the lmfdb/genus2_curves and another in the scripts/genus2_curves directory.  Presumably we only need the one in the scripts directory?

nmascot:
@AndrewVSutherland corrected.
I have also slightly modified the html page: local solvability is now part of a new section called "Rational points".

JohnCremona:
Any obstacles to merging this?

AndrewVSutherland:
I think it's ready to merge.

PR #2072, opened 2017-06-12
add jacobian variety decomp + dim of corresponding shimura variety

jenpaulhus:
This commit adds a line for Jacobian variety decompositions, as well as the dimension of the corresponding Shimura variety to the higher genus data.  You can see the new information here, for example:
http://127.0.0.1:37777/HigherGenus/C/Aut/7.16-4.0.2-4-4-4.4

There should be a knowl for "Shimura variety", but I wasn't sure which prefix to put it under (i.e. ag.shimuravariety  or something else?) so I haven't added that yet.  I will write details on where this data comes from under the "Source of the data" entry, but if you need more info now, just let me know and I can elaborate.

Also, test.sh produced some pyflakes errors but they were all from the file ./data_mgt/utilities/browse.py so I chose to ignore them since I have never touched that file.  If this is something I did wrong, just let me know and I'll fix it.

### DIFF
--- a/lmfdb/higher_genus_w_automorphisms/main.py
+++ b/lmfdb/higher_genus_w_automorphisms/main.py
@@ -63,6 +63,8 @@ def sign_display(L):
 
 def cc_display(L):
     sizeL = len(L)
+    if sizeL == 1:
+        return str(L[0])
     stg = str(L[0])+ ", "
     for i in range(1,sizeL-1):
         stg =stg + str(L[i])+", "
@@ -110,7 +112,8 @@ def decjac_format(decjac_list):
             entry = entry + "^" + str(ints[1])
         entries.append(entry)
     latex = "\\times ".join(entries)
-    ccClasses = [ints[2] for ints in decjac_list]
+    ccClasses = cc_display ([ints[2] for ints in decjac_list])
+    #ccClasses = [ints[2] for ints in decjac_list]
     return latex, ccClasses
 
 @higher_genus_w_automorphisms_page.route("/")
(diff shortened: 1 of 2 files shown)


davidfarmer:
Current labels, and talking to some people, make it look like
ag.shimura_variety
would be an appropriate label .  It could be ag.shimuravariety, but other things in ag.
use the underscore.

jenpaulhus:
Added the Shimura variety knowl per David's suggested naming scheme.

davidfarmer:
The spacing of "corresponding conjugacy class(es):" looks odd.  Should it be flush left?
I'd also make both appearances of "refined passport" into knowls.
Otherwise, looks good.

jenpaulhus:
Added knowl to second "refined passport".
Got rid of the indentation of "corresponding..." and capitalized it and (most importantly!) changed the wording to actually represent what the data was (a number representing an irreducible C-character, not a conjugacy class).

Both changes can be seen here:
http://127.0.0.1:37777/HigherGenus/C/Aut/3.48-48.0.2-4-6.1

I'm in the process of adding data up to genus 15 to the database, and I need to edit a couple of knowls, but other than that, I think this is ready to be considered for the release version.  Do I make a separate pull request for that?

AndrewVSutherland:
Yes please make the switch from beta to production in a separate PR.

First wait for this PR to be merged (I'll take a quick look now and give David Farmer a chance to take another look before merging).  Once this PR has been merged, you can create a new PR in which the only change should be removing that status: beta line for higher genus curves in sidebar.yaml.

AndrewVSutherland:
In the Jacobian variety decomposition display I think you need to put braces around the subscript to handle multi-digit subscripts, see http://127.0.0.1:37777/HigherGenus/C/Aut/11.46-2.0.2-23-46.1 for example.

jenpaulhus:
Ah, nice catch.  Thanks.  I've fixed the issue of subscripts, and also dealt with a similar issue for superscripts, which you can see in the Jacobian decomp here:

http://127.0.0.1:37777/HigherGenus/C/Aut/10.360-118.0.2-4-5.1

AndrewVSutherland:
Looks good!

PR #2080, opened 2017-06-13
allow trailing slash on URLs for ec isogeny classes

JohnCremona:
Before, the URL EllipticCurve/Q/11.a worked (and was converted to .../Q/11/a/) while EllipticCurve/Q/11.a/ did not.  Now it does -- and gets converted to the same thing.
As a side-effect one can also have EllipticCurve/Q/11.a1/ instead of EllipticCurve/Q/11.a1, both getting redirected to EllipticCurve/Q/11/a/1 (as both are handled by the same route function which determines whether the label is that of a curve or a class).
This is a one byte change!


JohnCremona:
Closing since the same change is being made by @lemurell together with some similar changes to elliptic curve L-function URLs.

PR #2092, opened 2017-06-13
Fix #2073

### LINKED ISSUE #2073: hmf: large eigenvalue display when asking for more
In HMF pages, the large eigenvalues are not displayed. However, when one changes the number of eigenvalues to display (for instance to check for which primes there are Hecke eigenvalues in the DB), all the eigenvalues get displayed and the page takes a long time to render. Maybe we should also hide them in this case?

Example: /ModularForm/GL2/TotallyReal/3.3.321.1/holomorphic/3.3.321.1-293.1-b

AurelPage:
This PR fixes #2073. It removes the implication numeigs set to a value => display_eigs. Now HMF pages keep the numeigs argument when switching display_eigs using the buttons, and keep the display_eigs argument when setting numeigs using the form.

Example of page to play with: /ModularForm/GL2/TotallyReal/3.3.321.1/holomorphic/3.3.321.1-293.1-b

### DIFF
--- a/lmfdb/hilbert_modular_forms/hilbert_modular_form.py
+++ b/lmfdb/hilbert_modular_forms/hilbert_modular_form.py
@@ -296,6 +296,7 @@ def render_hmf_webpage(**args):
         numeigs = int(numeigs)
     except:
         numeigs = 20
+    info['numeigs'] = numeigs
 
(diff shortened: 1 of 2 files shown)


AurelPage:
This PR passes the tests (i.e. fails the same ones as master).

Issue #1456, opened 2016-05-25
Remove obsolete L-function code


AndrewVSutherland:
As noted in https://github.com/LMFDB/lmfdb/issues/1431#issuecomment-221508874, there is some obsolete L-function code that uses the mongo db database Lfunction which we would like to remove (it has been superseded by Lfunctions).

The purpose of this issue is to remove the obsolete code and perform any other associated cleanup/reorganization of the L-function code that is appropriate.  @lemurell has offered to do this.

davidlowryduda:
I think this issue can now be closed.

edgarcosta:
I agree

PR #2094, opened 2017-06-13
Fix lcode and change url of EC L-functions

lemurell:
Removing some obsolete L-function code and some more cleaning up. Also changed the url of the L-function of EC/Q (and it's symmetric squares) to match the corresponding EC, so now it's "EllipticCurve/Q/<cond>/<iso>/".  The old version with LMFDB label redirects to the new scheme. Checked that it passes the tests.


lemurell:
I would say that this closes issue #1456.

PR #2096, opened 2017-06-14
Script to fix #1656

### LINKED ISSUE #1656: Insufficient Hecke eigenvalues
@danyasaki reported that certain HMFs do not have as many Hecke eigenvalues as prime ideals; he found the following:

2.2.85.1-180.2-m
2.2.85.1-180.2-n
2.2.85.1-180.2-o
2.2.85.1-180.2-p
2.2.85.1-180.2-q
2.2.85.1-180.3-m
2.2.85.1-180.3-n
2.2.85.1-180.3-o
2.2.85.1-180.3-p
2.2.85.1-180.3-q

(There does not appear to be any problem with forms with level 180.1.)

AurelPage:
This script started by Dan Yasaki and edited by me checks that HMFs satisfy the condition from #1656: if there is the Hecke eigenvalue for a prime P, then there are eigenvalues for all primes of the same norm as P. The script is ready to be used, but since it deletes data from the DB I have used it on a single form so far: compare
http://www.lmfdb.org/ModularForm/GL2/TotallyReal/3.3.321.1/holomorphic/3.3.321.1-293.1-b?numeigs=10000&display_eigs=False
and
http://beta.lmfdb.org/ModularForm/GL2/TotallyReal/3.3.321.1/holomorphic/3.3.321.1-293.1-b?numeigs=10000&display_eigs=False
The eigenvalue for the prime of norm 647 was deleted because the DB does not contain the eigenvalues for the other two primes of norm 647 in that field.

### DIFF
--- a/scripts/hmf/.gitignore
+++ b/scripts/hmf/.gitignore
@@ -1,0 +1,3 @@
+hmf-count-aps.sage.py
+hmf-numap-Np-missing.txt
+numap2-done-fields.txt
--- a/scripts/hmf/hmf-count-aps.sage
+++ b/scripts/hmf/hmf-count-aps.sage
@@ -26,11 +26,35 @@ C['hmfs'].authenticate(username, password) ## read/write on hmfs
 
 #Caching WebNumberFields
 WNFs = {}
+nf_data = {}
 
 def get_WNF(label):
     if not label in WNFs:
         WNFs[label] = WebNumberField(label)
     return WNFs[label]
+
+def get_nf_data(label):
+    """List of allowed numbers of eigenvalues for an HMF over that field.
+    """
+    if not label in nf_data:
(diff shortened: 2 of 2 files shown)


AurelPage:
The results will be easier to check on the webpages when PR #2092 will be merged.

davidfarmer:
Asked Dan Yasaki to give the script a test run.

danyasaki:
Adding a line 
assert abs(nb - num_ap) <= int(f['label'].split('.')[0]) - 1
in the truncation_bound
will ensure that not many eigenvalues are dropped.
I think it looks good.

AurelPage:
I incorporated Dan's suggestions.

davidfarmer:
Looks good!  Merging.

PR #2104, opened 2017-06-15
fixed bug when an ecnf has cm but no Galois image data

JohnCremona:
Test EllipticCurve/2.0.8.1/32.1/a/1 before and after.


JohnCremona:
This fix introduced a new bug whereby when a curve has cm but no Galois image data, while it does not crash, it incorrectly displays that there is no CM.  ANother fix is on its way...

PR #2112, opened 2017-06-15
Maass coefficient precision

### LINKED ISSUE #2076: Tiny coefficients should be zero
There are imaginary parts of size 10^-16 which should be 0,
and the same for some real parts (except that they are shown as 10^-11).

/ModularForm/GL2/Q/Maass/4f62407288aece3d9f000010

davidlowryduda:
This commit resolves #2076, so that now small coefficients are displayed as zero. 

For future choices of precision, there is now a kwarg in the `render_one_maass_waveform_wp` template that gets passed to the coefficient table, and this can be chosen appropriately. As suggested by David Farmer in #2076, the default choice of "small" is 1e-9, so that real parts or imaginary parts smaller than 1e-9 in absolute value are displayed as zero.


davidlowryduda:
I should note that a good page to check the changes on is /ModularForm/GL2/Q/Maass/4f62407288aece3d9f000010

davidlowryduda:
I've learned that there is a function defined in a utility file that facilitates the precision cutoffs. I am updating this pull request to use this utility, and I recommend not actually pulling this until then.

davidlowryduda:
I believe that this is now ready to be pulled.

davidfarmer:
Looks good.  Also makes things better when most of the coefficients
are complex, but some of them are real.  See
/ModularForm/GL2/Q/Maass/4f4c015d88aece438d00003b

PR #2126, opened 2017-06-16
Added code for converting sage curves to ecnf formated lines

koffie:


JohnCremona:
Explanation: Maarten adapted code of mine to make it easier for people who compute elliptic curves over number fields to output their curves in a format which exitsing scripts of mine can handle.  At the same time he refactored those scripts, separating those which access the database from those which do not.  As this creates some merge conflicts, I will resolve these, but only after I have finished current work which is also changing those scripts in minor ways.

koffie:
Hi @JohnCremona,
I just resolved the merge conflict for you.

JohnCremona:
Thanks Maarten.    It's hard to see from the diffs one file at a time that the functions which have been moved from one file to another have not actually changed.  That's what I was intending to check.  I expect I can trust you on that?  I may be uploading another batch over curves over Q(sqrt(-1)) soon so would like to be confident that nothing would go wrong!

scripts/ecnf/import_utils.py
```diff
@@ -0,0 +1,293 @@
...
+    """
+    cond_lab = ec['conductor_label']
+    output_fields = [ec['field_label'],
+                     cond_lab,
+                     ec['iso_label'],
+                     str(ec['number']),
+                     ec['conductor_ideal'],
+                     str(ec['conductor_norm'])
+                     ] + [",".join(t) for t in ec['ainvs']
+                          ] + [str(ec['cm']), '?' if ec['base_change']=='?' else str(int(len(ec['base_change']) > 0))]
```
koffie: hi @JohnCremona 
The `make_curves_line` is the only function I modified. I changed it so that it prints a '?' if `ec['base_change']` is set to '?'.

JohnCremona:
That sounds fine.  Thanks.

Issue #2136, opened 2017-06-21
Links to ECNF broken by label change


AndrewVSutherland:
On the production site the genus 2 curve page http://www.lmfdb.org/Genus2Curve/Q/2916/a/ contains a link to the elliptic curve isogeny class http://www.lmfdb.org/EllipticCurve/2.0.3.1/324.0.18/a/.

The elliptic curve link on the corresponding page http://beta.lmfdb.org/Genus2Curve/Q/2916/a/ on beta is broken because the label changed; it should link to http://beta.lmfdb.org/EllipticCurve/2.0.3.1/324.1/a/.

There are no doubt many more instances of this.  No code change is required, but the genus 2 data will need to change, and the migration of the updated genus 2 data will need to be coordinated with the migration of the updated ecnf data to the production site.

@JohnCremona are the ECNF labels now stable?  Is their any way to provide backward compatibility at the URL level (we should certainly update the genus 2 data regardless, I am thinking of links in published papers that cannot be changed, for example, the link http://beta.lmfdb.org/EllipticCurve/2.0.7.1/[242,136,1]/a/1 that appears as the third link on page 72 of https://doi.org/10.1017/fms.2015.33 is now broken (the curve equation is listed in the paper, so this is not the end of the world).

JohnCremona:
no no no no no
For imaginary quadratic fields:  currently in rapid flux (aka chaos).  I am trying to get the curves collection to match the bianchi newforms collection and it will be a while before that is done.  The fact that there really are newforms which have no curves ("fake curves") and CM curves with no cuspidal newforms does not help.  However the only *labelling* changes have been to use our new standard ideal labels for conductors.
For t.r. fields nothing should have changed except for some more curves being found (very few now missing).

I'm sorry.  I just discovered that in the range 20k-30k over 2.0.8.1 where I had found all (or almost all) the curves before the workshop, I had deleted some files by mistake and will have to re-find them;  and there are some other ranges where I do now have the newforms but not yet the curves.

Until about 2 weeks ago we only had conductor norm up to 10k for each field, so that is what you care about most.  For both URLs and labels all that changed is the conductor label which went from [N,c,d] to N.c.d to N.i where in all cases N is the norm, the first two encode the HNF of the ideal and the last one uses the new standard index.

JohnCremona:
[...]

PR #2137, opened 2017-06-21
added backwards-compatibility support for IQF conductor labels

JohnCremona:
This should fix Issue #2136.  Note that it requires a new collection elliptic_curves.IQF_labels which is now on beta.
Please test -- both the URLs listed at #2136 work OK (one using N.c.d format and one using [N,c,d]).   When I changed the labels I had to change a whole lot of ecnf tests -- I should have done this then instead!
The conversion is *only* used on conductor_label parts of URLs.

### DIFF
--- a/lmfdb/ecnf/WebEllipticCurve.py
+++ b/lmfdb/ecnf/WebEllipticCurve.py
@@ -11,6 +11,7 @@ from lmfdb.sato_tate_groups.main import st_link_by_name
 ecnf = None
 ecnfstats = None
 nfdb = None
+IQF_labels = None
 
 def db_ecnf():
     global ecnf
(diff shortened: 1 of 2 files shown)


AndrewVSutherland:
This looks good, thanks!  Is the collection elliptic_curves.IQF_labels ready to be copied to the cloud or are there more to be added?

Pyflakes is unhappy about many things (presumably detritus left over from last week), but none are relevant to this PR and the ecnf tests all pass.

JohnCremona:
Good -- yes for copying the IQF_levels collection to the cloud.
I know about the pyflakes stuff, did not want to clean that up myself during the workshop but we should remember to check on the before merging (I often don't).   I would like to be able to run just the pyflakes part of test.sh without having to leap in with a Crtl-C when the testing proper starts up.

AndrewVSutherland:
I can clean up the pyflakes issues (but this may not happen until tomorrow after my talk).

AndrewVSutherland:
The IQF_lables collection is now in the cloud: http://www.lmfdb.org/api/elliptic_curves/IQF_labels/ (it won't get used until we push to prod).   I understand the nfcurves is not yet ready to copy over, but is the code currently in master OK to run on the version of nfcurves in the cloud (i.e. is it safe to push to prod)?

JohnCremona:
@AndrewVSutherland Yes, should be no problem.  But do we have a record of when the nfcurves collection was last copied over?  The IQF label switch in the data from [N,c,d] to N.c.d was done a while ago, while the second swicth to N.i only last week.
Yesterday I noticed something weird about the Q(sqrt(-11)) curves whic I want to fix, but then (I'll tell you when) copying that collection to the cloud might be a good idea.  I have also not yet recreated the galreps information for the IQF curves -- when I re-upload them this does not happen automatically -- and that data is also missing for quite a lot of sextic fields.

AndrewVSutherland:
Let me know when nfcurves is in a stable state and I can compute all the missing galreps.

JohnCremona:
Will do.

AndrewVSutherland:
Regarding your question about when the file was last copied over, I don't actually know -- I thought I could easily determine this from the mongo db logs, but they get rolled over after two weeks.  I will update the copying scripts so they record their actions in a separate (much smaller) log file that we can retain indefinitely.

JohnCremona:
+1 for the log file!

Issue #2142, opened 2017-06-24
Bianchi modular form pages


JohnCremona:
After the June 2017 workshop these are nearly ready for beta, with lots of data already in the beta database.  Here is a list of things to do before I'll make a pull request:
   . (data) complete dimension table data for the same range as we have newforms: fields 2 not finished.
   . knowls needed on browse page, individual form page, newspace page (done some)
   . more search fields on browse page (done some)
   . source of data / labels / completeness knowl pages (done)
   . unit tests
That should be good enough for the first beta appearance.
Anyone impatient can run the bianchi branch from LMFDB

JohnCremona:
Closing this now it is on beta.

PR #2148, opened 2017-06-27
Bianchi forms

JohnCremona:
It's time for some more people to start to look at this.  From my checklist at #2142 I have not yet done unit tests.  The data is still begin added to but we have 122440 newforms, all but a few have elliptic curves, and for those for which there is no curve for an interesting reason, some information about that is visible.  (I will add more -- there are few enough that these exceptions could each have a top knowl).


davidfarmer:
Since you wrote "data is still begin added", it looks like I have to wait for that to see the
actual home pages.

I found it confusing to be faced with two choices:
Browse Bianchi modular forms spaces 
Browse Bianchi modular forms

For both I clicked on Qsqrt-7.  and the resulting pages were quite similar (to me),
and when I clicked on something from either page, I went to the same place.
Maybe it makes sense to only have one option (titled Browse Bianchi modular forms),
but you go to the spaces page?   Then people can just keep clicking, which is what
we mean by 'browse'.

I may be able to say more once I can browse to home pages.

JohnCremona:
You can browse home pages

AndrewVSutherland:
I only had time to take a quick look (I can take a more careful look later, but it might be a few days).  My main comment  is that I think to the extent possible we should try to make the layout of the Bianchii modular form pages consistent with that of the Hilbert modular forms pages.  For example, compare the layouts of the pages:

http://127.0.0.1:37777/ModularForm/GL2/ImaginaryQuadratic/2.0.8.1/32.1/a/

http://www.lmfdb.org/ModularForm/GL2/TotallyReal/2.2.8.1/holomorphic/2.2.8.1-9.1-a

I would expect common invariants like level to appear in the same place on both pages, and for the Bianchi modular forms Atkin-Lehner eigenvalues come before Hecke eigenvalues, while on the Hilbert modular forms they are listed in the opposite order.

Also, I presume the "primes" listed under the Atkin-Lehner and Hecke eigenvalue tables are really prime ideals generated by the elements listed -- shouldn't we use be using the ideal labels, as we do for Hilbert modular forms?  (and similarly for the level?)

Many of the headings need knowls, some of which may already be available from the Hilbert modular forms pages, but some not (e.g. "Atkin-Lehner eigenvalues" and "Hecke eigenvalues" should certainly have knowls, both on the Hilbert modular forms pages and the Bianchi modular forms pages, currently they don't in either case).

JohnCremona:
All comments so far are 100% predictable and of course inconsistencies will need ironing out and extra knowls added. 

@davidfarmer seemed to think that I was asking for a first review without actually having any data to look at.  No, I am not so stupid.  There are >100000 newforms in there.  I will continue to add more as I compute more   A small number are highlighted individually on the home page as being interesting, and there is a random form facility to click on.

The alternate view of newform spaces as well as individual newforms is partly there because for the last 3 years that was the only data I had uploaded (dimension tables but no actual newforms) and I left it in since I thought that was a useful view, but it can be deleted or differently labelled.  For classical modular forms we have views on new spaces, per level and with links to individual newforms, and I like that.  Hilbert forms do not have that (yet).

Some of the additional data I have computed for each newform is not currently present for Hilbert forms but after discussion with @jvoight and others am sure that it is desirable to show.

Some inconsistencies between Bianchi and Hilbert might therefore be eliminated by changes to the latter rather than the former.  I display AL eigenvalues  first since there is a fixed number of them while the Hecke eigenvalues are the start of an infinite table so I thought that table should come at the bottom..  Possibly the Hilbert pages could swap them over, but I don't mind too much.

I hope that when people have had time to give more time to a serious look at what is there I will see some better comments.

davidfarmer:
> @davidfarmer seemed to think that I was asking for a first review without actually having any data to
> look at.

My comment was me trying to reconcile the fact that for every
page I tried to visit, it said

   No Bianchi modular form in the database has label 2.0.4.1-106.1-a

[or whatever label it was]

I mistakenly assumed there was some data in the midst of being moved
somewhere.

I should have just reported the symptom, which I will do now:
For every link, including Examples of base change forms or Random,
I get a message like above.

What am I doing wrong?

JohnCremona:
That is weird since for me I have no problem with any of the examples or random forms.  (Obviously I would not have asked anyone to waste their time looking at this if such basic things were not working!)

You did find a bug, which I will fix: putting 2.0.4.1-106.1-a into the "find a specific form by label" box gives me a run-time error, but that form is there -- try putting 2.0.4.1 and 106 into the search boxes.  And a run-time error is different from what you reported.  

I assume that other (non-Bianchi) pages all work OK for you, so you are connected to the database properly (sorry to ask but...)?

JohnCremona:
I have fixed the trivial but in the find-by-label but will now write and include a whole range of unit tests and update the pull request when they are all working properly.

davidfarmer:
> I assume that other (non-Bianchi) pages all work OK for you, so you are connected to the database
> properly (sorry to ask but...)?

Other pages are working fine.

JohnCremona:
Can you see the bmfs database at http://beta.lmfdb.org/api/ with 130743 forms in it?

davidfarmer:
I see the database in the api, but still no home pages (same error message as above).

Could that be because I do not have the GAP database on my laptop?
A warning about that occurs when I click on the first basechange example.

JohnCremona:
That would be odd since these pages do not use GAP at all, but some do and it is a prerequisite.  But it makes no sense to me.
Another possibility is that with these new collections, somehow the authorization is set wrongly: I created the bmfs database a few years ago though the data is new.  I will investigate.  What a pain.

AndrewVSutherland:
For what it's worth, I had no problems accessing the pages (running on Atkin).

On June 28, 2017 7:19:56 AM EDT, John Cremona <notifications@github.com> wrote:
>That would be odd since these pages do not use GAP at all, but some do
>and it is a prerequisite.  But it makes no sense to me.
>Another possibility is that with these new collections, somehow the
>authorization is set wrongly: I created the bmfs database a few years
>ago though the data is new.  I will investigate.  What a pain.
>
>-- 
>You are receiving this because you commented.
>https://github.com/LMFDB/lmfdb/pull/2148#issuecomment-311631019

-- 

JohnCremona:
Thanks @AndrewVSutherland -- it means that there is nothing wrong in the authentication setup for that database & collections.  And also that there is 0 chance of me trying to work out what is wrong with @davidfarmer 's setup, so I will stop trying.

JohnCremona:
I have added a test file in lmfdb.bianchi_modular_forms and these all work for me.

davidfarmer:
Let's see if someone else can access the Bianchi forms on their laptop.

sanni85:
I just finished installing LMFDB on a new (second hand) laptop and as a test I just tried John's bianchi branch. It works and it loads data with no problem (tried 5 times, clicking on random, and the examples).

JohnCremona:
@sanni85 thanks for checking.

JohnCremona:
I have address many of the preliminary concerns expressed, including making the pages much more consistent with Hilbert pages, bearing in mind the differences.
I added ideal labels but kept in the generators which were already displayed since we do not yet have a way to explain what the ideal behind each label is.  When there is such a way (one ideal suggested is to have dynamic ideal-knowls similar to the number field knowls) this part of the display can be trimmed.
I know that more knowls are needed, and will work on them.  T show good faith and to prevent people wasting their time by pointing out the obvious I'll make another commit with a whole lot of blank (as yet) knowls where needed, so that they can be added later without more code changes.  I do **not** consider the fact that not all possible knowls have yet been written as a basis for delaying this getting to beta.

JohnCremona:
I nervously invite more comments, excluding any of the form "the knowl for x has no content yet".

JohnCremona:
Was it something I said?  There are 155,135 Bianchi newforms feeling very lonely with only me looking at them.

AndrewVSutherland:
I can look at this again later today.

JohnCremona:
Thanks @AndrewVSutherland as it's a couple of weeks since I last worked on this and I don't want it to rot.

JohnCremona:
I just fixed a couple of bugs but will wait for feedback, bug reports, etc.  This is not yet perfect by any means but the best way to improve this is for more people to be able to see it.

AndrewVSutherland:
OK, you wanted feedback so here you go:

Issues that you might (or might not) want to address before merging:

* It might be nice to be able to search on parity (sign of the functional equation).
* On http://127.0.0.1:37777/ModularForm/GL2/ImaginaryQuadratic/2.0.4.1/9125.3/a/ It seems weird to say that the analytic rank is both >= 1 and odd, since odd implies >=1.
* Captions are inconsistently capitalized, both on search pages (e.g. "dimension" vs "Level norm") and on object pages (e.g. "base change" vs "Sign of functional equation").
* I found http://127.0.0.1:37777/ModularForm/GL2/ImaginaryQuadratic/gl2dims/2.0.3.1 confusing.  What do the headings "Weight" "all" "2" mean?  I presume it means that all the forms on the page are weight 2, but the layout is confusing, since they look like column headings.  I would either move this to the text above or add a column for weight and fill it with 2's (I think I like the later better).
* Again on  http://127.0.0.1:37777/ModularForm/GL2/ImaginaryQuadratic/gl2dims/2.0.3.1 Is the n in d(n) the same as the n in n (I think both are the new dimension).  It the answer is no then I wouldn't call them both n, and if the answer is yes, why list n twice?
* Again on http://127.0.0.1:37777/ModularForm/GL2/ImaginaryQuadratic/gl2dims/2.0.3.1, the "For x out of y levels starting at level z" format seems weird to me, and the "...see previous levels..."  and ",,,see next levels..." links are different from what we usually do; why not just follow the format and conventions we use on most browse pages, as you did on http://127.0.0.1:37777/ModularForm/GL2/ImaginaryQuadratic/?field_label=2.0.4.1.  I realize that browsing spaces is slightly different than browsing forms, but there is no reason to use a different UI.

Comments/questions that you probably do not want to address in this PR:

* Would it be useful to be able to search on Hecke field?
* How about a download Hecke eigenvalues button on the form pages that would return a list of pairs of prime ideals and eigenvalues?  As of Magma V2-23.1 there are built in LMFDBIdeal and LMFDBLabel functions, so the "make_data" function in the Magma download script could even create the ideals.

AndrewVSutherland:
I should have added that I like all the changes you made and I think the form pages look quite good!

JohnCremona:
Thanks a lot!  I am hoping that by replying to this in an email rather than
on github I can insert my responses in the appropriate places, which does
not seem possible on github.

On 13 July 2017 at 01:08, Andrew Sutherland <notifications@github.com>
wrote:

> OK, you wanted feedback so here you go:
>
> Issues that you might (or might not) want to address before merging:
>
>    - It might be nice to be able to search on parity (sign of the
>    functional equation).
>
> Good idea, I will implement that.  Note that for the (relatively few)
dimension>1 forms we have (just field 2.0.4.1, level norm <=5000, dimension
2) the sign is not computed so I'll have to decide whether to include them
all, or exclude them all, from a sign-specific search.  I'll implement the
latter but that could be changed.


>
>    - On http://127.0.0.1:37777/ModularForm/GL2/ImaginaryQuadratic/2.0.4.1/
>    9125.3/a/ It seems weird to say that the analytic rank is both >= 1
>    and odd, since odd implies >=1.
>
>
Well yes.  I tend to like the extra emphasis involved -- we are saying that
the rank is positive! -- even though it is certainly logically redundant.
But I was expecting this comment and will trim the text.  Thanks for not
asking for the exact analytic rank -- the information given here is all
thet one can easily compute from te modular symbols without any analysis.


>
>    - Captions are inconsistently capitalized, both on search pages (e.g.
>    "dimension" vs "Level norm") and on object pages (e.g. "base change" vs
>    "Sign of functional equation").
>
>
I'll go through those to make them consistent.


>
>    - I found http://127.0.0.1:37777/ModularForm/GL2/
>    ImaginaryQuadratic/gl2dims/2.0.3.1
>    <http://127.0.0.1:37777/ModularForm/GL2/ImaginaryQuadratic/gl2dims/2.0.3.1>
>    confusing. What do the headings "Weight" "all" "2" mean? I presume it means
>    that all the forms on the page are weight 2, but the layout is confusing,
>    since they look like column headings. I would either move this to the text
>    above or add a column for weight and fill it with 2's (I think I like the
>    later better).
>
> Until the June workshop this was simpler since I only had data on weight
2, but then Alexander Rahm produced a lot of dimensions for higher weights
so I adapted the layout to handle that when present.  However, we soon
discovered that his dimensions were for SL2 forms while mine are for GL2
forms (which form a subspace, and are most relevant for elliptic curves --
effectively the forms which are on SL2 but not GL2 have a nontrivial
quadratic character).  So that is why we now have /gl2dims/ urls and
/sl2dims/ .  But at some point in the last month I reloaded the gl2dim data
and lost the sl2dim data so right now there are no sl2dims -- obviously I
can put that back once AR has sent me his raw data files.  [Aside: we have
no actual forms for any higher weights, just dimensions.]

Also relevant in answering your question is the fact that most dimensions
are 0!   I had some trouble in dealing with this: it seemed stupid to show
a whole page of dimensions which are 0, leaving the user to press "next"
several times before tey found anything nontrivial.  So I included the
option to only show lines where the dimension is positive.  Since we (in
principal) have several data for several weights that required the code
working out for each level whether any of the dimensions we had were
positive, and that went into the "all" column.  I think what I will do is
to omit the "all" column when only one weight is present, which will be
most of the time anyway.

Related to this is a problem I did not solve but would welcome suggestions
on.  As you can see I decided that the default for these dimensions pages
should be to only display nontrivial spaces, which the option of seeing all
of them.  But this makes it hard to control how many levels you see on the
page since at the time you do the database search you pick up all spaces
even the trivial ones.  A typical database entry is

{'field_absdisc': 4, 'field_label': u'2.0.4.1', 'gl2_dims': {u'2':
{u'cuspidal_dim': 5, u'new_dim': 3}}, 'label': u'2.0.4.1-12770.3',
'level_label': u'12770.3', 'level_norm': 12770}

and I will try to add to the query something which will only return records
for which gl2_dims.<w>.cuspidal_dim is nonzero for some w.


>
>    - Again on http://127.0.0.1:37777/ModularForm/GL2/
>    ImaginaryQuadratic/gl2dims/2.0.3.1
>    <http://127.0.0.1:37777/ModularForm/GL2/ImaginaryQuadratic/gl2dims/2.0.3.1>
>    Is the n in d(n) the same as the n in n (I think both are the new
>    dimension). It the answer is no then I wouldn't call them both n, and if
>    the answer is yes, why list n twice?
>
> All the n's are the same, since there is only one weight, but if & when
there is more than one weight the n in the "all" column would be the sum of
the n's for individual weights.  I will (1) omit the "all" column when
there is only one weight (and perhaps always), (2) have 2 columns per
weight one with the full dimension d and one with the new dimension n,
since the d (n) in one column is clearly not clear!


>
>    - Again on http://127.0.0.1:37777/ModularForm/GL2/
>    ImaginaryQuadratic/gl2dims/2.0.3.1
>    <http://127.0.0.1:37777/ModularForm/GL2/ImaginaryQuadratic/gl2dims/2.0.3.1>,
>    the "For x out of y levels starting at level z" format seems weird to me,
>    and the "...see previous levels..." and ",,,see next levels..." links are
>    different from what we usually do; why not just follow the format and
>    conventions we use on most browse pages, as you did on
>    http://127.0.0.1:37777/ModularForm/GL2/ImaginaryQuadratic/?field_
>    label=2.0.4.1
>    <http://127.0.0.1:37777/ModularForm/GL2/ImaginaryQuadratic/?field_label=2.0.4.1>.
>    I realize that browsing spaces is slightly different than browsing forms,
>    but there is no reason to use a different UI.
>
>
Quite right.  That is alegacy from the first version of this which I wrote
3-4 years ago and had not yet got round to changing.  I'll do it now.


>
>
> Comments/questions that you probably do not want to address in this PR:
>
>    - Would it be useful to be able to search on Hecke field?
>
>
Yes when we have more data for dimension > 1.   See above for why almost
all the data has dimension 1 -- my own code can only compute these.  I just
did forms.distinct('hecke_poly') and find that there are 12 real quadratic
fields currently present, and that is it for now.


>
>    - How about a download Hecke eigenvalues button on the form pages that
>    would return a list of pairs of prime ideals and eigenvalues? As of Magma
>    V2-23.1 there are built in LMFDBIdeal and LMFDBLabel functions, so the
>    "make_data" function in the Magma download script could even create the
>    ideals.
>
>
Certainly, data download is on the agenda.  No code snippets are planned
though Magma can compute all this data thanks to Dan Yasaki's code (though
having tried some examples I would not expect to reach level norm 120k as I
have done for 2.0.3.1!)

There will be an update later today.  Thanks again!

JohnCremona:
I have fixed all the things which @AndrewVSutherland mentioned, I hope (apart from the "not for this PR" things).
   . search on parity added and this + bc & cm are now listed in the search results.
   . I no longer point out that a positive odd integer is >=1.
   . capitalizations are consistent, or at least more so (which required changing some tests).
   . I removed the "all" column in the dimensions tables.  Other improvements there are now a 3-way switch between showing all spaces / all with positive cuspidal dimension / all with positive new dimension, in a cycle;  the search really finds 50 (or however many you want) with these properties rather than filtering them after the search;  the Next/Previous buttons are now standard.

There are some GL2/SL2 issues outstanding.   You can see dimension tables for SL2 since it turns out that I only deleted the SL2 dimension data for fields other than the 5 where I have lots of GL2 data. e.g. ModularForm/GL2/ImaginaryQuadratic/sl2dims/2.0.19.1 for what is available.  For the SL2 spaces we have no individual newforms; I also do not give a link from the level label to the newspace(s) -- there would be no additional data to show, and at present the code & template for level spaces is hard-wired for GL2.  That could be changed though that would require changing a URL such as /ModularForm/GL2/ImaginaryQuadratic/2.0.11.1/276.4.   (One might in fact argue that the URL /ModularForm/GL2/ImaginaryQuadratic/sl2dims/2.0.19.1 is contradictory since it contains first "GL2" then "sl2dims", though this is in fact no more contradictory than describing a newform for Gamma_1(N) as a newform for Gamma_0(N) with character.)

AndrewVSutherland:
I'm happy with all the changes you made and I think this is ready to be merged (which will make it easier for more people to look at it and to add/edit knowls).  Shall I go ahead?

JohnCremona:
Yes please -- Was hoping for @davidfarmer 's blessing but he still cannot see the data when he runs it locally.  To be useful it will need to be pushed to beta too; before pushing to master (no hurry for that of course) we'll need to copy some data.
I just noticed -- running without --debug to see the non-beta version -- that although there's no link or mention of Bianchi on the sidebar, as is right, the top level modular form browse page http://localhost:37780/ModularForm/ does show the link to Bianchi forms.  That link should be wrapped in a test for BETA -- I'll do that right now so hold on...

davidfarmer:
I'd like to see it on beta, especially since I still can't get the
Bianchi data to work on my laptop.

JohnCremona:
OK done.  I noticed that the link to Maass Forms on the Picard group is not controlled by the BETA flag but by whether the user is authenticated, and that almost nothing else is controlled by this outside the sidebar and the top page (for the logo and the Login button).  I can switch to the authentication test if preferred,

davidfarmer:
The Maass forms on the Picard group are not ready for beta,
so requiring a login seems reasonable to me.

AndrewVSutherland:
There is no reason to require login authentication for Bianchi modular forms since we are giving them beta status.

On http://127.0.0.1:37777/ModularForm/ would it make sense to list Bianchi modular forms immediately below Hilbert modular forms and add a parenthetical (on GL(2) over Euclidean imaginary quadratic fields) ?

JohnCremona:
OK, good idea.  Done!

AndrewVSutherland:
@davidfarmer Bianchi modular forms should be available on beta in < 15 minutes.

JohnCremona:
Thanks a lot.  I'll make filling in the knowls my top priority tomorrow morning.  I just added one for weight, only 99 to go...

PR #2150, opened 2017-06-29
move curves with automorphisms data from beta to prod

jenpaulhus:
This pull request is to move the "Higher Genus/Familes" from beta to production.  Two comments:

1. The genus 15 data is still loading into the database. Barring more thunderstorm induced power interruptions (a peril of living in the midwest US in June), it should be finished in the next 24 hours.

2. A small performance issue I'm a bit concerned about. A few pages like the following are very long pages (i.e. many, many lines).   They load from the database quickly, but MathJax spends a long time going through each line.  Is there a way to tell it to stop running MathJax once it hits the "Generating Vectors" line, as there is no special math notation beyond that point on any of the pages?
http://127.0.0.1:37777/HigherGenus/C/Aut/15.10-1.0.2-2-2-2-2-2-2-2-5.2

### DIFF
--- a/lmfdb/sidebar.yaml
+++ b/lmfdb/sidebar.yaml
@@ -143,11 +143,9 @@
          - title: Belyi
            status: future
          - title: "Higher genus:"
-           status: beta
            part3:
              - title: Families
                url_for:  "higher_genus_w_automorphisms.index"
-               status: beta
                colspan: onecolumn
            colspan: onecolumn
      - title: 2-d


AndrewVSutherland:
Before this PR is merged, we should copy the current version of the passports collection to the production database, but this should only be done once the database is stable.  Also, now that the passports collection is getting large, we should add a random object index to speed up the random object access time; once you are done importing the data you can do this via:

from data_mgt.utilities.rewrite import create_random_object_index
create_random_object_index (db, "passports")

where db should be the curve_automorphisms database where you have authenticated as editor or admin.  Your code will automatically take advantage of this index once it is created (via your call to the random_value_from_collection function), you don't need to make any code changes.

Let me know when the curve_automorphisms database is ready to go and I will copy it to the cloud.

Regarding the performance issue, I can't really test it now, since your loading is slowing things down (mongo db locks the curve_automorphisms database each time you update it), in fact I can't even get the example you gave to load (on a slow internet connection), but once things have stabilized I can take a look.
But in general, we should avoid automatically loading large amounts of data unless the user specifically asks for it; in other parts of the LMFDB where we have long lists of eigenvalues, for example, we typically truncate them and put in a more button that the user can click to see the rest.  Keep in mind that a casual user could very easily hit one of these long genus 15 pages by clicking on the random passport button (indeed, they are likely to hit one of the higher genus pages simply because there are more of them).

jenpaulhus:
Only loading some of the data for large examples sounds good.  I'll code that.

jenpaulhus:
The curve_automorphisms database is ready to be moved to the cloud.  All the data through genus 15 is up, and I ran the code above to create a random object index.

AndrewVSutherland:
The new data has been copied to the cloud, see http://www.lmfdb.org/HigherGenus/C/Aut/?genus=15

jenpaulhus:
I've now finished adding the last knowl from PR #2152  and so this PR should be ready for consideration.

AndrewVSutherland:
As far as I'm concerned, this is now ready for prime time -- unless anyone objects, I'll merge it tomorrow.

PR #2152, opened 2017-06-30
capped display of generating vectors default to 20

jenpaulhus:
These changes address an issue in #2150 and should be merged before that PR.

I've set a default of 20 generating vectors, and added a button to change that number as needed.  You can see an example of it here:
http://127.0.0.1:37777/HigherGenus/C/Aut/8.28-3.0.2-2-2-2-2.1

There are 24 total generating vectors in this example.

### DIFF
--- a/lmfdb/higher_genus_w_automorphisms/main.py
+++ b/lmfdb/higher_genus_w_automorphisms/main.py
@@ -38,6 +38,8 @@ def split_passport_label(lab):
     return passport_label_regex.match(lab).groups()
 
 
+credit ='Jen Paulhus, using group and signature data originally computed by Thomas Breuer'
+
 
 def get_bread(breads=[]):
     bc = [("Higher Genus", url_for(".index")),("C", url_for(".index")),("Aut", url_for(".index"))]
@@ -130,7 +132,7 @@ def index():
     learnmore = [('Source of the data', url_for(".how_computed_page")),
                 ('Labeling convention', url_for(".labels_page"))]
     
-    return render_template("hgcwa-index.html", title="Families of Higher Genus Curves with Automorphisms", bread=bread, info=info, learnmore=learnmore)
+    return render_template("hgcwa-index.html", title="Families of Higher Genus Curves with Automorphisms", bread=bread, credit=credit, info=info, learnmore=learnmore)
 
 
 @higher_genus_w_automorphisms_page.route("/random")
@@ -239,7 +241,7 @@ def higher_genus_w_automorphisms_search(**args):
         else:
             info['report'] = 'displaying all %s matches' % nres
 
-    return render_template("hgcwa-search.html", info=info, title="Families of Higher Genus Curves with Automorphisms Search Result", bread=bread)
+    return render_template("hgcwa-search.html", info=info, title="Families of Higher Genus Curves with Automorphisms Search Result", credit=credit, bread=bread)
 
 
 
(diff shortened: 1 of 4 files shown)


AndrewVSutherland:
This is definitely an improvement!  Pages like http://127.0.0.1:37777/HigherGenus/C/Aut/15.10-1.0.2-2-2-2-2-2-2-2-5.2 that were previously very slow now load instantly.

I think it would be helpful to indicate to the user (e.g. in the heading "Generating vector") how many vectors are being displayed out of how many in total (e.g. as in the search results "(displaying m of n matches)"), otherwise the user has no clear idea what  they should type into the "display number" text box at the bottom.

jenpaulhus:
Added a line as suggested above.  Two cases for if there is only one generating vector or >1:
http://127.0.0.1:37777/HigherGenus/C/Aut/3.168-42.0.2-3-7.1
http://127.0.0.1:37777/HigherGenus/C/Aut/8.28-3.0.2-2-2-2-2.1

However, now I've noticed another issue that I don't know how to solve.  The knowls on my pages which are inside of flask "if statements" are displaying at the very bottom of the page.  On the first page listed above, you can see this by clicking on the "refined passport" knowl by the equation of the curve,  or either knowl in the "Displaying the unique generating vector for this refined passport." sentence.  

On the elliptic curve pages, the torsion generator knowl is inside an "if statement" too, but I can't figure out why that knowl is showing up where it should, while mine don't.

AndrewVSutherland:
This behavior is not new with this PR, see http://beta.lmfdb.org/HigherGenus/C/Aut/3.168-42.0.2-3-7.1, for example.  One simple fix for the refined passport knowl is to not repeat it (you already have the passport knowl linked elsewhere on the page).

One other comment: I think it would be worth linking to the knowl for hyperelliptic curve (http://beta.lmfdb.org/knowledge/edit/ag.hyperelliptic_curve) on the main search page.  I would also suggest putting in a knowl for cyclic trigonal curve.

davidfarmer:
> http://127.0.0.1:37777/HigherGenus/C/Aut/3.168-42.0.2-3-7.1
> 
> However, now I've noticed another issue that I don't know how to solve. The knowls on my pages which are
> inside of flask "if statements" are displaying at the very bottom of the page. On the first page listed
> above, you can see this by clicking on the "refined passport" knowl by the equation of the curve, or

A knowl opens after the nearest enclosing block element.

The words "Equations(s) of curve(s) in this refined passport" are
not in a "p" paragraph.  Therefore the nearest enclosing block element
is the #content div.

The clue is the use of <br>.  That is bad html code (in almost all cases),
so it should be removed and instead proper p tags (open and then closed at
the end of the paragraph).

If the p tags result in poor layout, then that is  CSS problem, which I
am happy to fix.

jenpaulhus:
Fixed the paragraph tags so knowls show up in the right place now (thanks, David!). Also added links to hyperelliptic curve knowl and created cyclic trigonal knowl. See:
http://127.0.0.1:37777/HigherGenus/C/Aut/3.168-42.0.2-3-7.1

I added those knowl links on the main page and the search result pages too:
http://127.0.0.1:37777/HigherGenus/C/Aut/?genus=4&group=&signature=&dim=&inc_hyper=include&inc_cyc_trig=include&inc_full=include&count=20&Submit=Search

AndrewVSutherland:
Looks good!  One more small thing that I think should be added before moving this to production (and this PR would be a convenient place to do it).  In the footer there should be an indication of who created this data (i.e. you), as there is on http://www.lmfdb.org/EllipticCurve/Q/, http://www.lmfdb.org/NumberField/, http://www.lmfdb.org/Genus2Curve/Q/, and elsewhere.

-

jenpaulhus:
This latest commit adds credits to the footer of all pages in this collection.

PR #2161, opened 2017-07-13
Rational points data for genus 2 curves

AndrewVSutherland:
This PR incorporates data on rational points that was computed at the Rational Points 2017 workshop in Schney.  We searched for rational points on every curve in the database up to a height bound of 10^6 (on a y^2=f(x) model) and were able to prove the set of points we found was complete in almost all of the rank 0 and 1 cases where Chabauty's method can be applied (we also have probably correct results in some higher rank cases, e.g. when the curve is not locally solvable).

You can browse statistics on rational points on http://127.0.0.1:37777/Genus2Curve/Q/stats, and you can view rational points on a curve's home page, e.g. http://127.0.0.1:37777/Genus2Curve/Q/563011/a/563011/1.  In cases such as this where the list of rational points is not probably complete, this is indicated by the heading "Known rational points" with an explanatory knowl http://127.0.0.1:37777/knowledge/show/g2c.known_rational_points.  In cases where the set of rational points is provably complete such as http://127.0.0.1:37777/Genus2Curve/Q/4356/b/470448/1, this is indicated by the heading "All rational points" with an explanatory knowl http://127.0.0.1:37777/knowledge/show/g2c.all_rational_points.

There are magma code snippets that will let you create a list of the rational points on a particular curve.  I also added some tests to test_genus2_curves.py (all tests pass and pyflakes is happy as far as genus2_curves goes).  The code and data changes are backward compatible, so they can be migrated to the cloud in either order.

If you look at the code you may notice some lines relating to Mordell-Weil ranks that have been commented out.  With the kind assistance of Michael Stoll we have been able to compute provably correct Mordell-Weil ranks for more than 98% of the curves, along with complete sets of generators for the Mordell-Weil groups, but in the process of verifying these results I discovered a few errors due to bugs in Magma that are now in the process of being fixed.  Once this happens I will re-verify the results before adding them into the database.  I'll submit a new PR uncomenting these lines (and adding a display for the Mordell-Weil generators) once this happens.

Apologies of rhe large number of commits -- I wasn't setup to run the LMFDB locally in Schney so I had to commit/push/pull each time I wanted to test something on atkin.

### DIFF
--- a/lmfdb/genus2_curves/main.py
+++ b/lmfdb/genus2_curves/main.py
@@ -270,7 +270,7 @@ def genus2_curve_search(info):
         parse_bool(info,query,'locally_solvable','is locally solvable')
         parse_bool(info,query,'is_simple_geom','is geometrically simple')
         parse_ints(info,query,'cond','conductor')
-        parse_ints(info,query,'num_rat_wts','rational points')
+        parse_ints(info,query,'num_rat_pts','rational points')
         parse_ints(info,query,'num_rat_wpts','rational Weierstrass points')
         parse_bracketed_posints(info, query, 'torsion', 'torsion structure', maxlength=4,check_divisibility="increasing")
         parse_ints(info,query,'torsion_order','torsion order')
--- a/lmfdb/genus2_curves/main.py
+++ b/lmfdb/genus2_curves/main.py
@@ -424,7 +424,7 @@ class G2C_stats(object):
             avg = 0
             for value,n in counts:
                 prop = format_percentage(n,total)
-                if 'avg' in attr and attr['avg']:
+                if 'avg' in attr and attr['avg'] and (type(value) == int or type(value) == float):
                     avg += n*value
                 value_string = attr['format'](value) if 'format' in attr else value
                 vcounts.append({'value': value_string, 'curves': n, 'query':url_for(".index_Q")+'?'+attr['name']+'='+str(value),'proportion': prop})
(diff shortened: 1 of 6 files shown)


JohnCremona:
Sounds great -- I will try to review today.

JohnCremona:
This looks amazing.  I only have one query: in main.py you parse a new query field num_rat_pts but I cannot see any such search box on the page!  Similarly for "search again".  Is that an oversight?  If so do you want to fix it now or leave that for a new PR?

AndrewVSutherland:
I originally planned to add a search on num_rat_pts option but then decided that being able to browse via the statistics page was enough  (the search page is already pretty crowded and we don't support searching no number of integral points on the ec pages, for example).  But the query processing still needs to support it so that the browse from the statistics page works (and you can further narrow the search on the search again page if you wish to).

I could easily add a search box for number of rational points if people think it makes sense to do so but that can be a separate PR.

JohnCremona:
That does explain it.  So I can see curves with (say) 6 rational point from the stats page --ok-- and get a standard Search Results page inviting me to refine my search.  I can safely assume (and checked) that it will remember my previous requirement (visible in the URL bar as "&num_rat_pts=6") but that is a little invisible, and I can imagine someone (me) then forgetting how I got to that page and wondering why some curves I was expecting were missing.
I'll accept this now -- there such amazing data here, and well presented! but I do think that there should be another search box one day...

AndrewVSutherland:
@JohnCremona thanks, I will add a search box for number of rational points as soon as I get a chance.

FYI, the rational points data has been copied to the cloud and will be visible as soon as master gets pushed to prod.

PR #2164, opened 2017-07-14
Genus 2 curve search option for number of rational points

AndrewVSutherland:
As requested by @JohnCremona, this PR allows you to search genus 2 curves according to the number of (known) rational points on the curve.

### DIFF
--- a/lmfdb/genus2_curves/templates/g2c_browse.html
+++ b/lmfdb/genus2_curves/templates/g2c_browse.html
@@ -66,9 +66,9 @@ A <a href={{url_for('.random_curve')}}>random genus 2 curve</a> from the databas
 </select></td>
 </tr>
 <tr>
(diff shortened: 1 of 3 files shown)


JohnCremona:
It actually looks better with the extra search box, especially on the Search again page which used to have one by itself on the 3rd line.
On the Search again page there's room for "rational  points" in full, no need for the abbreviation.
(Not from this PR but I forgot to mention it earlier) On the Stats page some tables extend over 2 or 3 lines.  In those cases I think it would be a little better visually if there was a small gap between the continuation lines.  I think these tables are output by a script, so it would be a matter of inserting something between </tr> and <tr> to create an empty thin row with no vertical borders.  Feel free to ignore this suggestion.

AndrewVSutherland:
@JohnCremona Okay, changed pts to points, and (after a bit of fiddling), tweaked the stats page formatting along the lines you suggested.

JohnCremona:
+1

Issue #2158, opened 2017-07-11
Decorate the cursor with max_time_ms


edgarcosta:
I believe we should decorate the cursors associated with the API (and perhaps others) with max_time_ms (see: http://api.mongodb.com/python/current/api/pymongo/cursor.html#pymongo.cursor.Cursor.max_time_ms), as certain queries might be painful on the DB side, and even if the client dies after 30s, the mongodb server will keep processing the query.

AndrewVSutherland:
This is an excellent idea.

JohnCremona:
+1

Issue #2053, opened 2017-05-22
API whitelist


edgarcosta:
Hello,

Should we restrict the API queries to indexed keys?
Not many people use the API, but a bad script with a misspelled key on a big collection, might bring the mongodb server to halt.
For example, just like in #1599 or https://jira.mongodb.org/browse/SERVER-24833

Cheers,
Edgar

JohnCremona:
That sounds sensible -- but it would be useful to see a log of such failed requests, since some might be sensible (not typos) and a hint that a new index is worth adding.

edgarcosta:
Great. Could you label it as a feature request?

I think this would be Step 0 for a possible interface for Sage.

PR #2174, opened 2017-07-17
Sanitizing API

edgarcosta:
This deals with #2158 and #2053.

Try
http://127.0.0.1:37777/api/Lfunctions/Lfunctions/?a1=i1
and
http://127.0.0.1:37777/api/Lfunctions/Lfunctions/?degree=i4
.


AndrewVSutherland:
Looks good!

Issue #2128, opened 2017-06-16
Refactoring utilities


davidlowryduda:
In #2112, I tried to import a `truncatenumber` from lmfdb.lfunctions.Lfunctionsutilities.py to use to display Maass form coefficients with a correction factor. However, due to the imports within Lfunctionsutilities.py and within the Maass form files, this causes a circular import and cannot be done well.

The `truncatenumber` function is useful for database collections and python modules outside of L-functions. Similarly, there are other utility functions and utility files throughout the site. Often, these were created for a local purpose, but contain utilities that would be useful on a wider scale.

It would be a good idea to organize and refactor the utilities into a way that lets them be used and checked easier. I am beginning to work on this now.

AndrewVSutherland:
@davidlowryduda Is there more you want to do or can this be closed now?

davidlowryduda:
I intend on revisiting the utilities, but I think it is not worth keeping the issue open just for my memory. I'm closing this issue.

PR #2177, opened 2017-07-18
Unify an_list, partial utility refactoring

davidlowryduda:
As noted in issue #2128, understanding and refactoring utilities is a worthwhile (but ongoing) project. This pull request indicates some first steps towards that refactoring.

The line count of the difference appears large, but this is mostly just reorganization of the lmfdb/utils.py file. I've introduced some organization of these utilities to better understand which functions are used for what. I've also made two changes involving these utilities.

1. Firstly, symmetric power L-functions implemented their own `an_list` function in lmfdb/symL/symL.py. This very nearly duplicated `an_list` in lmfdb/utils.py, so I refactored this out of symL and just used the one from utils. See /L/SymmetricPower/2/EllipticCurve/Q/11/a/ to check, as an example.

    However, an_list is a good example of a function which may not ultimately belong in this utility file (perhaps it belongs in a central lfunction_utils.py file). It may be possible that it is ultimately better to split into a couple of utility files. This determination is a work in progress.

2. Secondly, I've slightly rewritten `truncatenumber` (now `truncate_number`) to not perform a series of elif comparisons and instead to perform a single computation and comparison. I've also added a docstring.

There is more refactoring to be done, but this is an indication of the direction I see this going.


davidfarmer:
I will take a look at this.

AndrewVSutherland:
@davidfarmer Thanks!  The large number of changes to utils.py make me nervous, given that it is imported in 109 different places in the lmfdb code base.  I realize that most of the changes are simply moving existing blocks of code around, but it will need to be carefully tested.

davidlowryduda:
@davidfarmer Thanks! These changes pass all the unit tests, but this also makes me aware that it would probably be a good idea to implement some more unittests, especially on the utility files themselves. I think that there are essentially no tests of utils.py itself.

JohnCremona:
The unit tests we have are all of the form: take this URL and test whether it works / it gives a page containing some string / it gives a page with a link which works.  What we don't have is test for individual functions in the style of Sage doctests.  Does anyone know if it would be possible to put test examples into the docstrings of these utility functions in such a way that they could be automatically tested easily from test.sh?

edgarcosta:
Yes, you could run the python file with
```
sage -t
```
and that should run the doctests, see:
http://doc.sagemath.org/html/en/developer/doctesting.html

davidlowryduda:
As an alternative to Edgar's comment, we could add in standard doctesting into our nosetest. This should be as simple as adding --with-doctest in the correct spot.

I'm experimenting with this now to see if this works like I think it should.

davidlowryduda:
@davidfarmer If you would like to hold off for a little bit, I will write a set of unittests for utils.py. My hope is to have tests that show that I don't accidentally delete some function in reorganization, as well as giving some evidence of consistency. How does that sound?

davidfarmer:
Feel free to make more changes.  I can;t spend serious time looking at 
this pull request until next week.

davidfarmer:
Looks good.  Poked around a fair amount of the website and seems seem okay.  Merging.

Issue #2077, opened 2017-06-13
Prohibit line break before ...


davidfarmer:
If you make this page just the right width:

/L/EllipticCurve/Q/304878.e/

then you can have the ... and the + before it on separate lines.

Those things need to be wrapped in a class='term' span.

mcdizzle:
This particular issue should be fixed but may appear elsewhere.

PR #2183, opened 2017-07-18
Preventing line break between + and ... at end of Dirichlet series of L functions

mcdizzle:
added class=term for '+ ...' on Dirichlet series in response to issue #2077.


JohnCremona:
This looks fine to me though I was not able to reproduce the original issue. Thanks!
@mcdizzle it's very helpful to include one or more URLs with a Pull Request so that reviewers can see what pages to test before & after.  Here you did give a link back to the original issue, which had such a URL, but copying the URL here would have saved a click or two.

PR #2184, opened 2017-07-18
cm_info wasn't handling '?'

edgarcosta:
try:
http://127.0.0.1:37777/ModularForm/GL2/ImaginaryQuadratic/?field_label=2.0.4.1


JohnCremona:
Thanks for fixing this.  All 'my own' dimension 1 forms should have this set but the dim 2 forms from Ciaran Schembri do not.  I thought I had checked that yhey displayed ok though.

Issue #1837, opened 2016-07-19
Separate import and database update scripts from website code and extend test.sh


AndrewVSutherland:
As discussed [here](https://github.com/LMFDB/lmfdb/pull/1819#issuecomment-233419972), we would like to put code that is not part of the LMFDB website into a separate directory structure and add some code validation (including pyflakes) to test.sh.

More specifically, I would propose the following:
- Move the directory lmfdb/import_scripts up one level so that import_scripts is parallel to lmfdb.  Inside import_scripts is a directory tree that effectively parallels the directory tree under lmfdb (we might want to shorten and/or generalize the name "import_scripts" to make our lives simpler).
- Move import/update scripts that currently reside in directories under lmfdb (e.g. those in lmfdb/hilbert_modular_forms, for example) into the parallel directory under import_scripts (e.g. import_scripts/hilbert_modular_forms).
- Extend test.sh to run pyflakes on the entire lmfdb/... subtree (pyflakes is fast enough for this to be quite feasible, it only takes a few seconds).
- Extend test.sh to search for uses of eval and sage_eval in all .py code in the lmfdb/... subtree.
- I think test scripts (e.g. test_....py) should stay in the lmfdb/... subtree even though they are not part of the website.

Please feel free to suggest modifications and additions to this proposal.  I have tagged this for Release 1.1.

JohnCremona:
I suggest that the very first step should be to action your first bullet point, i.e. just move the import_scripts directory.  I know that the import scripts I have (which are not in there) will not work after moving, without a small amount of editing, and the same might be true for the scripts already in in the separate directory.  So the question then is: does one of us try to make those edits at the same time, or do we recruit the "owners" of the 8 subdirectories  import_scripts/* to deal with their own?

edgarcosta:
One could use sys.path.append as a quick and dirty fix.

jwj61:
It is probably a combination of sys.path.append so scripts will be able to import lmfdb code and code to find password files.

It looks like the owners of the current import_scripts directories are me, Samuele, and maybe Stefan for Maass.  There is also code from Paul which is no longer used.

JohnCremona:
OK, so let's move them and ping them (Samuele and Stefan/Stephan?) on the pull request.

JohnCremona:
I'll do this now and make a PR.

JohnCremona:
[...]

PR #2185, opened 2017-07-19
Add doctesting, unittesting to tests

davidlowryduda:
**Note: This is not ready to be pulled yet.**

In #2177, John Cremona asked about adding in doctesting into our testing suite. This is possible, and the change to the testing suite is minimal (it now includes `--with-doctest`).

We also don't have any unittests in the code, but this is possible from our framework as well. So I've added an example unittest for utils.py, as well as a template unittest in case others felt inclined to see what this looks like. This is only good, I think.

However, there is a downside to incorporating doctests. Before, the only python files called during testing were files of the form test_<something>.py. An advantage and disadvantage of doctests is that now generic python files are called during testing, and docstrings are parsed to give doctests. The disadvantage here is that this also means that the various scripts and files that some have made for importing data to the database or testing for database coverage are called.

This has two negative side effects. 

1. Firstly, this causes files which are actually scripts (in the sense that their functionality isn't hiding behind functions) to be called. This is clearly a bad thing. As an example, this causes lmfdb/modular_forms/elliptic_modular_forms/emf_db_stats.py (which computes some statistics related to the database, and which takes some time) to run.

2. This also causes several scripts (such as lmfdb/local_fields/import_fields.py) to run. These should fail, and cause spurious errors. But it is dangerous to run random import scripts throughout the code.

Both of these are resolved by separating out scripts into the scripts/ directory, instead of lying somewhere within lmfdb/. I believe this is the intention of the scripts/ directory, but I wasn't certain enough to move those scripts first. It is also possible to avoid testing files which satisfy a REGEX, such as being named `import_<restOfFilename>.py`. But this wouldn't catch emf_db_stats.py, for example.

It is also possible to have a separate doctest call, which only calls certain files. We would probably want to doctest utility functions, but much of the frontend display code is not appropriate for doctesting.

Alternately, we may simply choose to not have doctests. It is still possible (and I think desirable) to include unittests, with or without doctests. But doctests have some nice properties.

So I ask others, what do you think is the best path to follow? I am happy to pursue any of these (or another) courses of action regarding tests (though I think that incorporating some unittests is a good idea).

### DIFF
--- a/lmfdb/test_unittest_template.py
+++ b/lmfdb/test_unittest_template.py
@@ -1,0 +1,27 @@
+# -*- coding: utf8 -*-
+# LMFDB - L-function and Modular Forms Database web-site - www.lmfdb.org
+# Copyright (C) 2017 by the LMFDB authors
+#
+# This library is free software; you can redistribute it and/or
+# modify it under the terms of the GNU Library General Public
+# License as published by the Free Software Foundation; either
+# version 2 of the License, or (at your option) any later version.
+
+import unittest2
+
+from lmfdb.base import DoctestExampleTest
+
+
+class ExampleUnitTest(unittest2.TestCase):
+    """
+    This is an example unittest. For this example, we will test
+    DoctestExampleTest, from lmfdb/base.py
+    """
+
+    # Note that the docstring to this function is printed during testing
+    def test_DoctestExampleTest(self):
+        r"""
+        Testing DoctestExampleTest.
+        """
+        ex = DoctestExampleTest(2)
+        self.assertEqual(ex.__str__(), "I am 2")
--- a/lmfdb/test_utils.py
+++ b/lmfdb/test_utils.py
@@ -1,0 +1,26 @@
+# -*- coding: utf8 -*-
+# LMFDB - L-function and Modular Forms Database web-site - www.lmfdb.org
+# Copyright (C) 2017 by the LMFDB authors
+#
+# This library is free software; you can redistribute it and/or
+# modify it under the terms of the GNU Library General Public
+# License as published by the Free Software Foundation; either
+# version 2 of the License, or (at your option) any later version.
+
+import unittest2
+
+from lmfdb.utils import truncatenumber
+
+class UtilsTest(unittest2.TestCase):
+    """
+    An example of unit tests that are not based on the website itself.
+    """
+
+    def test_truncatenumber(self):
+        r"""
+        Testing utility: truncatenumber
+        """
+        self.assertEqual(truncatenumber(1.000001, 5), "1")
+        self.assertEqual(truncatenumber(1.123456, 5), "1.123")
+        self.assertEqual(truncatenumber(-1.123456, 5), "-1.123")
+        self.assertEqual(truncatenumber(0.000002, 5), "0")
--- a/test.sh
+++ b/test.sh
@@ -42,7 +42,7 @@ else
   echo "pyflakes is happy"
 fi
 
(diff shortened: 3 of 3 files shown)


haraldschilly:
Yes, running doctests on the code as it is now is not a good idea. Back then when I disabled it, I started https://github.com/LMFDB/lmfdb-data as a separate repository related to all these ETL tasks. It's not so easy, though, because these scripts here depend on the lmfdb code as a library. Hence I think lmfdb itself needs to be a submodule inside that lmfdb-data repository and then everyone can start to move over their  scripts. Next step would be to also add tests for these import/load tasks, and later test scripts to validate/verify the data inside the database.

AndrewVSutherland:
Import scripts should reside in sub-directories of lmfdb/scripts, see open issue #1837.  So lmfdb/local_fields/import_fields.py should probably be moved to lmfdb/scripts/local_fields/import_fields.py (but @Jwj61 would be the one to confirm this).

I think doctesting across the board is likely to be more trouble than it is worth, but for utility functions it makes sense.  unittests are definitely desirable across the board (including utility functions), and should already be present in most modules (although there are certainly a lot of places where more could be added).   Many unittests are designed to check particularly annoying special cases that are data dependent and probably would not get hit by doctesting  (a good rule of thumb is that every time you fix a non-trivial bug you should add a unittest that verifies the fix, this also makes it easier for us to review pull requests).

The only constraint on unittests is that they should not be too time consuming-- it should be feasible to run https://github.com/LMFDB/lmfdb/blob/master/test.sh with no arguments (meaning it will run all unittests across the entire code base) on atkin in a reasonably short period of time (e.g. 10 minutes).  This is something we do anytime we are reviewing a non-trivial PR that is not completely localized (note that even apparently local changes to modules like numberfields or lfunctions, for example, can easily break code unittests outside of these modules).

jwj61:
I just moved two import scripts.  Let me know if there are others I overlooked.

davidlowryduda:
I like the the view of @AndrewVSutherland and @haraldschilly and, in particular, I now think that doctesting across the board is  probably more trouble than it is worth. I also think that unittests are desirable, including local unit tests for things like utility functions, as well as the end-to-end tests we have for many modules.

I'm going to close this pull request. I will later put a list of the scripts within the website code that I came across on #1837. And I will also begin to write some unit tests for utility functions.

JohnCremona:
Sorry for being late to comment but I wanted to make certain that we are all talking about the same location for "the scripts directory".  Relative to the lmfdb base (containing start-lmfdb.py for example) there is a scripts directory, with subdirectories for different sections (e.g. genus2_curves), and that is the one meant here.
There is also (by mistake possibly) a directory (relative to the same base) ./lmfdb/scripts, whoc contents I think should be moved to ./scripts, perhaps into ./scripts/Lfunctions.

AndrewVSutherland:
@JohnCremona Right, the "correct" directory is https://github.com/LMFDB/lmfdb/tree/master/scripts, the files in https://github.com/LMFDB/lmfdb/tree/master/lmfdb/scripts should be moved.  I will do this once a verify there are no code dependencies on the current location.

The key point is that only code that runs as part of the website should live under https://github.com/LMFDB/lmfdb; everything else should live somewhere else.  This will simplify unit testing, tracking down code dependencies and pyflakes issues, etc...

JohnCremona:
Agreed.  Note that we do allow code in the import scripts to import functions from the website directory, so that utility functions (e.g. constructing a number field element from a string representation, that sort of thing) is only implemented once.

AndrewVSutherland2:
Sure, we just don't want dependencies going the other way.

PR #2189, opened 2017-07-19
Hecke algebras

sanni85:
This pull request contains the implementation of the Hecke Algebras database which was presented in Warwick last June. 

Search and download functionalities work. Links to the modular forms database are work in progress.

Tests are going to be written after data upload. Knowls are going to be added later, once on beta.

The list of algebras in the favourite contains complete 2-adic and mod 2 data.
In the next days I will upload a complete set of levels and weight and l-adic data for those.


JohnCremona:
I had a quick look at the code (not scripts) in case anything jumped out at me.  I will also look at the pages (and hope that others will too!).  A couple of things:

   1.  Don't pass the connection C as a parameter.  There is no need, there is only one global instance.  And also, you have a lot if instances of C.<database_name>.<collection_name> in the code.  This is asking for trouble.  Follow a Good Example <TM> and put a function somewhere, to be imported where needed, so that you can use (something like) db_orbits.find(...) and not C.hecke_algebras.hecke_algebras.find(...).  So you don't need to refer to the connection name anywhere outside of these db_*() functions anyway.
   1. Your stats class does counting on-the-fly when created.  That's what I used to do before @AndrewVSutherland showed me a Better Way <TM> and provided utilities to help:  use the stats counting scripts in utilities to make new collections called xxxx.stats which hold these counts;  put the script to update these somewhere in your upload script files, so it is trivially easy to update the stats collections whenever you change the data; and initialize your stats class from there.  (It may even make having a stats class redundant.  Sorry, I was the one who started these...).

sanni85:
I will fix this and add commits.

I will do some clean up for the connection call. There are 3 collections in the same database and I am using all of them (also for stats), and that is why there are several calls...

JohnCremona:
OK, it's no problem to have 3 collections; the point is that the name of the collection should only appear once in the code.  One day you'll rewrite the collection x to x.new and to test it you'll then only need to change one occurrence of x in the code.  And if any other part of the system needs to access that database they just need to know one function to call.

JohnCremona:
... I like the cleaned up handling of the collections...

sanni85:
I am working on the stats now and doing more clean up.

sanni85:
Now stats should work as @JohnCremona  was suggesting.
Added tests, also run pyflakes and done more clean up.

JohnCremona:
@sanni85 and I have been discussing this and some changes will be made before this is merged, so the PR can remain open.

JohnCremona:
I am merging this since that will make it easier for @sanni85 and others to keep working on it, and it is invisible.

Issue #1959, opened 2017-01-10
Error handling HMF


sanni85:
Here an example of what is available now:
http://beta.lmfdb.org/ModularForm/GL2/TotallyReal/2.2.145.1/holomorphic/2.2.145.1-1.1

PR #2190, opened 2017-07-19
HMF url errors handling

sanni85:
This solves issue #1959, i.e. errors for url typing.

http://beta.lmfdb.org/ModularForm/GL2/TotallyReal/2.2.145.1/holomorphic/2.2.145.1-1.1
check now
/ModularForm/GL2/TotallyReal/2.2.145.1/holomorphic/2.2.145.1-1.1
/ModularForm/GL2/TotallyReal/4.4.9225.1/holomorphic/4.4.9225.jd1-176.6-e

Fixed also a test for hmf, now all hmf tests pass.

### DIFF
--- a/lmfdb/hilbert_modular_forms/hilbert_modular_form.py
+++ b/lmfdb/hilbert_modular_forms/hilbert_modular_form.py
@@ -284,7 +284,8 @@ def render_hmf_webpage(**args):
         label = str(args['label'])
         data = C.hmfs.forms.find_one({'label': label})
     if data is None:
-        return "No such form"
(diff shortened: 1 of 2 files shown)


JohnCremona:
Looks good!  With error messages like this I am never sure whether to try to distinguish a label which is validly constructed but happens to have no associated object (e.g. elliptic curve label 10.a1)  or one which is just garbage (e.g. elliptic curve label 'garbage').   Perhaps we should have one central place for a collection of regular expressions for all types of label, and use these?  so '10.a1' would pass while 'garbage' would fail (for elliptic curve over Q)?
That's for another issue of course.

PR #2194, opened 2017-07-20
Display data on ramified primes on number field pages.

jwj61:
This shows, when available, data on local decompositions of number fields at ramified primes.  Data is being computed, and will be progressively added.  This gives a connection between the local field pages and another part of the lmfdb.  The addition is near the bottom of number field pages.

For testing:
http://127.0.0.1:37777/NumberField/1.1.1.1 (Q is special by having no ramified primes, so it gets its own message)
http://127.0.0.1:37777/NumberField/11.1.54517760000000000.1 (has wild primes)
http://127.0.0.1:37777/NumberField/11.1.5999947987.1 (some primes have data, some do not because they are too big)
http://127.0.0.1:37777/NumberField/6.0.10816.1 (no data at all yet, so the section is not shown)

There is also code for making dynamic knowls with similar information.  This is currently not used, but doesn't hurt either.

### DIFF
--- a/lmfdb/WebNumberField.py
+++ b/lmfdb/WebNumberField.py
@@ -1,7 +1,7 @@
 # -*- coding: utf-8 -*-
 import lmfdb.base as base
 import sage
-from sage.all import gcd, Set, ZZ, is_even, is_odd, euler_phi, CyclotomicField, gap, AbelianGroup, QQ, gp, NumberField, PolynomialRing, latex, pari
+from sage.all import gcd, Set, ZZ, is_even, is_odd, euler_phi, CyclotomicField, gap, AbelianGroup, QQ, gp, NumberField, PolynomialRing, latex, pari, valuation
 import yaml, os
 import hashlib
 from sage.misc.cachefunc import cached_function
@@ -13,6 +13,21 @@ wnflog = make_logger("WNF")
 
 dir_group_size_bound = 10000
 dnc = 'data not computed'
+
+the_db=None
+
+def db():
+    global the_db
+    if the_db is None:
+        the_db = base.getDBConnection()
+    return the_db
+
+def nfdb():
+    return db().numberfields.fields
+
+def lfdb():
+    return db().localfields.fields
+
 
 # Dictionary of field label: n for abs(disc(Q(zeta_n)))
 # Does all cyclotomic fields of degree n s.t. 2<n<24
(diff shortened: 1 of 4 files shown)


JohnCremona:
Looks very impressive.  What is special about wild primes in the display?
Minor comments on code: you don't really need a helper function to extract a value from a dict with a default value if the key is missing:   d.get(key, default_value) does that.
Also can you not use a url_for() to put the links into the html you construct rather than hard-coding it?  That seems better to me, though I'm having a hard time remembering any good reasons.

jwj61:
I will take care of the code suggestions.

From the review point of view, there is nothing special about wild primes except that you actually get wild slopes.

jwj61:
Made changes

AndrewVSutherland:
I think this is a great addition to the LMFDB!  This is something that has been on my wish list for a while and its really nice to see the link between local and global fields.

One thing that might be a nice addition (but certainly does not need to be in this PR) would be to list the prime ideals corresponding to each factor of the Qp-algebra -- you can use our new prime ideal labeling scheme (there is pari/sage/magma code available to compute these labels), so this would not take up much screen real-estate -- you could put this in a column between p and the local field label.

One more thought: in the list of Frobenius cycle types for the unramified primes, you could link each cycle to the corresponding unramified local field (whenever it is present).  For example, on the page

 http://127.0.0.1:37777/NumberField/11.1.5999947987.1

the cycle type at 2 is (6,5), so you could link the 6 to http://www.lmfdb.org/LocalNumberField/2.6.0.1 and the 5 to http://www.lmfdb.org/LocalNumberField/2.5.0.1.

It's a small thing, but it would be easy to implement and save a few key strokes when one wants to browse the completions of a global field.  It complements what you are doing in the ramified cases and would complete the local algebra picture.

jwj61:
The unramified links sounded easy so I just tried.  Either you have to put an html anchor inside something processed by mathjax (I couldn't get that to work), or you start/stop math mode and the browser was inserting line breaks.  The best would be to put an anchor on a string inside a latex formula -- so if anyone knows how to do that, let me know.

AndrewVSutherland:
You should be able to put \href's inside mathjax, e.g.

> \(({\href{/LocalNumberField/2.6.0.1}{6}},{\href{/LocalNumberField/2.5.0.1}{5}})\)

works.

JohnCremona:
I'm happy whether or not you add the extra links in the PR.  Feel free to merge.

jwj61:
Thanks for letting me know about \href.  Currently it works if I put a static one in a document, but not when supplied as the value of a variable inserted by jinja (it doesn't show at all).  As a check, I looked at the page source, copied and pasted part produced by jinja into a template, and that part showed.  I will keep trying things so see if I can get it to work.  Meanwhile, I will merge.

jwj61:
Having the value of the href be invisible is not from jinja interaction.  The is from being in an html table, so maybe a mathjax bug.

Issue #975, opened 2016-03-31
HMF: simplify Hecke field polynomials


AurelPage:
From /roadmap: Simplify presentation of Hecke field (adjoin all elements to get an order, or mimic classical modular forms by working with the dual).

edgarcosta:
An example (for future reference):

The Heck eigenvalue field of 
http://www.lmfdb.org/ModularForm/GL2/TotallyReal/2.2.5.1/holomorphic/2.2.5.1-331.1-a
is
http://www.lmfdb.org/NumberField/6.6.50840384.1
and we should be using the integral basis computed for 6.6.50840384.1 to represent the eigenvalues.

SamSchiavone:
I wrote a script to express the Hecke eigenvalues for Hilbert modular forms in terms of an integral basis.  Given a hmfs label, the script extracts the polynomial defining the field containing the Hecke eigenvalues, runs polredabs on it, then looks up the number field in the LMFDB, extracts the integral basis stored there, and re-expresses the Hecke eigenvalues in terms of this integral basis.

1) How would we like to display this information on the pages for the Hilbert modular forms?  Maybe write alpha_1, ..., alpha_n for the elements of the integral basis and write the eigenvalues as a linear combination of them?  Currently the script outputs tuples of coefficients for each eigenvalue, the number field, the field generator, and the integral basis.

2) Should we keep track of the number fields that we find that do not have an LMFDB page?  Currently the script just outputs "No number field file found in LMFDB" when this happens.

edgarcosta:
The pr #2197 makes some progress towards this.
Perhaps we should write here, what are the next steps.

assaferan:
@edgarcosta @SamSchiavone - could one of you write here what are the next steps?

edgarcosta:
We should mimic what we did for classical modular forms https://arxiv.org/abs/2002.04717 Section 8.7

SamSchiavone:
The other thing that should probably be done first is `polredabs`-ing (or at least `polredbest`-ing) the polynomials for the coefficient fields of the HMFs. Then we can also link them up with the corresponding entries in the number fields database. I did some work on this a few years ago, but never finished it.

PR #2197, opened 2017-07-20
Added two scripts to express Hecke eigenvalues in terms of integral b…

SamSchiavone:
…asis


SamSchiavone:
#975

JohnCremona:
I like the idea but find the implementation extremely weird!
Why have a small amount of the conversion done in Magma?
Why are you reading from the cloud database?  
I would do this as follows:  write a function which takes a single newform (from the forms collection) and compute the new eigenvalues list, add that as an additional component to the form (i.e. a new key in its dictionary) and return it.  Then use the rewrite_collection utility to convert the whole forms collection to a new collection called forms.new.  There should be no problem running this, assuming that the time involved for each newform is quick small.  If the conversion cannot take place (e.g. if the Hecke field is too large and not in the database) the conversion function can simply return the form unchanged.
When that is done, you can experiment with the website code to display the new eigenvalues.

When you do the run to convert all the newforms you will need to be authenticated on the beta database.

I would recommend holding off for a while though since right now I and @jvoight and @danyasaki are doing a large-scale rewrite of the Hilbert forms database to sort out some issues there.  But you can still write and test your conversion function.

SamSchiavone:
> Why have a small amount of the conversion done in Magma?

At some point, I needed an isomorphism between the old field and the new field.  In Magma, IsIsomorphic(K_old, K_new) returns a bool, and if the bool is true, also returns the isomorphism.  In Sage, K_old.is_isomorphic(K_new) only returns the bool, but not the isomorphism.

> Why are you reading from the cloud database?

I can't authenticate because I don't have an LMFDB account.  Or do you mean why I used .xyz instead of .org?

> I would do this as follows: write a function which takes a single newform (from the forms collection) and compute the new eigenvalues list, add that as an additional component to the form (i.e. a new key in its dictionary) and return it.

Okay, that's more or less what it does now.  I guess I just need to write something that adds the output to a key in the dictionary.  But that is part of my question: what should be stored in the attribute 'hecke_eigenvalues'?  At the moment the eigenvalues are stored as a list of strings.  But would it be better just to store a list of tuples of coefficients and have whatever is rendering the html rewrite this as, e.g., 4*alpha_1 + 10*alpha_2 + ... ?

edgarcosta:
@jvoight Should one store the eigenvalues as tuples of integers or as strings in alpha_1, alpha_2, ... ?

JohnCremona:
If K1 and K2 are fields in Sage then K1.embeddings(K2) will give a list of maps from K1 to K2, empty if they are not isomorphic; in your case they have the same degree so any maps returned will be suitable isomorphisms.

On the storing of eigenvalues:  I think we should store these new coefficient lists in addition to the currently stored eigenvalues, at least in the first instance, as that will make any transition much easier to manage.  I can see no reason why not to store them as lists of ints, but there are often reason to store such lists as a single string such as '1,23,-4,5' from which the ints can easily be extracted if needed; this has the advantage that we don't mind if the ints get large in some cases, and I also think it is more efficient in mongo.

For the display, we can experiment.  I think the display might look too 'busy' if we displayed actual linear combinations such as \(\alpha_1+2\alpha_2-3\alpha_3+5\alpha_4\), so would try displaying a simple list instead '[1,2,-3,5]'.  We would have to define what the alphas are somewhere on the page anyway, and it would be no harder to say something like 'Each eigenvalue is expressed a list of its coefficients with respect to the integral basis *, *, *, * for the Hecke field ***'

I have some more thought on using a carefully chosen integral basis instead of the standard one, but I will work out an example first and then post here again.

JohnCremona:
I tried one manual experiment on an example of what you are doing, on the first form of dimension 10 which is '3.3.49.1-1049.1-b'.   The Hecke poly does not look too bad and the optimized one rather similar, but has a discriminant smaller by a factor of d^2 where d=97*8830823.   Not quite coincidentally this d is a common denominator for the eigenvalues when expressed in the power basis; after clearing the denominator, the max numerator is about 5.1*10^11.  Switching to the optimized polynomial for the field the common denominator is only 2*2069 and max numerator about 3.6*10^6 so that has gone down from 12 digits to 7.  Next we express with respect to the integral basis (in the optimized representation) and find the maximal coefficient is only 200124.  (This example has 169 coefficients).  The original 169 strings have total length 44323 while the integral basis coefficients use up 12052 (converting the list of lists to a string just for comparison).  Not as good a space saving as I was hoping for, but still a factor of 4.

Now comes the best part and my reason for this entire post.  The list of 169 eigenvalues expressed in the integral basis is a 169x10 matrix of integers, say M.  Using LLL on a 10x10 invertible submatrix I found a unimodular V such that MV is LLL-reduced -- the max entry is now only 22 (compared with 200124)!

Hence my suggestion: for each newform do what you propose (express the eigenvalues as Z-linear combinations of an integral basis of the optimised Hecke field) but choose the integral basis so that the resulting coefficients are LLL-reduced.  There's an additional cost in that this integral basis depends on the form and not just ts Hecke field, and will have to be displayed on the page;  but the benefit is that then the displayed (and stored coefficients) are a lot smaller.  If one did this additional step as part of the process in your script, we would need to store the customised integral basis in the database (as a dxd matrix of rationals where d is the dimension) as well as the (smaller) coefficients.

I have been suggesting something like this for years, and would love to have some feedback.

jwbober:
Sage has some functionality to do this already, I think, with its
compact_system_of_eigenvalues function. I think this is used somewhere in
the modular forms stuff, but not when the coefficients are actually stored
or displayed.

JohnCremona:
That's right.  See src/sage/modular/modsym/space.py where it is essentially
a one-liner, but does not do the LLL trick.

On 21 July 2017 at 15:55, Jonathan Bober <notifications@github.com> wrote:

> Sage has some functionality to do this already, I think, with its
> compact_system_of_eigenvalues function. I think this is used somewhere in
>

That's right: see src/sage

jvoight:
Just a comment about lists of strings or lists of ints: What is the max size of an int?  64 bits?  We've run into an issue like this before and had to revert back to strings because they were (eventually) too big.  JV

sanni85:
yes @jvoight  is right 64bits is the max check #1092

AndrewVSutherland:
What is the current status of this PR?  It has been sitting idle for 6 weeks now.  It only affects files in the scripts directory, so this isn't a major issue, but I'd like to see it either closed it or merged soon.

JohnCremona:
I see no harm in merging this, though it is only one step in what could be a major reorganisation of how we store (and possibly display) Hecke eigenvalue data.  ANy additional work on that can certainly start from this beginning.  I'm sure (as I said above) that there is no need for one function to be done in Magma.

JohnCremona:
If @SamSchiavone is reading this, would you like to comment?

SamSchiavone:
I've worked some more on the code, so I'll update soon, though it might
have to wait until after this weekend.  My compromise with the uncertainty
regarding what kind of basis would be best is to include the basis as an
input to the function.

JohnCremona:
OK thanks Sam.  We'll leave the PR open for now, so that if you push more to your branch the PR will automatically get updated.

SamSchiavone:
All right, I've updated to the most recent version of the script.  Thanks for explaining how to use Sage to find an isomorphism between the two fields.

At the moment the script uses the integral basis listed in the LMFDB (if the Hecke eigenfield has an entry), but I've included int_basis as an argument to one of the functions in case you want to use something fancier, e.g., an LLL-reduced basis.

edgarcosta:
Can I merge this?

Issue #1975, opened 2017-02-23
Base change HMF


sanni85:
http://beta.lmfdb.org/ModularForm/GL2/TotallyReal/4.4.13725.1/holomorphic/4.4.13725.1-1.1-a

on this page, it is claimed that the orbit is a base change, but it should not be (there is no definition of base change given, maybe I am misinterpreting it). 

The isogeny class of elliptic curves corresponding (http://beta.lmfdb.org/EllipticCurve/4.4.13725.1/1.1/a/)
contains curves which are not Q-curves nor base change, so the form should not be base change. Another argument it that If the form was base change then there should exists certain forms over Q(sqrt 5) which do not exist. 

I think the problem comes from the fact that the base field  (http://beta.lmfdb.org/NumberField/4.4.13725.1) is not Galois over Q.

JohnCremona:
I agree that we need precise definitions of both 'is_base_change' and 'is_CM'.  A page such as the one above has blank knowls for both terms (they link to nonexistent mf.cm and mf.base_change).  the cm should consistent with the existing knowl mf.elliptic.cm_form.  So a very easy first step would be for someone to add text to those knowls: easy in the technical sense, of course getting the definition right might not be so easy.

sanni85:
The Galois closure of the base field is this field http://beta.lmfdb.org/NumberField/8.8.700945700625.1 which does not have Q(sqrt(-5)) as subfield, but Q(sqrt(5)) (the Galois closure is totally real too). I think I checked the forms over Q(sqrt(5)) in the right level range (level norm 1 over the quartic field, means that the norm of the level for the form over Q(sqrt(5)) should only be divisible by 3,5,61 which are the factors in the discriminant) and none was giving the right base change, but I can be wrong.

JohnCremona:
Above p=11 tere are two primes of degree 1 (norm 11) both with eigenvalue 0, and one of degree 2 (norm 121) with eigenvalue +22.  But base-changing locally from GF(11) to GF(11^2)  would take eigenvalue a_11 to (a_11)^2-2*11 = -22 in this case.  That would be true whether the b.c. was coming from Q or the subfield Q(sqrt(5)), in which 11 splits.

Secondly, the 4 primes of norm 19 do not have the same eigenvalue except up to sign.

[...]

Issue #1972, opened 2017-02-17
Hilbert modular forms: more search parameters


sanni85:
It could be helpful to add search for base change and CM of the type include/exclude. 
This data is only shown on the page of single orbits.

JohnCremona:
This should be pretty easy, assuming that every form in the database has the fields 'is_base_change' and 'is_CM' set.  I found that doing forms.distinct('is_base_change') is very fast (the only values are 'yes' and 'no') while forms.distinct('is_CM') takes a long time, suggesting that there is no index for that, as confirmed by forms.index_information().  I am creating an index now (in the beta database).

The above comment underlines how helpful it would be to have https://github.com/LMFDB/lmfdb-inventory/blob/master/db-hmfs.md completed (including a list of indices).

JohnCremona:
Index finished, so I can confirm that each of the 368356 forms has both of these fields set, the only values being 'yes' and 'no'.

sanni85:
Adding the search is not a problem, I can write code for that, specially after the index addition you did before. Probably we should also add it for classical modular forms (I do not know how difficult that is because the database design is very different).

Anyway, there are two extra problems. On the one hand I am not so sure about the definition (see #1975  where I really think that the form is not base change), and on the other hand I think we should link objects: in this case the form with the base change if we do have it in the database.  Of course this would require more work (I think the same should be done for GL2 type genus 2 curves etc).

assaferan:
It seems like all the features requested in this issue have been added. @sanni85, @jvoight - if you agree, we can close this issue.

PR #2198, opened 2017-07-20
HMF: added base change and CM as search parameter

sanni85:
Added base change and CM as search parameter.
Added tests for this too.

check here:
/ModularForm/GL2/TotallyReal/
/ModularForm/GL2/TotallyReal/?start=0&field_label=&deg=5&disc=&cm=include&bc=exclude&count=100

The knowls for CM and base change need to be written, but I think @jvoight is going to do that.
Issues #1975 #1972 could be closed.

### DIFF
--- a/lmfdb/ecnf/templates/ecnf-index.html
+++ b/lmfdb/ecnf/templates/ecnf-index.html
@@ -72,13 +72,13 @@ Please enter a value or leave blank:
   <td>
          <select name='include_base_change'>
            <option value="off">exclude</option>
-           <option value="on" selected="yes">include</option>
+           <option value="on" selected>include</option>
          </select>
   </td>
   <td align=right>{{ KNOWL('ec.q_curve',title="Q_curves") }} </td>
   <td>
          <select name='include_Q_curves'>
-           <option value="include" selected="yes">include</option>
+           <option value="include" selected>include</option>
            <option value="exclude">exclude</option>
            <option value="only">only</option>
          </select>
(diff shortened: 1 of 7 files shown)


JohnCremona:
Good!
Small point: in the <option> tags it's enough to have the word 'selected' in the one you want to preselect, it is redundant to put 'selected="yes"', and I think clearer not too.
On the main browse page there's now a single  column of 9 boxes.  Time to make into 2 columns?  And in the search again section, instead of 3 rows of 3 you could have only 2 rows.
Incidentally do we have any HMFs not of parallel weight 2?  I found none by searching but perhaps I was doing it wrongly.  If only have weight 2 the weight search box could be commented out.

sanni85:
Take a look now. I have followed your advice more or less: I decided to make two section in the main browse page, one for the base field and one for the form. The part for the form has two columns. The search page result has 3 rows, the last only with the max number of results.

I have cleaned up the  'selected="yes"' and also from where I copied it....(that is the reason why you see changes in elliptic curves files).

JohnCremona:
Thanks!  Layout is better.  I'm sorry I gave you more work to do including changing code I wrote to matc my recommendations... I only recently saw that the 'selected ' tag could be simplified but never was organised enough to make changes except in the files I was working on.
@jvoight what about the weights?  We do only have parallel weight 2 forms in the database, and while that remains the case it seems misleading to allow it as a search field -- the only effect would ever be to reduce the number of hits to zero!  I would only comment out the boxes, in case of future expansion.

That's a separate issue though so I will merge this one now.

jvoight:
I remain optimistic that we will have higher weight forms at some point in the future, so I'd rather just leave it as is.  JV

PR #2199, opened 2017-07-21
bmfs: add link to base change forms where relevant

JohnCremona:
When a Bianchi newform is a base-change we add a link to the base-change form(s).  There may be 0, 1 or 2 of these.  We only show the link when the form downstairs exists in the database.  Usually there are 2, if any, which are twists of each other, but there can be just one if it is CM.  I did not find an example where one of the pair of twists is in the database and one not (I did not look hard).

Examples are:

bc forms not in db:  ModularForm/GL2/ImaginaryQuadratic/2.0.3.1/101124.1/a/

bc forms both in db:  ModularForm/GL2/ImaginaryQuadratic/2.0.11.1/1089.2/b/

bc of only one (CM) form: ModularForm/GL2/ImaginaryQuadratic/2.0.11.1/1024.1/c/

Note that I have done this in a sneaky way: when the Bianchi form has an attached elliptic curve and that curve is a base change then we use its label to get the label of the form over Q.  So this will not work for the exotic forms which are b.c. of forms with real quadratic coefficients, such as ModularForm/GL2/ImaginaryQuadratic/2.0.4.1/16384.1/d/
and other examples on the top browse page.  But I still thought it worth doing.  For the general case one would have to be able to search classical forms using their a_p in cases where those are not rational, and I do not know how to do that.


AndrewVSutherland:
@JohnCremona It looks like a merge conflict was created by #2209 which I just merged (sorry, I should have looked at this one first).  But it should be easy to fix -- if you can take care of this I will then review and merge.

JohnCremona:
No problem, I'll fix this now.

PR #2200, opened 2017-07-21
HMFs: localised collection names so each is in only one place to aid …

JohnCremona:
…testing (and clarity)

There are 3 collections inside hilbert_modular_forms: forms, fields, forms.search.  Each was referred to multiple times in the code which makes it rather hard to test things like switching from forms to forms.new (as I am now doing).  This PR just uses the model used elsewhere, in which for each collection there is one function which delivers the collection, so such changes only need to be done one.  (They are set in hmf_stats to avoid circular imports.)

As I wrote the above I noticed a few other places where these should be used so I will add those, after which a quick review would be welcome.

### DIFF
--- a/lmfdb/hilbert_modular_forms/hilbert_field.py
+++ b/lmfdb/hilbert_modular_forms/hilbert_field.py
@@ -9,17 +9,11 @@ from sage.all import ZZ
 from lmfdb.base import getDBConnection
 from lmfdb.WebNumberField import WebNumberField
 
-hmf_nfdb = None
-
-
 def db_hmfnf():
     """
     Return the fields collection from the HMF database
     """
-    global hmf_nfdb
-    if hmf_nfdb is None:
-        hmf_nfdb = getDBConnection().hmfs.fields
-    return hmf_nfdb
+    return getDBConnection().hmfs.fields
 
 def findvar(L):
     """
--- a/lmfdb/hilbert_modular_forms/hilbert_modular_form.py
+++ b/lmfdb/hilbert_modular_forms/hilbert_modular_form.py
@@ -2,12 +2,11 @@
 
 import pymongo
 
-from lmfdb.base import getDBConnection
 from flask import render_template, url_for, request, redirect, make_response, flash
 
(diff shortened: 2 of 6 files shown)


JohnCremona:
OK, ready for review.  Should be trivial since ./test.sh is happy!

davidfarmer:
I see that you use a global variable to hold the database connection,
and test whether it already exists before creating it (and just return it if
it already exists).  I have vague misgivings, but no suggested alternative for
how to do this.

Presumably this is done elsewhere in the code, so no harm to continue the practice.
Number fields seems to be just slightly different: it uses a global flag to record
if the connection has been made previously.

I'll merge in a day or so if nobody tells me not to.

AndrewVSutherland:
@davidfarmer You are right to have misgivings about this, there is no good reason to cache the return value from dbGetConnection() (or collections accessed via the connection) in a global variable.  The cost of calling dbGetConnection is low, and the cached value will likely never get reused -- worker threads are typically killed within a few minutes, and even if the user is on an hmf page and immediately clicks to another one, it will likely happen that a different worker thread will handle the second click and not have access to the cached global variable.

As we have seen in the classical modular forms code, caching generally gives no performance gains, and introduces the opportunity for very subtle bugs (e.g. bugs that cannot be reproduced when running in debug mode and which depend on which worker thread gets assigned to serve a given page request -- we had to completely disable all caching in the modular forms code to fix some of these).

I would recommend removing the global variable and simply calling dbGetConnection each time, as is done in https://github.com/LMFDB/lmfdb/blob/master/lmfdb/genus2_curves/web_g2c.py, for example (the caller is of course free to reuse the returned value, the point is to keep it in a local variable that will go out of scope once the caller is finished).

JohnCremona:
Im happy with drews suggested change and will consequently change several
other places where i use the caching trick as i have for years.

The main point is to have the collection name in exactly one place .  This
never had anything to do with misguided ideas of enhancing runtime
efficiency.

I'm travelling this week but will do that next week.

JohnCremona:
Done as requested.

Issue #2195, opened 2017-07-20
Yoshida_lift


rupertm2:
I've added a script, which takes a hilbert modular form defined over a totally real field and outputs some information (Level, weight, some eigenvalues) about the siegel modular form that is its Yoshida lift. I can also add functionality for adding the lifts of Bianchi modular forms. I also hope to be able to calculate the fourier coefficients, at some point in the future. You can find the script in my branch of the LMFDB under scripts/siegel_modular_forms. Should I make a pull request for this function? I'd appreciate feedback any feedback.

PR #2201, opened 2017-07-21
Yoshida lift

rupertm2:
This is the file which computes the data from Issue #2195 . There is documentation in the file.

### DIFF
--- a/scripts/siegel_modular_forms/Yoshida_Lift.py
+++ b/scripts/siegel_modular_forms/Yoshida_Lift.py
@@ -1,0 +1,59 @@
+# This script was created using the 2012 paper of Johnson-Leung and Roberts. 
+
+# INPUTS:
+# F is a real quadratic number field
+# hhecke_evals is a dictionary of eigenvalues of a hilbert modular form over F. It should contain primes up to norm of primeprec**2
+# hweight is the pair even integers that gives the weight of the hilbert modular form, no reason to default to [2,2]
+# primeprec is the uper bound on the size of the rational primes which will have eigenvalues calculated
+
+# OUTPUTS:
+# Paramodular Level: An int
+# Paramodular Weight: an Int
+# T(1,1,p,p): a dictionary which has keys all the primes up to primeprec, and has values which are hecke eigenvalues
+# T(1,p,p,p^2): See above
+
+
+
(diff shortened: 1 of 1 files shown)


JohnCremona:
Questions:  were you intending to use this once and for all -- in which case being in the scripts/ directory would be better -- or for it to be used on the fly at some stage?
More importantly, is there the intention of using the output to actually link to the Siegel MF (which would have to be in the database)?  That would be great, if possible, as we could list on the page of all or some HMFS the link to its lift.  If the SMFs are not in the database is your intention still to list these parameters in a section on the home page of a HMF?

rupertm2:
> were you intending to use this once and for all -- in which case being in the scripts/ directory would be better -- or for it to be used on the fly at some stage?

Once and for all was my intention, I thought I had put it in the scripts/ folder. I'll double check.

> More importantly, is there the intention of using the output to actually link to the Siegel MF (which would have to be in the database)? That would be great, if possible, as we could list on the page of all or some HMFS the link to its lift.

This is precisely what I was hoping to do, making and series of SMF home pages for Yoshida lifts, and maybe other lifts, and linking the pages together. I'm thinking it would be very useful to know if your SMF is some kind of lift or not. I'll have to wait for the HMFs to be worked out to accurately and consistently produce the data and format, but this is the goal. I also object to on the fly calculations since I'll *shorty* be able to calculate some fourier coefficients for Yoshida lift of HMFs, but I think it may take a substantial amount of time.

JohnCremona:
Good!  And your file is already in the scripts directory, sorry about that.

JohnCremona:
Before we merge this -- which is not at all controversial, being a script -- could you run pyflakes on it and fix the warnings?  (It should be enough to run ./test.sh since it runs pyflakes first, then kill it before it does the longer tests).  It's a little tedious you have to import from sage.all (various things) -- I think prime_range only perhaps, if you change occurrences of valuation(x,y) to x.valuation(y).  It also warns that you assign to I on line 25 but never use it.  Probably harmless but these checks can find typos which lead to bugs.

AndrewVSutherland:
+1 for fixing the pyflakes warnings, otherwise we get to see them again every time we test a new PR

rupertm2:
> could you run pyflakes on it and fix the warnings?

Done and now showing no warnings for me.

PR #2203, opened 2017-07-21
In Frobenius list, make the entries links to local field pages

jwj61:
This now works.

 http://127.0.0.1:37777/NumberField/17.1.133094720815439153.1

shows an example where some entries are there, and some have too big of a degree.


AndrewVSutherland:
Looks good, thanks for adding this!

PR #2207, opened 2017-07-27
Stop caching database connections

jwj61:
This removes caching of database connections from 3 areas.  Note, we keep the function which returns the database connection as a shorthand since its output is passed to dynamic knowl functions.


edgarcosta:
looks good to me

PR #2208, opened 2017-07-27
Weil poly

roed314:
Fix some bugs in abelian varieties over Fq and remove beta status.


JohnCremona:
Can you give some URLs to show fixed bugs?
I had a test failure:
Check that we can restrict to primitive or non-primitive abelian varieties only ... FAIL
(it's just that it looks for a label which only appears on the second page)

I would prefer the removal of beta status to be a separate PR, not mixed in with big fixes.

roed314:
Okay, I reverted the changes to the sidebar and added tests for the plots and the property box.

JohnCremona:
This looks really great -- and so many tests!  I like the added parameters in the encode_plot function which I think I wrote but could never get the plots to be trimmed suitably -- I need to apply this to elliptic curve isogeny graph plots which have much too much whitespace around them.

I'll merge this and push it to beta.  Then we can ask the community about removing beta status.

roed314:
Thanks!  Should I write to the lmfdb list or do you want to?

JohnCremona:
You -- make sure the others involved in this are in agreement first.
Since I pushed this to beta the new stuff is now visible e.g. at http://beta.lmfdb.org/Variety/Abelian/Fq/5/2/a_d_d_b_l

By the way when I click on the knowl for "ordinary" above the graph I do not see anything.  Any ideas why?

roed314:
The knowl does show up, but at the bottom of the page.  I'm trying to figure out why now....

JohnCremona:
OK -- that is usually caused by not enough <div>...</div> or <p>...</p> since the knowl appears at the end of the current block.

roed314:
I'm confused; not enough ellipses?

roed314:
I've been looking for about an hour and can't figure out why these show up at the bottom....

davidfarmer:
A knowl appears after the nearest enclosing block element.
In your case, that may be the whole page.

If you wrap it in "p" tags
(including the closing /p after it)
then probably that will work.

roed314:
Thanks for your help David.  I've created https://github.com/LMFDB/lmfdb/pull/2211 for this issue.

PR #2211, opened 2017-08-01
Weil poly

roed314:
Fix knowl that wasn't appearing on abvar/fq_isog pages by cleaning up <p> and <td> tags in the template.


JohnCremona:
Mostly great, but since you mentioned error parsing for input into the search boxes for dimension and firld cardinality I tried putting something  in there which is reasonable and works elsewhere, namely 1..6.  This crashes the server!  That is not good.  Could you (a) make dots work like a dash (as in 1-6) -- which must be nandled by a standard parsing function; and (b) catch incorrect input with a helpful message?  As an example, try putting x into the Conductor search box at http://www.lmfdb.org/EllipticCurve/Q/ which is handled well (not implemented by me!)

roed314:
Sure, I'll try to take a look today.  I got distracted by setting up a zulip server for Sage (which I'm about to announce on sage-devel).

roed314:
Where did you enter `1..6`?  When I put that in for dimension or q it works fine....

JohnCremona:
I put 1..6 into the dimension box (where the suggestion is 1-6).  I did just check that I am running commit cde02f3.  I get :

>   File "/home/jec/lmfdb/lmfdb/abvar/fq/main.py", line 68, in abelian_varieties
    return abelian_variety_browse(**args)
  File "/home/jec/lmfdb/lmfdb/abvar/fq/main.py", line 206, in abelian_variety_browse
    gD = parse_range(info['table_dimension_range'])
  File "/home/jec/lmfdb/lmfdb/search_parsing.py", line 156, in parse_range
    return parse_singleton(arg)
ValueError: invalid literal for int() with base 10: '1..6'

roed314:
Strange.  When I enter 1..6 into the dimension box (on localhost which is running the same commit you mentioned), I don't get any error.

roed314:
Oh, I think you're talking about a different box!

JohnCremona:
Yes sorry for the ambiguity - -I'm changing what's in the boxes in the Browse section which control the table displayed.

roed314:
Alright, I think this commit fixes things.  I still need to address Drew's suggestions from last week, which requires changing data.

JohnCremona:
This looks good now!  The other thing which requires new encoding of data can be in a new PR, I think, so I will merge this... as soon as you fix 2 pyflakes warnings:

> ./lmfdb/abvar/fq/main.py:10: 'parse_range' imported but unused
./lmfdb/abvar/fq/main.py:232: local variable 'err' is assigned to but never used

roed314:
Alright, done.  I also added some tests.

roed314:
Though I haven't run them yet; doing so now....

JohnCremona:
The test all work for me son unless you have added more since db6862a... this is good to go.

roed314:
I've added info on the Jacobian and principal polarizations to the property box (and a test for it).  Let me know if I should push it here or wait to the next PR.

JohnCremona:
You don't make it easy for a reviewer if you keep on adding things.  I thought I had finished this job!  I am going to bed.

roed314:
Sorry!  I was just trying to address all the comments raised in the thread.  I thought you might have already gone to sleep, which is why I pushed the new addition.

Sleep well, and we can continue this later.  I'll also work on addressing Drew's suggestion.

JohnCremona:
As I see it there are 4 things in the properties box which do not appear elsewhere on the page; this breaks conventions, whereby the properties box is a small subset of the info on the page.  (I hope I am right in claiming that this is our convention: I have not looked for counterexamples.)

There is a circle diagram which -- I can guess -- shows the Frobenius angles, but nothing on the page says so explicitly.  (OK, so most elliptic curves show a small plot in their properties box which appears nowhere else...)  I would like to hear more opinions about this: perhaps there is room for a small caption "Frobenius angles" in the properties box?

Then there are p-rank (should be shown on the main page, with a knowl), "contains a Jacobian" and "principally polarizable".  But not "simple" and "primitive" which are on the main page (I am looking at Variety/Abelian/Fq/2/199/b_ef as a random example).  For these I suggest that somewhere on the main page you list 4 properties (primitive yes/no, simple yes/no, contains Jac yes/no and pp yes/no) in one block rather than 4 separate small blocks; and that all four are in the properties box.

jwj61:
My recollection is that the properties box should be a subset of what is on the main page, with the possible exception of a picture.  But, sometimes the picture is in both places (e.g., isogeny class of an elliptic curve).

JohnCremona:
Thanks -- so by implication if you add something to the properties box it must also be added to the main page somewhere.

JohnCremona:
Apologies to @roed314 : on an elliptic curve page such as http://www.lmfdb.org/EllipticCurve/2.0.7.1/2668.5/b/1
the Properties box says "Q-curve: no" but there's no mention of Q-curviness in the main page.  I'm now adding to that Prop box anyway (when a curve is not base-change it says nothing, but at @jvoight 's suggestion it will now say "Base change: no").

roed314:
I'm happy for feedback on the properties box; I certainly don't mean for it to be breaking LMFDB conventions.  I'm a bit confused by which properties you're referring to, since I think that for any individual isogeny class, everything in the property box should also be listed on the page in some way, though possibly in a different format (for example, since containing a Jacobian implies being principally polarizable, we don't say "Principally polarizable" on the main page in this case).

* There is a heading with p-rank on the main page (in the Newton polygon section), with a knowl.  Should there be knowls in the property box?  My impression was no....
* So, it's better to have "Jacobian: yes" than "Contains a Jacobian" in the property box?  I can see how the first fits more with the list of field:value pairs, but the second contains a bit more information (since not every abelian variety in the isogeny class will necessarily actually be a Jacobian).  I also find yes/no fields a bit awkward.
* When the Jacobian/principal polarizability are unknown, we're not displaying anything on the main page, since we thought it would be a bit weird to include a section just to say "We don't know."  Correspondingly, we don't have anything appear in the property box in these cases.  I think I like this approach better than listing "unknown" in both places, but it does have the consequence that the items in the property box aren't the same across all isogeny classes.
* I don't like the idea of having the four yes/no fields in a single section of the main page, since they're really dealing with different things.  The only thing they have in common is that they are binary choices.
* I don't know what the right thing to do to label the picture is.  You're right that this is just a picture of the roots in the complex plane, and is equivalent to giving the Frobenius angles.  Any update on what you think should be done in regard to this?  I think that we could have a caption on the line above....

Anyway, I'm happy to make changes, but I'm not sure what the right outcome is.

JohnCremona:
Sorry about this.  You are right about p-rank already being on the page ( and no we don't want knowls the the P. box).
For people who are less expert, why not add "and hence is principally polarizable" after "This isogeny class contains a Jacobian."?  I can imagine someone scanning down the page to check the pp property, and it's not so user-friendly if all they see is something which implies it.
Other than that I'm ready to merge this.  The other things I raised would be good to discuss at a meeting at some point perhaps.

roed314:
No problem; I think it's a good change (I also added the case that we know it's not a Jacobian but are unsure about principal polarizability, which does occur when there are negative point counts).

JohnCremona:
Good!

PR #2217, opened 2017-08-07
Added api2 base

csbrady-warwick:
Added basics for datastructures for the API2 layer.
See lmfdb.api2.utils.build_api_structure as the key. Any search on a page should return an object of that type


JohnCremona:
I'm looking at this with Chris.  It's a first step to provide some utilities and we'll make a second PR with an example of their use.

PR #2219, opened 2017-08-11
ecnf: improved base-change and Q-curve handling

### LINKED ISSUE #2218: Some elliptic curves over number fields have incorrect "Q-curve" flag
This is more of a data issue not a code issue.  At some point I realised that my Q-curve test was wrong (it gives false negatives, never false positives) and pending a correction to that code stopped computing that flag.  Two problems with this:  (1) curves whose Q-curve status is unknown ended up with False in the database; (2) curves whose Q-curve status was incorrectly computed never had the flag corrected.

Extent of the problem:  fields of degrees 2, 3, 4.  I think I stopped computing the Q-curve status during the degree 4 curve computations.

Solution:  first, to avoid any wrong data being present, I suggest that I will change all 'q_curve'=False to 'q_curve'='?' or similar.  This will be very quick, and some people will object since it will include curves which cannot be Q-curves for easy reasons.
Second, recompute or compute the Q-curve status of all curves.  This will take a while.  I'll start with quadratic fields which are easier;   a general algorithm which works even over non-Galois fields has not been implemented by anoone (to my knowledge) and I am working on it.  The point is that the definition of Q-curve is that E should be isogenous to all its Galois conjugates over Q-bar, and it is __not__ enough to check the curves isogenous to E over its field of definition (which we already have in the database, of course).

JohnCremona:
This does 3 related things:

   1. Handles cases where the 'q_curve' field is '?' (see Issue #2218 )
   1. Shows "Base change: no" in properties box instead of omitting it and only showing it when True (suggestion of @jvoight )
   1. Adds a section at the bottom of ecnf home pages about Q-curviness and base-change-ness.

For examples see EllipticCurve/2.0.8.1/34166.3/b/2 (neither bc nor qc), EllipticCurve/2.2.5.1/45.1/a/5 (bc and hence qc) .  I cannot yet give an example of a qc not bc because of the problems raised at #2218 but there will be.  However I hope this PR can be merged first since the data changes under way to fix #2218 will make some q_curve fields '?' and that is not handled i the current code.

### DIFF
--- a/lmfdb/ecnf/WebEllipticCurve.py
+++ b/lmfdb/ecnf/WebEllipticCurve.py
@@ -408,12 +408,15 @@ class ECNF(object):
             self.ST = st_link_by_name(1,2,'SU(2)')
 
         # Q-curve / Base change
-        self.qc = "no"
-        try:
-            if self.q_curve:
-                self.qc = "yes"
-        except AttributeError:  # in case the db entry does not have this field set
-            pass
+        self.qc = self.q_curve
+        if self.qc == "?":
+            self.qc = "not determined"
+        elif self.qc == True:
+            self.qc = "yes"
+        elif self.qc == False:
+            self.qc = "no"
+        else: # just in case
+            self.qc = "not determined"
 
         # Torsion
         self.ntors = web_latex(self.torsion_order)
@@ -544,7 +547,8 @@ class ECNF(object):
             self.properties += [('base-change', 'yes: %s' % ','.join([str(lab) for lab in self.base_change]))]
         else:
             self.base_change = []  # in case it was False instead of []
-            self.properties += [('Q-curve', self.qc)]
+            self.properties += [('base-change', 'no')]
+        self.properties += [('Q-curve', self.qc)]
 
         r = self.rk
         if r == "?":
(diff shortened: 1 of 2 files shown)


AndrewVSutherland:
Looks good, just one small nitpick: the wording on the base change text at the bottom of http://127.0.0.1:37777/EllipticCurve/2.2.5.1/45.1/a/5 (for example) is not quite grammatically correct.  The phrase "so is also" following the comma would be better as either "hence also" or "so it is also".

JohnCremona:
OK I'll fix that -- thanks.

JohnCremona:
Thanks -- great timing since the rewrite of nfcurves just finished (as in #2218 ) so I'll now rename the collections there and make sure that all works OK.

JohnCremona:
Small remaining issue, sorry: for a curve where the q_curve flag is neither True nor False but '?' it says in the cody of the page that it is a Q-curve since the logic is 
`It {% if ec.qc=="yes" %} is {% else %} is not {% endif %} a  {{KNOWL('ec.q_curve', '\(\Q\)-curve')}}.`
So there will be small follow-up PR before this goes to beta.

PR #2222, opened 2017-08-11
L-functions associated to Elliptic curves over real quadratic number fields

edgarcosta:
Hello,

First of all, do not merge this!

This is just a hack to show you the data the data of the L-functions associated to Elliptic curves over real quadratic number fields.
If you are happy with the data, then I would move it to the right collections.

Here are some examples, and their corresponding objects in the database:
* 2.2.5.1-49.1-a1 (a quadratic twis of http://www.lmfdb.org/EllipticCurve/Q/175/a/1)
http://beta.lmfdb.org/api/Lfunctions/ecqd_Lfunctions/?origin=sEllipticCurve/2.2.5.1/49.1/a
http://127.0.0.1:37777/L/EllipticCurve/2.2.5.1/49.1/a/
However, I have no way to automatically detect the quadratic twist, thus it's hash is: '_3835819448407117806631286663382'.

* The only two curves with rank > 2 in this set:
http://beta.lmfdb.org/api/Lfunctions/ecqd_Lfunctions/?order_of_vanishing=i3
http://127.0.0.1:37777/L/EllipticCurve/2.2.93.1/197.1/b/
http://127.0.0.1:37777/L/EllipticCurve/2.2.93.1/197.2/b/

* A base change:
http://beta.lmfdb.org/api/Lfunctions/ecqd_Lfunctions/?origin=sEllipticCurve/2.2.5.1/80.1/a
http://127.0.0.1:37777/L/EllipticCurve/2.2.5.1/80.1/a/

Let me know what you think.

ps: After I'm done with CM quadratic fields I will push the code that I used to generate all the data

Cheers,
Edgar

### DIFF
--- a/lmfdb/ecnf/WebEllipticCurve.py
+++ b/lmfdb/ecnf/WebEllipticCurve.py
@@ -504,7 +504,7 @@ class ECNF(object):
         if totally_real:
             self.hmf_label = "-".join([self.field.label, self.conductor_label, self.iso_label])
             self.urls['hmf'] = url_for('hmf.render_hmf_webpage', field_label=self.field.label, label=self.hmf_label)
-            self.urls['Lfunction'] = url_for("l_functions.l_function_hmf_page", field=self.field_label, label=self.hmf_label, character='0', number='0')
+            self.urls['Lfunction'] = url_for("l_functions.l_function_ecnf_page", field_label=self.field_label, conductor_label=self.conductor_label, isogeny_class_label=self.iso_label)
 
         if imag_quadratic:
             self.bmf_label = "-".join([self.field.label, self.conductor_label, self.iso_label])
--- a/lmfdb/lfunctions/Lfunction.py
+++ b/lmfdb/lfunctions/Lfunction.py
@@ -77,6 +77,7 @@ def an_from_data(euler_factors,upperbound=30):
 # As of July 2015, some of the fields are hard coded specifically
 # for L-functions of genus 2 curves.  Need to update after the
 # general data format has been specified.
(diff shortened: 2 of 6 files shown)


JohnCremona:
I'm sorry I keep forgetting to add the number of curves in the class to each curve in the nfcurves collection.  I could have trivially done that yesterday when I did a rewrite of the collection for a different reason.  (However I don't know why your need it.)
I'll take a look at this (don't worry, I know that you not want it merged).

edgarcosta:
I made a mistake, I did not compute the floating point representation the zeros with enough precision. I forgot to take into account the bits of precision that one might lose between decimal and binary representation.
Perhaps, would make more sense to store these as a quotient, e.g.:
3162021536343784910358546392868/2^100
or
3162021536343784910358546392868/1267650600228229401496703205376
instead of
2.4943951714884926440088943760845.

For the quadratic CM fields, I'm only 48 isogeny classes away, these are the CM cases where the L-function is a square.

edgarcosta:
@JohnCremona I have no clue why we need the number of curves in the isogeny class. I just tried to recreate what was already there. Just in case someone wants take some of this code and do it this the right way.

JohnCremona:
I thought it might have been so that an elliptic curve L-function page could list as related objects each curve in the isogeny class.  Perhaps that was the case once, but now (see e.g. http://www.lmfdb.org/L/EllipticCurve/Q/14/a/) there's a link to the isogeny class, not to individual curves, so the L-function need not know how many curves there are.

edgarcosta:
I thought the same...
Most likely it's a relic...

PR #2226, opened 2017-08-15
master --> prod

JohnCremona:
The easiest way to discuss this is to make the PR so we can easily see what changes there have been.  It's about a month since we last merged into prod.  I'll create the PR and then make a short summary.


JohnCremona:
Here's a summary of what is in here:
   1. @sanni85 's Hecke algebras -- will not be visible on prod yet anyway.
   1. @roed314 's Abelian variety enhancements.
   1. Various code refactorings which so not affect the website by @davidlowryduda and @csbrady-warwick 
   1. Lattice work from recent Warwick mini-workshop/
   1. Some small improvements to ecnf by me.

AndrewVSutherland:
It looks like this doesn't change the status of abelian varieties, in which case the only changes that will be obviously visible on prod are the lattice and ecnf changes.  It seems safe/good to push (and we can then check that abelian varieties is working as expected on prod by accessing the URLs directly.

JohnCremona:
Thanks @AndrewVSutherland  -- I'll let you do the honours!

PR #2233, opened 2017-08-18
HMF: remove references to 'AL_eigenvalues_fixed' in website code

JohnCremona:
This will enable us to delete the field 'AL_eigenvalues_fixed' from the forms collection.  It was only ever intended to be temporary.

I also slightly improved (I think) the Atkin-Lehner Eigenvalue section of the web page for an individual HMF in two cases: /ModularForm/GL2/TotallyReal/2.2.173.1/holomorphic/2.2.173.1-1.1-a where the level is 1 so there are none (it used to just say "None." as in http://www.lmfdb.org/ModularForm/GL2/TotallyReal/2.2.173.1/holomorphic/2.2.173.1-1.1-a) and where there are no AL eigenvalues in the database despite the level not being one (currently true for 4990 forms), for example ModularForm/GL2/TotallyReal/3.3.49.1/holomorphic/3.3.49.1-512.1-a (compare with http://www.lmfdb.org/ModularForm/GL2/TotallyReal/3.3.49.1/holomorphic/3.3.49.1-512.1-a which displays a table with one row of '?', regardless of the number of bad primes).

I added these examples to the tests.

One small point of discussion (see the first example above) is the text which says "This form has no Atkin-Lehner eigenvalues since the level is (1)."  and if someone has a suggestion for a better way to say that the level is the unit ideal, please say so!

### DIFF
--- a/lmfdb/hilbert_modular_forms/hilbert_modular_form.py
+++ b/lmfdb/hilbert_modular_forms/hilbert_modular_form.py
@@ -387,9 +387,6 @@ def render_hmf_webpage(**args):
         for eig in info['eigs']:
             if len(eig['eigenvalue']) > 300:
                 eig['eigenvalue'] = '...'
-        for eig in info['AL_eigs']:
-            if len(eig['eigenvalue']) > 300:
-                eig['eigenvalue'] = '...'
 
     info['level_ideal'] = teXify_pol(info['level_ideal'])
 
--- a/lmfdb/hilbert_modular_forms/hilbert_modular_form.py
+++ b/lmfdb/hilbert_modular_forms/hilbert_modular_form.py
@@ -369,16 +369,16 @@ def render_hmf_webpage(**args):
 
     info['hecke_polynomial'] = web_latex_split_on_pm(teXify_pol(info['hecke_polynomial']))
 
-    if 'AL_eigenvalues_fixed' in data:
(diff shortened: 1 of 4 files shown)


JohnCremona:
Needs work:  breaks ModularForm/GL2/TotallyReal/2.2.296.1/holomorphic/2.2.296.1-29.2-c (which has the record dimension of 286).  To be fixed...

JohnCremona:
...fixed.   In large dimension (i.e. when the Hecke field has large degree) the display_eigs field is False and the code replaces the eig strings with "..." for display.  But it does this for the AL eigs too, which is unnecessary (they are only +1 or -1!) and the previous changes here broke that since instead of a list of '?' (which would have been changed into "..." if present).

Up for review again.

jvoight:
Thanks @JohnCremona , looks good to me!  

It's not true that the Atkin-Lehner eigenvalue is always +/-1 because of class number issues.  It is always a root of unity, but because the representation of the Hecke field isn't sensitive to this, it may display some horrible long expression in a primitive element.  But this is a minor issue.

JohnCremona:
Good, thanks.  Of course it's easy to find an example of AL eigs not +/-1, e.g. http://beta.lmfdb.org/ModularForm/GL2/TotallyReal/2.2.60.1/holomorphic/2.2.60.1-165.2-s

PR #2234, opened 2017-08-21
rewrite HMF stats creation and retrieval to use db collections

JohnCremona:
I first wrote the HMF stats code before we moved to using database *.stats collections for this sort of thing, together with the utility functions to create them in data_mgt/utilities.   The old way used a function make_stats() in scripts/hmf/import_data.py to create a yaml file which was part of the git code repository, and the data read in from this at run-time.
After this change, make_stats() creates or updates two stats collections hmfs.forms.search.stats and hmfs.fields.stats and the code in lmfdb.hilbert_modular_forms.hmf_stats.py uses that.  It's a lot simpler and almost certainly faster.

As I write this I realise that the information in hmfs.fields.stats can be obtained also from hmfs.forms.search, meaning that we can manage with just one stats collection which is simpler.

So I'll come back and say this is ready for review in a short while.


JohnCremona:
OK, ready for review.   To test -- the point is that there should be no changes visible -- not literally true since I added some minor enhancements.
   1. ModularForm/GL2/TotallyReal/ shows total number of fields, not just by degree;
   1. ModularForm/GL2/TotallyReal/browse/ (by clicking "all available fields" from the top page) shows number of forms and max level norm, for each degree;

AndrewVSutherland:
I get 3 errors when I run the HMF tests (they all look like mongo db query failures).

JohnCremona:
Thanks for checking -- obviously this didn't happen when I last tested, so I'll look at it and fix it.

AndrewVSutherland:
My bad, I was running the wrong branch, it looks good.

JohnCremona:
Phew

PR #2235, opened 2017-08-22
Hmf collections reorganisation

JohnCremona:
This PR should come after #2234 since I based this work on that.

This is about using a new collection withing the hmfs database, called 'hecke', containing only the Hecke eigenvalues, polynomial and A-L eigenvalues.  This code will work whether or not the collection 'hecke' exists as it uses the original hmfs.forms if it does not, but I will shortly copy the hecke collection to the cloud anyway.

To make it easy to use, no-one outside the hilbert modular forms code need use any collection names at all thanks to two access functions written for this.  There is one example of that here already, in lfunctions/LfunctionDatabases.

There are some consequential minor changes to scripts, and I took the opportunity to clear up one or two pyflakes issues.


AndrewVSutherland:
So this looks fine, but there is a conflict in hilbert_modular_form.py that needs to be resolved before it can be merged.

JohnCremona:
I will fix that.  It is bizarre since I based this on the previous PR precisely to avoid this issue -- but github cannot lie.

JohnCremona:
That was mysterious.  On the command-line git happily merged.  And as far as I could see the master commit was a parent of the one on the original PR!  Anyway, I did look at the apparent conflict (one line was duplicated) and it should be fine now.

JohnCremona:
Thanks!  Now I'll push this to beta and we'll check that all is well.  (Then I'll move hmfs.forms and check that all is still well).

PR #2237, opened 2017-08-23
remove duplicate definitions

davidfarmer:
I deleted the duplicate definitions.  In most cases the definitions were identical,
so I kept the one which was in the section of the page with similar functions. 
In one case I deleted the old version.  Also moved truncatenumber into the appropriate section.

All tests passed except
   test_hecke_algebra (lmfdb.hecke_algebras.test_hecke_algebras.HomePageTest) ... FAIL
but when I go to that page I see the two strings which are tested.
Not sure what to do about that.

And
   test_errorMessages (lmfdb.lfunctions.test_lfunctions.LfunctionTest) ... FAIL
That is 12 separate tests, but I guessed that it might be the Hilbert modular forms.
Sure enough, when I go to
   /L/ModularForm/GL2/TotallyReal/2.2.5.1/holomorphic/2.2.5.1-31.5-a/0/0/
I see 
   Unable to render L-function page due to the following problem(s):
      'NoneType' object has no attribute '__getitem__'
instead of the expected
    No Hilbert modular form with label

I assume someone else will address that as part of the ongoing hmf work
(and probably you already know).

(Actually, there also were other errors, but I don't think they could
have anything to do with the changes I made.  For example, on
my laptop
    /ModularForm/GL2/Q/holomorphic/1/12/1/
tells me that the space is currently unavailable,
but other classical modular forms spaces seem to be working)

### DIFF
--- a/lmfdb/utils.py
+++ b/lmfdb/utils.py
@@ -202,6 +202,30 @@ def splitcoeff(coeff):
     return answer
 
 
+def truncatenumber(numb, precision):
+    localprecision = precision
+    if numb < 0:
+        localprecision = localprecision + 1
+    truncation = float(10 ** (-1.0*localprecision))
+    if float(abs(numb - 1)) < truncation:
+        return("1")
+    elif float(abs(numb - 2)) < truncation:
+        return("2")
+    elif float(abs(numb - 3)) < truncation:
+        return("3")
+    elif float(abs(numb - 4)) < truncation:
+        return("4")
+    elif float(abs(numb)) < truncation:
+        return("0")
+    elif float(abs(numb + 1)) < truncation:
+        return("-1")
+    elif float(abs(numb + 2)) < truncation:
+        return("-2")
+    elif float(abs(numb - 0.5)) < truncation:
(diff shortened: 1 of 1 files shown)


JohnCremona:
I spotted the Hecke failure also.  It's because the text on the page depends on whether the BETA flag is set.  We could change the test "assert xyz in data" to "assert xyz in data or abcabc in data".

I'll sort out the HMFs now that all recent changes are in both master and beta...

AndrewVSutherland:
These changes all look good/safe, so I will merge (we should still follow up on the errors David noticed).

JohnCremona:
After merging I find (1) no pyflakes errors except the 3 which I've been assured pass later versions, and (2) no test failures except the Hecke one and test_workshoplinks which often fails.

davidfarmer:
Thanks for checking.  I think I have some database issues
on my laptop.

JohnCremona:
Please don't tell me you still cannot see my 178,239 Bianchi forms?!

davidfarmer:
Okay, I won't tell you!

PR #2239, opened 2017-08-23
make HMFs work without forms collection

JohnCremona:
This does something which the previous PR forgot.  Unfortunately the only way to test that it wirks is to run it locally, then rename:
sage: hmfs.forms.rename('forms.bak')
then see if all HMFs and their L-functions work, then rename back:
sage: hmfs.forms.bak.rename('forms')

Of course in between the two renamings anyone trying to view an HMF one beta will be disappointed, but it need only be like that for a minute.


JohnCremona:
It would be great to have this merged and pushed to beta and prod before the weekend since I'll be away for 5 days after that.  The production database has the required collections already (updated forms.search and hecke).   It still has the old forms collection, but after this that will be redundant (saving the 80G currently duplicated there between forms and hecke).

JohnCremona:
Thanks.  I'll now push to beta, then hide the forms collection and make sure that all is well...

PR #2242, opened 2017-08-25
Not found is a 404

edgarcosta:
If we don't have the modular form on the database, then the flask should return a 404 HTTP code.

### DIFF
--- a/lmfdb/modular_forms/elliptic_modular_forms/views/emf_render_web_newform.py
+++ b/lmfdb/modular_forms/elliptic_modular_forms/views/emf_render_web_newform.py
@@ -43,9 +43,12 @@ def render_web_newform(level, weight, character, label, **kwds):
     info = set_info_for_web_newform(level, weight, character, label, **kwds)
     emf_logger.debug("info={0}".format(info.keys()))
     ## Check if we want to download either file of the function or Fourier coefficients
-    if 'download' in info and 'error' not in info:
+    if 'error' in info:
+        return render_template("emf_web_newform.html", **info), 404
+    elif 'download' in info:
         return send_file(info['tempfile'], as_attachment=True, attachment_filename=info['filename'], add_etags=False)
-    return render_template("emf_web_newform.html", **info)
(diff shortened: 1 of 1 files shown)


edgarcosta:
Perhaps, I should have created an issue about this.
But there are more examples on LMFDB that we raise an error, and we display the error in a nice way without returning the proper HTTP code.

Another example is:
http://beta.lmfdb.org/ModularForm/GL2/ImaginaryQuadratic/2.0.643.1/1.0.1/a/

and I'm sure there are more examples out there.

JohnCremona:
What difference exactly does it make for us to return a 404 as well as showing a nice error message page?

edgarcosta:
For the regular user nothing changes.

But is is the standard way to say that we didn't the object that was not requested.
At the moment this might only impact web crawlers and such....

JohnCremona:
OK -- can you give a URL where the behaviour is now different, and can we actually tell?  I'm just wondering how to test this, but will be happy knowing that there are pages where the new code runs and which don't crash ;)

edgarcosta:
Here is an example:
http://localhost:37777/ModularForm/GL2/Q/holomorphic/1020/2/1/a/

perhaps the easiest way to see that you get a 404 is to wget that url

edgarcosta:
```
$ wget http://www.lmfdb.org/ModularForm/GL2/Q/holomorphic/1020/2/1/a/ --no-verbose
2017-09-03 15:36:37 URL:http://www.lmfdb.org/ModularForm/GL2/Q/holomorphic/1020/2/1/a/ [11914/11914] -> "index.html" [1]
$ wget http://127.0.0.1:37777/ModularForm/GL2/Q/holomorphic/1020/2/1/a/ --no-verbose
http://127.0.0.1:37777/ModularForm/GL2/Q/holomorphic/1020/2/1/a/:
2017-09-03 15:36:42 ERROR 404: NOT FOUND.
```

JohnCremona:
OK, I get it.

Issue #2182, opened 2017-07-18
missing L-functions instances


edgarcosta:
I have noticed that there is only one instance for the L-function  http://www.lmfdb.org/L/EllipticCurve/Q/27/a/ (see http://www.lmfdb.org/api/Lfunctions/instances/?Lhash=903291523394155081) and that the L-function http://www.lmfdb.org/L/ModularForm/GL2/Q/holomorphic/42/2/1/a/0/
is computed on the fly.

Should we add the missing instances?
```
{'Lhash': u'903291523394155081', 'type': ??? , 'url': u'ModularForm/GL2/Q/holomorphic/42/2/1/a/0/'}
```
Should we at some point loop over all isogeny classes of elliptic curves over Q and had the instance associated to the modular form?

Further, regarding elliptic curves over quadratic number fields, when uploading the data for the L-function associated to an elliptic curve for which we have the modular form in the database, should I add the instance for the associated modular form?

Thanks!

davidfarmer:
We need to do all the modular forms L-functions, and do it properly.
There is limited value in just doing some of them.

A possible exception (mentioned by Andy) is to only add those which
are needed for the non-primitive L-functions associated to a degree 4
object which is being put into the database.

I think it is a mistake to add the modular forms L-function which are not
needed for a specific purpose.

I guess the ones with integer coefficients and weight 2 could be done
relatively easily, because we know the identifier of the modular
form.  But if someone has the time to do that, I suggest it would be
better for them to do something else.  (Those will not be used
on home pages, so there is not actually any benefit.  Once all the modular
forms L-functions are in the database, then we will change how the home 
pages are displayed.)

edgarcosta:
Let me practical then, should I add an entry to instances for the Hilbert modular form/Bianchi modular form, if such form exists in the database, when adding the L-function data for an elliptic curve over a quadratic number field?

Later on, I will also add the Hilbert/bianchi modular forms L-functions, however, in many cases we do not have enough Dirichlet coefficients to say something rigorous about the L-function.
(The same will happen for classical modular forms)

AndrewVSutherland:
This issue as been addressed for CMFs and L-functions of elliptic curves and genus 2 curves over Q by #2717, and there is a seperate issue for HMFs/BMFs and Artin reps #2768.

PR #2246, opened 2017-08-30
L-functions associated to Elliptic curves over generic number fields

edgarcosta:
Hello,

I have implemented and adapted existing code to deal with L-functions associated to Elliptic curves over generic number fields.
I tried to it in a generic enough way, so in the future we could perhaps convert the class Lfunction_EC to something like Lfunction_database, where one uses the database to figure out  connections, e.g., "Origins" and "Related objects", instead of hard coding them. (see set_bread_and_friends() in main.py).

Here are some examples:

- http://localhost:37777/L/EllipticCurve/2.0.11.1/256.1/a/
It would have been very hard for me to figure out those 6 origins without using the DB.

- http://localhost:37777/L/EllipticCurve/2.0.1879.1/1.0.1/a/
In this case, in the instances collection, there is an object that points to the Bianchi Modular form, even if this Modular form is not on the DB (yet). 
http://beta.lmfdb.org/api/Lfunctions/ecqd1_instances/?Lhash=s1703562946121056506229482417896
While the page is being generated, it checks if the  links to the possible origins/related objects really work, if not, we do not link to them and add an n/a.

- I tried to uniformize L-functions associated to modular forms and its corresponding Elliptic curve (by removing listing all the curves in the isogeny class)
http://localhost:37777/L/ModularForm/GL2/Q/holomorphic/15/2/1/a/0/
http://localhost:37777/L/EllipticCurve/Q/15/a/
Compare against: http://beta.lmfdb.org/L/ModularForm/GL2/Q/holomorphic/15/2/1/a/0/

- We display its factors as Related objects http://localhost:37777/L/EllipticCurve/2.2.5.1/80.1/a/

- We handle EC over Q almost the same way as before, but now there is code to deal with the corner cases with DB calls:
http://localhost:37777/L/EllipticCurve/Q/271914/a/
If you let me deal with #2182, by adding the respective instances, even if the modular form does not exist, then we could avoid the code dealing with corner cases

A couple of things missing:
- On the bread crumb, I always point to : http://localhost:37777/L/degree2/EllipticCurve/ 
Do we really want a template for each degree? I don't think so, but we should discuss that.
See line 603 in main.py.
- Remove Lhash collisions from the DB (I will do that once everyone is happy with the code and the data)
- Merge the collections
- On the long run, recompute L-functions for elliptic curves and add classical modular forms to instances.

Cheers,
Edgar


JohnCremona:
I'll be happy to look at this form a user / ec / mf viewpoint but someone who knows about L-functions should look at it even more carefully

JohnCremona:
I am enjoying browsing through your sample links.
   1. In the first example I assumes that we had a base-change since the origins include both a Bianchi form and a Hilbert form, but no!  They are twists of base-change.  Interesting.
   2. In the example over Q(sqrt(-1879)) the text " If p=1879, then Fp is a polynomial of degree at most 3. "  could be better worded!  There's only one such prime and there the degree is 2.  Details...
   3. I agree with the decision not to list individual curves -- the number in an isogeny class can get quite large.  But some people will not like having to click a second time to see an actual curve in detail.  We'll see what others think.
   4. In examples such as http://localhost:37777/L/EllipticCurve/2.2.5.1/80.1/a/ it would be nice to say on the page (somehow) what the link is to those two isogeny classes over Q.  Can we not give links to the degree 2 L-function factors themselves?
   5. Breadcrumbs: it is strange to me that the breadcrumb from a degree 4 L-function should link to a page which is specifically about degree 2.  That page (http://localhost:37777/L/degree2/EllipticCurve/) could include examples of higher degree, surely?  Despite the URL?  Also the Degree 4 L-functons browse page http://localhost:37777/L/degree4/ should have some examples like these.

This looks really great:  these are very small issues.

edgarcosta:
2. The code doesn't really know why p=1879 is a bad factor... I don't think we should try to accommodate for these cases

4.  Yes we can give the links. We could also have a sidebar box for the factors.

5. Perhaps would be more natural to get rid of that breadcrumb. Just like in genus 2 curves. e.g.: http://beta.lmfdb.org/L/Genus2Curve/Q/151672/d/

AndrewVSutherland:
@davidfarmer Do you want to take a look at this?

edgarcosta:
3. Sounds reasonable to me.  I need to look into how to template for that.

4. It will deleted before the final merge. It is just an easier way to see what I am replacing, without using diffs.

davidlowryduda:
Edgar and I chatted a bit about displaying the factors of L-functions within the properties box, and he was more than happy to let me figure that out. I've now done this, and it's in https://github.com/davidlowryduda/lmfdb/tree/gaga_additional_props

This sits on top of Edgar's pull request.

edgarcosta:
I have merged your commits.

davidfarmer:
I think it wold be more natural to link to the L-functions that are the factors.

When you are looking at /L/EllipticCurve/2.2.5.1/80.1/a/ ,
except for the URL you only know that you are on a purely L-function page.
So it is more natural to link to those L-function factors, and not just to one
of the sources of those factors.

Also, I think it would be nice to show the factorization in the main
part of the page (both because we generally like the properties box to just
repeat information from the main page, and also because the factorization may
be the most important thing about it!)

Dedekind zeta functions have a nice way to show factorizations:
http://beta.lmfdb.org/L/NumberField/4.0.189.1/

davidfarmer:
If I am on
/L/EllipticCurve/2.2.5.1/80.1/a/
and I click to the Hilbert modular form, and then click to its L-function,
I am not back where I started: I am at a generate-on-the-fly L-function home page.

I assume you do not intend to fix that in this pull request.  If so, please make an
issue, including some clues for the person who does that.

edgarcosta:
@davidfarmer I have created two issues regarding your comments. #2268 and #2269

JohnCremona:
I'm happy for this to be merged now.  I think the remaining issues are now Issues so will not be forgotten.  Do the others reviewers agree?

davidfarmer:
I agree it is ready and work can progress from there.

PR #2252, opened 2017-09-04
Fixing test_gamma0/1

edgarcosta:
The mongodb query for these tests was ill-composed. The 'weight' key was defined twice and used `$ge` instead of `$gte`.
I fixed both issues.

### DIFF
--- a/lmfdb/modular_forms/elliptic_modular_forms/emf_test_pages.py
+++ b/lmfdb/modular_forms/elliptic_modular_forms/emf_test_pages.py
@@ -107,13 +107,13 @@ class EmfTest(LmfdbTest):
         errors = []
         spaces = getDBConnection().modularforms2.webmodformspace
         wmax = 40; Nmax = 25
-        data = spaces.find({'weight':{'$ge':int(2)},'weight':{'$lt':int(wmax+1)},'level':{'$lt':int(Nmax+1)},'character':int(1),'version':float(version)})
+        data = spaces.find({'weight': {'$ge':int(2), '$lt':int(wmax+1)},'level':{'$lt':int(Nmax+1)},'character':int(1),'version':float(version)})
         print "Checking %d spaces with trivial character of weight w <= %d and level N <= %d"%(data.count(),wmax,Nmax)
         errors = self.check_spaces(data)
         if errors:
             print "Errors occurred for the following labels: ", errors
         wmax = 12; Nmax = 100
-        data = spaces.find({'weight':{'$ge':int(2)},'weight':{'$lt':int(wmax+1)},'level':{'$lt':int(Nmax+1)},'character':int(1),'version':float(version)})
+        data = spaces.find({'weight':{'$ge':int(2), '$lt':int(wmax+1)},'level':{'$lt':int(Nmax+1)},'character':int(1),'version':float(version)})
         print "Checking %d spaces with trivial character of weight w <= %d and level N <= %d"%(data.count(),wmax,Nmax)
         errors = self.check_spaces(data)
         if errors:
@@ -123,13 +123,13 @@ class EmfTest(LmfdbTest):
         errors = []
         spaces = getDBConnection().modularforms2.webmodformspace
         wmax = 10; Nmax = 50
-        data = spaces.find({'weight':{'$ge':int(2)},'weight':{'$lt':int(wmax+1)},'level':{'$lt':int(Nmax+1)},'version':float(version)})
+        data = spaces.find({'weight':{'$ge':int(2), '$lt':int(wmax+1)},'level':{'$lt':int(Nmax+1)},'version':float(version)})
         print "Checking %d spaces of weight w <= %d and level N <= %d"%(data.count(),wmax,Nmax)
         errors = self.check_spaces(data)
         if errors:
             print "Errors occurred for the following labels: ", errors
         wmax = 20; Nmax = 16
-        data = spaces.find({'weight':{'$ge':int(2)},'weight':{'$lt':int(wmax+1)},'level':{'$lt':int(Nmax+1)},'version':float(version)})
+        data = spaces.find({'weight':{'$ge':int(2), '$lt':int(wmax+1)},'level':{'$lt':int(Nmax+1)},'version':float(version)})
         print "Checking %d spaces of weight w <= %d and level N <= %d"%(data.count(),wmax,Nmax)
         errors = self.check_spaces(data)
         if errors:
--- a/lmfdb/modular_forms/elliptic_modular_forms/emf_test_pages.py
+++ b/lmfdb/modular_forms/elliptic_modular_forms/emf_test_pages.py
@@ -107,13 +107,13 @@ class EmfTest(LmfdbTest):
         errors = []
         spaces = getDBConnection().modularforms2.webmodformspace
         wmax = 40; Nmax = 25
-        data = spaces.find({'weight': {'$ge':int(2), '$lt':int(wmax+1)},'level':{'$lt':int(Nmax+1)},'character':int(1),'version':float(version)})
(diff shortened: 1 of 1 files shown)


JohnCremona:
Code changes look reasonable.  Can you give a sample URL+query to illustrate at least one of the fixes?

edgarcosta:
There is no URL. 
This some test code that isn't used in the website.
You can print the query (or the dictionary) to see that 'weight':{'$ge':int(2)} always got overwritten by 'weight':{'$lt':int(wmax+1)}.


Here is an example:
```
        data = spaces.find({'weight': {'$gte':int(2), '$lt':int(wmax+1)},'level':{'$lt':int(Nmax+1)},'character':int(1),'version':float(version)})
        print
        print data.explain()['queryPlanner']['parsedQuery']
        print data.explain()['queryPlanner']['winningPlan']
        print
        data = spaces.find({'weight':{'$ge':int(2)},'weight':{'$lt':int(wmax+1)},'level':{'$lt':int(Nmax+1)},'character':int(1),'version':float(version)})
        print
        print data.explain()['queryPlanner']['parsedQuery']
        print data.explain()['queryPlanner']['winningPlan']
        print
```
Output:
```
{u'$and': [{u'character': {u'$eq': 1}}, {u'version': {u'$eq': 1.3}}, {u'level': {u'$lt': 26}}, {u'weight': {u'$lt': 41}}, {u'weight': {u'$gte': 2}}]}
{u'inputStage': {u'filter': {u'$and': [{u'character': {u'$eq': 1}}, {u'version': {u'$eq': 1.3}}]}, u'inputStage': {u'direction': u'forward', u'indexName': u'level_1_weight_1_chi_1', u'isUnique': False, u'isSparse': False, u'isPartial': False, u'indexBounds': {u'chi': [u'[MinKey, MaxKey]'], u'weight': [u'[2, 41)'], u'level': [u'[-inf.0, 26)']}, u'isMultiKey': False, u'stage': u'IXSCAN', u'indexVersion': 1, u'keyPattern': {u'chi': 1, u'weight': 1, u'level': 1}}, u'stage': u'FETCH'}, u'stage': u'KEEP_MUTATIONS'}


{u'$and': [{u'character': {u'$eq': 1}}, {u'version': {u'$eq': 1.3}}, {u'level': {u'$lt': 26}}, {u'weight': {u'$lt': 41}}]}
{u'inputStage': {u'filter': {u'$and': [{u'character': {u'$eq': 1}}, {u'version': {u'$eq': 1.3}}]}, u'inputStage': {u'direction': u'forward', u'indexName': u'level_1_weight_1_chi_1', u'isUnique': False, u'isSparse': False, u'isPartial': False, u'indexBounds': {u'chi': [u'[MinKey, MaxKey]'], u'weight': [u'[-inf.0, 41)'], u'level': [u'[-inf.0, 26)']}, u'isMultiKey': False, u'stage': u'IXSCAN', u'indexVersion': 1, u'keyPattern': {u'chi': 1, u'weight': 1, u'level': 1}}, u'stage': u'FETCH'}, u'stage': u'KEEP_MUTATIONS'}
```

edgarcosta:
At the end of the day, I didn't really change the query, as we don't modular forms of weight < 2.

jwj61:
I thought there was work in progress for adding weight 1 modular forms.

edgarcosta:
Scratch that. 
Indeed there are 2 modular forms of weight 1 on the db
```
[{u'_id': ObjectId('57325a0e051af53345eb9f47'),
  u'bitprec': 53,
  u'character': 30,
  u'character_naming_scheme': u'Conrey',
  u'character_orbit_rep': 1,
  u'character_used_in_computation': 1,
  u'creation_date': datetime.datetime(2016, 5, 10, 16, 21),
  u'cuspidal': 1,
  u'dimension': 2,
  u'dimension_cusp_forms': 1,
  u'dimension_modular_forms': 2,
  u'dimension_new_cusp_forms': 1,
  u'hecke_orbits': [u'a'],
  u'level': 31,
  u'modification_date': datetime.datetime(2016, 5, 10, 23, 21, 17, 831000),
  u'prec': 10,
  u'space_label': u'31.1.30',
  u'space_orbit_label': u'',
  u'sturm_bound': 0,
  u'version': 1.3,
  u'weight': 1,
  u'zeta_orders': []},
 {u'_id': ObjectId('57328897051af54b829054eb'),
  u'bitprec': 53,
  u'character': 46,
  u'character_naming_scheme': u'Conrey',
  u'character_orbit_rep': 1,
  u'character_used_in_computation': 1,
  u'creation_date': datetime.datetime(2016, 5, 10, 18, 11),
  u'cuspidal': 1,
  u'dimension': 2,
  u'dimension_cusp_forms': 2,
  u'dimension_modular_forms': 3,
  u'dimension_new_cusp_forms': 2,
  u'hecke_orbits': [u'a'],
  u'level': 47,
  u'modification_date': datetime.datetime(2016, 5, 11, 1, 19, 19, 379000),
  u'prec': 10,
  u'space_label': u'47.1.46',
  u'space_orbit_label': u'',
  u'sturm_bound': 0,
  u'version': 1.3,
  u'weight': 1,
  u'zeta_orders': []}]
```

edgarcosta:
Perhaps no has run the test_gamm1 since May 2016, or these two forms do not break the test...

Anyway, this edits are just for some test code (that takes very long time to run) that had some ill-composed queries.

JohnCremona:
Running test.sh with no specific filename only runs tests in files with filename test*.  This might be deliberate since lmfdb/modular_forms/elliptic_modular_forms/emf_test_pages.py 
 (which I am currently running) seems to be quite long.

For automatic testing perhaps we should have standard tests as they are now, in files starting test* and longer ones which would not be auto-tested, perhaps starting longtest*.  I can already hear someone say that any test which is not run automatically will never be run.... but perhaps that is true of emf_test_pages.

AndrewVSutherland:
This is deliberate.  I like your idea of having an automatic test that 
runs longer tests, but perhaps we should just have it run anything named 
*test* rather than requiring a longtest prefix.

davidfarmer:
If it takes even longer to run the test we are supposed
to run before making a pull request, people are not going to do it.

I think it is reasonable to have some special extra-long tests
that are run occasionally, but which do not burden everyone every time.

AndrewVSutherland:
@davidfarmer Agreed, that is why I named the file emf_test.py rather than test_emf.py, so that it would not get run every time someone runs test.sh.

One of the virtues of using an automated testing procedure that gets run automatically is that it would remove this burden and make it feasible to run longer tests on a more regular basis (this is something that Edgar, John, and I have been discussing -- there are standard tools available to do this that also provide integration with github).

This would not impact our existing use of test.sh.  We could also set it up so that test.sh is run automatically when a pull request is submitted, with the results posted so that both the person submitting the PR and the reviewer could see them.

edgarcosta:
Be aware that tests in 
`lmfdb/modular_forms/elliptic_modular_forms/emf_test_pages.py`,  as they are written at the moment, will never be run by automatic testing. They take at least a couple of hours (I quit after a then after a couple of hours, as they were running on my laptop).
And the automated testing only gives us around 50 min of testing per task, where this includes booting the virtual machine, untaring sage, etc...

We could work around this by having  parallelizing the tests somehow. At the moment I'm only focused on getting something equivalent of tests.sh to run.

Cheers,
Edgar

AndrewVSutherland:
@edgarcosta, thanks, this is good to know.  I don't think it's urgent to automate emf_test_pages.py right now, we can continue to run them manually when appropriate.

AndrewVSutherland:
I'm merging this (the change is tiny and fixes a bug), but we should continue to think about automated testing (which is unrelated to this PR but came up as a tangent).

edgarcosta:
@AndrewVSutherland, thanks. 
I'm working on it. Most of my recent pull requests have been motivated by that.

PR #2254, opened 2017-09-06
sorting search results for abelian varieties

edgarcosta:
The search results for abelian varieties weren't sorted.
This made some tests works/fail depending from what server you were reading the data.

I sort them by: (dimension, finite field cardinality, label, p-rank).
I have also fixed the test to accommodate this and added the corresponding index to the development database.

### DIFF
--- a/lmfdb/abvar/fq/main.py
+++ b/lmfdb/abvar/fq/main.py
@@ -4,6 +4,7 @@ import time
 import ast
 import StringIO
 import lmfdb.base
+from pymongo import ASCENDING
 from lmfdb.base import app
 from lmfdb.utils import to_dict, make_logger, random_object_from_collection
 from lmfdb.abvar.fq import abvarfq_page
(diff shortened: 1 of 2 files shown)


edgarcosta:
Ups, I forget to add some examples:
Here is what I'm trying to fix:
http://www.lmfdb.org/Variety/Abelian/Fq/?q=&simple=any&g=2&p_rank=&newton_polygon=&initial_coefficients=&abvar_point_count=&curve_point_count=&decomposition=&count=
vs
http://beta.lmfdb.org/Variety/Abelian/Fq/?q=&simple=any&g=2&p_rank=&newton_polygon=&initial_coefficients=&abvar_point_count=&curve_point_count=&decomposition=&count=

or

http://www.lmfdb.org/Variety/Abelian/Fq/?q=&simple=any&g=&p_rank=&newton_polygon=[0%2C1]&initial_coefficients=&abvar_point_count=&curve_point_count=&decomposition=&count=
vs
http://beta.lmfdb.org/Variety/Abelian/Fq/?q=&simple=any&g=&p_rank=&newton_polygon=%5B0%2C1%5D&initial_coefficients=&abvar_point_count=&curve_point_count=&decomposition=&count=

AndrewVSutherland:
Looks good.  I will update the abelian varieties collection on the cloud to make sure it has all the same indexes.

roed314:
So, I'm not completely convinced that we should sort by label, since leading as encode negative values.  Wouldn't it be better to sort lexicographically by polynomial?  Of course, we should do this in an indexed way....

AndrewVSutherland:
@roed314 Sorry David, I may have merged too soon, shall I revert? (alternatively, you can change this to work the way you think it should in your next PR, let me know which you prefer)

edgarcosta:
I agree that sorting lexicographically by polynomial might be more natural. But for that, we would need to store the polynomial as a string (to be able to index it for sorting).

Anyhow, I'm not sentimentally attached to any sorting, my main goal was to have some sorting, so tests wouldn't fail.

roed314:
I can fix it along with other stuff.

PR #2256, opened 2017-09-08
bmfs efficiency changes

JohnCremona:
This avoids some slower searches.  Two new indexes are added (as shown by the additional entries in the import scripts).  Some complicated searches are replaced by post-processing of the search results.  Anything better seems to require more than some entries in a stats collection (which I did start to create but which is *not used* currently).   I need something like "for one field_label, give me all the database entries for which gl2_dims exists and for at least one weight has a postive new (or cuspidal) dimension".

I think this is fast enough (and faster than before) though the time has moved from a complicated search to the post-processing.  It only happens when dimension tables are created, not for anything to do with the forms themselves.  If people want it faster than this I could do a collection rewrite (for the dimension collection) which adds 4 new flags to each entry saying whether that entry was positive dimension for GL2/SL2 and cuspidal/new in any weight present.

### DIFF
--- a/lmfdb/bianchi_modular_forms/bianchi_modular_form.py
+++ b/lmfdb/bianchi_modular_forms/bianchi_modular_form.py
@@ -6,7 +6,7 @@ from pymongo import ASCENDING
 from flask import render_template, url_for, request, redirect, flash
 from markupsafe import Markup
 
-from sage.all import latex, Set
+from sage.all import latex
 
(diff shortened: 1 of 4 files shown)


edgarcosta:
Looks good to me.

JohnCremona:
Thank -- happy to merge then?

edgarcosta:
Yup. I just had to run the tests.

PR #2257, opened 2017-09-08
enable automatic testing through travis

edgarcosta:
This will allow us to use https://travis-ci.org to automatically run something equivalent to 
`tests.sh`.

Here is an example:
https://travis-ci.org/edgarcosta/lmfdb/builds/273341808

A key thing is that I had to built a special package of Sage that comes already with GAP. Otherwise, we would waste all our available run time building Sage.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,81 @@
+language: sage
+sudo: false
+dist: trusty
+
+addons:
+    apt:
+        packages:
+            # needed for sage
+            - binutils
+            - gcc
+            - g++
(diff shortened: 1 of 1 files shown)


JohnCremona:
Thanks for working out how to get this to work

edgarcosta:
Now, someone with admin rights must go here: https://travis-ci.org/profile/LMFDB 
and enable it for LMFDB/lmfdb

JohnCremona:
Done.  Will be triggered by next push.

PR #2264, opened 2017-09-11
added base field info for ecnf iso class and bmfs space pages

### LINKED ISSUE #2261: Isogeny class pages should define the generator
Pages like this

http://beta.lmfdb.org/EllipticCurve/2.0.11.1/108.4/b/

should say what "a" is.

JohnCremona:
This fixes #2261 (see EllipticCurve/2.0.7.1/16.1/CMa/) and similarly for Bianchi modular form space pages (see ModularForm/GL2/ImaginaryQuadratic/2.0.3.1/219.2).

### DIFF
--- a/lmfdb/bianchi_modular_forms/bianchi_modular_form.py
+++ b/lmfdb/bianchi_modular_forms/bianchi_modular_form.py
@@ -312,15 +312,11 @@ def render_bmf_space_webpage(field_label, level_label):
             else:
                 data = data.next()
                 info['label'] = data['label']
-                nf = WebNumberField(field_label)
-                info['base_galois_group'] = nf.galois_string()
+                info['nf'] = nf = WebNumberField(field_label)
                 info['field_label'] = field_label
                 info['pretty_field_label'] = pretty_field_label
                 info['level_label'] = level_label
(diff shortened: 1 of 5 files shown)


JohnCremona:
Anyone?  Merge this and I'll push the last several things to beta.

davidfarmer:
Checking now

davidfarmer:
Looks good.  Will merge.

Out of curiosity, on this page
ModularForm/GL2/ImaginaryQuadratic/2.0.3.1/219.2
why is the factorization
(17a−7)=((−2a+1)).((9a−8))
written with double parentheses on the right?

JohnCremona:
I don't know about the parentheses and always mean to look into it.  It's coming from Sage where the code is latex(I.factor()) where I is the ideal.  I think that the latex() function on a factorization object puts the "prime" factors in parentheses, and in this case they are ideals so already have parentheses.  It would be easy to so a replace op "((" to "(" and "))" to ")"; surely that is always OK.

davidlowryduda:
In the utils.py file, there is a utility called web_latex_ideal_fact that is used elsewhere exactly for this purpose. I will write a short change to use that here.

PR #2265, opened 2017-09-11
Add testing for some dynamic knowls

jwj61:
Adds 6 tests, 1.5 seconds

### DIFF
--- a/lmfdb/test_dynamic_knowls.py
+++ b/lmfdb/test_dynamic_knowls.py
@@ -1,0 +1,32 @@
+# -*- coding: utf8 -*-
+from base import LmfdbTest
+import unittest2
+
+class DynamicKnowlTest(LmfdbTest):
+    """
+    These tests check the functioning of some dynamic knowls.
+    """
+
+    def test_Galois_group_knowl(self):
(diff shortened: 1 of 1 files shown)


edgarcosta:
We souldn't ignore whenever the automatic test fails:
https://travis-ci.org/LMFDB/lmfdb/builds/274340819?utm_source=github_status&utm_medium=notification

JohnCremona:
As far as I can tell, whenever there's a pull request to master this triggers travis to run the check, so all it takes to check that the tests pass (which includes pyflakes) is to wait a bit.

We rather got used to having a few tests which failed for known reasons.  There's small danger now that we'll suspend those tests rather than fixing them.

edgarcosta:
The goal should be to always have the automatic tests to pass automatically.
If a known tests fails for known reasons, we should disable it or fix it.
We should really avoid a pull request to be merged if it triggers a failure on the automatic tests.
Otherwise after the merge, unrelated pull requests will start to fail...

JohnCremona:
I agree entirely -- what I meant was that people might be *tempted to* disable a test rather than fix it...

jwj61:
Sorry, I should have waited for the tests to finish.

edgarcosta:
no problem!

Issue #2260, opened 2017-09-09
Don't hard code width of coefficient table


davidfarmer:
The width of the coefficient table of Maass forms coefficients is set inline at 400px.

This forces the table below the properties box if the screen is narrow.
For example:

/ModularForm/GL2/Q/Maass/4f5556fe88aece241f000006

Unless there is some other cases where the fixed width is needed, it should be eliminated
or done properly with sitewide CSS.

I think this just involves deleting line 41
$('#coeffs').css('width','400px')
of
modular_forms/maass_forms/maass_waveforms/views/templates/mwf_one_form.html

Making an issue as a reminder to look into this later, or something
easy for someone to do.

JohnCremona:
I merged the fix at #2267, and hope that it does what David F wanted.

PR #2267, opened 2017-09-12
Remove CSS width of Maass coeff table

davidlowryduda:
As David Farmer noted in #2260, the coefficient table was hard coded for
Maass form pages. The width of the box for more coefficient information
at the bottom of the page was also hard coded. This removes both,
so that both CSS defaults to auto.

For a moment I considered setting the bottom table to have `width: 100%;`
so that it scales with the page, but this looks bad on small screens.

### DIFF
--- a/lmfdb/modular_forms/maass_forms/maass_waveforms/views/templates/mwf_browse_all_eigenvalues.html
+++ b/lmfdb/modular_forms/maass_forms/maass_waveforms/views/templates/mwf_browse_all_eigenvalues.html
@@ -15,10 +15,10 @@ var evTable = $("table#eigenvalues").dataTable(
             "sAjaxSource": "{{url_for('mwf.get_table')}}",
              "fnServerData": function ( sSource, aoData, fnCallback ) {
                     jQuery.ajax( {
-                        "dataType": 'json', 
-                        "type": "POST", 
-                        "url": sSource, 
-                        "data": aoData, 
+                        "dataType": 'json',
+                        "type": "POST",
+                        "url": sSource,
+                        "data": aoData,
                         "success": fnCallback
                     } );
                 },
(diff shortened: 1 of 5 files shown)


edgarcosta:
Travis failed because of pyflakes, all the others tests passed:

```
The command "${SAGE} -sh -c 'parallel --joblog joblog --jobs 4 --results nosetests  --progress --verbose -a list ${SAGE_DIR}/local/bin/nosetests -v -s --testmatch=${TESTMATCH} ${COVERAGE}'" exited with 0.
```

davidlowryduda:
Interesting. I do like the automated testing. I noticed in the pyflakes PR codecov then displayed a graph of code coverage. Is that going to happen on every PR (which I suppose pass travis)?

edgarcosta:
Yes, but we can change that:
https://github.com/codecov/support/wiki/Pull-Request-Comments

JohnCremona:
@edgarcosta which verson of pyflakes is running on Travis?  This may matter.

edgarcosta:
Travis always installs the latest version: `pyflakes-1.6.0-py2.py3-none-any.whl`
In either case, the issue introduced in  #2265:
```lmfdb/test_dynamic_knowls.py:3: 'unittest2' imported but unused```
would have been detected by lower versions.

JohnCremona:
Sadly the 3 LMFDB machine in Warwick (atkin, lehner, lmfdb) have version 0.8.1 as they are running ubuntu 14.04.  Even ubuntu 16.04 only has version 1.1.0.  I mention this since we have seen things which pass newer version of pyflakes but not older ones.

edgarcosta:
For the record, travis is also running on a ubuntu 14.04, however, I
install pyflakes through pip.

JohnCremona:
@edgarcosta I had not thought of that (running pyflakes via a pip install instead of a system-wide one).  The test.sh script just uses whatever's in the path though.  How did you get it to use a different one -- did you customise test.sh?

edgarcosta:
I'm not running tests.sh. I'm running something equivalent (see:
https://github.com/LMFDB/lmfdb/blob/master/.travis.yml under `script:` )
tests.sh is unsuitable for travis. It takes too long to run and does not
return the right exit code, etc...

JohnCremona:
Looks good to me.

PR #2271, opened 2017-09-13
More tests, commenting out unused functions

jwj61:
Also, imported a function from another file where it already existed rather than having a copy.


JohnCremona:
I suppose that the automated test passing means I don't have to, but I will anyway...

JohnCremona:
Another 0.43% on code coverage -- well done!

PR #2274, opened 2017-09-14
Correct double parentheses in bmf ideal display

davidlowryduda:
The bmf pages displayed double parentheses around factored ideals. This short request corrects that by having the bmf pages use the `web_latex_ideal_fact` utility function from utils.py (which really just removes the double parentheses).


JohnCremona:
Looks good .  Thanks.

PR #2276, opened 2017-09-16
read from m0 by default

edgarcosta:
We are allowing anyone to read directly from `m0.lmfdb.xyz`.
To reflect this, I'm changing the default values for port and host.

I'm also changing the Travis config file. Please wait for the tests to pass before merge.


edgarcosta:
Feel free to merge it.

Issue #2277, opened 2017-09-18
Simplify URLs after a search


JohnCremona:
I just did a search on cubic fields which are unramified outside {2,37}, of which there are 3 (one cyclic and 2 not).  This is beautifully simple, just go to http://www.lmfdb.org/NumberField and put 3 into the degree box and 2,37 into the ramified primes, selecting "subset of".  The result is the page whose URL displays as

http://www.lmfdb.org/NumberField/?degree=3&signature=&discriminant=&galois_group=&ur_primes=&ram_quantifier=contained&ram_primes=2%2C37&class_number=&class_group=&count=20&search=Search

which is very ugly as it contains blanks for lots of search fields I did not specify.  But I want to email that link to a colleague, and also put it in a paper!  I can manually edit the URL, deleting the irrelevant bits, to get

http://www.lmfdb.org/NumberField/?degree=3&ram_quantifier=contained&ram_primes=2%2C37

which is quite a lot better, only the encoding of the comma near the end being a bit weird.  (You can edit %2C into , and that works, at least on one browser).

Can we do better?  I can see many people who might want to show in a paper the complete list of objects in the database with some property like this, and having a more user-friendly URL would then be useful.  Is it possible to eliminate the blanks fomr the URL?

AndrewVSutherland:
Yes, we can do better.  Indeed, the genus 2 code already does this.

JohnCremona:
That is good to hear.  I don't have time to look into how right now, hence this issue.  Similar things would be good elsewhere for the same reason.

JohnCremona:
@jwj61 Would I be justified in writing in a paper that there are exactly three cubic fields unramified outside {2,37} _because_ the LMFDB only has these three?  Just wondering, I do have in independent proof (it is only an example).  I assume not.

davidfarmer:
If you go to the Jones-Roberts database and do the same search,
it says "Results below are proven complete."  Maybe some day we
will see those words in the LMFDB!

JohnCremona:
Thanks @davidfarmer

jwj61:
@davidfarmer is right.  I was recently thinking about porting that feature to the LMFDB.  It took me a long time the first time around.

[...]

PR #2278, opened 2017-09-18
Remove unneeded search parameters from some urls

jwj61:
Deals with global fields, local fields, Galois groups, and Artin representations.  In each case, look at the url after do a search or refining a search.  With Galois groups, there is almost no change (if you specify only the degree, then the only difference is whether or not t is in the url).  This addresses issue #2277 for these areas.

### DIFF
--- a/lmfdb/artin_representations/templates/artin-representation-index.html
+++ b/lmfdb/artin_representations/templates/artin-representation-index.html
@@ -52,7 +52,7 @@ Artin representation: <input type='text' name='natural' size='50' example='3.2e3
 <h2> Search for an Artin representation </h2>
 Enter values into one or more boxes to restrict the search.
 <p></p>
-<form>
+<form id="search" onsubmit="cleanSubmit(this.id)">
   <table>
     <tr>
         <td align="right">{{ KNOWL('artin.dimension',title="Dimension") }}</td>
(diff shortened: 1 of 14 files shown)


jwj61:
Codecov shows a difference on web_ec.py even though this pull request doesn't touch that file.  Maybe it has been changed since the last check done by codecov?

edgarcosta:
Most likely last time, the `test_random(self)` grabbed an curve with `len(isodegs) > 3`.

JohnCremona:
Fantastic (and fast!).  I will take a look.

I have also noticed some strange tiny fluctuations in coverage in files not touched by a PR.  @edgarcosta must be right about it coming from random tests.  We can ignore it surely.

AndrewVSutherland:
Looks good, just one comment: it might be nice to also strip out options that are set to there default value, for example, if I search for Galois groups of degree 3 I get the shortened URL

http://127.0.0.1:37777/GaloisGroup/?start=0&paging=0&parity=0&cyc=0&solv=0&prim=0&n=3&count=50

but the URL

http://127.0.0.1:37777/GaloisGroup/?n=3

does exactly the same thing.

jwj61:
I will look into it.

PR #2279, opened 2017-09-19
Edit GettingStarted.md to refer to m0.lmfdb.xyz

davidlowryduda:
Since #2276, the default configuration is to read from `m0.lmfdb.xyz`. I've edited the GettingStarted to account for these changes.

In particular, it is now much easier for someone to get up and running with a local copy. They install sage and necessary packages as before, but one can now immediately start up the server and go (as opposed to setting up an SSH key on lmfdb.warwick.ac.uk). The GettingStarted page now reflects this.

### DIFF
--- a/GettingStarted.md
+++ b/GettingStarted.md
@@ -1,9 +1,10 @@
 Installation
 ============
 
-* To develop and contribute new code, see below on [Sharing Your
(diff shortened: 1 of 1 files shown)


JohnCremona:
Thanks for doing this.  I always think that these instructions should be updated by someone who has relatively recently had to follow them!
It's possible we might want to add/subtract names of people who can install ssh keys (though of course fewer of these will now be needed).  Harald probably no longer wants to do it.  @AndrewVSutherland and @jwj61 also have these rights.  But I'll merge now anyway.

PR #2280, opened 2017-09-20
First try at number field statistics

jwj61:
This adds a stats page for number fields.  On one hand, this is a glaring omission.  On the other hand, it quickly becomes unwieldy.  Consider this a rough draft open for comments/suggestions.

### DIFF
--- a/lmfdb/genus2_curves/main.py
+++ b/lmfdb/genus2_curves/main.py
@@ -9,7 +9,7 @@ from operator import mul
 from flask import render_template, url_for, request, redirect, send_file, abort
 from sage.all import ZZ
 
-from lmfdb.utils import to_dict, comma, random_value_from_collection, attribute_value_counts, flash_error
+from lmfdb.utils import to_dict, comma, format_percentage, random_value_from_collection, attribute_value_counts, flash_error
 from lmfdb.search_parsing import parse_bool, parse_ints, parse_bracketed_posints, parse_count, parse_start
 from lmfdb.genus2_curves import g2c_page
 from lmfdb.genus2_curves.web_g2c import WebG2C, g2c_db_curves, g2c_db_isogeny_classes_count, list_to_min_eqn, st0_group_name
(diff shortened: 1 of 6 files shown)


AndrewVSutherland:
Nice, I like it!  A few suggestions:
- It would be nice to have column totals (with links) in addition to the row totals you currently have.
- Class numbers have been computed for (%) of the fields in the database is missing a number before the %.
- I might be interesting to have the class number stats broken down by field degree, as you do with all the others.
- Breaking down the Galois groups by t-number seems a bit over the top (you have to scroll horizontally a long way, and it makes the page pretty slow to load).  I might suggest either putting this on a separate page, or maybe just summarize by cardinality of the Galois group?
- An additional table that would be very interesting to me is one with columns for primes p < 100 (say) and rows for degrees that lists fields of a given degree  ramified only at a given prime.  Ideally it would  also identify table entries where the list is known to be complete (most of them I would guess).

jwj61:
Thanks for looking at this.

Would column totals be of interest for signatures?  You pretty quickly get a mix of totally complex fields for one degree, and just slightly complex fields for other degrees.  For Galois groups, it makes less sense.  The first column would basically be cyclic groups, but after that you are getting an odd mix.

I hope to work on it this week and will comment when I think it is ready for another look.

AndrewVSutherland:
You're right, totals on t-number make absolutely no sense, I was thinking of totals by cardinality of the Galois group (which would make sense), sorry about that.  As far signatures, well, I don't know, but certainly fields with no real places are collectively interesting, and I could imagine being interested in fields that have a fixed number of real places (maybe I have a theorem that lets me do something with each real archimedean place).

jwj61:
I think this is now a realistic candidate for merging.  Tables are big, but not 1900 columns wide.

There are more row/column totals in tables, but some searches like multiple signatures are currently not supported by the number field search page.  (A cure for that would be to replace the signature box with r_1 and r_2 boxes, but that is a different issue.)

I don't think people really think of fields by size of Galois group since that is so heavily influenced by the degree of the field.  Instead, we have small degree fields where people might know the t-numbers, and for familiar families of groups.

AndrewVSutherland:
Looks good, I like the changes you made.

PR #2285, opened 2017-09-26
test read from userbd -> knowledge

edgarcosta:
We now do a test read on knowledge database instead of the userdb.

ps: travis will still fail, but that is another issue


AndrewVSutherland:
This seems to work so I will merge it, but let's test it carefully on beta before pushing to prod.

edgarcosta:
I'm running tests on atkin, just to be sure

PR #2287, opened 2017-09-26
Weil poly

roed314:
This change means that the abelian varieties URLs aren't filled with lots of empty fields.

### DIFF
--- a/lmfdb/abvar/fq/main.py
+++ b/lmfdb/abvar/fq/main.py
@@ -132,15 +132,11 @@ def abelian_variety_search(**args):
                 query['is_simp'] = True
             elif info['simple'] == 'no':
                 query['is_simp'] = False
-        else:
-            info['simple'] = 'any'
         if 'primitive' in info:
             if info['primitive'] == 'yes':
                 query['is_prim'] = True
             elif info['primitive'] == 'no':
                 query['is_prim'] = False
-        else:
-            info['primitive'] = 'any'
         if 'jacobian' in info:
             jac = info['jacobian']
             if jac == 'yes':
@@ -151,8 +147,6 @@ def abelian_variety_search(**args):
                 query['is_jac'] = {'$lt' : 1}
             elif jac == 'no':
                 query['is_jac'] = -1
-        else:
-            info['jacobian'] = 'any'
         if 'polarizable' in info:
             pol = info['polarizable']
             if pol == 'yes':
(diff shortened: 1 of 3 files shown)


AndrewVSutherland:
Looks good, just one small issue: If you do an empty search on the main browse page, it keeps you on the main page rather than taking you to the results page.

roed314:
Great; I'll take a look once I get home.  Do you know why the Travis CI build failed?

roed314:
This is actually an annoying issue to solve.  Once you start using `onsubmit="cleanSubmit(this.id)"`, then the code for determining whether to load the browse page or the search results page has a tough time distinguishing between a fresh load of the page and an empty search.  The genus 2 code solves this by having another field where you can specify the maximum number of curves to display, and prepopulates this field with the value 50.  You can check that if you delete this 50 then clicking "Search" will take you back to the browse page.  But having that field means that you have a longer URL for your search results (`Genus2Curve/Q/?count=50`), countering the goal of https://github.com/LMFDB/lmfdb/issues/2277.

I don't oppose adding such a field for abelian varieties, but I wanted to check that it was the right plan first.  Also, I have a 6am flight, so I should go to sleep. :-)

AndrewVSutherland:
@roed314 You make a fair point, let me think about this.

roed314:
So, one possible solution would be to add a text link for the empty search (something like "All classes" near "A random isogeny class from the database.")  Then having the empty search keep you on the browse page feels okay, since the user didn't actually enter any data.

AndrewVSutherland:
@roed314 I don't think we want to change the existing behavior (clicking search with no inputs takes you to the results page with everything included), which is common throughout the LMFDB.

What I might suggest instead is to modify the cleanSubmit function in https://github.com/LMFDB/lmfdb/blob/master/lmfdb/templates/homepage.html to check if it has actually cleared out all inputs and selects, and if so, add an input "all=1".  This will be ignored when the query is processed, provided no one actually uses the name "all", which I think is safe to assume.  With this approach all non-empty queries can be made as short as possible (the "all=1" won't get added to non-empty queries), and the empty query will be just about as short as it can be (and easy to understand).

Assuming you agree that this is a reasonable approach, I'm happy to merge this PR and then make the change to cleanSubmit in a separate PR (I don't think it is a good idea to include it in this one).

roed314:
Yeah, that sounds like a reasonable approach.  Once you merge this pull request, should I make a new one for adding abvar/fq_isog to the sidebar on www.lmfdb.org and e-mail the lmfdb list?

AndrewVSutherland:
OK, I will merge this, and then make the cleanSubmit change.  Why don't you wait for the cleanSubmit PR to be merged, then create a PR to move abvar to the sidebar on www.lmfdb.org.

roed314:
Sounds good.  Can you let me know here once the cleanSubmit change has been merged?

AndrewVSutherland:
OK the cleanSubmit change has been merged and I just pushed this PR and it to beta.  Just let me sanity check everything on beta, and then you should be good to go.

roed314:
Great.  I just got the number field data from John Jones, so I'm going to try to add that before the sidebar change.

roed314:
See https://github.com/LMFDB/lmfdb/pull/2290

PR #2288, opened 2017-09-29
Cleansubmit

AndrewVSutherland:


JohnCremona:
I'm arriving (or returning) late at this party.  Where should I look for how to use this cleanSubmit facility elsewhere?

AndrewVSutherland:
@JohnCremona All you need to do is bind cleanSubmit to onsubmit in your browse and search results templates.  For example, if you change line 10 of https://github.com/LMFDB/lmfdb/blob/master/lmfdb/elliptic_curves/templates/ec-search-results.html to

<form id="re-search" onsubmit="cleanSubmit(this.id)">

the cleanSumit javascript wrapper (implemented in the homepage.html template) will automatically get called when the user clicks the search again button and strip out any empty query parameters.

Note that for query parameters that automatically get set to a default value will not get stripped.  This means that it's actually preferable *not* to explicitly set the default value (e.g. in the elliptic curve pages, the default for optimal_only is no, but the value "no" is not actually set, so this parameter will get stripped, which is good, in constrast, the default for CM is "include" and the value "include" is actually set, whereas leaving it blank would produce exactly the same behavior and result in a shorter URL).

JohnCremona:
Thanks.  For future reference the first blank line in the above post contains html which (for me) is rendered and hence unreadable -- it says

`<form id="re-search" onsubmit="cleanSubmit(this.id)">`

lmfdb/templates/homepage.html
```diff
@@ -257,12 +257,13 @@ <h1 class="properties-header">Properties</h1>
...
-            var item, i;
+            var allInputs = myForm.getElementsByTagName('input');
+            var allSelects = myForm.getElementsByTagName('select');
+            var item, i, n = 0;
 
-            for(i = 0; item = allInputs[i]; i++) if(item.getAttribute('name') && !item.value) item.setAttribute('name', '');
-            for(i = 0; item = allSelects[i]; i++) if(item.getAttribute('name') && !item.value) item.setAttribute('name', '');
+            for(i = 0; item = allInputs[i]; i++) { if (item.getAttribute('name') ) { if (!item.value) { item.setAttribute('name', ''); } else { n++ }; } }
+            for(i = 0; item = allSelects[i]; i++) { if (item.getAttribute('name') ) { if (!item.value) { item.setAttribute('name', ''); } else { n++ }; } }
+            if (!n) { var all = document.createElement('input'); all.type='text'; all.name='all'; all.value='1'; myForm.appendChild(all); }
```
edgarcosta: @AndrewVSutherland , do you recall the purpose of this line?

AndrewVSutherland: Not off the top of my head (it was a year ago), but I do remember it being necessary in some edge cases.

edgarcosta: Thanks!
@alexjbest figured it out.
It is the difference between http://lmfdb.xyz/EllipticCurve/Q/?all=1 vs  http://lmfdb.xyz/EllipticCurve/Q/?

PR #2297, opened 2017-10-04
Added bcrypt based password hashing

csbrady-warwick:


lmfdb/users/pwdmanager.py
```diff
@@ -146,14 +160,34 @@ def authenticate(self, pwd):
...
+            for i in range(rmin, rmax + 1):
+                if self._data['password'] == hashpwd(pwd, str(i)):
+                    # log "AUTHENTICATED after %s!!" % (time() - t1)
+                    bcpass = bchash(pwd)
+                    if bcpass:
+                        logger.info("user " + self._uid  +  " logged in with old style password, trying to update")
+                        try:
+                            self._data['bcpassword'] = bcpass
+                            get_users().update_one({'_id': self._uid},
+                                               {'$set':{'bcpassword':bcpass}})
```
edgarcosta: Shouldn't we also remove the old `password` entry, i.e., the old hash?

csbrady-warwick: I considered that, but that would prevent people from logging on using older versions of the LMFDB source that they were running themselves. I have no idea if that would be a problem, but I though it best to not unilaterally make it happen. My plan was to leave the dual system running for a while and then purge the unused ones later.

edgarcosta: I wouldn't worry very much about that. I have the impression that no one logs in into the LMFDB using their local version. @JohnCremona, what do you think?

edgarcosta:
It looks good to me. 
I believe that the reason behind the switch is that `bcrypt` is theoretically slow and not GPU friendly, thus we would be much safer having these hashes public.

JohnCremona:
As far as I know the only reason for being logged in is to edit knowls, which as we know (though I cannot remember the reason) can only be done on beta, not a local version.  And also for those with admin rights to generate a key for new registrations, but again there is no reason for that on beta.

Hence it does not matter very much that local versions might not allow any login any more, but unless we explicitly all agree that login on a local version is deprecated I would be in favour of as much backwards-compatibility as possible.

Am I right that after this code change users will still be able to login to beta using their old password and not notice that anything behind the scenes has changed at all?

csbrady-warwick:
Absolutely John. The end user wouldn't even know that anything happened at all, the transition should be completely seamless.

edgarcosta:
> I would be in favour of as much backwards-compatibility as possible.

Then I would merge it. We just need to remember ourselves one day to purge the old hashes.

JohnCremona:
We have not yet merged this change into beta.  Before we do I must make sure that the sage/python installation used by the beta server has brypt.  I'll report back here when that is done.

JohnCremona:
OK I did that and started running the test script on lmfdb.warwick.  It is running OK but I just found a bad side-effect of us having changed the defauilt mongo connection from localhost to m0.lmfdb.xyz !!  Right now when I run lmfdb locally on lmfdb.warwick.ac.uk itself it uses m0 which is silly.  I think that this does not effect the real beta and prod servers on lmfdb.warwick but I am not sure about that, and we need to be sure.
[Sorry this is off the topic of this PR]

JohnCremona:
We do have something which needs fixing on lmfd.warwick.  The beta and prod servers there run from /home/lmfdb/lmfdb-git-beta and lmfdb-git-prod respectively, which have no file mongoclient.config so use the default which connect to m0.  Can someone confirm to me that I need to put mongoclient.config files in both those places using localhost and port 37010?   Otherwise beta will be slower than necessary, and anything like editing knowls will be impossible?  (But I just checked and I can edit knowls so I am puzzled.)

AndrewVSutherland:
We should wait for @edgarcosta to confirm/clarify.  Let's not merge anything to beta until we are certain we understand the situation.

edgarcosta:
@JohnCremona, they both have a mongoclient.config 
```
lmfdb@lmfdb:~$ ls lmfdb-git-*/mongoclient.config
lmfdb-git-beta/mongoclient.config  lmfdb-git-prod/mongoclient.config
```
and the file comes from here:
https://github.com/LMFDB/warwick-nt-cluster/blob/master/config/mongoclient.config

edgarcosta:
Also a relevant side note, on atkin I have 
```
host = lmfdb-ib
```
to use the infiniband structure.

JohnCremona:
@edgarcosta Of course, not sure how I missed that when I looked -- my terminal shows symbolic links in a not-very-visible colour.  Ignore the comment.

Issue #2269, opened 2017-09-12
Display the L-function factors


edgarcosta:
With the new standard for the (analytic) `Lhash` function we should be aim to display its the L-function factors as it is done in Dedekind zeta functions see:
http://beta.lmfdb.org/L/NumberField/4.0.189.1/

However, we should be pointing at, `L/Lhash` and not `L/<url for a possible origin>`. (this depends on #2268)

davidlowryduda:
That makes sense. When I added the factors to the properties bar (now in #2246), I hadn't noticed that we did this for a class of L functions already. Using `\rho_{<object label}` is a good thought that readily generalizes.

edgarcosta:
I agree, but sometimes it will not be clear which one is the right label to pick. For example, `L/EllipticCurve/2.0.11.1/256.1/a/` has four possible labels, and 8 possible origins:
Isogeny class 2.0.11.1-256.1-a
Isogeny class 2.0.11.1-256.1-b
Isogeny class 2.2.44.1-16.1-a
Isogeny class 2.2.44.1-16.1-c
Bianchi modular form 2.0.11.1-256.1-a
Bianchi modular form 2.0.11.1-256.1-b
Hilbert modular form 2.2.44.1-16.1-a
Hilbert modular form 2.2.44.1-16.1-c

davidlowryduda:
Yes, the lack of a canonical label is unfortunate, and unrectifiable. I wonder what would be a good alternative? We would want to display something of the form `L(s, ?)`. It doesn't seem practical to me to list all possibilities.

Is choosing any label (perhaps the first label in the list?) a bad idea? To then see where else this L-function appears, one would have to click on it and then check the origins on that page.

davidfarmer:
There is a proposal for labeling L-functions, purely as an L-function.

It encodes more meaningful information than the Lhash.

I think I wrote that up once.  I'll look.

davidlowryduda:
@davidfarmer Did you happen to find your proposal for labeling L-functions as an L-function?

AndrewVSutherland:
This issue is duplicated by #2653.  I think the discussion on that issue is more current so I am closing this one.  We are all agreed that we want to show the factorization of imprimitive L-functions and that we need a way to refer to L-functions in an origin-independent way (there is a proposal for how to do this that is currently under discussion, see #2876).

Issue #2268, opened 2017-09-12
Try to first read an L-function from the database, before creating it on the fly


edgarcosta:
Overall, I believe we should change our approach regarding L-functions.

When handling a request of the type: `L/url` we should first try to see if `url` is in the database.
If it is, get all the data from the database, the code should not need to know anything about the `url`, indeed, the code should only depend on the `Lhash`.

Only, if the `url` is not in the database we should use lcalc.

Perhaps, as first step we should handle requests of the type: `L/Lhash`

davidlowryduda:
I agree that it would be nice to be able to refer to `L/Lhash` instead of constructing the urls. If this is something that others agree on (and unless this is something that you are doing), I can implement handling of `L/Lhash`. I'm particularly interested in doing this to facilitate #2269 (though there are some details there that remain to be hashed out).

edgarcosta:
I can help you to sort out things, but I won't be doing any major LMFDB development for a while.

JohnCremona:
I'm not sure I understand what is being suggested here.  When a user is viewing an L-function after clicking on a link from an actual source (e.g. an elliptic curve) then the URL they see should show that source.  But now if that L-function "knows" that it is a product and wants to show links to the factors, is that what you are wondering how should it be displayed (which URL)?  No easy answer -- there may not be a source of the same kind as the original, for sure.  Is that the situation where people envisage showing a technical URL involving the hash?  OK, perhaps, as long as that page itself shows all the sources of that L-function it knows about (without having to decide on one of those).

edgarcosta:
I created two issues to try to keep the discussions separate.

Here I'm proposing, that if possible, the code itself only uses the `url` of the origin to fetch the `Lhash`, and then build the L-function and everything associated to it from the data on the database.
and if that fails, then it should use it's lcalc approach for it.

At the moment, there is a lot of repeated and hacky code in https://github.com/LMFDB/lmfdb/blob/master/lmfdb/lfunctions/Lfunction.py
and the collection Lfunctions.Lfunctions is not consistent all throughout.
If our long term goal is to have every L-function being served from the DB we should stop writing a class for each case. We should instead force ourselves to upload whatever it is needed to the DB and then handle it all in a abstract way.

Having something `L/Lhash` or `L/L-function label` working would be a proof of concept that the abstraction is indeed working.

AndrewVSutherland:
I believe this issue has been addressed in all the situations where it is still relevant (e.g. Artin L-functions).

PR #2299, opened 2017-10-05
Create pages at /L/lhash/<Lhash>/

davidlowryduda:
This is a progress update on having pages directly for each L-function.
We began thinking about doing this in #2269 and #2268.

There are some design choices that need to be made, which is why this
is a good place to stop and consider. This pull request is not ready to be
pulled right now.

With this pull request, it is possible to go to, for example, the L-functions

    http://localhost:37777/L/lhash/1938322253992393114/
    http://localhost:37777/L/lhash/dirichlet_L_6253.458/

These pages are arrived at from the lhash and values stored in the
database. For initial comparisons, other places where this exact
L-function appear in the "Instances" box in the gaga box. [In the
future, we may want to have a canonical page for each L-function, and
so this Instances box would need to be removed. I have indicated the
lines to remove in the code with `#TODO DLD` marks.] It is good to
compare the Lhash version against another instance, to see some
differences.

In the example above, there are a few obvious differences. Most of these
are because this "canonical" Lhash page doesn't know what sort of
L-function it is, and so it omits things displayed using this
information.

Let's see some of these for the first example page above.

1. The "Related objects" box on the side doesn't appear.

  This is because the related objects (created in `lfunctions.main.py`)
  populated the topics on the gaga box differently based on the type of
  the L-function. But Lhash doesn't have a type.

2. The Euler Product is displayed incorrectly.

  The Euler product is essentially gathered in `lfuncEPtex` in
  `lfunctions.Lfunctionutilities.py`. This has different cases depending
  on the type of the L-function. So an elliptic curve Euler product is
  constructed differently than a Dirichlet L-function's Euler product,
  for instance.

  I note that the good and bad factors are stored in the database. So it is
  possible to rewrite the Euler product constructor to produce the
  correct output using only the contents of the database. There is a
  question of presentation, though. Do we want a Dirichlet L-function's
  Euler product to have the correct `chi` displayed, for instance?

3. The plot of the Z-function doesn't appear.

  The plot is presented by modifying the url of the L-function to arrive
  at the url for the plot. It is possible to still do this, and just
  handle urls for the plots under Lhash's, but as this is an initial set
  of changes I thought that handling the plot is less important.


edgarcosta:
As you can see my definition of "week" is quite wide.

1. Having some missing origins is okay by me, you can always force those to show up by adding an entry to the instances DB (even though the object doesn't exist).

On the same note, the related objects in one case are the symmetric square/cube, I'm not sure why we can't have those for every page, but if I recall correctly those are computed on the fly.

2. This is the trickiest one. I would still want to show the Euler factors, even if they are straightforward as in the case of the Dirichlet L-function. However, in that case I also want to write that each factor as a really simple description, given by chi. 
I'm not sure how to solve this one.
Perhaps, this could be an optional field on the DB? Where else you think one would like to have such description for the Euler factors?

3. Yes, I would not worry about the plot at the moment.

4. The name should be: "L-function of degree D, weight W, conductor N, and X character" with D, W, N, and X replaced accordingly.

AndrewVSutherland:
@davidlowryduda What is the status of this PR.  Are you planning to make changes in response to @edgarcosta's comments?  Is their a timeline for when you think this will be ready to be merged?  It makes me nervous having PRs open for such a long period of time.

@davidfarmer do you have any feedback on this PR?

davidlowryduda:
@AndrewVSutherland I think I got bogged down in thinking about how things should ideally be a bit too much. Thank you for the prompt. But I have in mind a path that should work.

I will make some changes per @edgarcosta's comments. I'll do this within the next week, and I think this can be made ready for merging shortly after. Now that you mention it, this PR has been open for a long time.

davidlowryduda:
I believe this is ready to merge. Let me describe how I handled Edgar's comments.

1. I agree that having some missing origins is okay. It's more important to have access to the other instances of this L-function, and then one can determine their origins from their own pages.

2. I've removed the Euler product from the page entirely. To show the Euler product would require some additional information to be held in the database. Instead of displaying the Euler product, I have a bit of text encouraging viewers to check out other instances of that L-function.

3. Good. Instead of being blank, it now displays "Plot not available."

4. The name is now what you suggest.

@edgarcosta Would you mind taking another look?

davidlowryduda:
Thanks to Edgar for pointing something out to me, plots are now available. So my point (3) has changed, and now plots are handled appropriately.

edgarcosta:
I took a look and all looks good.

The failure is because: http://icms.org.uk/workshops/onlinedatabases from http://www.lmfdb.org/acknowledgment/activities got moved or deleted. We perhaps should stop testing that. @JohnCremona, what do you think?

JohnCremona:
Note that master now has a fix for the onlinedatabases link, so as long as the tests do now pass I am happy with this.

davidlowryduda:
Ok. I'll just pull the new changes to master and travis will rerun.

JohnCremona:
@davidlowryduda Could you resolve the merge conflicts?

davidlowryduda:
@JohnCremona Which merge conflicts? Is this not fully merged already?

I would be happy to resolve any conflicts, but I'm not quite sure how to see these conflicts. My local branch is fully merged without conflict, though.

JohnCremona:
Apologies, last time I looked here there was a red X saying it coudn't be merged, but I had already merged it so I have no idea what that was about.  Ignore!

Issue #2281, opened 2017-09-22
Cloud DB contains publically readable users table


hratcliffe:
The new cloud mirror on m0(dot)lmfdb(dot)xyz appears to contain the users table with publicly accessible emails and password hashes of everybody signed up for the webpage. For example using nothing but a new clone of the project:

```from lmfdb import base 
users = base.getDBConnection().userdb.users
a = users.find()
#Print emails for all records
for el in a:                               
    print(el['email']) 
```
This table should either not be included, should be locked down or should be fully encrypted.

edgarcosta:
Thanks!

I agree we should not have those public. We should do this within the useradmin of MongoDB.

I closed down the port temporarily, until someone edits the permissions of the default user.

However, no one should expect their passwords and emails to be secret. At the end of the day, we do not even use SSL when authenticating....

edgarcosta:
I created a new role with the following code:
```
roles = []
for db in C.database_names():
    if db not in ["local", "config", "userdb", "admin", "contrib", "upload","test"]:
        roles += [{u'role':u'read', u'db': db }]
privileges = [{ u'resource': { u'db': u"", u'collection': u"" }, u'actions': [ u"listDatabases" ] }]
role_dict = {u"role" : u"ReadUncensoredDatabases", u"privileges" : privileges, u"roles" : roles}
```
which gives:
```
{u'privileges': [{u'actions': [u'listDatabases'],
   u'resource': {u'collection': u'', u'db': u''}}],
 u'role': u'ReadUncensoredDatabases',
 u'roles': [{u'db': u'HTPicard', u'role': u'read'},
  {u'db': u'Lattices', u'role': u'read'},
  {u'db': u'Lfunction', u'role': u'read'},
  {u'db': u'Lfunctions', u'role': u'read'},
  {u'db': u'MaassWaveForms', u'role': u'read'},
  {u'db': u'SL2Zsubgroups', u'role': u'read'},
  {u'db': u'abvar', u'role': u'read'},
  {u'db': u'artin', u'role': u'read'},
  {u'db': u'bmfs', u'role': u'read'},
  {u'db': u'curve_automorphisms', u'role': u'read'},
  {u'db': u'elliptic_curves', u'role': u'read'},
  {u'db': u'genus2_curves', u'role': u'read'},
  {u'db': u'halfintegralmf', u'role': u'read'},
  {u'db': u'hgm', u'role': u'read'},
  {u'db': u'hmfs', u'role': u'read'},
[...]

PR #2301, opened 2017-10-06
Removing all mention or usage of emails

edgarcosta:
Starting to address #2281 

@JohnCremona Could you check that the registration page looks okay?

### DIFF
--- a/lmfdb/base.py
+++ b/lmfdb/base.py
@@ -256,14 +256,6 @@ def nl2br(s):
     return s.replace('\n', '<br>\n')
 
 
-@app.template_filter('obfuscate_email')
-def obfuscate_email(email):
-    """
-    obfuscating the email
-    TODO: doesn't work yet
-    """
-    return u"%s…@…%s" % (email[:2], email[-2:])
-
 
 @app.template_filter('urlencode')
 def urlencode(kwargs):
--- a/lmfdb/users/main.py
+++ b/lmfdb/users/main.py
@@ -246,7 +246,6 @@ def register_token(token):
             return flask.redirect(url_for(".register_new"))
 
         full_name = request.form['full_name']
-        email = request.form['email']
         next = request.form["next"]
 
         if pwdmanager.user_exists(name):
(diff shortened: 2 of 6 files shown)


JohnCremona:
OK I'll take a look.

JohnCremona:
I cannot connect to m0.lmfdb.xyz.  I don't think I am the only one.  What's up?

JohnCremona:
Using the ssh tunnel to lmfdb.warwick (from home).  I cannot login when running the website locally.  I know the username and password are correct since they work on beta.  I don't know the reason.  This is not caused by this branch, I also cannot login when running the master branch locally.  I think that this used to be possible -- can anyone confirm or otherwise?

edgarcosta:
1- I have closed the m0 port as a response to #2281.
Only now I had time to work on this.
Once I manage to get to my gate and get some stable internet I will replace
all the emails in the DB by email@example.org
And later on when this is merged I will delete those entries as we agreed.

2- what is the error that you get? Are you trying to login as webserver? I
will also try to see how it goes.

JohnCremona:
So right now anyone running master with default config will fail since the new default is to use m0 but that fails!  

The login failure I get is just "wrong password" but the same password is OK on beta and that is using the same database.  But the current master branch has the new bcrypt code which beta does not so that might be the problem.

In the short term I suggest replacing the default config back to using localhost and 37010 so the tunnel works as before.  OK?

JohnCremona:
@csbrady-warwick is behind the log message which appears when I failed to log in:
WARNING:users@2017-10-06 19:48:02,606: user cremona logged in with old style password, but update was not possible [pwdmanager.py:188]
I think this must be an authentication failure.  I am using a passwords.yaml file which still has the old admin password in it.

JohnCremona:
I still cannot check the registration page since when running locally I am not authenticated on the userdb.  Unless someone has a way round that (temporairly for this test) we'll have to merge this into master and beta and then see how the new user registration looks then.

edgarcosta:
Port opened and all emails set to email@example.org, just in case I forgot
something stupid.
I saved the old emails, but I don't plan to use them for anything.

Cheers,
Edgar

edgarcosta:
@JohnCremona I would just merge it and test it afterward. It is not like we have someone registering every second :)

PR #2304, opened 2017-10-07
remove beta status from Bianchi forms

JohnCremona:
Just two things here: remove beta status in sidebar so you see Bianchi under Modular Forms even when running without beta (and the colour is standard under beta); and the page ModularForm/ includes Bianchi forms in the list always, instead of only in beta.

There was a discussions a few weeks ago which agreed that this was ready.

### DIFF
--- a/lmfdb/bianchi_modular_forms/bianchi_modular_form.py
+++ b/lmfdb/bianchi_modular_forms/bianchi_modular_form.py
@@ -162,6 +162,8 @@ def bianchi_modular_form_search(**args):
         v_clean['field_label'] = v['field_label']
         v_clean['short_label'] = v['short_label']
         v_clean['level_label'] = v['level_label']
+        v_clean['level_norm']  = v['level_norm']
+        v_clean['level_number'] = v['level_label'].split(".")[1]
         v_clean['label_suffix'] = v['label_suffix']
         v_clean['label'] = v['label']
         v_clean['level_ideal'] = teXify_pol(v['level_ideal'])
@@ -172,6 +174,7 @@ def bianchi_modular_form_search(**args):
         v_clean['cm'] = cm_info(v['CM'])
         res_clean.append(v_clean)
 
+    res_clean.sort(key=lambda x: [int(x['level_norm']), int(x['level_number']), x['label_suffix']])
     info['forms'] = res_clean
     info['count'] = count
     info['start'] = start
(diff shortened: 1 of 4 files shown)


AndrewVSutherland:
I think you should be able to type the label of a level into the label search box and have it take you to the page for the level.  You can search by level norm, or go to a form by label, but there is no way to go directly to the page http://beta.lmfdb.org/ModularForm/GL2/ImaginaryQuadratic/2.0.4.1/50562.2, for example, other than typing in the URL, even if you know the level label.

AndrewVSutherland:
It would be nice if the SL2 browse pages (like http://beta.lmfdb.org/ModularForm/GL2/ImaginaryQuadratic/sl2dims/2.0.4.) linked the level labels to their pages.  Indeed, I don't see anyway to get to a page like http://beta.lmfdb.org/ModularForm/GL2/ImaginaryQuadratic/2.0.4.1/5.2 without manually typing in the URL.

Also, what is the sort order on these pages?  I was surprised to see 5.2 listed before 5.1.

JohnCremona:
I have updated the PR to handle points 1 and 3 above (jump to a level space, and sort levels properly).

I have not yet worked out how to display more information for the SL2 levels since our URL schema was not designed for having both GL2 and SL2 data.  A Url such as  http://www.lmfdb.org/ModularForm/GL2/ImaginaryQuadratic/2.0.7.1/88.2 refers to the level Gamma_0(N) where N is the ideal with label 88.2 in the field 2.0.4.1 and the definition of Gamma_0(N) allows arbitrary unit determinant.  The corresponding SL2 level is this group intersected with SL(2,O_K), whose space is larger.  I can certainly display the information about the dimensions of such a space, but not of the newforms in it (which we do not have except for the subspace corresponding to the larger group), and I do not know what URL to use.

It would be nice if this change (which still includes the moving from beta status) could be merged now and we can continue to think about how to deal with the other levels later.

AndrewVSutherland:
I'm happy to merge this, does anyone else want to chime in before I do?

JohnCremona:
@AndrewVSutherland I know you did not mean me in your invitation for more comments, but I have thought of a way to deal with the issue labelled #2 above (but in a different PR as it will take a little work):
Simply, to use the same URL such as http://www.lmfdb.org/ModularForm/GL2/ImaginaryQuadratic/2.0.7.1/88.2 to show both the GL2 and SL2 dimension data by expanding the first table there in some way.  The table of individual newforms will still be GL2 only but there would be a note about SL2.  And the "level" section would be clearer about the GL2/SL2 levels.  The SL2 dimension tables could then include links to these pages just as the GL2 ones already do.

jwj61:
Looks good to me.

AndrewVSutherland:
@JohnCremona That sounds like I good plan.  I will go ahead and merge this now.

JohnCremona:
Thanks for all the feedback.

PR #2312, opened 2017-10-18
Remove Lkey from lfunctions code

### LINKED ISSUE #2298: In lfunctions, what is Lkey?
I see that many (not all) of the L-functions have implemented a method called Lkey(), but that none of these are ever used. I wonder --- is this a previous attempt at making something like the Lhash we now have? Or is there a deeper, longer plan in the works that hasn't yet been realized?

davidlowryduda:
In #2298, we noticed that many (not all) of the Lfunction objects in the
code implement Lkey. But this is never used and no one seems to remember
what it was for. To prevent others in the future from losing time while
trying to figure out if these have a purpose, this commit removes all
instances of Lkey from the code.

Resolves #2298.


edgarcosta:
Looks good to me!

If later on some realized why they wanted Lkey around it will also be trivial to go back.

PR #2313, opened 2017-10-19
reading robots.txt from a file

edgarcosta:
Two things:
1- Trying to make it easier to edit the robots.txt for the LMFDB servers not running on www.lmfdb.org.

2- Making sure that SemrushBot stops crawling into the beta server, as their bots don't seem to understand the wildcard concept.

### DIFF
--- a/lmfdb/website.py
+++ b/lmfdb/website.py
@@ -145,6 +145,11 @@ def robots_txt():
         fn = os.path.join(os.path.dirname(os.path.abspath(__file__)), "static", "robots.txt")
         if os.path.exists(fn):
             return open(fn).read()
+    # not running on www.lmfdb.org
+    else:
+        fn = os.path.join(os.path.dirname(os.path.abspath(__file__)), "static", "default_robots.txt")
+        if os.path.exists(fn):
+            return open(fn).read()
     return "User-agent: *\nDisallow: / \n"
 
 @app.route("/style.css")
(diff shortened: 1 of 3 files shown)


lmfdb/website.py
```diff
@@ -145,6 +145,11 @@ def robots_txt():
         fn = os.path.join(os.path.dirname(os.path.abspath(__file__)), "static", "robots.txt")
         if os.path.exists(fn):
             return open(fn).read()
+    # not running on www.lmfdb.org
+    else:
+        fn = os.path.join(os.path.dirname(os.path.abspath(__file__)), "static", "default_robots.txt")
+        if os.path.exists(fn):
+            return open(fn).read()
     return "User-agent: *\nDisallow: / \n"
```
haraldschilly: well, the final return statement is no longer hit. 

the idea here was that any other lmfdb setup is signalling to not be indexed. this should lead to establishing the main lmfdb.org website as the sole one to be indexed and there is no duplicated page at a random place elsewhere on the web. search engines detect this, and then they think this is just a non-unique content being ripped off and somehow decides who is the real one and who is not. that might lead to being removed from the index for the main site. 

tl;dr: in case anyone has concerns about SEO, the idea is to disallow indexing any copy of the main lmfdb page for all search engines

haraldschilly:
Maybe that bot should also be added to https://github.com/edgarcosta/lmfdb/blob/36d0988afe9a006840cdd7fe85e2f5a725f9d6b7/lmfdb/static/robots.txt ?

lmfdb/website.py
edgarcosta: > well, the final return statement is no longer hit.

I agree with you, I left it there, just in case either file doesn't exist.

edgarcosta:
@haraldschilly Yup. Just took care of that.

PR #2319, opened 2017-11-07
removing ecqd1 references

edgarcosta:
I had forgotten for a while to remove the references to the ecqd1 collections.
I have also forgotten to request an update of the Lfunctions database in the cloud.
Drew, could you take care of the update?

Cheers,
Edgar


AndrewVSutherland:
The Lfunctions database has been updated, can you double check that everything looks OK on www.lmfdb.org and let me know (then I will delete the old versions).

edgarcosta:
Looks good to me.

edgarcosta:
Thank you!

PR #2322, opened 2017-11-10
Hgm update

jwj61:
Hypergeometric motive pages are not public on the main site, and had not been touched in years.  This updates them, fixing a bug, adding new features, and adjusting the data format.  It still has a ways to go before it is ready for the main site, primarily in the areas of adding knowls, writing knowls, and adding data.

Some visible changes:

 - the page http://beta.lmfdb.org/Motive/Hypergeometric/ (essentially empty) now redirects to hypergeometric motives over Q
 - on the page for a family, there is a plot of the defining parameters on the unit circle.  It was broken, now fixed
 - on the same page, there is now more information about local monodromy groups (although it needs explanation)
 - there are more search boxes, although they need to be explained
 - added family hodge vector, which differs from the hodge vector of an individual HGM when t=1
 - on the page for an individual motive, we now have factored local factors

The changes outside of the hgm directory are small.  We added an option to one of the search parsing routines (when updating the search parsing), updated the name of the database inside L-functions, and adjusted tests to go with the new data.  Also added the new import scripts.


AndrewVSutherland:
I'm sorry I've been completely swamped this week, but if no one else gets to it I can review this over the weekend.

jwj61:
Thanks.  To a certain extent, there is not much to review since it is not ready for production.  On the other hand, it would be good to get it to beta since that is the best way to make it so that people can fill in knowls.

JohnCremona:
Thanks for the work on modernising this.  I'm looking at it now, mostly random clicking around.  I see no harm in getting this into master and beta soon.  The following minor points came up but can be dealt with in a later round.

 - In the top page Motive/Hypergeometric/Q/ there are some HTML strings in {{}} that need | safe, I think, since they are rendering verbatim.
 - In the Labels explanation page Motive/Hypergeometric/Q/Labels the final L in the breadcrumbs is invalid and crashes. Similarly for Source (S) and Completeness (C).
 - Quite a few of the green numbers in the browse table on the top page give searches with no hits. Jut flagging this up in case it's not intended.

AndrewVSutherland:
In addition to @JohnCremona's comment that some of the browse table links give searches with no hits, the numbers in the browse table cells don't match the number of results in most cases (e.g. the entries for weight 0 and degrees 1,3,5 are displayed in the table as 1,2,4, but the numbers of results are 2,6,14).

jwj61:
We are aware of both issues with numbers in the browse table.  Yes, it is worth remembering that it needs fixing.  

Computations of families is in progress on the Warwick machines, currently doing degree 8.  The other problem will require a coding change, but I am still thinking of different ways to resolve the problem.

JohnCremona:
I'll do another push to beta, as soon as I have added an acknowledgement I missed from last week's work, since the intention of this was (partly) to allow easier knowl editing.

BTW, it would be very convenient if we could edit knowls from local builds.  There should be no reason why not in principle -- the new inventory management and editing system (soon!) can edit the inventory database locally as long as the suitable yaml file is in the LMFDB root dorectory.

Issue #2191, opened 2017-07-19
Add search box for local fields


davidpeterroberts:
A good search box to add on pages such as

http://www.lmfdb.org/LocalNumberField/?p=3&n=9

would be "top slope".   This is the largest of the numbers appearing within the square brackets in the column slope content. 

A complication is that sometimes there are no slopes.  To get these fields the user could type perhaps "tame".

jwj61:
Done

PR #2325, opened 2017-11-15
Add top slope as a search parameter for local fields.

jwj61:
This resolves issue #2191 .  Since slopes can be rational numbers, these are stored in the database as strings.  To make searches work with ranges, we encode the rational number in a form explained on the inventory page (basically, a fixed width for the decimal expansion followed by the rational number in its natural form).  This allows searches to mix decimals and things like 7/3.

If there is tame ramification and no wild ramification, we call the top slope 1; unramified means it is 0 (this is explained in the top slope knowl).

The bulk of the changes are to add the parsing routines to the search-parsing file.

Some searches to try, put in 0 or 1 for a given prime in for top slope, or 2.333..2.334 (with p=2 since 7/3 is a top slope which occurs), or just 7/3, or mix/match making a list of items and ranges.

### DIFF
--- a/lmfdb/local_fields/main.py
+++ b/lmfdb/local_fields/main.py
@@ -8,7 +8,7 @@ from lmfdb.base import app, getDBConnection
 from flask import render_template, request, url_for, redirect
 from lmfdb.utils import web_latex, to_dict, coeff_to_poly, pol_to_html, random_object_from_collection, display_multiset
 from lmfdb.search_parsing import parse_galgrp, parse_ints, parse_count, parse_start, clean_input, parse_rats
-from sage.all import PolynomialRing, QQ, RR, gp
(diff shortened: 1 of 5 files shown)


JohnCremona:
The additional parsing is certainly rather complicated!  I am not sure why people might want to use decimals here but I'm not an expert in this.  Lots of things I tried worked as expected, but when I put 2..3 in the top slope search box I get an error:
`TypeError: unable to convert '2-3' to a rational`

jwj61:
Thanks for looking at it.  I zeroed in on the problem, and should be able to finish it off soon.

As an aside, allowing decimals seemed natural since someone could reasonably put 1.5 in for a slope and expect it to match, or if they wanted 2<s<3, they could use 2.001..2.999.

jwj61:
Should be ready for testing again

Issue #2324, opened 2017-11-13
Add torsion growth data for elliptic curves over Q


JohnCremona:
Enrique Gonzalez has provided data for torsion growth of elliptic curves over Q in number fields of degree up to 7 (so far).  This week (beginning 13 Nov) we are uploading this data and will display it on each curve's home page, with a link to the number field over which the torsion grows when we have it.  In many cases we do not, even for cubic fields where we appear to only have fields of discriminant up to 1 or 2 million (depending on the signature).

davidlowryduda:
I think this can be closed too.

JohnCremona:
The code has not yet been pushed to prod, that's the only reason it is still open.  But the data is on prod so that could happen soon.

JohnCremona:
Closing after the push to prod.  Enrique is still working on the degree 8 data.

PR #2326, opened 2017-11-16
added scripts and code for torsion growth data for elliptic curves

JohnCremona:
This addresses issue #2324

### DIFF
--- a/lmfdb/elliptic_curves/templates/ec-curve.html
+++ b/lmfdb/elliptic_curves/templates/ec-curve.html
@@ -549,7 +549,58 @@ This curve has no rational isogenies.  Its isogeny class <a href={{ data.class_u
 {% endif %}
 {% endif %}
 
+{% if data.data.torsion_growth_data_exists %}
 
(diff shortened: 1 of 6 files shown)


JohnCremona:
This code adds a new section at the bottom of elliptic curve pages (over Q), listing all fields of degree up to 7 over which the curve has more torsion than over Q.  When the field is in the LMFDB it is identified by label, otherwise by defining polynomial.  When the base-change curve is in the database we give a link to it.

Same URLs:

http://localhost:37777/EllipticCurve/Q/392/c/1
http://localhost:37777/EllipticCurve/Q/90/c/4

An example where the extension field is not in the database:  http://localhost:37777/EllipticCurve/Q/391915/c/1

An example where torsion grows from Z/3 to Z/21:  http://localhost:37777/EllipticCurve/Q/162/c/3 (this is the unique curve over a cubic field with a point of order 21.

davidlowryduda:
This looks good to me.

PR #2331, opened 2017-11-24
Online Inventory Editor

hratcliffe:
Adds a web-based inventory viewer and editor. Every database and collection has a viewable name and description info, all keys have description and type info and there is also info. on the schema and number of records in each collection and their indices. 

Editing uses webpage login and passwords.yaml for db authentication. 

@JohnCremona any more essential description?


JohnCremona:
Those who were at the June Warwick workshop saw an early version of this which is now ready for use.  It makes creating and updating inventory information easy;  the inventory info is all stored within the database itself (as you can see from http://beta.lmfdb.org/api/) where the database 'inventory' has several collections in it which are managed by the code here.

Testing so far has been by the authors and by me, using a locally running branch, where the authentication comes from the user having the usual database passwords in a file called passwords.yaml in their LMFDB root directory.  When merged fully such editing should also be possible on beta.lmfdb.org by anyone logged in, but this will need further testing.  On lmfdb.org the inventory information will be viewable but not editable.  All rather like editing knowls.

A longer-term benefit over and above the ease of use for maintaining inventory information is in the development of a more sophisticated API.

Information in the inventory database right now comes from (1) scripts which look at the actual content of the beta database and collections, and also at the inventory repository, and merge the two; (2) manual editing while testing, by me.  The result of (1) is only ever going to be a starting point for the process.  There may be newly added collections, and data fields within existing collections, which are not yet listed here, and one thing still to be done (? @hratcliffe ) is a way to refresh the inventory by repeating (1) without affecting manual changes done by (2).  I'll be doing that soon since (1) was last carried out on 11 Oct and on the following week I added 3 data fields to elliptic_curves/curves.

hratcliffe:
Yes, there's still a little work to do to finish off the uploader code, and a few convenience features I'd like to add to incorporate changes to LMFDB itself into the inventory. I'll get to those ASAP.

AndrewVSutherland:
@hratcliffe @JohnCremona What is the status of this PR, are we waiting on more commits?  Does it need further review?  If so, do you have any guidance for reviewers?  I wasn't at the June workshop, so I have no idea how this works or what I should be testing, and I expect the same is true of anyone else who wasn't there (and other than J.C. no one who was there has given any feedback in the 2.5 months since this PR was submitted).

Would it be easier for people to review if we just merged this?  Is there any harm in doing that?

JohnCremona:
@AndrewVSutherland I have tested this a lot running on a local copy but the way Heather did the authentication meant that the version here would not work on beta as we intend.  Heather is going to switch from using authentication based on having a local password to the same system as we use for knowls.  At that point I'll merge to beta -- I was waiting until then before posting a mini tutorial on how to use it (though it is fairly self-explanatory).

AndrewVSutherland:
@JohnCremona Thanks, I'll be sure to check it out once it's merged.

hratcliffe:
That tweak should fix the authentication, but the CI build seems to be failing to find the mirror db. I don't know why that might be.

edgarcosta:
@hratcliffe , could you merge your branch with master? 
Then it will be fixed by: https://github.com/LMFDB/lmfdb/commit/84339964af6d9d673be8db25a76db9ef44f89b58

edgarcosta:
Thank you.

I have just noticed that the pull request is for beta branch and not for the master branch, as usual. This might create conflicts in the future. We should be careful when merging this.

AndrewVSutherland:
@edgarcosta Good catch!!  I notice that this PR has a bunch of changes in it that do not have anything to do with the online inventory change (e.g. the changes to genus 2 curves, artin representations, L-functions, etc)  they are differences between master and beta.

This entire PR needs to be recreated against a current version of master (this should not be hard to do, as almost all the changes are the creation of completely new files).  It will be an absolute mess if we try to merge this into beta.

hratcliffe:
@JohnCremona  and I originally agreed to do this against beta so it could be tested before going to master, since there's likely to be interim changes to the system.

AndrewVSutherland:
@hratcliffe Apologies if I closed this prematurely, but let me confer with @JohnCremona before we reopen.  We have had problems in the past when master and beta are out of synch, which led to our current workflow in which beta is always a lagging copy of master.  I just want to make sure we are all on the same page before we make a change like this.  In principle there is no reason beta can't diverge from master, but in practice, with many different people merging PRs, it makes it easy to unwittingly merge changes that are going to create conflicts down the road but are not visible at the time the merge is done.

hratcliffe:
No problem. This branch doesn't affect any code from lmfdb, it's a stand alone flask app apart from using things like the DB connection functions. So hopefully it'd be OK to temporarily put into beta until it's been verified a bit.

JohnCremona:
Everyone: you can blame me for not explaining clearly enough to @hratcliffe our work flow which is that "master" is a development branch and the first place anything gets merged into, this having no immediate effect on either beta or production webservers;  we periodically merge master into beta so the new code gets to run on beta.lmfdb.org and (possibly later) ro prod whic hruns on www.lmfdb.org.

Best for now would be: @hratcliffe takes her local branch and merges the current master (done recently but just in case, do it again), then makes a new PR from that to the master branch here.

Then I'll merge it to master and push to beta so we can test whether the authentication is working as planned; and if it is, people can start to use the (fantastic new) inventory-editing system.

Issue #2332, opened 2017-11-28
bib link on citations page broken


JohnCremona:
As reported on the Bug Report spreadsheet:  On the page 
http://beta.lmfdb.org/citation/citations (also on www.lmfdb.org) clicking on the bib links does not work.

I will fix this.

JohnCremona:
It was a triviality: bibtex processing left a line in an html file like this:
`  title = {{R}iemann's hypothesis},
`
and the double { confused jinja.  Inserting a space fixes it.

davidlowryduda:
I believe that this can be closed now.

But I wonder: how are citations added? I see that there is some automatic generation. The header in `citations_content.html` indicates that one uses bibtex2html on a citations.bib file. But I don't see a citations.bib file anywhere.

The reason why I ask is that I wonder if the automatic processing will always course doubled `{{` in cases like this one.

JohnCremona:
See the lmfdb/citations repository.  I'm open to better ways of doing this.  When we add new things by cutting and pasting from Google Scholar into the .bib file we have to be on the lookout for things like this, and up to now my testing has picked up on them.  This time I did not test properly.

One good thing to do would be to add a couple of unit tests, since that would have caught this problem.

PR #2334, opened 2017-11-29
Citation test

davidlowryduda:
The problem in #2332 went unnoticed because of a lack of testing. This
commit adds tests for the base citations page and the "How to Cite" the
lmfdb page. These tests really just check that these pages exist. It
also tests a specific reference in the citation list and tests for its
bib entry in the associated bibliography list. This would catch the
problem from #2332.

Note that I've also submitted a pull request to the LMFDB/citations
repository to prevent this issue from coming up. This is
LMFDB/citations#9


JohnCremona:
Thanks!

PR #2335, opened 2017-12-05
Plot rational points on genus 2 curves

nmascot:
Now that we display a list of rational points on genus 2 curves, I thought that I'd add code to display these points on the plot (execpt the points at infinity, of course).

Here is a nice example (in my opinion): Genus2Curve/Q/432248/a/864496/1

I had to make two non-trivial implementation choices:

- I chose to display the red points above the blue curve, because the other way round was ugly. However, in some cases the points hide the tiny component of the locus of the real points they lie on, so that the picture seems to show a point which is not at all on the curve (whereas it really is). Examples: Genus2Curve/Q/673305/a/673305/1 give you an idea of what's going on (you have to magnify the plot a bit to see it); and with Genus2Curve/Q/581923/a/581923/1, the component of the curve is completely masked (it was barely visible in the first place).

- I also decided that the rational points that do not fit in the plotting window chosen to represent the curve do not get diplayed. Examples: Genus2Curve/Q/830804/a/830804/1, Genus2Curve/Q/70304/a/140608/1 .
This is because the plotting window is determined by the code so as to show the interesting bits of the curve (such as Weierstrass points) without showing too much of the infinite branches (else the interesting bits get squashed). The code that does this is empirical but seems to work well in practice. I did some tests to see what happens if we enlarge the plotting window so that all rational points fit in, and in many cases it did not look good at all (the nice part of the curve was completely squashed).

But of course, these are just personal preferences; your opininon is welcome !

### DIFF
--- a/lmfdb/genus2_curves/web_g2c.py
+++ b/lmfdb/genus2_curves/web_g2c.py
@@ -9,7 +9,7 @@ from lmfdb.elliptic_curves.web_ec import split_lmfdb_label
 from lmfdb.number_fields.number_field import field_pretty
 from lmfdb.sato_tate_groups.main import st_link_by_name
 from lmfdb.genus2_curves import g2c_logger
-from sage.all import latex, ZZ, QQ, CC, NumberField, PolynomialRing, factor, implicit_plot, real, sqrt, var, expand, nth_prime
+from sage.all import latex, ZZ, QQ, CC, NumberField, PolynomialRing, factor, implicit_plot, point, real, sqrt, var, expand, nth_prime
 from sage.plot.text import text
 from flask import url_for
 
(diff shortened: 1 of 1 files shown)


AndrewVSutherland:
I like this a lot!  A particular nice example is:

http://127.0.0.1:37777/Genus2Curve/Q/563011/a/563011/1

And I agree with the decision not to rescale the plot in order to fit all the rational points.

I'm happy to merge this but I'll wait a bit to give others a chance to comment before I do.

edgarcosta:
Also looks good to me.

A small comment, (perhaps deserves a new issue) I would like to be able to click on the plot to enlarge it (or to simply open the full resolution image in a new window).

nmascot:
@edgarcosta Good idea ! I'd normally do this by right clicking -> view image, but maybe some browsers don't offer that possibility.
PR #2337 is my simple-minded attempt at improving this. Personally I'd like something similar to
https://www.w3schools.com/howto/howto_css_modal_images.asp
but that is beyond my html skills.

PR #2337, opened 2017-12-09
Clickable plots

nmascot:
Make plots of elliptic and hyperelliptic curves enlargeable by clicking on them

### DIFF
--- a/lmfdb/ecnf/WebEllipticCurve.py
+++ b/lmfdb/ecnf/WebEllipticCurve.py
@@ -540,7 +540,7 @@ class ECNF(object):
         # Plot
         if K.signature()[0]:
             self.plot = encode_plot(EC_nf_plot(K,self.ainvs, self.field.generator_name()))
-            self.plot_link = '<img src="%s" width="200" height="150"/>' % self.plot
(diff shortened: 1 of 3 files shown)


edgarcosta:
Thanks!

It works on Firefox, however, not in Chrome.
Chrome doesn't allow the user to (automatically) navigate to data URL.
A workaround is to use fancybox or something like that to have a popup effect, see: http://fancyapps.com/fancybox/3/ .
Anyway, this is not worth the effort at the moment.

I'm happy to merge this.

edgarcosta:
Another alternative is to use what you suggested in #2335:
https://www.w3schools.com/howto/howto_css_modal_images.asp
but fancybox already has all that builtin, and it is easy to use.

AndrewVSutherland:
@nmascot @edgarcosta What is the status of this PR?  It looks like there is a conflict that needs to be resolved before it can be merged.

edgarcosta:
@AndrewVSutherland Tonight I will double check that I have indeed fixed the conflict

edgarcosta:
I have resolved the conflict.

AndrewVSutherland:
@edgarcosta Thanks.  BTW, it does work in chrome if you use open link in a new tab (or open image in a new tab).

Issue #2223, opened 2017-08-13
Error in genus 3 curve family data


koffie:
The page http://www.lmfdb.org/HigherGenus/C/Aut/3.4-2.0.2-2-2-2-2-2.1 says that the curves in that family are Hyperelliptic. But this makes no sense given that the equation there is a smooth plane quartic, and smooth plane quartics are never Hyperelliptic. 
I also think that the Jacobian decomposition data is incorrect. This curve has three different involutions. Each involution yielding a map to a different genus 1 curve whose pullback to the jacbonian is distinct from the elliptic curves obtained from the other involutions. Hence the jacobian should decompose as three (possibly non isogenous) elliptic curves.

JohnCremona:
@jenpaulhus ?

AndrewVSutherland:
Jen should chime in here, but I believe that a family is tagged as "hyperelliptic" if it contains a hyperelliptic curve -- this does not imply that all curves in the family are hyperelliptic.  For example, in genus 3 there is only one family listed for the cyclic group of order 2 (see http://www.lmfdb.org/HigherGenus/C/Aut/?genus=3&group=%5B2%2C1%5D).  But this family necessarily contains both hyperelliptic and non-hyperelliptic curves.

jenpaulhus:
Drew is basically correct.  But there is still a small error.  The issue is in the equation.   If you go back to the family: http://www.lmfdb.org/HigherGenus/C/Aut/3.4-2.0.2-2-2-2-2-2  you can see there are 7 inequivalent actions.  Six of them are hyperelliptic (1,2,3,5,6, and 7 in my numbering scheme) and each of those individual pages is displaying the wrong equation.  (And the Jacobian decomp is correct.)

Only one of then is non-hyperelliptic:
http://www.lmfdb.org/HigherGenus/C/Aut/3.4-2.0.2-2-2-2-2-2.4
For this one the equation is correct (and notice the Jac Decomp  is ExExE here).

I'll have to go back into the code and see why that one equation is recording for all curves in the family.  Thanks for pointing this out.

koffie:
[...]

PR #2341, opened 2017-12-15
Add a search by group order for Higher Genus Families + small phrasing/knowl fixes

jenpaulhus:
1. Change top matter at /HigherGenus/C/Aut/ to be separate from the Completeness knowl, and  to add more info (stats page is coming in the next pull request!).  I will add “Completeness” information to that knowl once this pull request is accepted. This should deal with the remaining suggestions in issue #2223 and so that issue can be closed once this code is accepted.

2. The primary thing this code does is implement a  search by group order, instead of just Magma/Gap group number.  The new code allows for searching by genus as well.  So if you want all Hurwitz curves in the database, you can type 84(g-1) in the Group Order search field.  Also, rearranged the order of searching to make more sense.   /HigherGenus/C/Aut/

3.  Implement a Download Magma or GAP code button for displayed search result. 

4. On a page like /HigherGenus/C/Aut/3.14-2.0.2-7-14.1, this gets rid of extra s in “Equations(s) of curve(s)...” line and changed wording to “Jacobian variety (partial) decomposition” to be more clear that the decompositions may not be the finest possible. I will clarify the knowls to explain this once this pull is merged.


jenpaulhus:
I can't figure out what this Travis CI error is.  My test.sh worked fine locally before I made the pull request, and the error message from the Travis CI build is a bit of a mystery to me.

jwj61:
I am no expert, but it doesn't look like the error is your fault.  Maybe try pushing an innocuous change to trigger it to run again?

jenpaulhus:
Added the names of my students to the comments in the main.py file, but still looks like the Travis CI failed again.

edgarcosta:
@jenpaulhus could you do another dummy commit?

We have fixed the issue with the test script: https://github.com/LMFDB/lmfdb/commit/84339964af6d9d673be8db25a76db9ef44f89b58

jenpaulhus:
Ok, tried another dummy commit.  Looks like a new issue, though.  If this is something on my end, please let me know.  Thanks!

edgarcosta:
This is the issue:
```
test_stuff_not_computed (lmfdb.number_fields.test_numberfield.NumberFieldTest) ... FAIL
test_url_bad (lmfdb.number_fields.test_numberfield.NumberFieldTest) ... ok
test_url_label (lmfdb.number_fields.test_numberfield.NumberFieldTest) ... FAIL
test_url_naturallabel (lmfdb.number_fields.test_numberfield.NumberFieldTest) ... FAIL

FAIL: test_stuff_not_computed (lmfdb.number_fields.test_numberfield.NumberFieldTest)
  File "/home/travis/build/LMFDB/lmfdb/lmfdb/number_fields/test_numberfield.py", line 34, in test_stuff_not_computed
    assert 'Not computed' in L.data
AssertionError

FAIL: test_url_label (lmfdb.number_fields.test_numberfield.NumberFieldTest)
  File "/home/travis/build/LMFDB/lmfdb/lmfdb/number_fields/test_numberfield.py", line 54, in test_url_label
    assert '0.481211825059603' in L.data # regulator
AssertionError

FAIL: test_url_naturallabel (lmfdb.number_fields.test_numberfield.NumberFieldTest)
  File "/home/travis/build/LMFDB/lmfdb/lmfdb/number_fields/test_numberfield.py", line 58, in test_url_naturallabel
    assert '0.481211825059603' in L.data # regulator
AssertionError

Ran 22 tests in 60.884s
```
I'm triggered a new test of the branch master to be sure that we don't have some data missing.

edgarcosta:
The automatic testing failed also for the master branch, thus I assume there is some data missing.

@jwj61 Could you confirm this?

jwj61:
It looks like it is the other way around.  A test of a degree 23 field had previously said that something (like the class group) was not computed, but I have been filling in data.  Currently

 http://www.lmfdb.org/NumberField/23.1.459177995857290463522143518056811108.1

shows not computed, but the same thing on beta does have it computed.

The fix for that is to change the test. 

The other two failures are related.  Previously, the regulators for those fields were being computed on the fly, which gave a certain level of precision (used in the test).  Now they are precomputed and stored in the database (on beta, not prod), and that has a little less precision.  The test expects the higher precision, which is why it fails.

I can make a quick pull request modifying the tests.

jwj61:
Pulling from master, then pushing to github should now fix things.

jenpaulhus:
Done!  And looks like it worked.

AndrewVSutherland:
I really like the new search (being able to include g as a parameter is wonderful!).

There are a few consistency issues with your implementation of the code download options:

* the standard way to organize a list of results is to create a list called "data" (see any of elliptic curves, genus 2 curves, or number fields, for an example), rather than using the name result_record you currently use.

* We generally use the flask method "send_file" to return the results so that it will automatically download a file rather than returning text context that the user has to explicitly save (or cut and paste).  For small downloads being able to cut and paste directly from the browser is actually kind of nice, but for large lists I think a file is preferable (and faster).  See the "download_search" function at line 462 of https://github.com/LMFDB/lmfdb/blob/master/lmfdb/genus2_curves/main.py for an example of how to use send_file (which you need to import from flask).

* I notice that you put the download buttons at the top next to the "search again" button, whereas everywhere else the download buttons are at the bottom (as in http://www.lmfdb.org/Genus2Curve/Q/?start=0&count=50&cond=1-300, for example).  But I have to say I like the way it looks to have them at the top as you have done, so that you don't have to scroll down; putting them next to the search again button is quite intuitive and does not waste any vertical space.   I do think you should should shorten the button captions (if you had 4 download options as we do in some other places, they would likely not all fit on the screen).  I would be open to changing this convention, but then we would need to make a lot of changes elsewhere; otherwise I think you should change your implementation to be consistent. 

Does anyone else have an opinion about this? (e.g. @JohnCremona or @jwj61 or @davidfarmer)?  What do people think about the idea of having download buttons at the top next to "search again" on the search results page instead of (or in addition to) the bottom?

jenpaulhus:
I, too, am happy to have a search with genus as a parameter!  All credit for the actual implementation of that search goes to my student @lexmart.  

1. I've changed the download files to create a record called "data".  This required me to change a test also, hence why test_hgcwa.py has also been changed slightly.

2. The download buttons should all now download files, instead of opening text files.   For example you can click the buttons related to a search:
/HigherGenus/C/Aut/?start=0&paging=0&genus=3&signature=&groupsize=2..5&group=&inc_hyper=include&inc_cyc_trig=include&inc_full=include&dim=&count=20

or the download links on the right for one refined passport:
http://localhost:37777/HigherGenus/C/Aut/3.3-1.0.3-3-3-3-3.2


3. I've changed the buttons to download a search  to match with the standard of other lmfdb pages. So now the "Download displayed results for" is before the buttons and the buttons now say just "Magma" or "GAP". For example:
/HigherGenus/C/Aut/?start=0&paging=0&genus=12&signature=&groupsize=42&group=&inc_hyper=include&inc_cyc_trig=include&inc_full=include&dim=&count=20
(Sidebar: the next pull after this will deal with the extra long urls searches are producing for this page.)

I did, however, keep the Download button above the results for now.  If the consensus is that it should  be at the bottom, that is an easy one line change for me.

My two cents are that the top of the page is a better place for the button.  I think new users may not even notice the buttons at the bottom. We don't want people to miss this feature, as it is a key feature of the database in terms of utility of the database for other researchers. 

A brief, non-scientific, survey on my part suggests that not many pages have buttons to download code from search results, so I don't think changing this would be much work.  Also, there is some inconsistency in what the buttons look like anyway, so it appears some code change will need to happen on those pages regardless.

See here:
http://www.lmfdb.org/EllipticCurve/Q/?conductor=389&include_cm=include&surj_quantifier=include&count=100
or here:
http://www.lmfdb.org/EllipticCurve/?include_base_change=on&include_Q_curves=include&conductor_norm=31&include_isogenous=on&include_cm=include
(both have grey buttons)

versus here:
http://www.lmfdb.org/Genus2Curve/Q/?start=0&count=50&cond=1-300
(green buttons)

If others think it's better to have the download buttons before the search results, and the only issue is time to do it, I'll volunteer to make those changes on various pages, as long as it can wait until late February!

edgarcosta:
> What do people think about the idea of having download buttons at the top next to "search again" on the search results page instead of (or in addition to) the bottom?

I think it certainly makes sense to have it at the top, next to search again, and there is no harm to have it also at the bottom.

JohnCremona:
I think we should try hard to maintain consistency, so what is being discussed here has implications all over:  
   1.  I am happy with adding the download buttons at the top and bottom, just as we have Next/Previous buttons in both places.  So we need to do this for all search results pages, and some of those may not yet have the Next/Previous options.
   2. I agree with @AndrewVSutherland that using send_file consistently is better than the current behaviour where sometimes people implement that and sometimes the download text appears directly in the browser.
   3. There are further issues and discussions about the actual content of the downloaded data, but that can be discussed elsewhere.

jenpaulhus:
Ok, the latest commit adds the "Download" buttons to the bottom as well.

PR #2342, opened 2017-12-16
Hgm

jwj61:
Various updating for hypergeometric motives.  The only changes outside the hgm pages are to update the corresponding bits in L-functions.  Also, noticed a css class was being used, but not defined.  Changes should only be visible when exploring hgm's (whose status is beta).

Counts in the families table should now match the corresponding search results.  There were two issues here.  In the data, we had repetition from isomorphic families (now removed), and the previous counts were over \C, but we are using \Q as our base field.  Now twists are being counted separately.

There is a change in the data/searching.  The problem is that the HGM for [A,B,t] is isomorphic to the one for [B,A,1/t].  We now impose a normalization to know which comes first between A and B.  But, one can take p-parts (and prime to p-parts) of the A and B.  Users can search on that, but the p-part operation does not preserve the normalization.  So, for a single motive, we store the A,B related data in both orders, and search both.

Related to the isomorphism above, we take a label which does not match the normalization and renormalize it for the user.

Now better handling of errors.  In particular, there was a problem with not found errors appearing in the flask log.

Note, on the last hgm pull request, there was a comment about raw html appearing.  This was due to a knowl not existing, and the user being logged in.  It was fixed independently by creating the corresponding knowls.


AndrewVSutherland:
Well this is definitely an improvement, so I'm happy to merge, but I did notice while playing around with it that searching by conductor does not work:

* Both of the suggested inputs (2^13 3^7 and [1..1000000] are considered invalid.
* Even if you type in the conductor of a motive in the database, e.g. http://beta.lmfdb.org/Motive/Hypergeometric/Q/A1.1_B2.2/t-1.1 with conductor 32, it does not find it.

These problems are present in the current version as well, so there is no need to fix them in this PR unless you want to.  Just let me know if you want to go ahead and merge now or wait.

I would also suggest explaining the ordering of A,B and the normalization in the labels knowl (but I like the fact that you search on both orders).

jwj61:
Yes, adding the information on how we order A,B should, and will, go into information on labels.

You can hold off on merging until I fix the conductor part of the search.

AndrewVSutherland:
@jwj61 Any update on the status of this?

jwj61:
Afraid not.  My lmfdb time has been used elsewhere

jwj61:
This fixes the basic conductor search.  I say basic because the page says that it will accept conductors in factored form.  Fixing that will take changes to the search parsing, and may be of interest for other pages, so I left that to a different pull request.

JohnCremona:
Could you (pending the additional code needed to search by factored conductor, which I don't see as at all urgent) remove the sample and hints in the conductor search box at http://localhost:37777/Motive/Hypergeometric/Q/
since they give examples of input which does not yet work?

JohnCremona:
Looks good -- as far as I am concerned this can be merged once the tests have passed (and don't wait for me to do the merging).

PR #2350, opened 2018-01-19
Genus 2 curves download buttons at top and bottom

AndrewVSutherland:
As proposed in the thread on PR #2341, this changes the genus 2 search results page to have download buttons both at the top and bottom of the page.  I've made the change in just one place for the moment so people can give feedback, but if/when this is merged I am happy to go make the same change elsewhere.

### DIFF
--- a/lmfdb/genus2_curves/templates/g2c_search_results.html
+++ b/lmfdb/genus2_curves/templates/g2c_search_results.html
@@ -143,7 +143,17 @@
 
 </tr>
 
-<tr><td class="button"><button type='submit' value='refine' style="width: 110px">Search again</button></td></tr>
+<tr>
+<td class="button"><button type='submit' value='refine' style="width: 110px">Search again</button></td>
+<td align=left colspan="-2">
+Download all search results for&nbsp;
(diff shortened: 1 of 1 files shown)


JohnCremona:
Looks very good.  Something we should do everywhere!

AndrewVSutherland:
@JohnCremona In the next few days I can go and add this everywhere else I can think of, I was waiting for this to get reviewed and merged before proceeding.

JohnCremona:
Thanks -- that would be great, but also rather tedious to do.  Perhaps somewhere we should make a list of other things which should be done to make our pages more consistent across collectons.

AndrewVSutherland:
A possibly useful project would be to put together a "style guide" that identifies common features and behaviors that we expect certain standard pages (browse, search results, home pages, ...) to have, based on what is currently done (given that our style/UI evolves by consensus, I think this is the best approach).  In the process of doing so, I'm sure we would find many inconsistencies.

edgarcosta:
The style guide is certainly worth it. 
However, I wouldn't fix it right away, but instead I would raise an issue pointing at it, and mark the issue as "good first issue", in the same flavour as "Open Beginner Tickets" in sage.

PR #2351, opened 2018-01-22
Online Inventory Editor

hratcliffe:
Recreating Inventory PR against master.

Adds a web-based inventory viewer and editor. Every database and collection has a viewable name and description info, all keys have description and type info and there is also info. on the schema and number of records in each collection and their indices.

Editing uses webpage login and the same authentication setup as knowls


JohnCremona:
I have tried this locally and it works as expected, given the appropriate file in the root directory as the beta webserver will have.  The inventory info is visible (at /inventory) and editable when logged in, not otherwise.
Hence I'll push this to master and beta now.  Further changes can be made later once more people have tried it out.  Note that the databases/collections visible now are those which existed at some specific date in the (quite recent) past.  This can be updated, currently on request; at some point we will make it possible for developers to cause new collections to be processed (which takes a little time).

JohnCremona:
Before I push this to beta I want to record one thing:  on beta I can log in and out perfectly, and thanks to cookies I am automatically logged in when I first go to beta.lmfdb.org.  When running locally (this is before this branch is merged) if I log out then I cannot log in again.  I do not know how long this has been the case since normally I let myself be automatically logged in and do not logout.  The first thing I will do on beta after pushing this is to see if the same happens.

edgarcosta:
@JohnCremona for your issue on not being able to login, see #2351

PR #2353, opened 2018-01-22
warning if bcrypt failed to import or generate salt

edgarcosta:
@JohnCremona This might the reason why you aren't able to login while running locally.

Alternatively, I strongly believe that we should add bcrypt to requirements.txt.
I don't see any drawbacks, and not being there causes confusion.


JohnCremona:
I can confirm that install brcypt solved the problem (and the system Sage now has it on atkin and lehner).
@edgarcosta can you add it to requirements.txt and update this PR?

edgarcosta:
Done

Issue #2356, opened 2018-01-26
The page icms.org.uk/workshops/onlinedatabases no longer exists, causing tests to fail


davidlowryduda:
In our testing of the acknowledgements page, we check that the LMFDB is noted on some external pages. It appears there used to be a page at https://info.maths.ed.ac.uk/workshops/onlinedatabases which referenced the LMFDB, but this page has either moved or been removed.

I looked to see if I could find where it moved within their site, but I think it has instead been taken down.

I suppose we should remove this test?

JohnCremona:
Before we do that, let's try to find out if the ICMS (in Edinburgh) has
done this on purpose or by mistake.  The previous ICMS director has an
office down the hall from me (and you DLD) now, so he might know the policy!

Right now the website http://www.icms.org.uk is down anyway.

John

davidfarmer:
That workshop was joint with AIM, so you could instead (or also)
use this link (which also mentions the LMFDB):

    https://aimath.org/pastworkshops/onlinedata.html

It is not easy to maintain permanent links to things like
past activities, especially when you update your website.

PR #2358, opened 2018-01-26
Lfunfix

### LINKED ISSUE #2346: Missing elliptic curve L-function?
From the feedback page:

In reference to http://www.lmfdb.org/L/EllipticCurve/3.3.148.1/61.1/a/ , paweleckrzysztof1@gmail.com says:

Unable to render L-function page due to the following problem(s):
No L-function instance data for "EllipticCurve/3.3.148.1/61.1/a" was found in the database.

==== End of quote

I can see that that url produces the error as described, but if I remove the "L", I get an elliptic curve, and the link to the L-function takes it to a different, and functioning, page.  Probably the first step is to contact the person and ask how they got to that url.

edgarcosta:
This PR:
1- fixes #2346.
2- Adds urls for the L-functions to the pages of elliptic curves over imaginary quadratic fields, and all the pages of the isogeny classes of elliptic curves defined over a field of degree 2.
4- If the the L-function is not on the database or the conductor of L-function is too large, now it displays "L-function not available".  I used the same cut that is used in Hilbert Modular forms. 

ps: A posteriori, I have realized that no elliptic curve over a field of degree >= 3 makes the cut.
pps: This will fail the tests, due to #2356

### DIFF
--- a/lmfdb/ecnf/WebEllipticCurve.py
+++ b/lmfdb/ecnf/WebEllipticCurve.py
@@ -512,10 +512,11 @@ class ECNF(object):
         if totally_real:
             self.hmf_label = "-".join([self.field.label, self.conductor_label, self.iso_label])
             self.urls['hmf'] = url_for('hmf.render_hmf_webpage', field_label=self.field.label, label=self.hmf_label)
-            if sig[0] > 2:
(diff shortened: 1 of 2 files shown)


edgarcosta:
I'm sorry for the title.

JohnCremona:
Looks good to me.  I think Travis is rerunning since I merged #2358 fixing #2356.

JohnCremona:
The tests all pass when merged with current master.  I did see a passing Travis build but for some reason it does not appear here,

PR #2359, opened 2018-01-29
minor fix to avoid deprecation warning with Sage-8.1

JohnCremona:
Small change to avoid getting a deprecation warning with Sage-8.1.  I tested this works also on 8.0 (as now run on prod and beta servers).  Let's see if Travis is happy too.


davidlowryduda:
Oh wow --- I know the deprecation warning just said to use double underscores instead, but I didn't think that would be the whole story.

Indeed this seems to work and it clears up the warnings.

PR #2361, opened 2018-01-30
fixed deprecation warning for p-adic expansions

JohnCremona:
This fixes the other deprecation warning when run with Sage 8.1


JohnCremona:
Sorry the code fails using 8.0, it's 8.1 which introduces the deprecation warning and where the new code works.  I'll add in a try/except block so it works with both.  Coming up...

PR #2364, opened 2018-02-01
small EC/NF tidy-up

JohnCremona:
Three things here.

   1.  I added fields 'class_size' and 'class_deg' to the records in elliptic_curves.nfcurves where they were missing (about half the records), so I could simplify the code by reducing some trivial on-the-fly work and database lookups.
   2   See http://localhost:37777/EllipticCurve/2.2.12.1/3432.4/o/ or http://localhost:37777/EllipticCurve/2.0.4.1/67600.4/d/ or http://localhost:37777/EllipticCurve/2.0.4.1/233.1/a/ -- before the list of curves in the isogeny class it says how many there are and the degrees -- information which is already lower down the page but looks nice here.  (Note that individual curves have have had similar information about their isogeny class visible for some time).
   3. Fixed a breadcrumb issue: see http://localhost:37777/EllipticCurve/2.2.17.1/512.7/d/ where there's blank where the field label/name should be on beta.

While writing this I saw that the breadcrumb is still not quite right so the PR will be updated soon.

### DIFF
--- a/lmfdb/ecnf/WebEllipticCurve.py
+++ b/lmfdb/ecnf/WebEllipticCurve.py
@@ -491,13 +491,7 @@ class ECNF(object):
 
         # Isogeny information
 
-        if self.number==1:
-            isogmat = self.isogeny_matrix
-        else:
-            isogmat = db_ecnf().find_one({'class_label':self.class_label, 'number':1})['isogeny_matrix']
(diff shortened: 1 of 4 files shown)


JohnCremona:
OK to go now.

AndrewVSutherland:
Looks good, just waiting for Travis before merging.

Issue #2273, opened 2017-09-14
Fix bread crumb for L-functions associated to Elliptic curves over number fields


edgarcosta:
At the moment the `Elliptic Curve` in the bread crumb 
```
 ⌂ → L-functions → Degree 4 → Elliptic curve → 2.2.5.1.31.1.a
```
points to http://beta.lmfdb.org/L/degree2/EllipticCurve/.
Which is clearly wrong :)

There is no http://beta.lmfdb.org/L/degree4/EllipticCurve/, but I am also not sure if we really want to have a page per degree....

JohnCremona:
One suggestion:  instead of having any degree-specific pages /L/degree<d>/EllipticCurve, just have one /L/EllipticCurve.  That could have a general opening paragraph, a more detailed section for degree 2 as now, and possibly more to be added later (if the curves over sextic fields get L-functions one day they will have degree 12).

This would however go against the other scheme we have whereby /L/degree<d> is a page linking to all known sources of degree d L-functions.  So maybe we do need a page per degree (of e.c. L-fns) after all.

edgarcosta:
Alternatively, one could entirely skip that crumb, as it is done in the genus 2 curves:
http://beta.lmfdb.org/L/Genus2Curve/Q/791597/a/

JohnCremona:
This is simply a matter of commenting out line 573 in lfunctions/main.py (which is already tagged FIXME) so I am doing that, hoping that it will not conflict with PR#2369

PR #2369, opened 2018-02-05
refactor elliptic curve (over Q) and lhash L-function code

davidlowryduda:
In this pull request, I reorganize the code in set_bread_and_friends, to make that method less monolithic and to make the overall display of L-functions a bit more understandable.


davidlowryduda:
See #2370 for a discussion and explanation behind this pull request.

JohnCremona:
I suggested one tiny change in #2273 (to fix that issue at the same time) but more eyes would be welcome on this one.

edgarcosta:
I agree with the proposed change

davidlowryduda:
The proposed change makes sense to me. I've updated the pull request to incorporate @JohnCremona's suggestion.

JohnCremona:
Thanks for the quick action.

PR #2372, opened 2018-02-06
master --> prod

JohnCremona:
It's been a long time since we updated prod.  No hurry -- if there are reasons for waiting, that's fine, but otherwise this would be a good idea.


AndrewVSutherland:
I'm happy to merge this, but I notice that beta is missing a few commits that are included in this PR, I'd like to push to beta first (doing so now).

JohnCremona:
Good point. Thanks.

JohnCremona:
Since @jwj61 just added a couple more commits to master I'm about to do another push to beta, then I will merge this (after the tests pass), and keep a close watch on the websites today to make sure that all looks well.

PR #2382, opened 2018-02-11
code changes for hmf collection name changes

JohnCremona:
After this, all HMF code should work whether there exist collections forms or forms.search or both (and similarly with the rand and stats derived collections).  Currently the beta database has both; after this has been tested & merged and pushed to beta I will try temporarily renaming the forms.search* collections and if that works OK as expected they can be dropped.  (forms is now  a direct copy of forms.search).

### DIFF
--- a/lmfdb/hilbert_modular_forms/hilbert_modular_form.py
+++ b/lmfdb/hilbert_modular_forms/hilbert_modular_form.py
@@ -20,16 +20,13 @@ from lmfdb.search_parsing import parse_nf_string, parse_ints, parse_hmf_weight, 
 
 def db_forms():
     hmfs = getDBConnection().hmfs
-    if 'hecke' in hmfs.collection_names():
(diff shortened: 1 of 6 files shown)


JohnCremona:
Strange -- one failing test:  test_download (lmfdb.genus2_curves.test_genus2_curves.Genus2Test) ... works fine for me locally.  I clicked on "restart build" on travis so the tests will re-run, but I'm not sure if this github page will get updated.  We'll see.

JohnCremona:
OK that must have been some one-off glitch.

edgarcosta:
We fixed that bug over the weekend

JohnCremona:
Thanks for the explanation! 
I will push this to beta and then  move (i.e. rename) the hmfs collections forms.search* since after this code change that should work.  If something is wrong I'll move them back so there's a chance that HMF functionality on beta will break.  Not for another 25 minutes though as I just missed the 15-minute slot.

PR #2384, opened 2018-02-12
tests reflecting #2379

edgarcosta:
@AndrewVSutherland, could you please look into this?


AndrewVSutherland:
Looks good, thanks for the fix.

PR #2388, opened 2018-02-14
Updating Travis script

edgarcosta:
Lately we had some automatic tests failing because it takes more than 10 minutes to untar sage. 
I have extended the timeout to be 30 minutes.
Further, I added a line to check the connection with our server, this time using `nc` instead of `printf` and `mongo` (that gave us some trouble mid december).
```
nc -vz  m0.lmfdb.xyz 27017
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -39,7 +39,8 @@ install:
     - export SAGE_INSTALL_GCC=no
     - export MAKE='make -j4'
     - wget  --no-verbose https://storage.googleapis.com/sage-lmfdb/sage-${SAGE_VERSION}-Ubuntu_14.04-x86_64.tar.bz2
-    - time tar xf sage-${SAGE_VERSION}-Ubuntu_14.04-x86_64.tar.bz2 -C ${HOME}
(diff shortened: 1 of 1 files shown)


davidlowryduda:
Do you know how long it usually takes travis to untar sage? I don't remember noticing this before --- is this something to do with sage 8.1?

edgarcosta:
It oscillates quite a lot. It is mostly due to the server a its current load.
On a good day/server it takes 3 min, on a bad day it can go up to 7 min, on a really bad day goes over 10min and then travis by default cuts the cord.

For example. in this build https://travis-ci.org/LMFDB/lmfdb/builds/340322461 it took 7 minutes
```
time tar xf sage-${SAGE_VERSION}-Ubuntu_14.04-x86_64.tar.bz2 -C ${HOME}
real	7m9.807s
user	4m43.815s
sys	0m41.898s
```
and I think we got lucky, as this 32 core server was clearly  overloaded 
```
$ uptime
 15:55:45 up  7:03,  0 users,  load average: 189.16, 181.34, 202.70
```

PR #2393, opened 2018-02-25
Stop lexicographical sorting of record numbers on inventory record pages

davidlowryduda:
Right now, on the inventory record pages, like

    http://www.lmfdb.org/inventory/numberfields/fields/records/

the records are sorted lexicographically by record number, so that 100 and 101 come before 10 or 1. While it may be desirable to sort this data (perhaps it would be useful to be able to sort by count, for instance), it is certainly not desirable to sort by record number in dictionary order.

This small pull request undoes the lexicographical ordering. I note that by removing the sort function, I am implicitly using that the database returns the information by record number.

### DIFF
--- a/lmfdb/inventory_app/inventory_helpers.py
+++ b/lmfdb/inventory_app/inventory_helpers.py
@@ -146,7 +146,7 @@ def empty_null_record_info(records):
     """Empty null name or description fields in list of records"""
     try:
         for item in records:
-            print item
+            #print item
             item = empty_all_if_null(item)
     except Exception as e:
         log_dest.error(e)
--- a/lmfdb/inventory_app/static/inventory_general.js
+++ b/lmfdb/inventory_app/static/inventory_general.js
@@ -122,6 +122,7 @@ function fetchAndPopulateData(blockList, pageCreator, startVisible=true){
     }else{
 	    populateBlocklist(XHR.blockList, data);
     	pageCreator(XHR.blockList, startVisible=startVisible);
+      $( document ).trigger("dataPopulated");
     }
   });
 
--- a/lmfdb/inventory_app/static/inventory_general.js
+++ b/lmfdb/inventory_app/static/inventory_general.js
@@ -85,7 +85,7 @@ function setNotEditableById(id){
 
 function getBoxTitles(blockList){
   //Returns all the unique box names. Assumes name is all the bit before the final
-  var keys = Object.keys(blockList.blockList).sort();
+  var keys = Object.keys(blockList.blockList);
   var uniq_keys = {};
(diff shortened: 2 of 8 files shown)


hratcliffe:
Does the database guarantee that and are things always unpacked from JSON the same way? I know the conversion to and from JSON doesn't guarantee ordering is preserved in general. I have on the to-do list to fix the ordering, but I'm not sure it's reliable to not do anything with it.

JohnCremona:
This does look superficially slightly better (in the example, where there are apparently a very large number of different record types) but surely it would be preferable to decide what order we want to see the records in (e.g. in reverse order of "count" after any dummy base record if present) and number them like that?  [By the way, quite soon I understand that all the oldpolredabs* keys will be redundant and dropped, which could help.]

davidlowryduda:
I've iterated once more. Now I've added functionality that if you click on the "Record Num" or "Count" row headers, then the list is sorted by them. If you click one twice, then it's sorted in reverse order.

I have not made this obvious to the user in any way yet --- I'm first trying to nail down the desired functionality.

@JohnCremona Could you take another look?

hratcliffe:
Hmm if you sort and then refresh the page, will anything persist? It might be better to sort the underlying  list and regenerate the table afterwards. At that point we can just use built-in JS sort with a custom comparator which would be more flexible too.

davidlowryduda:
No, if you sort and refresh the page, nothing persists. But the sorts are essentially instantaneous as the number of rows is not very large. This doesn't sort the editor pages --- just the record viewing pages.

 It's not clear to me why one should worry about about refreshing the record viewer pages. I know @hratcliffe put good effort into caching the editing pages so that they persist after refresh, and that makes sense to me. But perhaps I missing a bigger idea?

hratcliffe:
Only that it gets harder and harder to know when a page might be refreshed. For instance, I use Opera which has the ability to pause tabs in the background and refresh them when you come back, and it gets irritating if changing tabs undoes things, or the page relays completely if you catch the back button.  

Mainly I am thinking it could be worth doing sorting 'properly' across all the view/edit pages. I put in something cursory, but it could be done better without too much work.

davidlowryduda:
Looking at this once more, I fell down a small rabbit hole. The first of these new commits corrects a problem that John pointed out to me, which is that I wasn't keeping the dummy row fixed. But I then noticed that LMFDB already has a tablesorting javascript library in the main javascript resources, so then I shifted to use that instead (and then smoothed out some of the resulting CSS so that it correctly colors rows after sorting them).

Along the way, I noticed #2395.

JohnCremona:
This looks good to me, judging by appearance and function rather than the code since I am no js expert.

PR #2396, opened 2018-02-28
Elliptic curve code tidy up

JohnCremona:
Nothing here has any visible effect on the website, it is only tidying up some older obsolete code and making sure that there are no more references to obsolete keys in a curve record so that the records can subsequently be tidied up.
Note: as well as running the standard tests I also ran ~100,000 random curves with this code.

### DIFF
--- a/lmfdb/elliptic_curves/elliptic_curve.py
+++ b/lmfdb/elliptic_curves/elliptic_curve.py
@@ -4,7 +4,6 @@ import time
 import ast
 from pymongo import ASCENDING
 from operator import mul
-import lmfdb.base
 from lmfdb.base import app
 from flask import render_template, url_for, request, redirect, make_response, send_file
 import tempfile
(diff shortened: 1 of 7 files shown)


JohnCremona:
The Travis failure was a timeout on a workshop link.

jwj61:
That machine is currently down while being upgraded/replaced.

JohnCremona:
@jwj61 Shall we disable that test then?

jwj61:
It is up to you.  I hope it will be up and running today.  Then again, I was hoping they would get it set up properly yesterday.

edgarcosta:
I would just wait for the link to be back online and move along with the
merge if that is desired.

JohnCremona:
OK, so I will merge this having checked manually that the only test failure was that one.

jwj61:
For the record, hobbes is back up

PR #2398, opened 2018-03-01
Added command line options for mongo port and host

csbrady-warwick:
Allows you to easily specify an override for the mongodb port and hostname. Useful for debugging purposes, but shouldn't need to be used normally.

### DIFF
--- a/lmfdb/website.py
+++ b/lmfdb/website.py
@@ -242,6 +242,8 @@ def get_configuration():
                                            "debug",
                                            "help", 
                                            "mongo-client=",
+                                           "mongo-port=",
+                                           "mongo-host=",
                                             # undocumented, see below
                                             "enable-reloader", "disable-reloader",
                                             "enable-debugger", "disable-debugger",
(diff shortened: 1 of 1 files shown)


JohnCremona:
@csbrady-warwick There are already command-line options for these, either -p or --port and -h or --host.  Are your additions necessary?

csbrady-warwick:
Those options are to flask, not to mongo. --port sets the flask server port and --host sets the flask bound host (localhost by default). I've checked fairly carefully and there are no options to control mongo hosts or ports other than through the mongo-client config file. That's rather heavyweight for just setting the port and host info.

edgarcosta:
yup, the other ones are just for flask.

davidlowryduda:
It's true that the current ones are just for flask. This seems like a pretty light and painless addition. But I think it would be a good idea to also include some small documentation for others who might want to do this in the `Running` section of the GettingStarted.md document.

JohnCremona:
I stand corrected!  I'll merge this now but agree that having the two new command-line flags should be added to the GettingStarted doc.

Issue #2399, opened 2018-03-01
Inventory Display bugs


hratcliffe:
New inventory has some bugs in the display logic. Any small ones can go in this thread for fixing.

* edit of inventory has collapsible sections: in some cases the wrong part is collapsed/expanded
* edit of records has similar but separate issue with collapsible sections
* button styling is not all consistent in some browsers

AndrewVSutherland:
@JohnCremona can we close this issue?

JohnCremona:
@AndrewVSutherland Yes.  I think the sections expand and contract OK and I am not bothered about the 3rd point.

PR #2400, opened 2018-03-01
Fix error in expanding sections in inventory

hratcliffe:
Fixes part of Issue #2399 where the '+' buttons on inventory edit pages open up too much data


JohnCremona:
I confirmed that the multiple opening up when one key occurs as a substring of another no longer happens.

PR #2401, opened 2018-03-01
Edited documentation to describe new keys

csbrady-warwick:
Edited gettingstarted.md to refer to new keys in section on running a local copy of the database (previous instructions would not have worked) 

There was a previous 'dbport' parameter to set the database port, but without the ability to specify the hostname as well, this would not work since the change to using the cloud database. Parameter has been removed in favour of new ones.


davidlowryduda:
This looks good to me.

JohnCremona:
Thanks

PR #2403, opened 2018-03-01
Change table width CSS in inventory

davidlowryduda:
When I made the inventory records sortable, I used class for which the LMFDB has some CSS. This is a minor tweak to make the width *not* take the whole screen, resolving something John pointed out to me.

### DIFF
--- a/lmfdb/inventory_app/templates/inv_style.css
+++ b/lmfdb/inventory_app/templates/inv_style.css
@@ -131,17 +131,18 @@ div.hide, span.hide{
     width: 0px; /* unsets global CSS  width */
 }
 .viewerTableHeaders, .viewerTableEls{
-  border: 1px solid black;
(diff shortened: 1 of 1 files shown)


JohnCremona:
Maybe I am doing something wrong but it looks just the same to me now!  There is a space prefix on every line (if multiline) in the top row only.

davidlowryduda:
@JohnCremona Ah, I see. Initially I had thought that you were seeing something width-related since I know the monitor in your office has a vastly wider resolution than my screens. So I switched to a high-resolution viewer and noticed that 100% width was a silly idea.

But now I notice inline spacing. And I think I've corrected it now. Would you mind looking again?

JohnCremona:
Fine now -- thanks.  I was only testing on a laptop.

PR #2407, opened 2018-03-16
Inventory Rereading

hratcliffe:
Adds ability to trigger re-reading of the inventory data and some small bugfixes etc. 

I've tested on my local machine running over the tunnel, and everything appears to work. On the beta server, scraping and uploading of the resulting data should be fine. However, to use the progress feature, we need the role to be able to read operations. If this does not work, I'll work out what the role needs adding. The progress feature shows as a % done on the page you are sent to after clicking any collection on the /rescrape page. If this sits at 0 for a time then jumps to 100, this is the likely culprit.


JohnCremona:
Thanks.  I have tested this locally and all seems well.  I read the updated instructions (in Inventory.md) to see what to do.  One minor cosmetic thing would be to sort the database names on the /inventory/rescrape page.
I will merge this to master and beta; we may then need to add to the permissions for the progress meter.
Soon, maybe now, I suggest (1) having a place to click to get to the top-level /inventory page, at least on beta (for now) but not requiring any login (to reassure people, the ability to edit the inventory would be greyed out to anyone not logged in), and (2) from that top inventory page and visible only to people logged in and some higher level of authority (e.g. people whose logins have the "admin" tag set) a button to go to /inventory/rescrape.
Before I merge to beta it would be nice to have clear reassurance that the rescrape facility will not be possible for normal users, whether or not logged in, but requires further authorization: can you tell us what that is, for the record?

JohnCremona:
While I was running this locally there were various logging messages appearing on my screen.  Some of these look as if they may be an indication that something is not right.  (In my testing I did rescrape some collections, only small ones which I was pretty sure had not had any changes of content since the last time).  Here are some of the lines which appeared (am sample):

`
WARNING:Inv_log@2018-03-17 13:31:08,982: In upload with fa197c25-631e-432e-bf21-8bc2f4852b86 
ERROR:Inv_log@2018-03-17 13:31:09,800: Error removing fields not master 
not master
ERROR:Inv_log@2018-03-17 13:31:11,235: Error updating record {'_id': ObjectId('59bbfcb90e318aaecb10372e')} not master 
ERROR:Inv_log@2018-03-17 13:31:12,731: Error inserting new record not master 
`

JohnCremona:
As well as the decorate login_required there is also admin_required and I suggest that this is what the rescraping functions should have.
There are currently 7 admin users, and this attribute has only been used so far for giving some people the ability to create accounts for others: this needs to be looked at.  They are
['Jonathan Bober', 'Alina Bucur', 'Ralph Furmaniak', 'Harald Schilly', 'John Cremona','David Farmer', 'John Jones']
of whom at least one is no longer active, and I can think of several who would need this power.   Definitely all those who are allowed to edit knowls though.

AndrewVSutherland:
@JohnCremona Did you really mean to say all users who are allowed to edit knowls, or is there perhaps a negation missing in your last sentence?

JohnCremona:
Well spotted @AndrewVSutherland !  We can discuss who off-list perhaps.

JohnCremona:
Both users 'rse' and 'drew' now have admin status.  I hope @hratcliffe and @csbrady-warwick and @AndrewVSutherland  can still login OK.

hratcliffe:
I'll add the admin_required decorator to the relevant sections. 

@JohnCremona , I think the error messages about master is because getDBConnection is returning the cloud connection. That's not writable, so the editing and scrape-uploading wont work. Locally I was using the tunnel and authenticating using password.

JohnCremona:
@hratcliffe You are right -- I was testing from my laptop at home, with the default db connection rather than from atkin or lehner (or using the tunnel).  This is an additional layer of authentication, in effect, and hence a good thing.  On the other hand, the rescraping web pages themselves gave no indication that anything was wrong.   Can we tell at run-time whether or not we have a writable connection to the database?
I will test again this time making sure that I am connecting to the Warwick db.  I am also planning to see what happens when I create a little new collection (in an existing database) to see if the rescrape picks it up, and then if that works, what happens when that collection is dropped.

JohnCremona:
I am now testing this locally on atkin, connecting to the main (Warwick) beta database as we should.  First thing people need to know for such a test is that you need to have the appropriate p/w file in your LMFDB root directory so that you get limited read/write access: this is the same as you need if you want to edit knowls on a local instance of the website, and for the same reason.  Anyone not knowing what "the appropriate p/w file" is can contact me directly; it is the same file which the beta webserver itself has.

Now rescraping small existing collections seems ok ("seems" only because I had not changed any contents since the last scrape).  As predicted the progress monitor spins but jumps from 0 to 100% in one go.

I then created a new collection called test_collection in the elliptic_curves database, with a single small record.   The /inventory/rescrape page showed it right away -- good. *But* clicking on the rescrape button started something which 10 minutes later has apparently not finished.  I killed that webserver instance.
I can provide more information if required to debug this.

hratcliffe:
@JohnCremona not sure why the process kept running, but it does seem to have succeeded. Anything in the log files?

For dropping collections, I have added a page under inventory/controlpanel which lets you do this sort of thing. The 'mark gone' button should mark any that no longer exist in the main DB.  Right now I don't have the 'remove collections' button hooked up, because I wasn't certain if this was wanted. That will remove any inventory for collections detected as 'gone'.

JohnCremona:
I will send you LMFDBinventory.log by email.
I went to /controlpanel (I suggest that for users logged in with admin status there is a button to clock to go there from the main /inventory page).  Clicked on "mark collections gone" but got no indication that anything had happened.  Then clicked on Cleanup removed collections and the confirmation button but then a pop-up box appears saying "localhost:37777 says: An error occurred processing request. Action TBC".

hratcliffe:
There might be a slight delay since it has to communicate with server, but mark_gone worked for me before. 
As I said, the remove_collections button isn't yet hooked up, hence the error message.

jwj61:
I tried this for the first time on the numberfields.fields.  I was running the copy of lmfdb from lehner, using the standard database connection there.  It quickly said that the rescraping was complete, but there was no change (and there should be), and no error message given on the browser page.

I have a feeling I need to do something so that the local copy runs with a higher level of permissions, but I don't know how to do that.  But, if there is a permissions error, it should say so.

JohnCremona:
@jwj61 I certainly agree with the last point.  I will contact you offline (well, online but not in public)  to tell you how to run with the necessary permissions.

csbrady-warwick:
Yes, you are right it should warn you if you're running without appropriate permissions, although you do now get caught in a bit of a web of conditions unless you are careful (if authenticated with database, if logged in, if admin etc.)

Since many of the elements of LMFDB behind the login system require write access to the database to be useful (and there are similar problems with editing knowls etc.) you could make an argument that you shouldn't be able to access the ```users``` table at all without having write permissions to the database. This would simplify matters more than trying to catch all possible states in the specific operations behind the login.

JohnCremona:
Progress monitoring:  even when  I am authenticated on the admin database (with username admin and p/w what it is) I find that the admin.command(u) (as in scrape_frontend.py) gives an error, but admin.current_op(u) works fine (with the same u).  I don't know why that is.
According to https://docs.mongodb.com/manual/reference/method/db.currentOp/ you can use current_op to find your own ops without further authentication, from mongo 3.2.9 on.  We are using 3.2.1.  We could upgrade, but that might cause more problems.

hratcliffe:
The progress monitor isn't vital, it's just nice for the scrapes which take a long time. I'll have a look tomorrow at whether current_op does anything different.

I may have been running the scrapes on the cloud and only writing over the tunnel, and I think the cloud Mongo install is newer, so that may be why it worked in my tests. If that's it, then the webserver role needs to be granted inprog permission, per https://docs.mongodb.com/manual/reference/privilege-actions/#inprog

JohnCremona:
Two question for @edgarcosta then: first, what version of mongo is running on the cloud, and second, can you grant (or explain how to grant) to the webserver role  inprog permission?  This remains a mystery to me however much documentation I read.

JohnCremona:
I have attempted to grant the webserver user 'inprog' permission but what I did has not helped show the progress meter.  Since mongo will not allow addiing privileges to builtin roles such as readWrite, I created a new role called 'rescrape' which has 'inprog' privileges on all collections in the inventory db (and then also on the admin db).  I will record the commands used to do this since the documentation is very obscure, and it seems that the webserver user really does have the new rolw which does have the extra privileges. *But* you still see the same error messages "not authorized on admin to execute command { currentOp: 1, $ownOps: 1 }".  I don't know why (unless one has to restart mongo for the changes to take effect -- @edgarcosta do you know if that is the case?)

JohnCremona:
Some success: I did two things and both made a difference.  First, I read that the clusterMonitor role already has the  inprog privilege so added that role to the webserver user.  That led to the old error message not appearing; a different one appeared from line 21 of scrape_frontend.py, namely "bad query: BadValue unknown top level operator: $ownOps".
I fixed that by changing .command(u) to .current_op(u).
This first test I then did, on a small collection, briefly showed 43% under the progress meter before finishing.  So I rescraped some more collections -- but in every case the rescrape finished instantaneously saying there was nothing to report.  This seems suspicious.

hratcliffe:
@JohnCremona most of the collections are very small: numberfields.fields takes 10-15 mins, but I didn't find anything in between. 

Note that the progress meter page does a time-doubling refresh rate to avoid polling too frequently in the long cases, so the meter wont be terribly steady. (I think it starts at 1 second delay and doubles up to 128 seconds then stays there)

I'm at a workshop today/tomorrow, but will be back Wednesday to finish the changes we already agreed.

JohnCremona:
Actually I think the rescrape *is* taking a measurable length of time, it's just the web page which reports back instantly even though the process is continuing.  I just started elliptic_curves.curves and again it instantly reported that it had finished with nothing to report, but from http://localhost:37777/inventory/elliptic_curves/ right now it is locked (Edit inventory etc greyed out).  The ones I was testing yesterday are correctly showing that they were rescraped yesterday.  And now it has finished and is unlocked -- it took between 1 and 5 minutes.

hratcliffe:
Very odd,  that never happened in my tests. Possibly the changes mean the monitor page is not getting the data or format it expects. What happens if you add /monitor to the end of the progress url? You should get a JSON document containing 3 keys.

JohnCremona:
It looks like 
`
{
  "curr_coll": 0, 
  "n_colls": 1, 
  "progress_in_current": 12
}
`
and the last number (12) goes up on refresh.  Then: removing the /monitor gives me what I was expecting to see, a spinning thing and a %-age increasing as I watch!

hratcliffe:
That's the expected behaviour. Not sure what was happening before. 

It is possible that the upload after the scrape is taking a significant time: the progress monitor really does monitor the scraping process. Probably something should be tweaked to make sure it doesn't report completion until the upload part is also done, which is when everything unlocks.

edgarcosta:
@johncremona,

regarding your two questions:
1-
```
$ mongod --version
db version v3.2.19
git version: a9f574de6a566a58b24d126b44a56718d181e989
OpenSSL version: OpenSSL 1.0.2g  1 Mar 2016
allocator: tcmalloc
modules: none
build environment:
    distmod: ubuntu1604
    distarch: x86_64
    target_arch: x86_64
```

2-
regarding roles, you must first create the desired role, and then add the
role to the user.
I showed a bit of that in #2281
Alternatively, see
https://docs.mongodb.com/manual/tutorial/manage-users-and-roles/

To make it clear, you would like the webserver user to have the access to
privilege action `inprog`?

I think, you don't need a new role, if you just want to monitor its own
actions:
https://docs.mongodb.com/manual/reference/privilege-actions/#inprog

JohnCremona:
It was the pymongo version I was asking about.  But the version we are running at Warwick is only 3.2.1, and anyway the link you give (which I had read) says that you don't need inprog permission only from 3.2.9.
Anyway I gave the webserver user the (predefined) role clusterMonitor and that worked.  I had also created a new role but that turned out not to be necessary.

What's left for this PR:  (1) I had to change one line in scrape_frontend.py (line 21, change "command" to "current_op") before it worked for me; that is something I thought might be the pymongo version since the code in @hratcliffe 's PR clearly works for her.  (2) After that it works for me except that agter clicking OK to run a rescrape, I see the "all finished" page and only after refreshing that web page to I see the progress monitor.

edgarcosta:
Here is the pymongo version:
```
sage: import pymongo
sage: pymongo.__version__
'3.6.0'
```

hratcliffe:
I was testing using 3.5.1 locally. 

I'll look into the all-finished report, because that shouldn't happen

hratcliffe:
That last commit should fix the progress meter. You should no longer get a report of completion until the upload is done. It will sit at 99% while it uploads.

JohnCremona:
Thanks -- will retest as soon as I get a chance.

edgarcosta:
on a side note, I'm a bit confused why the pymongo version in the cloud is relevant, at the end of the day we will never use this feature on those servers.

JohnCremona:
@edgarcosta I only meant to ask  Heather which pymongo version she had tested on in case it was different and that could be the cause of the command_op issue.

JohnCremona:
Still only works for me when I change that one line in scrape_frontend.

hratcliffe:
Yeah that change will still be needed. Those changes just (hopefully) fix the premature completion while still uploading.

JohnCremona:
I made a tiny PR to the inventory branch on csbrady-warwick/lmfdb.  If you merge that this will update, then I will merge this.

csbrady-warwick:
Hi John,

I think that if you're using current_op like that then you don't have to pass it the BSON object at all. I was only using .command because PyMongo hasn't been updated to allow you to use use ownOps. Would you mind checking if it works for you with just

progress = connection['admin'].current_op()

Chris

PR #2408, opened 2018-03-21
master --> beta

JohnCremona:


JohnCremona:
The only test failure I can see is 
`
FAIL: test_big_degree (lmfdb.artin_representations.test_artin_representation.ArtinRepTest)
`
(It should be easier to find a single failure i nhundreds of lines of output which the Travis test produces!)
which clearly has nothing to do with this PR.  I am tempted to ignore it.

JohnCremona:
The problem is that one Artin rep tests tries for http://beta.lmfdb.org/ArtinRepresentation/2.1951e2.120.1c1
and this no longer exists.  I'm going to merge a small patch ti skip that test -- @jwj61 do you know what happened with this and can you suggest a better fix?

jwj61:
No, I don't know where it went.

I am testing some recomputed data in Artin representations on beta (which seems like a likely answer), but this representation does not seem to be on the main server either.  So, I am still looking into it.

jwj61:
OK, I think I have it sorted out.

We have one set of artin representations which have been on the site for a while, which I will call our core set.  Then, I added some to beta for a paper by Andy Booker et. al., but this involved a high degree field, and that required a change to the lmfdb code.  So, I put in new code along with a test for it.  The new data did not go to the cloud.

Then I discovered a bug which ran thoughout the Artin representation data, and recomputed the core set.  That is what is currently on beta, and I am looking at to see if the bug is fixed.

I am also recomputing the "new data", and will add that when it is ready.  Then the test can be reactivated.

edgarcosta:
@johncremona,
> It should be easier to find a single failure i nhundreds of lines of
output which the Travis test produces!)

The tests are run in parallel.
At the end, if something fails we print all the outputs.
To find the error one needs to search in the log. If desired,  one could
easily have a grep at the end displaying all the appropriate lines.

JohnCremona:
@jwj61 thanks for the explanation.

@edgarcosta A final grep might be helpful since scrolling back it is easy to miss the failures.  There is some colour coding but not one which helps (me) much.  But it's not a big issue.

PR #2409, opened 2018-03-21
added code download for ecnf

JohnCremona:
Home pages of elliptic curves over Q have download buttons (on the right-hand side) for downloading all data and also for downloading all the code snippets.  This PR adds the same for elliptic curves over number fields.
To test, go to a random elliptic curve page and then click on each of the 4 Download links.  Even better, run Magma/Sage/GP on the files thus obtained.  I did this for a few random curves and perhaps I was lucky since there must be many curves here for which Sage or Magma will have difficulty computing the rank (for example); but that is no reason not to have this.  There are far fewer code snippets for GP (and perhaps new versions can do more).
I added a new test function too.

### DIFF
--- a/lmfdb/ecnf/WebEllipticCurve.py
+++ b/lmfdb/ecnf/WebEllipticCurve.py
@@ -574,6 +574,14 @@ class ECNF(object):
 
         self._code = None # will be set if needed by get_code()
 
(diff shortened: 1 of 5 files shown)


AndrewVSutherland:
Looks good to me.  If you want an example where computing the rank is hard to compute, try http://127.0.0.1:37777/EllipticCurve/6.6.300125.1/1.1/a/1

JohnCremona:
Thanks.  (If computing the ranks was never difficult we would already have computed them all in the database!  That's why I put the warning line comment in the download file.)

PR #2411, opened 2018-03-22
Prevent user login if database cannot be written to

csbrady-warwick:
Fix as suggested in #2407. Prevent user from logging in at all if database cannot be written to. Since almost all features behind the login system are useless without write access to the database and there are oddities with the user information change, the knowl editing and a few bits of the inventory system this fixes several problems.

### DIFF
--- a/lmfdb/users/main.py
+++ b/lmfdb/users/main.py
@@ -7,7 +7,8 @@ import pymongo
 ASC = pymongo.ASCENDING
 import flask
 from functools import wraps
-from lmfdb.base import app, getDBConnection, _mongo_user
+from lmfdb.base import app, getDBConnection
+from bson.son import SON
 from flask import render_template, request, Blueprint, url_for, make_response
 from flask.ext.login import login_required, login_user, current_user, logout_user
 
@@ -50,7 +51,11 @@ def ctx_proc_userdata():
     userdata['userid'] = 'anon' if current_user.is_anonymous() else current_user._uid
     userdata['username'] = 'Anonymous' if current_user.is_anonymous() else current_user.name
     userdata['user_is_authenticated'] = current_user.is_authenticated()
-    userdata['user_can_write'] = lmfdb.base._mongo_user != 'lmfdb'
+    try:
+        roles = getDBConnection()['admin'].command(SON({"connectionStatus":int(1)}))
+        userdata['user_can_write'] = 'readWrite' in [el['role'] for el in roles['authInfo']['authenticatedUserRoles'] if el['db']=='inventory']
+    except:
+        userdata['user_can_write'] = False
     userdata['user_is_admin'] = current_user.is_admin()
     userdata['get_username'] = get_username # this is a function
     return userdata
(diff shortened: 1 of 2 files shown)


JohnCremona:
Thanks -- I will try this out but it would be a good idea if others would too.  I emailed @AndrewVSutherland  @jwj61 and @edgarcosta  about what is needed for enabling this locally.
One thing I want to do is remind myself what other changes to how the website looks are controlled by being logged in, in case this has more unintended consequences.

JohnCremona:
I checked that this works as expected -- on a local instance without the required password file (as exists on the beta server) it will not allow login, but with that file in place it will.  However whatever cookie system we have which normally (for me at least) makes me logged in whenever I start up still makes me logged in without the password file in place, so that's loophole.   I don't know how important people think that is.  I can see the place where the cookie is set (in lmfdb/users/main.py in login()) but not where it is found and used -- presumably something flask_login does.  I guess that does not matter since if the cookie is set then you must have previously logged in ok.
One other issue though: we do need to check that the host to which pymongo has connected is *not* m0.lmfdb.xyz or the cloud server.  Right now I can log in locally (with the password file in place) but can still not edit since I am using the default mongo config (on my laptop).  Fixing this too would be good.  I think the get_configuration() function in website.py is where this hostname is set and stored with other stuff in the variable configuration.  Possible?

csbrady-warwick:
Fixing it so that the cookie login is prevented would be quite tricky, but the second part can be done easily enough. I think that you're right about checking with get_configuration, although I'll also looking into using the mongodb connectionStatus ( https://docs.mongodb.com/manual/reference/command/connectionStatus/ ) command to actually check if you can write to the database. I'll do some more work on it.

JohnCremona:
It looks as if connectionStatus is the thing to use.  Ignore the cookie thing.
BTW there's a redundant import in users/main.py causing pyflakes to fail on Travis.

csbrady-warwick:
That should have set it to use connectionStatus to check the write permissions on the inventory database as a check. Also fixed the pyflakes warning

JohnCremona:
I am still allowed to login when connected to m0.lmfdb.xyz: I have the authentication necessary to allow writing to the db, but something else is (intentionally) preventing any writes to the secondary replicaset.  An additional test of connection.is_primary (with no brackets) seems to be necessary.

csbrady-warwick:
OK. I'll have to have a proper look at this. It shouldn't report you as having write permission to a replica set. Testing for is_primary might only be a workaround for a deeper problem.

edgarcosta:
Please see #2415

JohnCremona:
@edgarcosta could you conform that your PR#2415 (just merged) includes what is here, so this one can be closed?

JohnCremona:
OK, this works as wished.
People might be surprised that they cannot login to a locally-running server when they used to, but (as far as I can tell) there is nothing you can usefully do while logged in when connected to a read-only database.  If this turns out to be wrong we can revisit.

edgarcosta:
I agree, this is different and there is no drawback to also merge this

PR #2413, opened 2018-03-23
Reenable tests

jwj61:
There were two tests which were disabled because of problems in the Artin representation data.  They should now both be fixed.


jwj61:
I spent at least 10 minutes looking through the travis output, but cannot find what test(s) failed.

JohnCremona:
I'll see if I can find it.

JohnCremona:
It's jusy pyflakes:
lmfdb/artin_representations/test_artin_representation.py:3: 'unittest2' imported but unused

You can blame me since that import was only included to allow skipping a test...

jwj61:
Thanks for finding it.

PR #2415, opened 2018-03-26
read only access if connected to a secondary server

edgarcosta:
This forces the application to login with the default user if not connected to a primary server.

@JohnCremona, could you please double check if this works as you would expect.

### DIFF
--- a/lmfdb/base.py
+++ b/lmfdb/base.py
@@ -106,15 +106,19 @@ def makeDBConnection():
 
         logging.info("_mongo_C.nodes = %s" %  (_mongo_C.nodes,) )
         logging.info("_mongo_C.read_preference = %s" %  (_mongo_C.read_preference,) )
-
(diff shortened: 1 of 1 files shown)


JohnCremona:
This works for me and does what it says.  Before merging I want everyone to be quite clear what the effect is:  none at all for the beta webserver;  or for anyone running the server locally who is connecting to the beta database, i.e. the mongo instance  at lmfdb.warwick.ac.uk -- this includes those with accounts on the Warwick machines and is also how all developers used to connect, using the ssh tunnel.  But developers running a local copy using the default m0.lmfdb.xyz will only have read access (whatever passwords they know).  This makes sense since the m0 replica is read-only anyway.
As far as I can see it must therefore already be true that uploading data (or doing anything which requires writing to the database)  is  already only possible when connected to the primary beta webserver at Warwick.  This is documented in the GettingStarted guide ("A read-only copy of the database is hosted at m0.lmfdb.xyz. This is what is used by default").

Issue #1572, opened 2016-06-20
Genus 2 server error on search download


jwj61:
From Jen Balakrishnan via the feedback page:

"Clicking the magma/sage/pari links at the bottom of this page:
http://lmfdb.org/Genus2Curve/Q/?analytic_rank=0 (not just analytic rank 0, but also for any analytic rank query)

results in the following message:

The page you are looking for has produced a server error.
If you want to help improving the page behind ""http://lmfdb.org/Genus2Curve/Q/?analytic_rank=0"", please report that page's URL!"

Include @jbalakrishnan (and maybe directly jennifer.balakrishnan@gmail.com) in replies to this issue.

AndrewVSutherland:
I will fix this later today (I want to wrap up a PR I am working on now for Siegel modular forms first).

AndrewVSutherland:
Fixed by #1578

AndrewVSutherland:
And I'll e-mail Jen directly

edgarcosta:
@jbalakrishnan has reported the same issue again, and I have verified it

AndrewVSutherland:
@edgarcosta I just checked this and it works for me (using chrome on ubuntu), can you give more details on how you reproduced it?  I used the query http://www.lmfdb.org/Genus2Curve/Q/?analytic_rank=0&count=50 and clicking any of the magma, sage, pari buttons seems to work fine.

JohnCremona:
Same for me (also chrome ubuntu).  I'll try on a different machine in a while.

jbalakrishnan:
The issue for me (chrome on OS X) is that the downloaded file seems to be
all ~66,000 curves in the database, not the particular list of curves found
by the query.

Jen

AndrewVSutherland:
@jbalakrishnan Thanks, I get the same problem.

edgarcosta:
I also tested it in OSX with Chrome.

AndrewVSutherland:
Fixed by PR #2416.  As  @jbalakrishnan  indicated, the problem is not that the download buttons caused a server error, but that the download results contained all the curves in the database rather than the subset matching the query.  This has been fixed.

jbalakrishnan:
@AndrewVSutherland Thanks!

edgarcosta:
#2416 has been merged

PR #2416, opened 2018-04-03
Fix for #1572 (again)

AndrewVSutherland:
Fixes issue with download on genus 2 search results page.  To test do a query to select a subset of genus 2 curves (e.g. http://127.0.0.1:37777/Genus2Curve/Q/?analytic_rank=4&count=50) then download the results by clicking on any of the sage/pari/magma buttons at either the top or bottom of the list.  Then check that the download file contains only the curves that it should (e.g. just 10 curves for the query above), not all the curves in the database.

### DIFF
--- a/lmfdb/genus2_curves/main.py
+++ b/lmfdb/genus2_curves/main.py
@@ -459,6 +459,7 @@ def download_search(info):
     lang = info.get('download','text').strip()
     filename = 'genus2_curves' + download_file_suffix[lang]
     mydate = time.strftime("%d %B %Y")
+    print info
     # reissue query here
     try:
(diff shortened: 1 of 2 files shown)


edgarcosta:
The test fails because:
'http://www.maths.bris.ac.uk/~maarb/public/lmfdb2013.html' times out

edgarcosta:
this seems to solve the issue for me.

JohnCremona:
Does anyone have an idea for how we can avoid these timeouts when accessing external websites?  They happen regularly.  Perhaps we can just exclude these from the Travis tests.

edgarcosta:
We can easily remove `lmfdb/test_acknowlegments.py` from the travis battery test if desired.

Issue #2420, opened 2018-04-19
Number field polynomial input crashes on non-integral input


JohnCremona:
I was looking for the field defined by the polynomial x^3-8*x^2-x+8/9 which is mentioned in Lemma 4.13 of https://arxiv.org/abs/1804.06777 but entering that into the search box causes a crash.  This is different from the better (not perfect) message which appears if you enter an integral polynomial defining a field not in the database.
It would not be hard to extend the parsing of that box to allow for this.  Perhaps a job for @koffie ;) unless he wants to change the Lemma instead.

JohnCremona:
Thanks @edgarcosta I managed to find it using Sage's optimzed_representation() method for number fields.
Thanks @jwj61 for offering to fix it.

koffie:
Thanks for fixing the lookup of the troublesome polynomial already on the same day that it appeared in an article on arxiv!

JohnCremona:
I just merged the fix so am closing this.

PR #2421, opened 2018-04-19
Better handling when user inputs a polynomial when searching for a number field

jwj61:
This fixes issue #2420 .  The program was putting the polynomial in the url, which caused a problem because the polynomial contained a slash.

If a user inputs a polynomial in Z[x] which is in the database, the url is converted to the label anyway.  So, if the polynomial is in the database (after polredabs), we switch to the label right away.  This change also makes the process a little more efficient.  Before, the polynomial would go through polredabs several times.  Now it only happens once.

To test, enter a polynomial such as x^3-1/3 (or the one from the issue) in the number field search page.

### DIFF
--- a/lmfdb/number_fields/number_field.py
+++ b/lmfdb/number_fields/number_field.py
@@ -615,7 +615,7 @@ def number_field_search(info):
         query = {'label_orig': info['natural']}
         try:
             parse_nf_string(info,query,'natural',name="Label",qfield='label')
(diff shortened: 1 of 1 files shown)


JohnCremona:
Good!  I'll test this tomorrow.  It's funny that the solution was to remove the clean_input() call...

jwj61:
The real change is that the line in question uses a label returned by our own function, not the raw input supplied by the user.  The result of that function is already clean.

Issue #2422, opened 2018-04-20
Search for number fields by signature times out


JohnCremona:
From the bug report system: "Global Number Field search results did not work for signature for [0,3], [0,4]..."
I confirmed this on prod and beta.  It times out, even if you include the degree too (e.g. degree 6 and signature [0,3]).
The existing indices are viewable at http://beta.lmfdb.org/inventory/numberfields/fields/indices/ showing that there is an index on signature and a joint index on signature and degree, so it is not that.  And the signature is stored as a string like '0,3' so I do not know what is wrong, unless something is not right with the code which turns the search box entries into an actual search.  From the command line in Sage,
`fields.find({'signature':'0,3'})`
works fine right away.

edgarcosta:
I'm not sure what is going on. It looks like a database issue, and some mismatch between m0 and ms.

On the ms.lmfdb.xyz (where www.lmfdb.org reads from) the timings are very painful, 30 mins to do the find command and 2.4s for the count.
```
2018-04-20T12:43:27.745+0000 I COMMAND  [conn49589] command numberfields.fields command: count { count: "fields", query: { degree: 6, signature: "0,3" } } planSummary: COUNT_SCAN { signature: 1, degree: 1 } keyUpdates:0 writeConflicts:0 numYields:16968 reslen:62 locks:{ Global: { acquireCount: { r: 33938 } }, Database: { acquireCount: { r: 16969 } }, Collection: { acquireCount: { r: 16969 } } } protocol:op_query 2415ms
2018-04-20T12:52:29.123+0000 I COMMAND  [conn49541] command numberfields.fields command: find { find: "fields", filter: { degree: 6, signature: "0,3" }, sort: { degree: 1, disc_abs_key: 1, disc_sign: 1 }, limit: 20 } planSummary: IXSCAN { signature: 1, ramps_all: 1, degree: 1 } keysExamined:2170099 docsExamined:2170098 hasSortStage:1 cursorExhausted:1 keyUpdates:0 writeConflicts:0 numYields:84402 nreturned:20 reslen:22429 locks:{ Global: { acquireCount: { r: 168806 } }, Database: { acquireCount: { r: 84403 } }, Collection: { acquireCount: { r: 84403 } } } protocol:op_query 1842030ms
```

While from m0.lmfdb.xyz things are way faster, the count takes 0.6s and the find 1.4s.
```
[...]

PR #2423, opened 2018-04-22
Timeout Decorator for search queries

edgarcosta:
Hi,

This is related to #2422.

With the goal of being able to test if searches are performed in a reasonable amount of time and give a better feedback to the user whenever the query takes too long, I wrote a small script that decorates our db queries with `.max_time_ms(timeout)` and should replace the following two lines
```
nres = res.count()
res = res.skip(start).limit(count)
```
present in almost every search handling code.
I have only replaced these two lines for numberfieds, but should be easy to do the same for rest of the search fields.

try it out with: 
http://localhost:37777/NumberField/?start=0&ur_primes=2%2C3%2C5%2C7&ram_quantifier=contained&count=20

In this case, the query times out while trying to do the count:
```
2018-04-22T17:27:03.625+0000 I COMMAND  [conn4343] command numberfields.fields command: count { count: "fields", query: { ramps: { $nin: [ "2", "3", "5", "7" ] } }, maxTimeMS: 60000 } planSummary: IXSCAN { ramps: 1, degree: 1 } keyUpdates:0 writeConflicts:0 exception: operation exceeded time limit code:50 numYields:2373 reslen:89 locks:{ Global: { acquireCount: { r: 4748 } }, Database: { acquireCount: { r: 2374 } }, Collection: { acquireCount: { r: 2374 } } } protocol:op_query 60662ms
```

This isn't perfect.  But I think it is much more helpful to see:
```
Error: The search query took longer than expected! Please help us improve by reporting this error here.
```
than Google's `Error: Server Error`.
Also, if the server is overloaded, the timeout will not help us, as it only limits the CPU cumulative time.

Let me know what you think.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -76,7 +76,7 @@ script:
     # run nosetests in parallel with the aid of parallel
     # we cannot use the option --processes in nosetests, as MongoDB doesn't support forking
     # usually we have at most 4gb of RAM per test, we should avoid running many jobs at once
-    - ${SAGE} -sh -c 'parallel --joblog joblog --jobs 4 --results nosetests  --progress --verbose -a list ${SAGE_DIR}/local/bin/nosetests -v -s --testmatch=${TESTMATCH} ${COVERAGE}'
+    - ${SAGE} -sh -c 'parallel --joblog joblog --jobs 3 --results nosetests  --progress --verbose -a list ${SAGE_DIR}/local/bin/nosetests -v -s --testmatch=${TESTMATCH} ${COVERAGE}'
 
 after_failure:
     # if some process failed, it will be clear in joblog which one failed
--- a/lmfdb/elliptic_curves/elliptic_curve.py
+++ b/lmfdb/elliptic_curves/elliptic_curve.py
@@ -10,7 +10,7 @@ import tempfile
 import os
 import StringIO
 
-from lmfdb.utils import web_latex, to_dict, web_latex_split_on_pm, random_object_from_collection
+from lmfdb.utils import web_latex, to_dict, web_latex_split_on_pm, random_object_from_collection, search_cursor_timeout_decorator
 from lmfdb.elliptic_curves import ec_page, ec_logger
 from lmfdb.elliptic_curves.ec_stats import get_stats
 from lmfdb.elliptic_curves.isog_class import ECisog_class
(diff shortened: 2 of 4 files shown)


edgarcosta:
@davidlowryduda , let me rephrase your questions:

- What is the added value of raising the ValueError?
There is no added value, and I feel I should pass the method `search_input_error` and its args, but then someone somewhere must initialize `info['err']`.

- Why does `info['err']` must be initialized?
No clue, I'm just imitating what the `SearchParser` does.
If I don't do it, the template freaks out.

edgarcosta:
I cleaned the code a bit, and have not incorporated it into elliptic curves.
try: 
http://localhost:37777/EllipticCurve/Q/?start=0&include_cm=exclude&surj_primes=2%2C3%2C5%2C7&surj_quantifier=include&count=100

you might need to lower the timeout to for example 1s, to observe it working.

edgarcosta:
(and I broke elliptic functions, I will fix this later on)

edgarcosta:
The sad thing is that the wrapper cannot really handle the timeout exception, and must pass an exception outside, as the outside function `whatever_search` must still return something.

I will try to do the other pages over the next couple of days.

edgarcosta:
@JohnCremona, @jwj61 and @AndrewVSutherland, do you agree with this idea of limiting the time to do a search  query on the website? I would like to hear from one of you before trying to integrate this function with the remainder of the LMFDB.

JohnCremona:
Yes I do like this.  I am no expert on decorators so thought that I could not comment sensibly on the suggested code changes.  Is it not the case that gunicorn (or some aspect of our configuration) already has a timeout for requests?  So is the benefit of this change that we can catch the timeout in our own code and hence display something useful, rather than having the system timeout and displaying some unhelpful error message?

AndrewVSutherland:
@edgarcosta Yes, I think this is a good idea.  I might suggest merging and pushing what you have to beta and prod so we can try this out in production before you make changes everywhere else (do that in a separate PR). 

I presume you are still planning to keep the gunicorn timeouts in place, but possibly make them longer?

JohnCremona:
Are you planning any changes in gunicorn config on the warwick servers, beta and lmfdb.warwick.ac.uk?

edgarcosta:
Very good question! 

I had set the search query timeout to 
- 60 seconds for beta (and development)
- 27 seconds for production (the reason behing this number is the 30s timeout of google's load balancer)

But now looking at your config files, it looks likes I should've just set 27 seconds for both of them.
As the gunicorn timeout for http://beta.lmfdb.org/ is 30s. 
Do you think 27 seconds is enough for development?


Regarding http://lmfdb.warwick.ac.uk, I see that the gunicorn timeout is 20s, this perhaps should be raised to 30s.

AndrewVSutherland:
I'm happy to merge this and #2426 and push to beta where we can see it in action.  Is there any reason to hold off?

edgarcosta:
I have just pushed some small changes.
Now the timeout is the same for both versions and it got reduced 25
seconds.
This is mostly because gunicorn's for beta is set at 30s.

Feel free to merge it.

edgarcosta:
tests have failed, I will look into those

edgarcosta:
The current issue is: 
- the link http://hobbes.la.asu.edu/lmfdb-14/ is timing out

On the first ran, these searches also timed out (most likely because we were running several tests at the same time):
  - http://www.lmfdb.org/EllipticCurve/Q/?rank=0&include_cm=include&torsion=2&sha=4&surj_quantifier=include&count=100
  - http://www.lmfdb.org/EllipticCurve/Q/?start=0&include_cm=include&surj_primes=2&surj_quantifier=include&optimal=on&count=100

These weren't an issue when I reran the tests. To avoid the same issue in the future, I have fixed the first one with a new index, but the second one can't be fixed with any index.

PR #2425, opened 2018-04-23
removing old links to code.google.com

edgarcosta:
I'm not sure if these links are ever seen, but just to be safe I decided to point them to the update urls.

### DIFF
--- a/lmfdb/modular_forms/elliptic_modular_forms/views/templates/experimental/emf_new.html
+++ b/lmfdb/modular_forms/elliptic_modular_forms/views/templates/experimental/emf_new.html
@@ -198,7 +198,7 @@ in the field of complex numbers. The values are shown in the table below.
           <th>Eigenvalue
         </tr>
       </thead>
-      <tbody>
+      <tbody
         {% for Q,c,ev in atkinlehner %}
         <tr>
           <th> {{Q}} </th>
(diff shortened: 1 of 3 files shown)


lmfdb/modular_forms/elliptic_modular_forms/views/templates/experimental/emf_new.html
```diff
@@ -198,7 +198,7 @@ <h2>Further Properties</h2>
           <th>Eigenvalue
         </tr>
       </thead>
-      <tbody>
+      <tbody
```
davidlowryduda: I think an extra closing `>` was deleted here.

davidlowryduda:
Other than the one (super minor) change I commented on above, this is ready to be merged.

It doesn't seem like any of the templates in this folder are used anywhere, but  I suppose it makes sense to do this in case they are used somewhere else. I notice that these files haven't been touched for 4 years (except for a minor CSS update that appears to have been lmfdb-wide search-and-replaced). I wonder if these files have a future, or if they're not worth keeping? Maybe @fredstro would know.

lmfdb/modular_forms/elliptic_modular_forms/views/templates/experimental/emf_new.html
edgarcosta: Fixed it

edgarcosta:
Great question, I just noticed the outdated links.

Just to be clear, the template `404.html` is still currently in use. When one calls `flask.abort(404, ... )` flask calls `not_found_404`. 
But perhaps the template needs to cleaned, as it looks like we always pass a message.


Further, `lmfdb/characters/main.py` calls directly the template and this perhaps should be fixed.

PR #2426, opened 2018-04-24
Two new tests to test signature searches

### LINKED ISSUE #2422: Search for number fields by signature times out

edgarcosta:
This combined with #2423 should close #2422.

### DIFF
--- a/lmfdb/number_fields/test_numberfield.py
+++ b/lmfdb/number_fields/test_numberfield.py
@@ -97,5 +97,13 @@ class NumberFieldTest(LmfdbTest):
         L = self.tc.get('/NumberField/stats', follow_redirects=True)
         assert 'Class number' in L.data 
 
+    def test_signature_search(self):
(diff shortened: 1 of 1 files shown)


davidlowryduda:
This also seems good to me. I suggest we merge this and #2423 at the same time.

PR #2437, opened 2018-05-08
Update completeness page for number fields.

jwj61:
These are all text changes to the page showing completeness claims for number fields.  Mostly this is updating to account for recent (and not so recent) additions.


JohnCremona:
Comment -- to save work in future and to get such changes appearing right away on beta and prod, you can put this info into a knowl and use the static knowl funcionality.  For example, http://www.lmfdb.org/EllipticCurve/Q/Source just shows the content of the knowl http://www.lmfdb.org/knowledge/show/dq.ec.source, so editing is easier.

PR #2445, opened 2018-05-11
Hilbert modular eigenvalue spacing

### LINKED ISSUE #2444: Line up eigenvalues on Hilbert modular forms pages
On Hilbert modular forms pages (like [this random one](http://beta.lmfdb.org/ModularForm/GL2/TotallyReal/6.6.1229312.1/holomorphic/6.6.1229312.1-169.1-b)), the column containing the coefficients appears ragged since some entries have minus signs and others don't. We could make it not appear ragged if we inserted the space for a minus sign on each row without it.

davidlowryduda:
On lines for eigenvalues that are positive, insert the space for a minus sign to make the column line up nicely. For comparison, look at the eigenvalues on a page like

http://beta.lmfdb.org/ModularForm/GL2/TotallyReal/6.6.1229312.1/holomorphic/6.6.1229312.1-169.1-b

before and after this change.

This resolves #2444

### DIFF
--- a/lmfdb/hilbert_modular_forms/hilbert_modular_form.py
+++ b/lmfdb/hilbert_modular_forms/hilbert_modular_form.py
@@ -18,7 +18,6 @@ from markupsafe import Markup
 from lmfdb.utils import to_dict, random_value_from_collection, web_latex_split_on_pm
 from lmfdb.search_parsing import parse_nf_string, parse_ints, parse_hmf_weight, parse_count, parse_start
 
-from . import hmf_logger
 
 def db_forms():
     hmfs = getDBConnection().hmfs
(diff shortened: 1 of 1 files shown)


edgarcosta:
What about aligning that column on the right?

davidlowryduda:
There is generally no hope of also having right-alignment, since there will be eigenvalues of varying numbers of digits. Or polynomial ones like

http://beta.lmfdb.org/ModularForm/GL2/TotallyReal/2.2.233.1/holomorphic/2.2.233.1-36.1-b

If you're suggesting right-alignment instead of left, I find it more aesthetically pleasing to have all three columns match alignment. Do you prefer right-alignment?

edgarcosta:
Sorry about that. I agree, that over a number field, it would be horrible to have right alignment! 

I believe that your fix takes care of it in a nice and minimal way.

PR #2446, opened 2018-05-11
Allow up-to-date version of flask-login

### LINKED ISSUE #2173: flask-login version
We use a fixed version of flask-login, should we now allow to update it? 
If we do this, everybody would need to update their version of flask-login

davidlowryduda:
When flask-login updated to version 0.3.0, they changed `current_user.is_authenticated()` (and similar) to `current_user.is_authenticated`, and one cannot use the `()` version anymore. As noted in #2173, we've used a fixed pre-0.3.0 version for the LMFDB for a long time.

In this pull request, I update the code to work with both pre 0.3.0 and post 0.3.0 versions of flask-login. If one runs the site or the tests with a version of flask-login that is pre 0.3.0, then a WARNING is logged suggesting that it may be a good idea to update flask-login.

This closes #2173

### DIFF
--- a/lmfdb/inventory_app/templates/control_panel.html
+++ b/lmfdb/inventory_app/templates/control_panel.html
@@ -17,7 +17,7 @@ Inventory Control Panel
   <div id='dialog'></div>
   <div class='listing'>
   <div class='unbreaking'>
-      {% if current_user.is_authenticated() %}
+      {% if user_is_authenticated %}
       <h2>General Inventory Control</h2>
         <input type='button' id = 'markgone' title = 'Check which collections are gone and mark' value ='Mark Collections Gone' onclick='sendControlRequest("mark_gone")'>
         </p>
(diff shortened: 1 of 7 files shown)


davidlowryduda:
Thank you for noticing --- this isn't quite ready for another check yet. I see another thing I must check for, but I need to think for a moment about the best way to do this.

JohnCremona:
I tested this using both the old flask-login and the new and both seem to work fine.

edgarcosta:
What is the impact that this will have on the beta and prod servers?
Should we upgrade flask once the code gets pushed to prod and beta?

JohnCremona:
@davidlowryduda wrote the code (and I tested it) so that it works equally well on the older version is flask-login, so no action is necessary, but since the requirements.txt file no longer forces the version number we can expect developers to be running the new version before long (e.g. after their next Sage upgrade).  Hence, I do suggest that we upgrade flask-login on beta and prod after the code is pushed there -- perhaps this would be most conveniently done when we upgrade to Sage 8.2, recently released (and runs fins when I checked).

PR #2450, opened 2018-05-15
Get rid of oldporedabs for numberfields.  

jwj61:
Sage 8.1 computes the new version, so we don't need the extra lookup.

To test, go to the number field web page and search for x^4 - 2*x^3 + x^2 + 2*x + 1, and it will still be found even after this change (this is the old polredabs version of a polynomial).  To create a situation where it is not found is now complicated -- you would need to test against some old version of sage.

There will be a data pull request as well where we delete the two entries from the number field database, but the code and data pull requests can be done in either order.


JohnCremona:
After this is fully merged, will the keys oldpolredabscoeffs and oldpolredabscoeffhash be removed from the database?

PR #2458, opened 2018-05-23
Knowl show quality

JohnCremona:
This PR is to generate discussion rather than asking to be merged now.  With this, when you view a knowl (not as a pop-up, but on its own page such as http://www.lmfdb.org/knowledge/show/mf.bianchi.bianchigroup ), under the list of Authors there is a new section "Knowl status" showing its review status (beta, ok or reviewed) and the last author with datestamp.

Currently there is jusy one "reviewed" knowl (http://www.lmfdb.org/knowledge/?reviewed=on&filter=) and around 15 "ok" ones, but a process has started to review (and where necessary correct) more.  Personally I think the "ok" status is redundant.

### DIFF
--- a/lmfdb/knowledge/knowl.py
+++ b/lmfdb/knowledge/knowl.py
@@ -254,6 +254,14 @@ class Knowl(object):
             a = users.find({"$or": a_query}, ["full_name"])
         return a
 
+    def last_author(self):
+        """
+        Full names for the last authors.
+        (lookup for all full names in just *one* query, hence the or)
+        """
+        users = getDBConnection().userdb.users
+        return users.find_one({'_id': self._last_author}, ["full_name"])["full_name"]
+
     @property
     def id(self):
         return self._id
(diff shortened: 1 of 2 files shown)


JohnCremona:
The extra commit implements @davidlowryduda 's suggestion by appending " (beta status)" to the title of any knowl whose status is beta (i.e. neither 'ok' nor 'reviewed').  This appears the the tiny title line of pop-up knowls as well as in the main knowl view.

Again, this is only a proof of concept and whether we actually want this is open for debate.  In particular the added wording can be changed trivially (it's around line 310 of knowl.py)

JohnCremona:
The last commit only adds " (beta status)" to the title of knowls when in beta mode.  Right now most knowls are still beta but we would not want to advertise that on the main website.  After most knowls have been reviewed this can change again.

JohnCremona:
Note: there is some more on this currently upstream as a PR to my branch so I'll leave this open but there will be more.  It could also be merged now if people liked the new stuff already.

Issue #1147, opened 2016-05-03
top/bottom knowls on ecnf pages


AndrewVSutherland:
Currently we have top/bottom knowls for elliptic curves over Q but not over number fields.

Note that this will not work completely until https://github.com/LMFDB/lmfdb/issues/1140 is fixed, but issue 1140 does not prevent us from putting the links in now.

AndrewVSutherland:
This issue depends on #1333, which is scheduled for Release 1.1.

JohnCremona:
I have just added this (PR in a minute).  Both #1140 and #1333 are fixed.

PR #2459, opened 2018-05-23
add top & bottom knowls for ecnf home pages

JohnCremona:
As requested in Issue #1147.
I found it hard to think of curves with any iinteresting properties to test this out on, ending up with http://localhost:37777/EllipticCurve/2.0.11.1/11.1/a/1 and http://localhost:37777/EllipticCurve/6.6.300125.1/1.1/a/2 (the latter is the only curve we have with a rational point of order 37).  With this patch you should see a top knowl for each, and (if logged in) the ability to create top knowls for all other ecnfs and bottom knowls for all.

### DIFF
--- a/lmfdb/ecnf/templates/ecnf-curve.html
+++ b/lmfdb/ecnf/templates/ecnf-curve.html
@@ -30,6 +30,10 @@
   </style>
 #}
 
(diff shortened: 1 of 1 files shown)


AndrewVSutherland:
Looks good, I like the curve http://localhost:37777/EllipticCurve/6.6.300125.1/1.1/a/2 :)

PR #2460, opened 2018-05-23
Minor changes (mostly making breadcrumbs consistent)

ngonhi:
- statistics for Genus 2 curves
- Delete Global number field labels in Learn more about in Global number field label page


JohnCremona:
I am investigating the tests which fail.

JohnCremona:
It's good to have consistency of capitalisation.
I like thte extra layer of breadcrumbs "Modular forms" on HMF and BMF pages.

In some places one of the "learn more about" links was omitted deliberately since that information is already displayed on the page itself: Artin rep main page; Bianchi mf main page, and several others.  In fact that is the reason why there is a function for omitting one element from that list!  I think we would like that put back to how it was (if others agree).

There is a problem with the new ecnf breadcrumbs, which is raising errors and causes te test suite to fail. It is debatable whether it is worth having the extra layer of bread crumbs "Elliptic Curves ->Number Fields -> ..." when the first two point to the same page, but I don't have strong feelings.  However, the breadcrumbs for some ecnf pages crash -- please fix!

JohnCremona:
I was puzzled as to why pyflakes does not pick up the missing commas.  It  sees something like 
`(string,url_for())(string,url_for())`
and thinks that is valid python but at run-time it tries to evaluate the first tuple at the second, which of course does not work.  -1 for pyflakes!

lmfdb/ecnf/main.py
```diff
@@ -342,7 +352,8 @@ def show_ecnf(nf, conductor_label, class_label, number):
         return search_input_error()
     label = "".join(["-".join([nf_label, conductor_label, class_label]), number])
     ec = ECNF.by_label(label)
-    bread = [("Elliptic Curves", url_for(".index"))]
+    bread = [("Elliptic Curves", url_for(".index"))
```
JohnCremona: This line needs a comma at the end

JohnCremona: This line needs a comma at the end

lmfdb/ecnf/main.py
```diff
@@ -314,7 +323,8 @@ def show_ecnf_isoclass(nf, conductor_label, class_label):
     label = "-".join([nf_label, conductor_label, class_label])
     full_class_label = "-".join([conductor_label, class_label])
     cl = ECNF_isoclass.by_label(label)
-    bread = [("Elliptic Curves", url_for(".index"))]
+    bread = [("Elliptic Curves", url_for(".index"))
```
JohnCremona: This line needs a comma at the end

jenpaulhus:
Sorry, John. We made a mistake while getting used to github. This is students working with me and this is an early version of code we  are not quite ready to  push  yet, but which ended up escaping us.   This pull request can be deleted.

PR #2464, opened 2018-05-25
ec/ecnf URL shortening for JC to review

AndrewVSutherland:
Per comment https://github.com/LMFDB/lmfdb/issues/2277#issuecomment-392056034, this PR makes the minimal changes suggested there, which only involves changing the html templates to set the value of "include" options to a blank string so that it won't appear in the URL (it actually isn't necessary to change the query processing at all).  It does not change the "surj_quantifier=include" attribute as this would require some minor changes to the query processing as well.

The purpose of this PR is solely so that @JohnCremona can review the change and decide if he wants to make it or not (so please do not merge this unless you are John Cremon).

### DIFF
--- a/lmfdb/ecnf/templates/ecnf-index.html
+++ b/lmfdb/ecnf/templates/ecnf-index.html
@@ -72,13 +72,13 @@ Please enter a value or leave blank:
   <td>
          <select name='include_base_change'>
            <option value="off">exclude</option>
-           <option value="on" selected>include</option>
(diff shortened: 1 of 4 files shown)


JohnCremona:
Thanks, I'll check it out.

JohnCremona:
Thanks! My name is JC and I claim the prize.  I see the logic now: when the code treats one value as a default then the form does not need to positive fill that field.  It looks good -- I tried many combinations.

Issue #2465, opened 2018-05-25
Kloosterman sum compute button on Dirichlet characters not working


AndrewVSutherland:
If you go to a Dirichlet character page, e.g. http://www.lmfdb.org/Character/Dirichlet/7/3 and enter a pair of numbers in the Kloosterman sum box (e.g. 1,2) and click the compute button nothing happens.  The compute buttons for Gauss and Jacobi sums do seem to work, and the slider for Kloosterman sums does work (but also see #71).

JohnCremona:
PR #2468 fixes this.  However there is more discussion at #71 suggesting that (1) the sliders are unnecessary and (2) we should display the default values when the page loads.  Hence I will keep this issue open even though I will merge #2468 on the grounds that having something working beats it not working, while we decide.

davidlowryduda:
Yes, I was just figuring out how to display default values. (I'm still a javascript neophyte, but I think this is fine).

I haven't removed the sliders. I was unsure whether opinions have changed on their utility. But if we do want to remove them, I know how.

AndrewVSutherland:
I vote that we remove them, I don't think they add any particular value and make the U/I different from the rest of the LMFDB.  If you want to go ahead and add this to your outstanding PR (or a new PR if you prefer), I will review it and then we can close both this issue and #71.

davidlowryduda:
Ok. I've updated my PR to remove the sliders as well.

AndrewVSutherland:
@davidlowryduda, I am reopening this issue, as the character sums are not being hidden for very large modulus characters as they should be, see

http://www.lmfdb.org/Character/Dirichlet/19265902/13392695

for example.

AndrewVSutherland:
Fixed via #2548

Issue #71, opened 2013-09-15
Gauss sum should display initially in Dirichlet char home page


jbalakrishnan:
URL or blueprint/codefile: The home page of any Dirichlet character.

In order to see a Gauss, Jacobi, or Kloosterman sum, I have to click Compute or move a slider.

What should it be?

It would be better if values were displayed when the page first loads.
The default should be a=1 for the Gauss sum.  The default should not be
1,1 for Kloosterman sums, but 1,2 would be ok.

Also, the Jacobi sums are written as sums of roots of unity, but it
probably should be numerical, like the others.

AndrewVSutherland:
I think the sliders should be removed; when the modulus is large they are pretty useless (see http://www.lmfdb.org/Character/Dirichlet/99999/2).  I took a quick look at automatically filling in the values when the page loads but I don't see an easy way to do it (I'm sure it can be done).

pascalmolin:
I can do this. So remove slider (I agree, I used the jacobi sums recently
and never the slider) and place default input: 2 for jacobi and 1,1 for
kloosterman sounds reasonable.

Pascal

2016-07-11 16:14 GMT+02:00 Andrew Sutherland <notifications@github.com>:

AndrewVSutherland:
@pascalmolin Thanks!  But let me merge PR 1703 before you create a new branch, it contains changes to some of the Character/Dirichlet files.

PR #2470, opened 2018-05-29
Display default values for characters on character pages.

davidlowryduda:
This was first mentioned in #71, although there is more discussion
there. See also #2465.

### DIFF
--- a/lmfdb/WebCharacter.py
+++ b/lmfdb/WebCharacter.py
@@ -132,7 +132,7 @@ class WebCharObject:
 ###  Dirichlet type
 
 class WebDirichlet(WebCharObject):
-    """ 
+    """
     For some applications (orbits, enumeration), Dirichlet characters may be
     represented by a couple (modulus, number) without computing the Dirichlet
     group.
@@ -337,7 +337,7 @@ class WebHecke(WebCharObject):
         numlabel = self.number2label( c.exponents() )
         if prim == None:
             prim = c.is_primitive()
-        return (modlabel, numlabel, self.char2tex(c), prim ) 
+        return (modlabel, numlabel, self.char2tex(c), prim )
 
     @staticmethod
     def ideal2tex(ideal):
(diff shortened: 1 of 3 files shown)


AndrewVSutherland:
So this works and the sliders are gone, which is great, but before we can merge this we are going to need to impose a bound on the modulus for which we try to compute Gauss/Jacobi/Kloosterman sums.  If you go to a random Dirichlet character, the modulus is likely to be large enough that these sums cannot be computed in a reasonable amount of time (i.e. a second or two).   Try going to http://127.0.0.1:37777/Character/Dirichlet/87497/76911 for example.  The compute buttons are then completely non-responsive.

I would suggest that we not try to compute these sums when the modulus exceeds 999 (it still works OK at 9999 but is sluggish, whereas at 999 everything is snappy).  I think we should simply hide the sums entirely when the modulus is 1000 or greater.  Is this a change you can make @davidlowryduda ?

pascalmolin:
There should be already this kind of mechanism. If the modulus is large the
WebCharacter class has some features disabled (character values, character
sums...).
Right now the limit is fixed at 100000, we could set it to 1000.

(see the lines "if modulus < 100000:" in characters/main.py)

Do you think there should be a different limit for character sums and
display of character values and generators ?

Pascal



2018-05-30 0:05 GMT+02:00 Andrew Sutherland <notifications@github.com>:

AndrewVSutherland:
@pascalmolin Yes, I am aware of this.  I think ideally there should different (lower) value for character sums.  But if this is a pain to implement, compromising on 10000 for both would be manageable.

pascalmolin:
Done in PR#2471 (sorry I should have merged with David's PR).

Also I read David's PR#2470, I think we could handle these default values
differently:
- using javascript calls makes the server recompute the dirichlet group for
each
- those values could be computed from the beginning and added to the
character object instead.

I can do this if you want (for q<1000 this is not a big deal).

Pascal

2018-05-30 11:14 GMT+02:00 Andrew Sutherland <notifications@github.com>:

davidlowryduda:
Yes, when I wrote my initial PR I hadn't thought about limiting the modulus, so it didn't occur to me to precompute values. But it's not clear to me that it's a good idea to precompute them still, unless I misunderstood your comment and you meant to add them to the database? Otherwise I think that precomputing a million Kloosterman sums on each page is a bit excessive.

I notice that Pascal's #2471 is almost exactly what I would have done, but also that we both modify some of the same parts of the same file in slightly different ways. In a little bit I will go ahead and absorb #2471 into #2470 and not worry about the merge later.

davidlowryduda:
Ah, i see now that @pascalmolin put in his own pull request to do the same thing. These two pull requests are now essentially duplicate. I shall close mine and merge his.

Issue #2456, opened 2018-05-20
Inventory Editor - sometimes can't change types


davidlowryduda:
On some of the inventory pages, it doesn't seem possible to change the types. For example, I cannot change the types on the Hilbert Modular Form `forms` collection (http://beta.lmfdb.org/inventory/hmfs/forms/) through the online interface (http://beta.lmfdb.org/inventory/hmfs/forms/edit/). But I can alter other types sometimes without any problem.

The interface appears to work as expected and displays the `Submitted edits successfully` page and redirect, but the types don't actually change.

hratcliffe:
Immediate bug should be fixed now, I'm not entirely sure whether the proximate cause is already fixed so I'll have to do some more checking.

JohnCremona:
@hratcliffe -- is there a PR or was this fixed by tweaking the database contents?

hratcliffe:
Just the contents. If there is a deeper problem there may be code.

JohnCremona:
OK -- so if @davidlowryduda can confirm that what used not to work now works, we can close this issue.

davidlowryduda:
I still can't seem to change the types on that page.

hratcliffe:
Might have missed some entries, I'll have another look

davidlowryduda:
Thanks --- I have no problems now.

PR #2472, opened 2018-05-30
Fix bug and add reporting

hratcliffe:
Fixes the issue leading to #2456 Adds some reporting to make such bugs easier to diagnose in future.


JohnCremona:
@davidlowryduda does this fix the bug for you?  If so I'll merge it.

hratcliffe:
Bug should have been fixed by data modifications: this will prevent a reoccurence.

Issue #2447, opened 2018-05-13
Search Galois groups by order


kedlaya:
The search results for Galois groups include order as one of the fields. So why not the option to search for Galois groups by order rather than (or in addition to) degree?

jwj61:
Putting it as an addition definitely makes sense.

I thought it would take 5 minutes, but it is a little more complicated because we store the group orders as strings (23! is pretty big).  So, we will need to use an alternate strategy so searching for ranges works.  After switching to postgres, then we can probably safely store the order as an integer, even if we expand the range in terms of degree.

kedlaya:
If waiting until after the postgres conversion would make this easier, then I propose to wait.

AndrewVSutherland:
This was addressed by PR #2473 and is now live on beta.

PR #2473, opened 2018-05-31
Add search by group order and gap id's to transitive groups

jwj61:
In response to issue #2447 this adds searching by group order to the transitive group section.

This raises the question of the best way to order search results.  With the old ordering (by (n,t)), if you search for groups of order 24, S_4 appears many times, but they are not grouped together.  So, I added gap id's to the database (to the extent they exist for these groups).  But, if you always sort results by (order, gap id), then a search with say n=6 will not produce the groups ordered by "t", which seems natural in that case.

So, the code now picks a sort order based on the search.  Usually it is the old order (n,t), unless "order" is specified and "n" is not.  Then it sorts by (order, gap id, n, t).  This is clearly a choice, so open to discussion and opinions.

Testing the search, try searching (n=6), (order=24), (n=6 and order=24), for example.

As long as gap id's are now in the database, we display them on the individual group pages down in the section of pure group invariants (as opposed to invariants of the transitive action).  There are 3 cases: (1) we have the id and it is in our small groups database (order of group < 512 I think) ; (2) we have the id, but the group is **not** in our small groups database; (3) we don't have an id (because magma doesn't have it), probably because there are no ids fixed for that group order.  In case (1) we display a dynamic knowl for the group, in (2) we just display the pair [order, id], and in (3) we say that there is no id.  To see the various options, search with n=9, and then pick groups of various orders (A_9 and S_9 are the only ones of type (3) with n=9).

### DIFF
--- a/lmfdb/galois_groups/main.py
+++ b/lmfdb/galois_groups/main.py
@@ -103,19 +103,6 @@ def index():
                 ('Galois group labels', url_for(".labels_page"))]
     return render_template("gg-index.html", title="Galois Groups", bread=bread, info=info, credit=GG_credit, learnmore=learnmore)
 
-# FIXME: delete or fix this code
-# Apparently obsolete code that causes a server error if executed
-# @galois_groups_page.route("/search", methods=["GET", "POST"])
-# def search():
-#    if request.method == "GET":
-#        val = request.args.get("val", "no value")
-#        bread = get_bread([("Search for '%s'" % val, url_for('.search'))])
-#        return render_template("gg-search.html", title="Galois Group Search", bread=bread, val=val)
-#    elif request.method == "POST":
-#        return "ERROR: we always do http get to explicitly display the search parameters"
-#    else:
-#        return flask.abort(404)
-
 # For the search order-parsing
 def make_order_key(order):
     order1 = int(ZZ(order).log(10))
@@ -240,7 +227,7 @@ def render_group_webpage(args):
         data['subinfo'] = subfield_display(C, n, data['subs'])
         data['resolve'] = resolve_display(C, data['resolve'])
         if data['gapid'] == 0:
-            data['gapid'] = "No gap id's for groups of this order"
+            data['gapid'] = "Data not available"
         else:
             data['gapid'] = small_group_display_knowl(int(data['order']),int(data['gapid']),C, str([int(data['order']),int(data['gapid'])]))
         data['otherreps'] = wgg.otherrep_list()
--- a/lmfdb/galois_groups/main.py
+++ b/lmfdb/galois_groups/main.py
@@ -143,6 +143,7 @@ def galois_group_search(**args):
     try:
         parse_ints(info,query,'n','degree')
         parse_ints(info,query,'t')
+        parse_ints(info,query,'order')
         for param in ('cyc', 'solv', 'prim', 'parity'):
             parse_bool(info,query,param,minus_one_to_zero=(param != 'parity'))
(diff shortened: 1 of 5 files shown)


AndrewVSutherland:
This looks great!  I especially like being able to view all the small group data by clicking on the GAP id (now I want to add this feature to Sato-Tate groups as well).

My only comment is that in cases where the group order is out of range of our current small groups database (i.e. > 256) I would simply have it say "not available" rather than "No gap id's for groups of this order", since for group orders less than 2000 there are GAP id's, we just don't have them in the LMFDB.

This PR motivates me to put more of the small groups data into the LMFDB -- I will try to do this next week while I am at AIM (I will be adding data on GL2(Fp) subgroups next week in any case, and it would be natural to do this at the same time).

AndrewVSutherland:
Actually, I have one more comment (a feature request really).  It would be nice to be able to search by GAP id as well.  A quick and easy way to do this would be to have the "search by label" box accept GAP ids as well as transitive group ids (e.g. if I type in [24,12] instead of 8T14 I should go to a search results page that lists all the transitive groups with isomorphic to [24,12]).  Alternatively you could create a separate input box for this on the main search page.

jwj61:
I was a little uneasy about "No gap id's for groups of this order", but I think you misunderstand the situation a little bit.  I put gap ids in the transitive group database for any group in the transitive group where magma knows the gap id.  This goes beyond the range of the current lmfdb small groups database.  If you have a group in this in between area (e.g., 9T32), then it displays the id, not the message.  The id is simply not clickable since there is no information to show.  It checks the small group database when generating the page to see if it should be a simple pair or a dynamic knowl.

So, the message "No gap id's ..." is only in cases where magma does not have an id.  I am not crazy about "Data not available" since that implies that the data exists, and for S_9, there is no full classification of groups of order 9!.  Thoughts on a third possibility for the wording?

I like the suggestion about searching by gap id in the label box and will get it done.  I may not get it done for a couple of weeks while I am traveling (I leave later today).

JohnCremona:
I have updated http://beta.lmfdb.org/inventory/transitivegroups/ since the groups collection has a new field.  Perhaps we should add this to our standard procedure for copying a collection to the cloud since the times when the inventory will need updating are precisely the times when we want to copy the collection.

AndrewVSutherland:
You are right, I did misunderstand, I see that you list gap id's in all cases where Magma supports small group identification, and then the order is less than 512 you provide a small group knowl.

However, it is not true that groups of order 512 do not have gap id's, they do, however the implementation of IdentifyGroup in Magma does not support them.  So it's not strictly true to say that there is no gap id in these cases, there is, you just don't know which one it is.  It will be time consuming, but there is no reason we cannot match the 249 transitive groups of order 512 we have in the LMFDB against the list of 10494213 groups of order 512 in the small groups database.  I will take a look at putting together a parallel computation to do this matching -- it would be a useful thing to have in the LMFDB.

For the moment, for groups of order 512 I would change "no gap id" to "gap id not available".

jwj61:
Changed the message.  I warmed up to "Data not available".

AndrewVSutherland:
@jwj61 Thanks.  Shall I go ahead and merge this?

jwj61:
Sure.  Searching by gap id can come in a future pull request.

Before this goes to the main server, the transitivegroups.groups database would need to be copied to the cloud.

AndrewVSutherland:
The transtiviegroups.groups database has been updated on the cloud.

Issue #434, opened 2015-10-27
Compute mod 2 galois representation, link to associated field explicitly


jvoight:

AndrewVSutherland:
The 2-torsion field of the Jacobian of a genus 2 curve is typically going to have degree well beyond the range of the LMFDB (as high as 720).  It would be possible to instead identify the 2-torsion field as the Galois closure of an etale algebra of degree <= 6, which could be identified by a list of number field labels in some agreed ordering.  But this would have the unfortunate disadvantage that non-isomorphic etale algebras may have the same splitting field, and one would really like to know when two genus 2 Jacobians have isomorphic 2-torsion fields.

John Jones points out that for any etale Q-algebra of degree <= 6 we can always construct a number field of degree at most 9 with the same splitting field K (worst case is two S3-cubics with a common quadratic subfield).  We could then sort the subfields of K with Galois closure K and pick a minimal representative (e.g. apply polredabs to a defining polynomial and sort lexicographically).  Most (possibly all) of these degree <= 9 fields are already in the LMFDB, and John says it would not be difficult to add any that are not.

One possibly confusing aspect of this approach is that even when we have a generic genus 2 curve y^2=f(x) with f an irreducible sextic, it might happen that the 2-torsion field of its Jacobian would be identified as the splitting field of an irreducible sextic g that does not define the same field as f.  I could live with this, but in the long term I'd love to see Galois closures of etale Q-algebras as bonafide LMFDB objects in their own right (with their own labels) that we might store up to some reasonably low degree bound (e.g. 6) and discriminant bound (e.g. 10^6).

I'll note that we don't do this for elliptic curves -- presumably we would also want this here.

AndrewVSutherland:
The two torsion fields have all been computed, what remains to be done is to add the image of the representation, as a subgroup of GSp4(Z/2Z)

jvoight:
Move the mod 2 Galois representation from Mordell-Weil to the Galois representations section.

shiva-chid:
I am working on it

PR #2483, opened 2018-06-08
Two torsion field

AndrewVSutherland:
This PR partially addresses issue #434 by displaying the 2-torsion field of the Jacobian of a genus 2 curve on its home page.  Some examples to look at:

http://127.0.0.1:37777/Genus2Curve/Q/256/a/512/1
http://127.0.0.1:37777/Genus2Curve/Q/1369/a/50653/1
http://127.0.0.1:37777/Genus2Curve/Q/3969/a/3969/1

Scroll down to the "Invariants of the Jacobian" section to see the 2-torsion field.

The 2-torsion field is specified by giving a number field whose Galois closure is the 2-torsion field (in general the 2-torsion field of genus 2 curve can have degree as large as 720, but it can always be specified as the Galois closure of a field of degree at most 9).

When mod-2 Galois representations for GSp(4) become available in the LMFDB we will add a link to the home page of the mod-2 representation as well.

### DIFF
--- a/lmfdb/genus2_curves/templates/g2c_curve.html
+++ b/lmfdb/genus2_curves/templates/g2c_curve.html
@@ -281,7 +281,7 @@ function show_invs(invstyle) {
 <p>
 {{place_code('two_torsion_field')}}
 </p>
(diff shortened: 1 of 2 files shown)


JohnCremona:
I did not look on Travis to see why some checks failed: possibly random.
I was expecting number field knowls, given the name of the new object on these pages, but they are not knowls, just links to the field's home page.  Is that what was intended?

PR #2484, opened 2018-06-08
Two torsion field

AndrewVSutherland:
This PR partially addresses issue #434 by displaying the 2-torsion field of the Jacobian of a genus 2 curve on its home page. Some examples to look at:

http://127.0.0.1:37777/Genus2Curve/Q/256/a/512/1
http://127.0.0.1:37777/Genus2Curve/Q/1369/a/50653/1
http://127.0.0.1:37777/Genus2Curve/Q/3969/a/3969/1

Scroll down to the "Invariants of the Jacobian" section to see the 2-torsion field, which should display a number field knowl if you click on it.

The 2-torsion field is specified by giving a number field whose Galois closure is the 2-torsion field (in general the 2-torsion field of genus 2 curve can have degree as large as 720, but it can always be specified as the Galois closure of a field of degree at most 9).

When mod-2 Galois representations for GSp(4) become available in the LMFDB we will add a link to the home page of the mod-2 representation as well.

### DIFF
--- a/lmfdb/genus2_curves/templates/g2c_curve.html
+++ b/lmfdb/genus2_curves/templates/g2c_curve.html
@@ -270,15 +270,17 @@ function show_invs(invstyle) {
 <p>
 {{place_code('tamagawa')}}
 </p>
-<h3 class="inline"> {{ KNOWL('g2c.tamagawa', title='Tamagawa numbers') }}: </h3>
-{{data.tama}}
+<h3 class="inline"> {{ KNOWL('g2c.tamagawa', title='Tamagawa numbers') }}: </h3>{{data.tama}}
 
 <p>
 {{place_code('torsion_subgroup')}}
 </p>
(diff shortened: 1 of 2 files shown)


AndrewVSutherland:
I'm not sure what Travis's problem is, it looks like all the tests (and in particular, all the genus 2 tests) ran without a problem.  @edgarcosta can you take a look and see what went wrong?

edgarcosta:
On it

edgarcosta:
test_ratpts fails. I will try to fix it

AndrewVSutherland:
@edgarcosta I see the problem know, thanks.  I can fix this.

edgarcosta:
Yup, some curves have `curve['two_torsion_field'] = None.`

AndrewVSutherland:
The issue is that some of the two torsion fields fall just outside the range of number fields we have in the database (the odd parts of the discriminant are all under 10^6, but the 2-part can push the absolute discriminant past 10^7).  For the two torsion fields not in the database I will compute the Galois groups and a Galois group knowl and display a defining polynomial when a number field page is not available.

I will close this PR and resubmit once this has been done.

PR #2490, opened 2018-06-08
add Inventory to sidebar

JohnCremona:
This adds a link to the Inventory on the sidebar near the bottom just above Knowledge.  As with knowledge it only appears on beta.  This will make it easier for people to get to see (and if authenticated, edit) the inventory.

### DIFF
--- a/lmfdb/sidebar.yaml
+++ b/lmfdb/sidebar.yaml
@@ -281,4 +281,11 @@
        url_for: "knowledge.index"
        type: knowl
 
+9Inv:
+   type: single
+   heading:
+       title: Inventory
+       url_for: "inventory_app.show_edit_root"
+       type: knowl
+
 ...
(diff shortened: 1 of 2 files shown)


edgarcosta:
Looks good to me.

Should we also add a link to API on the sidebar?

JohnCremona:
@edgarcosta I think not as it's easy to type and is going to change anyway.  I don't mind users of beta looking at the inventory (maybe on prod at some point when it is tidier and more complete) but the existing api is rather rude (though certainly useful for us developers) so let's wait until we have something we can be more proud of.

edgarcosta:
Makes sense.

JohnCremona:
Any objection to me merging this after @edgarcosta 's verdict?  I think Edgar should be given merge rights anyway.

edgarcosta:
@JohnCremona, I have merge rights, but when it comes to significant changes, I feel less inclined to use it.

PR #2492, opened 2018-06-11
Magma snippet assumed that x was defined

edgarcosta:
The Magma snippet defining the number field in elliptic curves over number fields assumed that the variable x is defined.

Compare:
http://www.lmfdb.org/EllipticCurve/2.2.5.1/31.1/a/1
vs 
http://localhost:37777/EllipticCurve/2.2.5.1/31.1/a/1


JohnCremona:
I see that when you download the whole lot of Magma code (from the right-hand side Downloads section), I did remember to add the line defining x.  So now we have that line but it isn't used, which is a bit inconsistent.  If I had been fixing this I would have added a second line before the one defining the number field to match that one.

edgarcosta:
I was trying to follow the number fields snippets.

We can either do that, add one extra line in the code snippet or delete the extra line in the downloaded code.

In hindsight, which one do you prefer?

JohnCremona:
@edgarcosta Compromise: here's what genus 2 curve pages have:
`magma: R<x> := PolynomialRing(Rationals()); C := HyperellipticCurve(R![-30, 0, 37, -14, -5, 1], R![1, 0, 1, 1]);`

edgarcosta:
Perfect! See if you approve.

JohnCremona:
Sorry to be a pain but while it works, the full version ("Download all magma code") now looks like

`Qx<x> := PolynomialRing(RationalField());
// Define the base number field: 
R<x> := PolynomialRing(Rationals()); K<a> := NumberField(R![-2, 4, 4, -5, -1, 1]);`

in which the first line is redundant.  Also, the exact same thing you fixed is also there in the Sage code, where it does not crash since Sage defines a symbolic x on startup, so it would be best to make a similar change to the one you just made in the Sage snippet, following which lines751-4 of ecnf/main/py could be deleted.

If you prefer I can do this myself?

edgarcosta:
I delete the lines regarding Magma, but I think the line
```
x = polygen(QQ);
```
makes sense, as a priori `x` is just a symbolic expression.
This doesn't break sage, to use the symbolic expression, but I don't consider it a good standard.
For example,
```
sage: z = polygen(QQ);
sage: L.<a> = NumberField(z^2 - 2);
sage: z == L.defining_polynomial().parent().gen()
True
sage: K.<a> = NumberField(x^2 - 2);
sage: x == K.defining_polynomial().parent().gen()
x == x
```

JohnCremona:
I agree.  If you really want to you can insert "x = polygen(QQ); " at the beginning of the code snippet for Sage as well, and then remove it from the preamble to the full code file.

edgarcosta:
I will take care of this tomorrow.
I will add instead
```
 R.<x> = PolynomialRing(QQ);
```
Should I do something similar for number fields?

edgarcosta:
@JohnCremona I think this is ready to be merged

Issue #2475, opened 2018-05-31
Transitive group lookup by GAP id


AndrewVSutherland:
Now that we have GAP ids in the transitive group database, it would be nice to be able to search by GAP id.  This could be done by allowing GAP ids to be entered into the the transitive group label search field.

PR #2494, opened 2018-06-15
Allow searching by gap-id

jwj61:
This addresses issue #2475 allowing searching by gap-id in the transitive group database.  Comments:

- This got its own search field.  Results may be more than one transitive action, so it goes to the search results page, whereas using the search by label box should take you to a individual home page 
- I added another field for the database for this, gapidfull.  Trying to search using the order and id (index) in separately was painful, especially since a user can enter values into the field for order which would have to be merged with the order from the gap id.  The upshot is that we would need another data upload to the cloud before this goes to the production server
- I added to the Browse section of the page, listing some groups which have multiple transitive actions
- There is a function in the general search parsing for gap id's, but it has too many values hard-wired.  Maybe it should be moved to the one file from which it is called.

### DIFF
--- a/lmfdb/galois_groups/main.py
+++ b/lmfdb/galois_groups/main.py
@@ -8,7 +8,7 @@ from lmfdb import base
 from lmfdb.base import app
 from flask import render_template, request, url_for, redirect
 from lmfdb.utils import to_dict, list_to_latex_matrix, random_object_from_collection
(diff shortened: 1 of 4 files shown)


AndrewVSutherland:
Looks good to me.  Can I copy the groups collection to the cloud now?  I'd like to do that before committing.

jwj61:
Yes.  Thank you.

AndrewVSutherland:
@jwj61 The updated groups collection is now on the cloud.

Issue #2414, opened 2018-03-26
Better error handling when Travis fails


edgarcosta:
Improve the block
```
after_failure:
    # if some process failed, it will be clear in joblog which one failed
    - cat joblog
    # print all the stdout and stderr of nosetests run
    - find nosetests -type f -exec printf "\n\n filename = %s\n" {} \; -exec cat {} \;
```

perhaps first find the files with the string "FAIL:" or "... FAIL" and only display those, with some coloring.

JohnCremona:
I am puzzled by this.  When I watch the test log as they happen, what I see is a few thousand lines scrolling past with success / failure for each one, until it gets to the end.  At that point if the exit code is 1 I can scroll back slowly (or start at the top) to look for a failure message.  Yet the script you quote above suggests that the job log is only displayed if there is a failure.  I see everything every time even when the tests pass.

edgarcosta:
All that output comes from a single line
```
${SAGE} -sh -c 'parallel --joblog joblog --jobs 4 --results nosetests  --progress --verbose -a list ${SAGE_DIR}/local/bin/nosetests -v -s --testmatch=${TESTMATCH} ${COVERAGE}'
```

let me break this a bit.

Let's start by the end, for each line `X` in the file `list` it will run in parallel:
```
${SAGE_DIR}/local/bin/nosetests -v -s --testmatch=${TESTMATCH} ${COVERAGE} X
```

every line printed by the commands above is printed, but given that we are running everything in parallel these might get garbled. (the tool parallel tries to group the output and errors)

that issue leads us to `--results nosetests` this option saves the `stdout` and `stderr` for every `X`  in `nosetests/X/stdout` and `nosetests/X/stderr`.

for last the option` --joblog joblog`, creates a file named `joblog` where it reports on each command executed.
If one of the tests fails, we prints this file.
See for example: https://travis-ci.org/edgarcosta/lmfdb/builds/340317335
where on line 2560 reports that genus2_curve tests failed.

In the case a tests fails, we also print all the `nosetests/X/stdout` and `nosetests/X/stderr` in a "ungarbled" way with:
```
    # print all the stdout and stderr of nosetests run
    - find nosetests -type f -exec printf "\n\n filename = %s\n" {} \; -exec cat {} \;
```
See again: : https://travis-ci.org/edgarcosta/lmfdb/builds/340317335 (these start at line 2581)

Now back to the issue.
```
[...]

PR #2500, opened 2018-06-18
fixing double period

edgarcosta:
I fixed the double period in:

"Data computed by Andrew Booker, Jeroen Sijsling, Andrew Sutherland, John Voight, Raymond van Bommel, Dan Yasaki.."

pyflakes will fail because of #2496, I fixed that in #2492


AndrewVSutherland:
@edgarcosta I notice that Travis still seems to output a lot of information on tests that did not fail, which I thought you had changed (see #2414), or am I mistaken?

edgarcosta:
@AndrewVSutherland at the moment, we do not omit any of the tests, but we summarize what went wrong at the end:
screen shot 2018-06-18 at 16 17 36

PR #2503, opened 2018-06-20
Allow filtering knowls by category and quality

davidlowryduda:
This makes it possible to click on a category in /knowledge/, and then choose some subset of qualities and filter them out. This could be useful for those interested in reviewing various knowls.


JohnCremona:
This works for me -- thanks.
Starting from the top level with no filter, if you click on a category first you see all the knowls in that category, with no filters checked.  Then you can apply any combination of the quality filters and click Filter, doing this over and over without losing the  category filter. Good.
In the other order if you first apply a quality filter (e.g. check beta and click Filter) you see all the knowls of that quality; then click on a category after that and it keeps the quality filter on.  Good!

PR #2507, opened 2018-06-20
move inventory logs and silence a couple of unnecessary INFO messages

JohnCremona:
The inventory creates and writes to a couple of log files which show up in "git status" and the like.  Rather than just add the names of these to .gitignore I created a new directory called "logs", added that to .gitignore (so it exists even though we'll never commit any files in it) and changed the place where these files are created to include "logs/" at the front of the filename.  This seems to work OK.

Anyone finding files called LMFDBinventory.log ot LMFDBtransactions_inv.log can confidently remove them.

At the same time I removed a couple of  INFO messages which always appear at startup when a couple of blueprints start up as these not necessary (probably were put in for some debugging purpose).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -27,3 +27,4 @@ lmfdb/.coverage
 .coverage
 .DS_Store
 .idea/
+logs/
--- a/.gitignore
+++ b/.gitignore
@@ -27,4 +27,5 @@ lmfdb/.coverage
 .coverage
 .DS_Store
 .idea/
-logs/
+LMFDBinventory.log
+LMFDBtransactions_inv.log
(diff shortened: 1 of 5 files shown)


JohnCremona:
The first version did not work because git will not commit an empty directory, and the logs/ directory needs to exist but the files in it were all in .gitignore.  I think this will work: there's an empty .gitignore in logs/ and the top-level .gitignore does not ignore all the logs/ directory, just the two filenames which we are trying to ignore.  Now we wait for Travis...

JohnCremona:
The first version did not work because git will not commit an empty directory, and the logs/ directory needs to exist but the files in it were all in .gitignore.  I think this will work: there's an empty .gitignore in logs/ and the top-level .gitignore does not ignore all the logs/ directory, just the two filenames which we are trying to ignore.  Now we wait for Travis...

JohnCremona:
The Travis failure is due to a different change in lattice pages which I will fix.

PR #2508, opened 2018-06-20
Attribute completeness of class number 1 data

kedlaya:
Adding an attribution for the statement that the table of primitive lattices of class number one is complete. (I gather this is a subtable of the Catalogue of Lattices.)

### DIFF
--- a/lmfdb/lattice/lattice_stats.py
+++ b/lmfdb/lattice/lattice_stats.py
@@ -1,6 +1,7 @@
 # -*- coding: utf-8 -*-
 from lmfdb.base import app, getDBConnection
 from lmfdb.utils import make_logger, comma
+from lmfdb.knowledge.main import render_knowl_in_template
 
 logger = make_logger("lattice")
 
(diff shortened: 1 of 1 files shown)


JohnCremona:
Should we give a citation (I found https://arxiv.org/abs/1208.5638)? Otherwise this is fine.
@kedlaya your branch names are most unimaginative!

kedlaya:
If we give a citation, we should give both arXiv and DOI, and maybe also MR3091733. 

Re the branch names, these are defaults generated by Github when I generate pull requests there, without creating a local copy of the repository. This is handy for changes that are so trivial that I don't need to recompile to see if they work correctly.

JohnCremona:
I agree with giving as many possible citation links (here and everywhere).

edgarcosta:
I added the arxiv, MR and DOI references.

But honestly, half of that text is statistics of the data, and the other half is addressing some completeness of the data, and the later should be a knowl.

edgarcosta:
ps: I'm happy to merge once someone agrees with my changes

kedlaya:
I think you're right, the Kirschmer-Lorch attribution should go in the "Source of the data" knowl instead. I just added it there, so you can take it out of here.

kedlaya:
I guess that makes this PR redundant...

PR #2516, opened 2018-06-27
Switch from nosetest to pytest

### LINKED ISSUE #2511: switch from nosetest to pytest
nosetest (package `nose`)  is no longer maintained. the goal of this ticket is to switch to pytest (sane choice) and adjust the `test.sh` file.

davidlowryduda:
As Harald noted in #2511, nosetest is no longer maintained. This sets up pytest.

There are three aspects to this pull request. Firstly, there are a few changes to the inventory that silence print statements and prevent the line `sys.stdout = os.fdopen(sys.stdout.fileno(), 'w', 0)` from running. (It appears this was used for initial debugging, but it's no longer necessary). This had caused pytest to complain about IOErrors.

Secondly, there are changes to test.sh. This now uses pytest (and pytest-cov for coverage). This works very well on my machine locally.

Thirdly, I've attempted to setup Travis. I haven't tried to set up Travis before, so I'm somewhat certain that this will be a work in progress (most likely with @edgarcosta heavily involved).

But this resolves #2511.


haraldschilly:
I didn't run it, but just reading the patch it looks good to me. I also noticed that travis is using an old trusty ubuntu? It could make sense to push this to xenial. Is travis activated for this repository? I could try running it for this PR here manually.

haraldschilly:
ok, actually, travis ran: https://travis-ci.org/LMFDB/lmfdb/builds/397380341

I wasn't paying attention to the "checks-passed" here :-)

JohnCremona:
I tested this on atkin( at Warwick) where I installed the new python modules according to the revised requirements.txt.  It runs fine.  I like the compact output for all the passing tests.  I have not yet simulated a failing test to see how that looks.  More people testing on more machines would be a good idea.

JohnCremona:
OK, the error messages on a failing test look easy to understand.
I tried the parallel version with -n20 and that was quite impressive: lots more output as it started up 20 database connections.  It was still easy to see a single failure which I rigged for testing purposes.
I tested the coverage option and that works fine.  (We are at 59% coverage and at some point should try to increase that.)

edgarcosta:
This looks good to me!

I wouldn't worry about parallelizing it at the moment.

Our tests are much faster than when I set this up travis for the first time, and now we are very far away from the 50min time limit enforced by travis.
(and most likely the database is the true bottleneck)

Furthermore, I'm not sure how psycopg and pystests will interact, so I would prefer to keep it simple.
In the past, nosetests built in parallelization was not compatible with pymongo not being fork safe, and that is why I went through such convoluted way to parallelize it (and because the tests were taking much longer).

@haraldschilly, travis only offers trusty or precise.

Should I sign us up for coveralls?

haraldschilly:
@edgarcosta oh, I see

I saw, there is codecov-io active in other PRs. I'm not against coveralls, any such util is nice to have :-)

and yes, the database via a remote connection is likely a bottleneck. if it's still setup as I assume, @JohnCremona is able to run it almost locally and hence parallelization is something he will value.

davidlowryduda:
@edgarcosta Yes, I think signing up for coveralls sounds like a good plan.

edgarcosta:
On my to-do list for tomorrow.

davidlowryduda:
Something is happening that I don't quite understand. Having travis run pytest without further commandline arguments works fine, but passing in arguments (like `--verbose`, but more relevant to us are the coverage commands).

PR #2524, opened 2018-07-05
Tidying up local fields

jwj61:
Added knowls, particularly to the page for an individual field.

Display Base field and disc root field as knowls.  This led to "pretty display" more generally of Q_p and quadratic extensions, which is also used for intermediate fields.

The unramified/totally ramified tower is now presented as polynomials in x.  A good way to look at different variations are degree 8, p=11, then pick fields with different values of f.  If f=1, then the unramified subfield is Q_p, so the Eisenstein polynomial doesn't need to have "t" in it, so it is eliminated in that case.

Realized that we don't need to pass a database connection to a dynamic knowl, and removed it from local fields dynamic knowl.  I plan to do other cases when I revise their pages.


JohnCremona:
Looking good.

Issue #2527, opened 2018-07-09
Number fields should show number of roots of unity


JohnCremona:
Number field pages give the unit rank and a torsion generator but do not say what the order of that generator is.  That would be helpful.  The simplest way would be to insert "(order 6)" or similar after the torsion generator.
It is not clear to me whether or not the units of finite order are stored.  I think not. So the fact that their number is not stored wither is not an issue!

jwj61:
Addressed in pull request #2528.  

It is computed on the fly.  For the time being, it does not slow things down since the same gp command which computes the root of unity also returns its order.

Yes, it could be stored in the database.  For the moment, it does not seem urgent.  As a test, I ran all of the degree 18 fields in gp through this command, and the root of unity command takes basically no time (3-20 ms).  This comes after the number field has been initialized, and that is a little longer (around 40ms), but is always this fast since we supply the integral basis to nfinit.

PR #2528, opened 2018-07-09
Add order of root of unity on number field page

jwj61:
Addresses issue #2527 by displaying the order of the root of unity on individual pages of number fields.

### DIFF
--- a/lmfdb/number_fields/number_field.py
+++ b/lmfdb/number_fields/number_field.py
@@ -449,10 +449,13 @@ def render_field_webpage(args):
     info.update(data)
     if nf.degree() > 1:
         gpK = nf.gpK()
-        rootof1coeff = gpK.nfrootsof1()[2]
-        rootofunity = Ra(str(pari("lift(%s)" % gpK.nfbasistoalg(rootof1coeff))).replace('x','a'))
+        rootof1coeff = gpK.nfrootsof1()
+        rootofunityorder = int(rootof1coeff[1])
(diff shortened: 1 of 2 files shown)


JohnCremona:
Thanks!  Checking it out.

JohnCremona:
Something is causing all totally real fields (I think) to fail -- not just this branch but on beta (and even legendre.mit.edu).  www.lmfdb.org/NumberField/1.1.1.1  

Surely nothing to do with this PR but something is broken somewhere.  In Artin reps I think.

AndrewVSutherland:
It seems OK on www.lmfdb.org (in particular, the link to Q you posted works for me) but not on beta, and the link http://beta.lmfdb.org/NumberField/2.2.5.1 also throws a server error.

Needless to say, we should *NOT* merge to prod until this is fixed.

AndrewVSutherland:
Just to be clear, this has nothing to do with this PR.

JohnCremona:
So it must be a data issue with some collection not copied to the cloud.  Since lots of number field tests fail this must be recent -- in fact since Travis pass this PR!  Perhaps @jwj61 can help.

jwj61:
I suspect that the current problems are with Artin representations as you guessed.  I was adding a feature which meant storing a new field in the database, but forgot that artin rep'ns are more complicated.  Reverting the change is in progress.

jwj61:
It should be back now.

PR #2529, opened 2018-07-09
Galois tidying

jwj61:
Addresses concerns from the review of this code.  Mainly, there are more knowls, and some layout is adjusted.  Also, some of the dynamic knowl code is simplified.  Incidentally, the import script for Artin reps is adjusted.

### DIFF
--- a/lmfdb/galois_groups/main.py
+++ b/lmfdb/galois_groups/main.py
@@ -127,7 +127,7 @@ def galois_group_search(**args):
         parse_ints(info,query,'n','degree')
         parse_ints(info,query,'t')
         parse_ints(info,query,'order', qfield='orderkey', parse_singleton=make_order_key)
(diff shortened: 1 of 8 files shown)


JohnCremona:
@davidlowryduda : we are seeing many failures from Travis coming from the coverage test -- is this only because the coverage has gone down by epsilon?  It would be nice if we kept the coverage test but in such a way that it never resulted in failure.

JohnCremona:
I have checked that everything raised in the review have been addressed here, but perhaps @davidfarmer should do also.  The only issue I saw (in addition to those) was on the page of a single group in te list of invariants there is an entry |Aut(F/K)|, where the knowl tells me what the automorphism group of a field extensin is, but nowhere does it tell me what F and K are here.

Other than that, I think that this is good to go

jwj61:
Good point.  I wrote a new knowl for it.

JohnCremona:
OK!  If you could fix the typo in the knowl that would be good ("snd") otherwise I'll do it (and become a coauthor).

jwj61:
Typo fixed

PR #2531, opened 2018-07-10
Artin container

jwj61:
This is mostly minor tweaking of search form layout, bringing it in line with other parts of lmfdb (lower case for search box labels, aligned left on the browse page).

We now have completeness results in a knowl.  They are based on group, dimension, and smallest containing permutation representation.  The last item is a good way to distinguish characters with the same group and dimension (does not distinguish Galois conjugates or representations which differ by a group automorphism).  So, this also adds a search box for smallest containing permutation representation.

This meant a new entry to the database, but there has to be a corresponding change in the artin code -- it will crash if the database contains fewer or more entries than it expects in each document.  So, this switches the artin databases to temporary names.  It should work on beta, but those databases would need to be copied to the cloud before this code goes there (but the pull request needs to be accepted first).  Then we go around again changing the names back.

One final note, I will be traveling for around 2 weeks and not have a laptop.  So responses to questions or comments, or changes to the pull request won't happen until then.

### DIFF
--- a/lmfdb/artin_representations/databases/Dokchitser_databases.py
+++ b/lmfdb/artin_representations/databases/Dokchitser_databases.py
@@ -1,7 +1,7 @@
 artin_location = ("artin", "representations")
 galois_group_location = ("artin", "field_data")
-#artin_location = ("artin", "representations_new")
-#galois_group_location = ("artin", "field_data_new")
+artin_location = ("artin", "representations_new")
+galois_group_location = ("artin", "field_data_new")
 
 from type_generation import String, Array, Dict, Int, Anything
 
@@ -57,6 +57,7 @@ Galois_Conjugate = Dict(
 Dokchitser_ArtinRepresentation = Dict({
     "_id": Anything,
     "Baselabel": String,
+    "Container": Anything,
     "Dim": Int,
     "Indicator": Int,
     "Conductor": TooLargeInt,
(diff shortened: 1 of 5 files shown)


davidlowryduda:
I see these unnecessary coverage failures, and am fixing it. These will go away shortly.

edgarcosta:
I set some limits:

screen shot 2018-07-15 at 11 56 38

and triggered Travis to run again, I hope this clears the failures from coverage

AndrewVSutherland:
@jwj61 I understand that the new code in this PR depends on new collections (field_data_new and representations_new) that are present on beta but not on prod.  I am confused as to why you say this PR should be merged before the new collections are copied to prod, I think the safest sequence of operations is:

1) copy field_data_new and representations_new to the cloud (the old code will not use them)
2) merge this PR and push to beta and prod
3) copy field_data_new and representations_new to field_data and representations (which won't be read by the new code)
4) change the new code to read from field_data and representations and submit a second PR
5) merge the second PR and push to beta and prod
6) delete the collections field_data_new and representations_new collections on beta and the cloud

Given the need to coordinate the timing of these operations (and the fact that I want to get this all done this week before the postgres conversion) I can take care of the the code change in step 4), which I believe only requires deleting lines 3 and 4 in lmfdb/artin_representations/databases/Dokchitser_databases.py.

Does that all sound good to you @jwj61 ?

jwj61:
That sounds good.  (Maybe I mis-wrote what was needed as far as the order of operations.  The main thing is the code can't go to the cloud until the new collections are there, but you seem clear on that.)

PR #2532, opened 2018-07-14
Include title on main page

tahulse:
This is to clarify the name of the site beyond the title in the browser tab. This was my first commit and done under the guidance of David Lowry-Duda.

### DIFF
--- a/lmfdb/pages.py
+++ b/lmfdb/pages.py
@@ -80,7 +80,7 @@ def index():
 
     return render_template(tmpl,
         titletag="The L-functions and modular forms database",
-        title="",
+        title="LMFDB - The L-functions and Modular Forms Database",
         bread=bread,
         boxes = boxes)
 


davidlowryduda:
For some additional background information: while at the Building Bridges automorphic forms conference, there were some people calling the LMFDB the "London Modular Forms Database". And I thought "Hmm --- isn't the title readily seen from the main page?" But no! It's in the window/tab title, but this isn't visible on phones/tablets (or when people have too many tabs open).

davidlowryduda:
@tahulse It's exciting that you're diving in, but I should have explained what branches are. What happens right now is that when you add a commit, it automatically is added here. But this last commit you made (concerning Maass forms) wants to be a separate pull request.

So for the moment, hold off on adding more, and we'll chat (away from github) and separate these. Does that sound reasonable?

tahulse:
Yup!

JohnCremona:
More comments on this?  It effects only the top page and adds the title  
LMFDB - The L-functions and Modular Forms Database

JohnCremona:
It is certainly mergable but I wanted more opinions about whether we want the title and the exact wording.

sanni85:
I think it is needed... I have been asked what LMFDB stands for lots of times... 
By the way, check here http://wwwf.imperial.ac.uk/~dhelm/M4P58/ps1_sols.pdf  for an instance of the London Modular Forms Database.

PR #2535, opened 2018-07-15
Deleting alternate to truncate_number

tahulse:
There were two similar functions in utils.py, truncate_number() and truncatenumber(). The later file is not used by anything else and is redundant, and it was suggested it be removed by David Lowry-Duda. It has been removed.


davidlowryduda:
Excellent, thanks! Welcome to LMFDB-ing. I'm merging.

PR #2536, opened 2018-07-15
extra period in the credit string

edgarcosta:


edgarcosta:
Since this is just a text edit, I will merge it myself

PR #2539, opened 2018-07-19
fixed Belyi knowls

michaelmusty:
We cleaned up some of the knowls code.


edgarcosta:
They missed one mistyped knowl.

I will merge this, even though the the tests will fail because of https://hobbes.la.asu.edu/lmfdb-14/ being down.

PR #2540, opened 2018-07-19
pushing changes from master to beta

edgarcosta:
I will merge this once I confirm that only test that fails is https://hobbes.la.asu.edu/lmfdb-14/ being down.


edgarcosta:
as expected, that was the only failure.

PR #2542, opened 2018-07-24
Small cleanup Higher Genus Curves w/Automorphisms

jenpaulhus:
This pull request is for @AndrewVSutherland or @jvoight to look over, per our conversation at ANTS.

* I temporarily hid the "Shimura Variety dimension" data until I get a chance to better describe this piece of data.
* I changed the "Jacobian variety (partial) decomposition" to "Jacobian variety group algebra decomposition" and created a knowl to explain the group algebra decomposition.  I'll wait for your approval on that name before I write a draft of the knowl.  Particularly, at ANTS I mentioned the term "isotypical decomposition" instead of "group algebra decomposition".  The isotypical decomposition is a coarser decomposition than what I am computing here.  The isotypical decomposition comes from considering the central idempotents of Q[G] (G =Aut(X)) corresponding to each simple Q-algebra in the Wedderburn decomposition of Q[G],  whereas the group algebra decomposition additionally considers the primitive idempotents for each simple Q-algebra, thus factoring each term further. Happy to elaborate if you need me to.

You can see both changes here:
http://localhost:37777/HigherGenus/C/Aut/4.12-2.0.4-6-12.1

According to my records these are the only immediately pressing changes for the page.  Other changes we discussed will show up after the back end changes to lmfdb happen.


jenpaulhus:
Oops. this is not quite what I wanted to pull.  I will delete and retry.

PR #2547, opened 2018-07-26
Minor fix to genus 2 curve page formatting

AndrewVSutherland:
For @JohnCremona, remove <h3> subheading level from bottom level one-line displays so that they are not displayed all in green (now only knowl links are).

### DIFF
--- a/lmfdb/genus2_curves/templates/g2c_curve.html
+++ b/lmfdb/genus2_curves/templates/g2c_curve.html
@@ -27,6 +27,7 @@ function show_invs(invstyle) {
   }
 </script>
 
+
 {#
 <script src="http://aleph.sagemath.org/embedded_sagecell.js"></script>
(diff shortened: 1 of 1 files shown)


JohnCremona:
Looks good.  For a curve with no rational points the whole phrase "There are no rational points"  is the knowl link; possibly better if only the last 2 or 3 words were (excluding "There are")?

e.g. "http://localhost:37777/Genus2Curve/Q/336/a/172032/1

AndrewVSutherland:
@JohnCremona I changed the knowl link as you suggested (and added a period).

AndrewVSutherland:
@JohnCremona I think this is ready to merge, but let me know if there is anything else you think should be changed.

JohnCremona:
Sorry -- thought I had done this already

PR #2557, opened 2018-08-10
Switch to postgres

roed314:


roed314:
This is the main switch to Postgres, though there will of course be some follow up bugfixes and changes.  Issues specific to the switch are being tracked at https://github.com/roed314/lmfdb_postgres/issues, though some may be eventually ported to the normal LMFDB issue tracker.

PR #2569, opened 2018-08-16
Fix latex

SamSchiavone:
In version 8.3 of Sage the latex command changed slightly, which broke the display for our equations. (Adding \left and \right to parentheses.) I made a very small change to fix this. I think this bug affected all curves, but here's one example: 

http://beta.lmfdb.org/Belyi/7T6/%5B7%2C4%2C4%5D/7/421/421/g1/b

P.S. I'm not sure why there are so many merges before my commit, but I only edited a single file.


AndrewVSutherland:
Hi Sam, if you look at the version of belyi_web.py on which your PR is based, it is not current (for example, it includes the line "from pymongo import ASCENDING"), this is the reason for all the extra commits.  You need to resynch your master by pulling from upstream and then create a new branch in order to fix this.

But I will go ahead and create a new PR that makes the change you submitted here, so you don't need to do anything right now, this is just for future reference.

SamSchiavone:
Hi Drew, sorry for the late response and thank you for fixing this!

Issue #2565, opened 2018-08-16
Hide Belyi from taskbar 


jvoight:
Belyi appears in the task bar at http://www.lmfdb.org/, but it's not officially been moved from beta.  We need to hide it.

AndrewVSutherland:
This has been fixed.

edgarcosta:
Thank you all for catching and fixing my error!

PR #2573, opened 2018-08-18
Adding warning to beta.lmfdb.org connected to mongodb

edgarcosta:
# This PR is meant to be merged at some specific time (TBD)

This PR is meant to add a warning on beta.lmfdb.xyz (and temporarily on beta.lmfdb.org) informing users about the staleness of the mongo database.

This PR should be merged into ***beta*** before we copy one last time the knowl and user collections to the postgres db.

Furthermore, I would appreciate someone could check the language and test it. (I'm on an airplane and can't really do it)

P.S: I'm proposing merging into `beta`, as `prod` and `beta` are not in synchrony anymore (due to #2565)

### DIFF
--- a/lmfdb/knowledge/templates/knowl-edit.html
+++ b/lmfdb/knowledge/templates/knowl-edit.html
@@ -2,6 +2,13 @@
 {% import 'color.css' as color %}
 {% block content %}
 
+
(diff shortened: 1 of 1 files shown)


JohnCremona:
OK

edgarcosta:
I'm merging this now, as the switch will happen in a couple of minutes

PR #2584, opened 2018-08-20
updating Travis to use Sage 8.3

edgarcosta:


JohnCremona:
Assuming that the Travis test passes this it can obviously be merged right away.

PR #2585, opened 2018-08-21
fix latex

AndrewVSutherland:
This replaces PR #2569


AndrewVSutherland:
This fixes the problem at

    http://beta.lmfdb.org/Belyi/7T6/%5B7%2C4%2C4%5D/7/421/421/g1/b

where the curve is not displayed.  It should display correctly at

    http://127.0.0.1:37777/Belyi/7T6/%5B7%2C4%2C4%5D/7/421/421/g1/b

PR #2587, opened 2018-08-21
Postgres

roed314:
Fixing various small bugs (including one that prevents any data changes at the moment) and lowering the threshold for reporting slow queries.  I've run tests on legendre and they all pass.


edgarcosta:
Can we also merge this to dev and web?

PR #2592, opened 2018-08-22
Handling a generic refused connection for hobbes

edgarcosta:
One more exception to handle connection refused from hobbes.
@jwj61, could you test this? (under the conditions that make this test fail)


jwj61:
Yes, I will later today (don't want to take my laptop in the rain.

PR #2593, opened 2018-08-23
Local gsm

jwj61:
There are two special cases of Global splitting models (in the normal case, it is a polynomial).  In one, such as 
  http://www.lmfdb.org/LocalNumberField/163.8.7.2
we have not computed one yet.  In
  http://www.lmfdb.org/LocalNumberField/2.8.0.1
it is a theorem that it does not exist.  Going to postgres changed how these special cases were stored in the database.  This updates the code to recognize the two special values.

It also makes a minor capitalization adjustment for consistency, and adds tests for the cases above.

### DIFF
--- a/lmfdb/local_fields/main.py
+++ b/lmfdb/local_fields/main.py
@@ -222,9 +222,9 @@ def render_field_webpage(args):
         else:
             rffriend = "/LocalNumberField/%s" % rflabel
         gsm = data['gsm']
-        if gsm == '0':
+        if gsm == [0]:
             gsm = 'Not computed'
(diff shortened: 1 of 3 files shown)


AndrewVSutherland:
@jwj61 I'm confused by the fact that on http://beta.lmfdb.org/LocalNumberField/163.8.7.2 in this PR it says the field is not Galois over Q_163 but on http://www.lmfdb.org/LocalNumberField/163.8.7.2 is says it is Galois.  Which is correct?

I realize this might not be related to this PR, but I'd like to understand what the issue is.

PR #2604, opened 2018-08-28
Fixing a series of bugs related to creation and edition of users

edgarcosta:


edgarcosta:
I don't understand why I am seeing previous PR merges on this PR

edgarcosta:
To avoid adding extra stuff to this PR I'm merging it, so far these are only backend changes

PR #2628, opened 2018-09-12
fixing identation

edgarcosta:
@alexjbest, does this look right?


alexjbest:
No, I guess there is some tabs vs spaces going on here or something (which python will also complain about), but now the if is more indented than the surrounding.

edgarcosta:
I have retabbed the file (aka removed all the tabs)

Now should look better

edgarcosta:
Tks @alexjbest 
I will wait for the tests to pass.

PR #2633, opened 2018-09-29
Hyperlink DOIs to preferred resolver

katrinleinweber:
Hello :-)

The DOI foundation recommends [this new resolver](https://www.doi.org/doi_handbook/3_Resolution.html#3.8). Yes, a bit ironic that they would change the URL of their service, but it's now [encrypted](https://www.ssllabs.com/ssltest/analyze.html?d=doi.org). Because the old links with the `dx` subdomain continue to work, there is no urgent need to do anything.

However, I'd hereby like to suggest to follow the new recommendation and update all static DOI links and any code that generates new DOI links.

Cheers!

### DIFF
--- a/lmfdb/knowledge/main.py
+++ b/lmfdb/knowledge/main.py
@@ -154,7 +154,7 @@ def ref_to_link(txt):
         elif ref.startswith("doi"):
             ref = ref.replace(":","")  # could be doi:: or doi: or doi
             the_doi = ref[3:]    # remove the "doi"
-            this_link = '{{ LINK_EXT("' + the_doi + '","http://dx.doi.org/' + the_doi + '")| safe }}'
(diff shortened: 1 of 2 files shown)


AndrewVSutherland:
Hi Katrin, thanks for the PR!  This seems like a reasonable change (http://dx.doi.org is still fully supported, but there is no reason not to switch to the https resolver, although I note that many/most DOIs will still resolve to a non-https link)

I don't think the Travis errors have anything to do with your change, but we may want to sort these out before merging (no action is required on your part).

edgarcosta:
I reran the tests and everything seems fine now.

katrinleinweber:
Thank you :-)

katrinleinweber:
You're welcome and thanks for merging :-)

Issue #2579, opened 2018-08-18
After postgres, Maass form pic does not display


jvoight:
(2) In the latter, no picture is displayed.

http://www.lmfdb.org/ModularForm/GL2/Q/Maass/4f55570e88aece241f00000c
http://dev.lmfdb.xyz/ModularForm/GL2/Q/Maass/4f55570e88aece241f00000c

edgarcosta:
Fixed: http://cmfs.lmfdb.xyz/ModularForm/GL2/Q/Maass/4f55570e88aece241f00000c

edgarcosta:
I will wait for the merge

AndrewVSutherland:
Fixed by PR #2642

PR #2642, opened 2018-10-02
Displaying mwf plots

edgarcosta:
Addressing #2579


edgarcosta:
You can already see the fix at: http://cmfs.lmfdb.xyz/ModularForm/GL2/Q/Maass/4f55570e88aece241f00000c

PR #2654, opened 2018-10-18
master -> dev

AndrewVSutherland:
I think now is a good time to push master to dev (which is the code running on beta.lmfdb.org -> dev.lmfdb.xyz).  Other than the change to the logo size, these are mostly internal changes not visible to the user or small bug fixes cleaning up issues stemming from the postgres conversion.  

Given that it has been 6 weeks since we last pushed and there are a bunch of changes, @edgarcosta @roed314 can one of you please sanity check this PR before I merge it?


edgarcosta:
I'll take a look

edgarcosta:
Everything looks good! 

@AndrewVSutherland, whenever we are done with CMFs I will have a gunicorn instance running on the master branch, so we have an easy place to check that everything is working as expected

PR #2657, opened 2018-10-21
Putting in new rcs knowls for Artin representations

jwj61:
This puts in the rcs knowls discussed in August for Artin representations.  I don't think we have them for other areas, so now would be the time to sort out changes to the names of the learnmore links, titles of the pages, and order of the learnmore links, etc.

### DIFF
--- a/lmfdb/artin_representations/main.py
+++ b/lmfdb/artin_representations/main.py
@@ -212,7 +212,7 @@ def reliability():
 
 @artin_representations_page.route("/Completeness")
 def cande():
-    t = 'Completeness of Artin representation Data'
+    t = 'Completeness of Artin Representation Data'
     bread = get_bread([("Completeness", '')])
     learnmore = learnmore_list_remove('Completeness')
     return render_template("single.html", kid='rcs.cande.artin',
--- a/lmfdb/artin_representations/main.py
+++ b/lmfdb/artin_representations/main.py
@@ -23,6 +23,16 @@ def get_bread(breads=[]):
         bc.append(b)
     return bc
 
+def learnmore_list():
+    return [('Source of the data', url_for(".source")),
+            ('Rigor of the data', url_for(".rigor")),
+            ('Completeness and extent of the data', url_for(".cande")),
+            ('Artin representations labels', url_for(".labels_page"))]
+
+# Return the learnmore list with the matchstring entry removed
+def learnmore_list_remove(matchstring):
+    return filter(lambda t:t[0].find(matchstring) <0, learnmore_list())
+
 def make_cond_key(D):
     D1=ZZ(D)
     if D1<1: D1=ZZ(1)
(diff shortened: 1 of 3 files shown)


AndrewVSutherland:
General comments:

For the link titles in the learn more about box, I suggest following the template we are using for CMFs (the knowls have yet to be written but the links are there), see http://cmfs.lmfdb.xyz/ModularForm/GL2/Q/holomorphic/

Completeness of the data
Source of the data
Reliability of the data

These headings are shorter (which keeps the box narrower, freeing up some screen real estate), and I think the ordering is natural (what is there, where it came from, why I should trust it).

Specific comments:

I think the source knowl should say more about the programs used..  Were these computations done using Sage, Magma, Pari, something else?  Is the code available anywhere?  Did you use any of the Artin representation specific code built into one of these packages (e.g.the Magma ArtinRepresentations function?)  There seems to be some discussion of this in https://arxiv.org/pdf/1610.01228.pdf, that could perhaps be summarized here.

Regarding the rigor knowl (which I suggest labeling "reliability"), in addition to the question of mathematical rigor, it would be useful to know whether any sanity checks were done (e.g. for CMFs we have spent a lot of time computing modular forms using different methods to compare the results, and have found bugs in both magma and sage while doing so).  Are there any external results in the literature that can be used as sanity checks (e.g. paper X proves that there are exactly N objects of type Y and we found them all).  I realize there might not be a lot you can say here, I'm asking more the purpose of setting the precedent that we should be thinking about these issues (I would take it as an axiom that just as every mathematics paper almost surely contains at least one typo, even if all the theorems are true, every mathematical database almost surely contains at least one error, even if only rigorous methods were used to create it -- it is worth asking what steps were taken to catch errors).

jwj61:
A mundane question: the titles of these 3 pages for classical modular forms only capitalize the first word, but many lmfdb page titles capitalize almost everything.  What is our style in this regard?

AndrewVSutherland:
I think we should only capitalize the first word (as on http://www.lmfdb.org/NumberField/) for example.

AndrewVSutherland:
@davidfarmer Agreed.  Nearly all mathematics journals follow this convention (capitalize only proper nouns and the first word of titles), notably including http://annals.math.princeton.edu/2010/172-1/p02.  I think we should do the same.

jwj61:
The title of that number field page is "Global Number Fields" with every word capitalized.  Maybe you thought I meant so other sort of title.

AndrewVSutherland:
Ah, I was actually referring to the links in the "learn more about" box, bot now I understand you are referring to the headers on the pages themselves, e.g. http://www.lmfdb.org/NumberField/Discriminants, and I can see that the header on the place-holder page http://cmfs.lmfdb.xyz/ModularForm/GL2/Q/holomorphic/Completeness (which currently has no content) is capitalized differently.

I personally think it looks better with only the first word (and proper nouns) capitalized, but the header on that place holder page was not intended to trigger a style change.  This is probably something we should discuss separately.

davidfarmer:
My previous comment applies to all of those cases.

jwj61:
Capitalizing all non-trivial words is the most common, so I will do that.  If we change style on this generally, then these pages will change with pretty much everything else.

jwj61:
Updated the Artin pages, and put rcs links in the global number field pages as well.

PR #2660, opened 2018-10-22
Siegel to beta and GL3Maass to future

davidfarmer:
The following changes were decided in August.

Remove from public TOC:
   Modular forms --> GL(3)
and
  Modular forms --> Other

Both remain on beta, but
  Modular forms --> GL(3) --> Maass
has been moved to Future.

Note that the "GL(3)" and the "Other" in the beta TOC will not render properly
(they will not be rotated).  I will fix that as a separate pull request, because my understanding
is that TOC changes should be done individually.  I hope the small CSS change in this pull
request is an acceptable violation of that policy.

### DIFF
--- a/lmfdb/sidebar.yaml
+++ b/lmfdb/sidebar.yaml
@@ -92,18 +92,20 @@
            colspan: onecolumn
      - title: GL(3)
        style: rotate
+       status: beta
        part2:
          - title: Maass
-           url_for: "l_functions.l_function_maass_gln_browse_page"
-           url_args:
-             degree: 'degree3'
+           status: future
+           colspan: onecolumn
          - title: Cohomological
            status: future
            colspan: onecolumn
      - title: Other
        style: rotate
+       status: beta
        part2:
          - title: Siegel
+           status: beta
            url_for: "smf.index"
          - title: U(2,1)
            status: future
--- a/lmfdb/templates/base.html
+++ b/lmfdb/templates/base.html
@@ -116,7 +116,12 @@ _gaq.push(['_trackPageLoadTime']);
 {%- endif %}
 </head>
 
+{%- if BETA -%}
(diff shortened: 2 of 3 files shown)


davidfarmer:
I think this can be accepted before the TOC is fixed, because the non-rotated items
don't look too bad.  See issue #2661 .

jwj61:
Would it be hard to keep GL(2) rotated on the main site, but not rotate it on beta?  That would make it look better on beta since GL(3) and Other are not rotated just below it.

davidfarmer:
One thing that might make that easy, and possible be useful for
other things, is to put a "beta" class on the body of the beta HTML.

I think that makes sense in general, because then the beta pages
know that they are beta.

Adding some fairly restrictive CSS to not rotate the GL(2) on
the beta page would be no problem.

The CSS part of this suggestion is a hack, but not a terrible one and
it avoids needing to make major changes to the TOC.  The beta class on
body is probably reasonable, even if it is not currently being used.

I'll do this in a couple days, in case there is some problem I missed.
(I checked that body.beta has no effect on the appearance.)

AndrewVSutherland:
@davidfarmer Are you planning to add any commits to this PR?  I'd like to merge this soon.  I think we can live with having GL(2) rotated but GL(3) not on beta for a period of time if we need to, but if there is any easy fix that you want to include in this PR, please go ahead.

davidfarmer:
Yes, I'll have one more commit, probably tomorrow.

davidfarmer:
I added the 'beta' class to body on beta, and used that to add a very specific CSS selector
to not rotate GL(2) in sidebar on beta, as suggested by @jwj61 .

AndrewVSutherland:
Looks good, Travis timed out but didn't hit any problems.

PR #2665, opened 2018-11-12
fix for Bianchi newforms which have either CM or SFE fields not set

JohnCremona:
We now display '?' instead of crashing.  To test, try ModularForm/GL2/ImaginaryQuadratic/?field_label=2.0.4.1
Note that '?' is only displayed in the search results table, for brevity.  On the forms's own page it is better.

Don't review this yet since (1) I want to add some tests for this, and (2) I found another bug while writing this text.

### DIFF
--- a/lmfdb/bianchi_modular_forms/test_bmf.py
+++ b/lmfdb/bianchi_modular_forms/test_bmf.py
@@ -10,7 +10,12 @@ class BMFTest(LmfdbTest):
         assert text in self.tc.get(path, follow_redirects=True).data
 
     def check_args(self, path, text):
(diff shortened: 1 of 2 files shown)


AndrewVSutherland:
@JohnCremona So I ignored your instructions and reviewed this, it works and fixes a server error on beta, so it would be nice to merge.  Let me know if you are OK with merging or still want me to wait.

JohnCremona:
@AndrewVSutherland Thanks -- I found another place where the disc needed to be multiplied by 4, not yet committed, and was working on adding some tests.  So not yet....

JohnCremona:
OK now -- please merge if tests pass properly.

PR #2670, opened 2018-11-16
master -> web

AndrewVSutherland:
Pushing master to web to catch up with all the changes that are now on dev (this will include the change to remove GL(3) Maass forms from production).

This will not impact www.lmfdb.org which is still running on mongo db, but it will update the code running on www.lmfdb.xyz.

We will be switching www.lmfdb.org to the web branch in the near future, this is in preparation for that transition.


AndrewVSutherland:
There is no rush so I will let Travis run (until it times out), just to see how far it gets.

edgarcosta:
ps: this changes will be reflected in http://prod.lmfdb.xyz/

AndrewVSutherland:
@edgarcosta Right, thanks for the correction.  This time Travis managed to get in under the 50 minute mark.

Issue #2671, opened 2018-11-28
WebNumberField.is_null behavior changed


davidlowryduda:
While tracking down a bug with the value fields showing for Dirichlet character pages, I noticed that the behavior of `WebNumberField.is_null()` has apparently changed.

To see the change, I run the following code on lmfdb/master and lmfdb/prod (which happens to be running a few months old right now)

```python
import lmfdb.WebNumberField
lmfdb.WebNumberField.WebNumberField.from_cyclo(12).is_null()
```

On lmfdb/prod, this returns False.

On lmfdb/master, this returns True.

I suspect this has to do with a change for postgres, but I haven't tracked down the root of the change yet.

davidlowryduda:
This has been resolved in #2672.

PR #2672, opened 2018-11-28
Fix issue #2671 on the output of from_cyclo in WebNumberField

jwj61:
The problem involves data types between sage and python.  Maybe the two servers were running different versions of sage.  To test, look at http://127.0.0.1:37777/Character/Dirichlet/13/2 and see if "Value field" is listed as a related object (it should be).

### DIFF
--- a/lmfdb/WebNumberField.py
+++ b/lmfdb/WebNumberField.py
@@ -231,7 +231,7 @@ class WebNumberField:
         if euler_phi(n) > 23:
             return cls('none')  # Forced to fail
         pol = pari.polcyclo(n)
-        R = PolynomialRing(QQ, 'x')
+        R = PolynomialRing(ZZ, 'x')
         coeffs = R(pol.polredabs()).coefficients(sparse=False)
         return cls.from_coeffs(coeffs)
 
--- a/lmfdb/characters/test_characters.py
+++ b/lmfdb/characters/test_characters.py
@@ -108,7 +108,7 @@ class DirichletCharactersTest(LmfdbTest):
         W = self.tc.get('/Character/Dirichlet/1/1')
         assert  '/NumberField/1.1.1.1' in W.data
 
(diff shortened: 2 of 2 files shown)


davidlowryduda:
@jwj61 Thanks for getting that so quickly. I know that polredabs has been changing a little bit. I suppose the change in polynomial ring is one of those changes? I'm very happy you knew this fix --- I think this would have taken me a while. This commit solves my issue.

------------

There is one thing I still didn't understand, so I went on a bit of a trip. I report it here for those interested. 

I'm using Sage 8.4 (2018-10-17) for both of my tests in #2671 (that is, on the branches lmfdb/prod and lmfdb/master), so there isn't a sage version difference. But lmfdb/prod worked as expected, and lmfdb/master didn't work as expected --- yet this fix would seem to imply that lmfdb/prod should also need this change.

But I figured out what was going on. The coefficients for the test field are `[1, 0, -1, 0, 1]`.
- In mongo (and thus on prod), one would look up the coefficients as the string `"[1, 0, -1, 0, 1]"`. In `from_cyclo`, these coefficients came from a polredabs over QQ, but integers came out and were ultimately converted into the correct coefficient string.
- But in the postgres setup (and thus on master), one would look up the coefficients as the list `[1, 0, -1, 0, 1]`. The key here is that **the types of the numbers in the list matter** in our postgres setup, since the types are given encodings in db_encoding and db_backend. In the database, they are lists of integers (since of course number fields are given by lists of integers). In `from_cyclo`, these coefficients come from a polredabs over QQ, and are presented as rationals, not integers. This led to the lookup key being a list of rationals instead of integers, and thus the db was returning nothing (causing the `is_null()` call to think there was no such numberfield).

And this is why John Jones's change resolves my issue, but wasn't necessary for the lmfdb/prod branch to function correctly.

Issue #2561, opened 2018-08-11
Evaluate whether to switch to katex from mathjax


roed314:
From William Stein:
```
Katex covers about 99.9% of functionality now. 
(When we switched cocalc, one person pointed
out some spacing changes, and it turned out katex
was right and mathjax was wrong.)

Katex is usually about 25x - 100x faster to render
than mathjax, and it's done synchronously.
Switching would make LMDB feel massively more
responsive, completely getting rid of the
mangled-for-a-moment look you get with mathjax...
```
See https://github.com/Khan/KaTeX.

davidfarmer:
What is the expected speedup when MathJax 3 comes out?

AndrewVSutherland:
I believe this comparison is with MathJax 2.7.2, we should do a comparison with the MathJax v3 beta, which is available at

https://github.com/mathjax/mathjax-v3/tree/beta

We should do a comparison using some of our own page between katex and mathjax v3.

We should look at this soon.  I can't think of any other change that has the potential to dramatically improve the performance of every section of the LMFDB.  With the switch to postgres mathjax rendering is now the bottleneck on most pages, and in many cases it takes several seconds.  Our goal should be to have every LMFDB page load in under 400ms.

haraldschilly:
@AndrewVSutherland I think the only real question here is if all formula processing is done on the server (and not in the web browser). KaTeX should be able to do this quickly, and on top of that there are certainly possibilities for server side caching. Maybe an interesting for step could just be a little bit of hacking how the templating system works in order to render the formulas. If that looks ok, work out how to do this performantly.

alexjbest:
I guess this should be closed now as we switched to katex!

PR #2674, opened 2018-12-08
Change MathJax to KaTeX

alexjbest:
This is an experimental proof of concept, for people to look and see what KaTeX can and cannot do for us (see: #2561). If people would like to check how their favourite page looks and report back the issues that have arisen that would be great.

At least for me KaTeX loads orders of magnitude faster and looks _different_ (but not _worse_ to my eye). To me it looks a little sharper in places and the main difference I saw is that it doesn't bold as much (e.g. math in headings like GL_2 is far bolder in mathjax).

Edgar has set up <http://katex.lmfdb.xyz/> where you can try this out?

Caveats as of right now:
- [x] I didn't test the knowl editing stuff as I couldn't log in, this is complicated and surely broken.
- [x] Some code I couldn't even find if / where it is used (like some of the various ajax functions and datatables things in characters), therefore I could not check if it is broken. It would be really great if we could delete all currently unused or duplicated functions btw.
- [x] I checked many pages but not all of them of course, much more testing is needed
- [x] More advanced macros/strange tricks are probably broken, I added a macro for mathstrut as vphantom( already to help with this, mathstrut is not supported <https://katex.org/docs/support_table.html> but I think tex defines it as "\vphantom(" anyway, we can keep the macro, or change all mathstruts in source to something else.
- [x] KaTeX seems to play with css a little different (maybe some of its defaults are not the same as mathjax) which can cause some odd behavior.
- [x] some code comments still refer to mathjax
- [x] the acknowledgements still refers to mathjax
- [x] not all ams macros are supported, e.g. we use smallmatrix in the knowl /knowledge/show/mf.gl2.history see <https://katex.org/docs/support_table.html> in this instance a plain old matrix is not so much worse but realistically all knowls should be checked (maybe this can be automated though)!
- [x] same as previous for eqnarray in /knowledge/show/lfunction.history
- [x] the knowl test page is very broken /knowledge/show/test.text
- [x] in particular asciimath is broken
- [x] equation doesn't seem to work on /L/ModularForm/GL2/Q/Maass/4f55571b88aece241f000013/
- [x]  same for align on /L/EllipticCurve/Q/11/a/ it should be changed for aligned which is supported
- [x] generally mathjax had a number of plugins/options specified that katex equivalents of may not exist for
- [x] spacing is different for mathjax and KaTeX when wrapping an array in brackets (to make a matrix when pmatrix doesn't handle alignment) e.g. in the multiplicative galois module structure knowl on http://www.lmfdb.xyz/NumberField/6.0.214375.1 same for transitive groups.

Well you get the idea, it does look good though!


edgarcosta:
katex.lmfdb.xyz is running this branch

http://katex.lmfdb.xyz/info

pascalmolin:
Really convincing. Rendering is immediate and makes browsing easier.

I note that functional equations and Euler products on L functions pages do
not render
(the environments begin{equation} and begin{align} do not exist in katex)

This makes a lot of pages, but I think this is what you noted as last two
caveats.

Pascal

Le sam. 8 déc. 2018 à 03:44, Edgar Costa <notifications@github.com> a
écrit :

alexjbest:
I fixed a few things, so there are less glaring problems now, still like Drew says, the more eyes the better, there are so many parts of the lmfdb I'm simply unaware of.

As for the last issue @AndrewVSutherland raised, we might have more of a problem there as https://github.com/KaTeX/KaTeX/issues/572 seems to say essentially "don't do that" in the first place. So if we commit to KaTeX we will need to  work around this by having a written version (potentially with fancy characters) I think.

alexjbest:
As for knowls, the worst change I believe is smallmatrix -> matrix, I can find 22 instances of this (see bottom. I say worst as these are functionally different and I can find no way around it. If this is merged these will need changing, and possibly some inline math would need to be "display"ed.

Eqnarray is used in ec.j_invariant ec.discriminant ec.q.j_invariant lfunction.history.euler, changing this to aligned works well though.

Equation is used on  crystals.crystal test.text specialfunction.k_bessel mf.quasi_modular, of these crystal uses a strange split command and already appears broken on mathjax, and the others can be changed to display math no problem.

Asciimath and "tag/label/eqref" are all only used in the test knowl, which is fortunate as none of these features work with katex (OK tag does actually work to give equations visible names, but you have to reference it manually)

*In short*: from what I've seen, changing all knowls to work with KaTeX will require removing some features which aren't currently utilised, and making about 20 matrices look a bit bigger than they used to, which personally I think is a small price to pay.



knowls with a smallmatrix: group.fuchsian.cusps.width mf.gl2.history.poincare mf.bianchi.bianchimodularforms mf.bianchi.hyperbolic3space mf.siegel.character mf.siegel.cusp_form mf.siegel.vector_valued mf.half_integral_weight mf.elliptic.eisenstein mf.upper_half_plane mf.siegel mf.jacobi mf.slash_action mf.elliptic.hecke-operator mf.quasi_modular specialfunction.poincare mf.siegel.family.gamma_2 mf.siegel.family.gamma0_3 mf.siegel.family.gamma0_3_psi_3 mf.bianchi.level mf.bianchi.sign mf.elliptic.fouriercoefficients)

edgarcosta:
Here is another small issue, we don't want the qexp to wrap in search results, e.g.:
http://cmfskatex.lmfdb.xyz/ModularForm/GL2/Q/holomorphic/18/14/c/

edgarcosta:
@jwj61, 
See:
http://cmfskatex.lmfdb.xyz/ModularForm/GL2/Q/holomorphic/?weight=1-14&level=14
vs
http://cmfs.lmfdb.xyz/ModularForm/GL2/Q/holomorphic/?weight=1-14&level=14

if your browser window is narrow, each result will take multiple lines due to the wrapping of the q-exp.

alexjbest:
In any case this can be accomplished using the normal tex syntax so we could have 

```
        \({{ mf.qexp_display | replace("+","+~") | replace("-","-~") }}\)
```

alternatively the qexp_display string can be saved in the db like this.

One other possibility is that displayed equations don't break  in katex, (note that the above is an inline equation and so tex would break at the + and - too). So we can use `\[` and `\]` instead, this causes centering so we would need some other trick to get it to left align, unsure how to do this

davidfarmer:
I think the issue is that KaTeX makes it easier to wrap long equations.

But that just means a decision has to be made:  What do we
want those long equations to do?  Take up only one line? Wrap?
Something else?

The CSS can be adjusted so that in the KaTeX case those long formulas
behave as we choose.

I don't think it looks good when the browser is narrow and and each
row in the table takes up several lines.

Two lines, or even three, looks okay to me.  But I am not sure of
all the use cases and would be happy to hear why it is good to have
one line per entry.

I didn't like seeing the "Field" entry take up two lines, as
happens with Q[x]/(x^{12} + ...) .

alexjbest:
Ignore the above the easier/best way is to just use `<td nowrap>` I think!

<https://www.w3schools.com/tags/att_td_nowrap.asp>

AndrewVSutherland:
@alexjbest Perfect, this is exactly the sort of solution I was hoping for -- we will give this a try.

I agree with @davidfarmer that there are situations (e.g. columns in a table) where we do not want  wrapping, we would rather permit horizontal scrolling, but there are other cases (e.g. a long polynomial displayed in its own div on an objects home page) where wrapping is fine, even preferable, and should be allowed.  The key point is that we need to be able to make an intelligent decision about this and it sounds like we now have an easy way to implement it.

@davidfarmer: I think for q-expansions listed in search results it is better, not to wrap them.   We are displaying a somewhat arbitrarily chosen prefix of an infinite series to display (the number of terms we show in the table depends on how long the coefficients are).  The point of showing the q-expansion in the table is just to give users a quick visual impression of what the object looks like, not to display anything mathematically meaningful.  On the forms home page we display a fixed number of coefficients (10 or 100) which we do allow to wrap, and users can download 1000 coefficients if they wish.

JohnCremona:
I agree that in a search results list we should only display a small snapshot and not have wrapping.  In the case of q-expansions, it is possible that a user may recognise one with integer coefficients so it is good to see the first few terms, but for anything more complicated the only information a user will get from that display is simply the fact that the coefficients are not rational.

davidfarmer:
> @davidfarmer: I think for q-expansions listed in search results it is better, not to wrap them. We are displaying
> a somewhat arbitrarily chosen prefix of an infinite series to display (the number of terms we show in the table

It would be good if we could identify everything as falling into
a handful of cases and do it 'right', meaning that a class and CSS
determines wrapping, and whether overflow scrolls or is truncated, and
whatever else needs to be decided.

But it is not unreasonable to just code what is decided for each
case, and maintain an issue where we list the KaTeX display choices
that need to be refactored later.  (Specifically, I want whoever is making
those decisions to be conscious that a decision is being made, and to
check that decision with others.)

Something that particularly bothered me, and I am not sure if we caught
all cases, is when a long equation forces content near the top of the
page to drop below the properties box.

jwj61:
We definitely did not catch all cases where wide content causes thing to drop below the property box, but some of them are cured by the katex switch (e.g. a number field with a long defining polynomial).

alexjbest:
Ok I believe this is ready to merge, we've changed all knowls that we found, and checked a lot of pages out: to the extent that I'm convinced if there are any lingering tough-to-find issues, they should be surmountable.

AndrewVSutherland:
@alexjbest Looks good, I also think this is ready to merge, any objections?

alexjbest:
None from me!

PR #2675, opened 2018-12-12
removed latex from 2 drop-down menus for elliptic curve torsion struc…

JohnCremona:
…ture

It was unnecessary and rendered badly epecially with katex.


AndrewVSutherland:
Looks good.

PR #2676, opened 2018-12-12
Update editorial board

AndrewVSutherland:
Updates to editorial board


jwj61:
Associate editors has a chair?

AndrewVSutherland:
@jwj61 Good catch, fixed.

PR #2682, opened 2018-12-18
fixed incorrect display of 2-adic data for ec/Q

JohnCremona:
A glitch in the code for displaying elliptic curves over Q was introduced during the postgres conversion, as you can see at 
http://beta.lmfdb.org/EllipticCurve/Q/8190/q/4
and 
http://beta.lmfdb.org/EllipticCurve/Q/64025/f/1
(first two lines of the Galois Representation section in each case).

This fixes that.


JohnCremona:
@roed314 I suspect that when you changed the code you were intending to change the keys (column labels) in the new database so they did not start with "2", but never did.

Issue #2663, opened 2018-10-24
Incomplete documentation in db_backend.py


csbrady-warwick:
Currently the functions in db_backend.py have docstrings but only some of them include full interface documentation.  In particular several public (non underscored, such as delete) methods of classes do not include interface docs. Several other docstrings are not consistent with their actual implementation (e.g. upsert has optional Parameter commit that is not documented, as does _swap_in_tmp but reload_revert does document the same parameter, cleanup_from_reload documents a commit parameter that is not present in the interface).

Attributes of classes are not documented, even nominally public ones such as PostgresTable.search_table, which would also be helpful.

AndrewVSutherland:
Resolved by #3525.

Issue #2502, opened 2018-06-20
Elliptic curves over number fields: pages should show local root numbers


JohnCremona:
Compare the table of local data for elliptic curves over Q (e.g. http://beta.lmfdb.org/EllipticCurve/Q/369054/o/2) and over a number field (e.g. http://beta.lmfdb.org/EllipticCurve/2.0.8.1/25398.7/b/1):  the latter omits the local root number.

This is not hard to fix but not quite trivial since the curves database contains the root numbers for each curve (added for this purpose) while nfcurves does not.  For primes of additive reduction computing the root number goes beyond what we would want to do on the fly, so this means adding something to the database.

PR #2683, opened 2018-12-18
show local root numbers for ecnf if present without failing if not

JohnCremona:
In preparation for adding local root numbers to ec_nfcurves.  This code displays the root numbers when they are in the database, otherwise omits the  column in the Local Data table.

### DIFF
--- a/lmfdb/ecnf/WebEllipticCurve.py
+++ b/lmfdb/ecnf/WebEllipticCurve.py
@@ -470,7 +470,6 @@ class ECNF(object):
             ld['p'] = web_latex(P)
             ld['norm'] = P.norm()
             ld['kod'] = ld['kod'].replace('\\\\', '\\')
-            ld['kod'] = web_latex(ld['kod']).replace('$', '')
 
         # URLs of self and related objects:
         self.urls = {}
--- a/lmfdb/ecnf/WebEllipticCurve.py
+++ b/lmfdb/ecnf/WebEllipticCurve.py
@@ -469,7 +469,8 @@ class ECNF(object):
         for P,ld in zip(badprimes,local_data):
             ld['p'] = web_latex(P)
             ld['norm'] = P.norm()
-            ld['kod'] = ld['kod'].replace('\\\\', '\\')
+            while '\\\\' in ld['kod']:
+                ld['kod'] = ld['kod'].replace('\\\\', '\\')
 
         # URLs of self and related objects:
         self.urls = {}
--- a/lmfdb/ecnf/templates/ecnf-curve.html
+++ b/lmfdb/ecnf/templates/ecnf-curve.html
@@ -284,7 +284,7 @@ Generators:
 {{pr.p}}
 </td>
 <td align=center>
-{{pr.norm}}
+\({{pr.norm}}\)
 </td>
 <td align=center>
 \({{pr.cp}}\)
@@ -304,11 +304,11 @@ Generators:
 {% endif %}
 </td>
 {% if pr.rootno %}
-<td align=center>{{pr.rootno}}</td>
+<td align=center>\({{pr.rootno}}\)</td>
 {% endif %}
-<td align=center>{{pr.ord_cond}}</td>
-<td align=center>{{pr.ord_disc }}</td>
-<td align=center>{{pr.ord_den_j}}</td>
+<td align=center>\({{pr.ord_cond}}\)</td>
+<td align=center>\({{pr.ord_disc }}\)</td>
+<td align=center>\({{pr.ord_den_j}}\)</td>
 </tr>
 {% endfor %}
 </table>
--- a/lmfdb/ecnf/templates/ecnf-curve.html
+++ b/lmfdb/ecnf/templates/ecnf-curve.html
@@ -271,6 +271,9 @@ Generators:
 <th>{{KNOWL('ec.tamagawa_number', title='Tamagawa number')}}</th>
 <th>{{KNOWL('ec.kodaira_symbol', title='Kodaira symbol')}}</th>
 <th>{{KNOWL('ec.reduction_type', title='Reduction type')}}</th>
+{% if ec.local_data.0.rootno %}
+<th>{{KNOWL('ec.local_root_number', title='Root number')}}</th>
+{% endif %}
 <th>{{KNOWL('ec.conductor_valuation', title='ord(\(\mathfrak{N}\))')}}</th>
 <th>{{KNOWL('ec.discriminant_valuation', title='ord(\(\mathfrak{D}\))')}}</th>
 <th>{{KNOWL('ec.j_invariant_denominator_valuation', title='ord\((j)_{-}\)')}}</th>
@@ -300,6 +303,9 @@ Generators:
     Good
 {% endif %}
 </td>
+{% if pr.rootno %}
+<td align=center>{{pr.rootno}}</td>
+{% endif %}
 <td align=center>{{pr.ord_cond}}</td>
 <td align=center>{{pr.ord_disc }}</td>
 <td align=center>{{pr.ord_den_j}}</td>


JohnCremona:
This is one step towards fixing #2502

jwj61:
I would review it, but it would help to have a pair of pages, one which has data and one which shows the fallback if there isn't data.

JohnCremona:
Right now none of the curves has this extra data, so every page uses the fallback code.
OK, we can leave this until I have put the data in (I am still moving slowly towards being confident about changing data in the new database), at which time no conditionals will be needed anyway.

JohnCremona:
OK I have implemented code to add local root numbers and run it so far on just one field, one with only 59 curves.  Try
EllipticCurve/6.6.1995125.1/29.1/b/1
(or any other curve over that field) to see the root numbers.
http://localhost:37780/EllipticCurve/6.6.1995125.1/1.1/a/1
has everywhere good reduction so there's no table.

BUT that rewrite has resulted in the random curve link failing (even on beta).  @roed314 or @edgarcosta  please help!  The rewrite command I gave was
nfcurves.rewrite(add_root_number, {'field_label':'6.6.1995125.1'})
where nfcurves=db.ec_nfcurves
Is this because of the output pessage "resort disabled" which I also see when I run nfcurves.resort() separately?

alexjbest:
@JohnCremona This is more a flaw with the random selection than anything else (I'd say). It picks a number from 0- the maximum id seen and tries to see if that curve is in the db. Right now there are _only_ 59 curves in the db with ids all around 65500 so there is no chance random will find one of the present ids in 100 tries.

Without really knowing what I'm talking about; perhaps resetting the table completely so that ids start from 1 again is the way to go when reloading all data. But it doesn't sound like you wanted to reset all data in the first place, just some columns on some entries?

JohnCremona:
@alexjbest I only saw your post after writing mine just now.  I certainly did *not* want to delete all the curves except 59, so please can you revert the table to how it was,  post here exactly the command which you used for that, and then I will try again.

The stupid thing is that I selected a small subset of curves for this as a test in case it went wrong.  I suppose tat what I should have done instead is not to put a query in the rewrite, and if I want to change only some rows, let the individual entry modification function do nothing unless some condition is satisfied.

alexjbest:
@JohnCremona Sorry I don't know how to revert! We'll have to wait for Edgar or David unfortunately.

JohnCremona:
Well there are instructions at https://github.com/LMFDB/lmfdb/blob/master/Postgres_FAQ.md but not everything works as stated there (e.g. resort, upsert).  It should be as simple as db.ec_nfcurves.revert().

If necessary it should be possible to re-upload the data for this table, which will test how successfully I have been in updating the upload scripts.  I also write the entire table out to a file recently (and made no changes since them until this).

edgarcosta:
I have fixed.

For the record: 
- I dumped all the ec_nfcurves tables with
```pg_dump --clean --if-exists --schema=public -t 'ec_nfcurves_*' -v --file /scratch/importing/ec_nfcurves.dump --format tar lmfdb -U lmfdb```
- Checked that ec_nfcurves_old1 had all the desired rows:
```
lmfdb => select count(*) from ec_nfcurves_old1;
count
--------
 661073
```
- Then ran: 
```
sage: db.ec_nfcurves.reload_revert()
Swapped backup ec_nfcurves with ec_nfcurves_old1
```
- and everything worked as expected:
```
lmfdb=> select count(*) from ec_nfcurves;
 count
--------
 661073
```

If this wasn't the desired outcome, please let me know, as I can restore the dump.

Regarding some of the things mentioned above:
- the resort is disabled, and given the tiny small size of ec_nfcurves, this shouldn't make a difference at all. (sorry we haven't had the time to work on it)
- `nfcurves.rewrite(add_root_number, {'field_label':'6.6.1995125.1'})`
You effectively rewrote the table with the function to `add_root_number` applied to only the rows matching the search `{'field_label':'6.6.1995125.1'}` (the famous 59)
I agree that rewrite could have better documentation...

Anyhow, anything like this or wrong things in the Readme.md, please add to #2663, and we will get it to it. Unfortunately, CMFs is the priority at the moment.

JohnCremona:
@edgarcosta Much appreciated. No worried about resort; at one point I entirely wrongly thought that had caused the problem since after my change I first check on a curve which had the new data (and which displayed fine) and then tried a random curve (which did not as it had been deleted).

It seems that if one wants to rewrite a subset of the rows then one can either use upsert for those rows (if the number is very small) or use rewrite with a function which knows which rows to do nothing with, i.e. let the single-row-rewrite function make the decision on which rows to change and not the rewrite method itself.

I am going to close this PR and will reopen it when the data has been changed, which will not now be for several days.

JohnCremona:
The data is all there and I have slightly updated this PR.  Review please!

JohnCremona:
ping

AndrewVSutherland:
I can look at this today.

AndrewVSutherland:
@JohnCremona As I mentioned when we spoke, I get the following error when I try to run your branch

~/github/lmfdb$ sage --python start-lmfdb.py --debug
INFO:LMFDB@2019-03-05 22:11:46,384: Configuration = {'postgresql_options': {'user': 'editor', 'host': 'localhost', 'password': '282a29103a17fbad', 'port': 5432, 'dbname': 'lmfdb'}, 'logging_options': {'slowcutoff': 0.1, 'logfile': 'flasklog', 'editor': 'drew', 'slowlogfile': 'slow_queries.log'}, 'flask_options': {'debug': True, 'host': '127.0.0.1', 'port': 41001}} 
    import lmfdb
  File "/home/drew/github/lmfdb/lmfdb/__init__.py", line 5, in <module>
    from website import main
  File "/home/drew/github/lmfdb/lmfdb/website.py", line 52, in <module>
    setup_logging()
  File "/home/drew/github/lmfdb/lmfdb/website.py", line 41, in setup_logging
    formatter = logging.Formatter(utils.LmfdbFormatter.fmtString.split(r'[')[0])
AttributeError: 'module' object has no attribute 'LmfdbFormatter'

I suspect this may be due to recent changes (but I'm not sure why that would impact your branch)

roed314:
This branch still has a bunch of files that were deleted in the recent merge.  When I try to run it, I get an error because I don't have pymongo installed.

It looks like some kind of failure in git merge, since it says there's nothing to merge when I try to merge in master.  Since there are only a few changes, perhaps it's best to start fresh from master, and either use `git cherry-pick` or make changes by hand?

JohnCremona:
OK, I'll try to cherry-pick my actual small changes onto a new branch.  Leave this PR open for now.

JohnCremona:
OK, so I cherry-picked the two commits and force-pushed so the branch in this PR consists of just those two on top of master.

For testing, look at any random curve from EllipticCurve/ and look in the Local Data section.  Also look at a curve with everywhere good reduction and global minimal model, for which the table is empty (e.g. EllipticCurve/2.2.152.1/1.1/a/2) and one with e.g.r. with no global minimal model (e.g. EllipticCurve/2.0.643.1/1.0.1/a/1) showing that the table does display the prime of good reduction at which the model is non-minimal correctly.

I am double-checking now but it should be the case that root numbers exist for all bad primes of all curves so the conditional code in the template is not actually necessary.

JohnCremona:
I can confirm that the root numbers all exist in the db and are all equal to +1 or -1.

By the way, this code also fixes some spurious multiple backslashes in the Kodaira symbols.  Compare http://beta.lmfdb.org/EllipticCurve/2.2.21.1/2180.1/l/1  with the same after this fix.  I think that the code which stores strings in the postgres database does not handle these properly, since when I did the rewrite to put in the root numbers I am certain that the Kodaira symbol strings only had double backslashes, i.e. were of the form '\\(I_1\\)'.  I understand the need to escape some characters but it should be that if one puts a string into the database and later extracts it then the result is unchanged!  (In this case I also see that storing the \(,\) in the database at all is unnecessary, they could be added in the template.  Next time I do a rewrite I will change that.)

JohnCremona:
This is ready for review

AndrewVSutherland:
Looks good.

roed314:
> I think that the code which stores strings in the postgres database does not handle these properly, since when I did the rewrite to put in the root numbers I am certain that the Kodaira symbol strings only had double backslashes, i.e. were of the form '\(I_1\)'. I understand the need to escape some characters but it should be that if one puts a string into the database and later extracts it then the result is unchanged! (In this case I also see that storing the (,) in the database at all is unnecessary, they could be added in the template. Next time I do a rewrite I will change that.)

There have certainly been some bugs related to backslashes in text.  I have fixed some of them, but perhaps after the ecnf data was moved from postgres.

We should check the data after a COPY TO/COPY FROM, since that's the source of some of the errors (and, in particular, would cause the problems to reappear on prod when we uploaded the data there).

It's certainly nicer to avoid backslashes in the database if possible.

PR #2687, opened 2018-12-21
Debounce knowl clicking

alexjbest:
Currently if a knowl is double clicked quickly it gets messed up (will stop sliding in and out properly and not be visible at all), especially if the internet connection is bad as it needs to finish loading before the second click happens. One can test this using google chrome by setting your connection speed to 3g or something in the developer toolbar. This can happen even with fast internet completely accidentally.

This PR fixes this by "debouncing" the knowl, enforcing a .5s wait before a knowl is clicked again, otherwise nothing happens.

There may be a different way of fixing this by changing the knowl code, but the debounce function might be useful elsewhere so I did it this way.

<http://www.labbookpages.co.uk/electronics/debounce.html>

### DIFF
--- a/lmfdb/static/lmfdb.js
+++ b/lmfdb/static/lmfdb.js
@@ -165,8 +165,7 @@ function knowl_click_handler($el) {
  * @see jquery's doc about 'live'! the handler function does the 
  *  download/show/hide magic. also add a unique ID, 
  *  necessary when the same reference is used several times. */
-$(function() {
-  $("body").on("click", "*[knowl]", function(evt) {
+function knowl_handle(evt) {
       evt.preventDefault();
       var $knowl = $(this);
       if(!$knowl.attr("knowl-uid")) {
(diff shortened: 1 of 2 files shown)


JohnCremona:
Is someone reviewing this?  I don't think I am competent to.

edgarcosta:
FYI, the pull request for CMFs will include this change.
Nevertheless, will be nice to know that people agree with this change.

JohnCremona:
In that case I suggest that this PR be closed -- and similarly any others with the same status, after adding a comment to say why.
Is the same true for #2706? If so I will not bother testing it separately.

alexjbest:
@JohnCremona I don't mind closing this, I was just trying to avoid duplication of effort, it happened once or twice that non-CMF things I fixed and pushed to Davids CMF branch were re-done over here by someone else (you I think!) so I made an effort to make general changes/fixes as PRs here instead. This is probably not something someone else will re-do in the mean time though.

edgarcosta:
I'm okay with that.

> Is the same true for #2706? 
No

I will add a note to our PR text about this RP.

PR #2692, opened 2019-01-08
Add RCS knowls for Maass forms

davidfarmer:
I made the three RCS knowls for Maass forms (rcs.rigor.maass, rcs.cande.maass, rcs.source.maass),
and added a "learnmore" to the Maass forms pages.

The learnmore with RCS links should now appear on every Maass forms page:
just start navigating from the home page.


AndrewVSutherland:
The same issue that arose with https://github.com/LMFDB/lmfdb/pull/2683 is affecting this branch (it references files that were deleted or moved in the recent CMF PR).  I'm going to create a new PR to replace this one by simply copying the one changed file (which has not been changed by anything else).

PR #2695, opened 2019-01-17
Fix Multiple Cross-Site Scripting Security Vulnerabilities

tch1bo:
Hello,

I noticed several Cross-Site Scripting [(XSS)](https://www.owasp.org/index.php/Cross-site_Scripting_(XSS)) vulnerabilities in your project.

Here are the Proof of Concept exploits for some of them:
http://www.lmfdb.org/Variety/Abelian/Fq/%3Cimg%0Asrc=x%0Aonerror=alert('XSS')%3E
http://www.lmfdb.org/EllipticCurve/2.2.5.1/%3Cimg%20src=x%20onerror=alert('XSS')%3E/
http://www.lmfdb.org/EllipticCurve/?jump=1&label=%3Cimg%0Asrc=x%0Aonerror=alert(%27XSS%27)%3E
http://www.lmfdb.org/ModularForm/GL2/ModL/%3Cimg%20src=x%20onerror=alert('XSS')%3E

Each of these vulnerabilities is a serious security issue as it allows attackers to execute arbitrary code in users' browsers. So, please consider my fixes in this pull request.

I found the bugs while testing DeepCode’s AI Code Review. The tool can help you automate the process of finding such (and many other types of) bugs. You can sign-up your repo (free for Open Source) to receive notifications whenever new bugs are detected. You can give it a try [here](https://deepcode.ai/app/gh/review?utm_source=github_python&utm_campaign=lmfdb_LMFDB_XSS).

Any feedback is more than welcome at chibo@deepcode.ai.

Cheers, Victor.

### DIFF
--- a/lmfdb/abvar/fq/main.py
+++ b/lmfdb/abvar/fq/main.py
@@ -13,7 +13,7 @@ from search_parsing import parse_newton_polygon
 from isog_class import validate_label, AbvarFq_isoclass
 from stats import AbvarFqStats
 from flask import flash, render_template, url_for, request, redirect, send_file
-from markupsafe import Markup
+from markupsafe import Markup, escape
 from sage.rings.all import PolynomialRing, ZZ
 
 logger = make_logger("abvarfq")
@@ -289,7 +289,9 @@ def by_label(label):
     try:
         validate_label(label)
     except ValueError as err:
-        flash(Markup("Error: <span style='color:black'>%s</span> is not a valid label: %s." % (label, str(err))), "error")
+        l = escape(label)
+        e = escape(str(err))
+        flash(Markup("Error: <span style='color:black'>%s</span> is not a valid label: %s." % (l, e)), "error")
         return search_input_error()
     g, q, iso = split_label(label)
     return redirect(url_for(".abelian_varieties_by_gqi", g = g, q = q, iso = iso))
--- a/lmfdb/ecnf/main.py
+++ b/lmfdb/ecnf/main.py
@@ -22,7 +22,7 @@ from lmfdb.ecnf.isog_class import ECNF_isoclass
 from lmfdb.number_fields.number_field import field_pretty
 from lmfdb.WebNumberField import nf_display_knowl, WebNumberField
 
-from markupsafe import Markup
+from markupsafe import Markup, escape
 
 LIST_RE = re.compile(r'^(\d+|(\d+-(\d+)?))(,(\d+|(\d+-(\d+)?)))*$')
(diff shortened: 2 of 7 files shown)


jwj61:
Scrolling through the changes to the number field file, 95% of the changes are white space in the python file.  Frustrating when looking through diffs.

tch1bo:
Sorry for that. I wouldn't say it's 95% though.

It might help if i say that all useful fixes are adding a call to `markupsafe.escape()` to sanitize the user input before using it in `markupsafe.Markup()`.

jwj61:
In the number field file, I only see 2 lines with and substantive changes, and one of them is the import line.  But, maybe I missed some?

tch1bo:
Yep, you're right. Only lines 14 and 585 are part of the fix.

edgarcosta:
@jwj61, I will look into it more seriously. But in short this is the wrong fix, and we should just use escape everywhere instead of Markup.

edgarcosta:
Or use flask.Markup

tch1bo:
If you just use escape instead of Markup, then the HTML-markup will not work properly. 
E.g.:

`Error: <span style='color:black'>%s</span>...`

will be escaped to:

`Error: &lt;span style=&#39;color:black&#39;&gt;%s&lt;/span&gt;...`

and the error will not be displayed with desired style.

My fix escapes the string properly and then inserts it into the HTML-markup, which then gets rendered properly.

tch1bo:
`flask.Markup` in this case will do the same as `markupsafe.Markup`. It will **not** escape the "<>" characters and will **not** fix XSS.

edgarcosta:
Sorry, I meant, to do:
```
>>> Markup('<strong>Hello %s!</strong>') % '<blink>hacker</blink>'
Markup(u'<strong>Hello &lt;blink&gt;hacker&lt;/blink&gt;!</strong>')
```
Instead of:
```
>>> Markup('<strong>Hello %s!</strong>' % escape('<blink>hacker</blink>'))
Markup(u'<strong>Hello &lt;blink&gt;hacker&lt;/blink&gt;!</strong>')
```
either case is better than what we do now:
```
>>> Markup('<strong>Hello %s!</strong>' % '<blink>hacker</blink>')
Markup(u'<strong>Hello <blink>hacker</blink>!</strong>')
```

AndrewVSutherland:
I am closing this PR as there are now several merge conflicts and some of these problems have already been addressed by other PRs.  I will create a separate issue to address remaining xss vulnerabilities.

PR #2713, opened 2019-02-08
several knowl links changed from ec.q.* to ec.* (and redundant ec.q.*…

JohnCremona:
… knowls deleted)

Partly in response to issues raised on the bug tracker by Vlad Dokchitser.  Several knowl links changed frrom ec.q.* versions to ec.*.  At the same time I have done a lot of related knowl editing, and some creation.

A few ec.q.* knowls which became redundant as a result have been *deleted*.  I should not have done that until this PR was merged and pushed to dev, sorry.  (One example is that ec.q.rank no longer exists, as it is covered by ec.rank.  But without this PR there are references to ec.q.rank.)

### DIFF
--- a/lmfdb/ecnf/templates/ecnf-curve.html
+++ b/lmfdb/ecnf/templates/ecnf-curve.html
@@ -189,7 +189,7 @@ cellpadding="5";
 </div>
 
 <div>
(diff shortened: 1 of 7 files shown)


davidlowryduda:
This looks good to me. I recommend that it be merged in.

Issue #2714, opened 2019-02-20
Top/bottom knowls for abelian varieties


kedlaya:
Isogeny classes of abelian varieties over finite fields do not currently support top/bottom knowls. I would like to add some of these during the IAS workshop in March.

JohnCremona:
Lines 58-59 of https://github.com/LMFDB/lmfdb/blob/master/lmfdb/elliptic_curves/templates/ec-curve.html show how to add this for the top knowl, and lines 633-634 for the bottom.

davidlowryduda:
Am I correct to think that this can be closed now that #2715 has been merged in?

kedlaya:
Yes. There will be a separate thread at some point to implement top/bottom knowls for arbitrary objects, which might actually render this redundant, but so be it.

PR #2715, opened 2019-02-21
Add top/bottom knowls for abelian varieties

kedlaya:
To address issue #2714, here is a patch to add top/bottom knowls to abelian varieties over finite fields.

By the way, when looking at the code, I see that some of the knowls are "av.etc" and some are "ag.etc". Is that discrepancy meaningful?


JohnCremona:
I suppose av = Abelian Variety is a subset of ag = Algebraic Geometry

JohnCremona:
The patch will not work.  There is no "data" object passed to the template, it should be cl (for 'class'), and the label is I think cl.label.  Probably changing data['lmfdb_label'] to cl['label'] in two places will work.

JohnCremona:
I corrected a typo and expect that tests will now pass (since I did actually test this...)

JohnCremona:
That seemed to do the trick.

Issue #2631, opened 2018-09-21
Hecke operators, cosets in subset not subgroup


jwj61:
From the feedback page (acx01c@gmail.com):

$\Gamma_n(N)$ and $M_2(\mathbb{Z})$ aren't groups.

Also it is worth mentioning how $(n,N) \ne 1$ makes the cosets a little different.

http://www.lmfdb.org/knowledge/show/group.gammann

AndrewVSutherland:
This is fixed in #2717.  The knowl in question is no longer referenced.

Issue #2630, opened 2018-09-15
Request for documentation on modular form web pages


jwj61:
From the issue tracker:

An explanation of what the notation for the nebentype characters in the eigenform database would help.  What is \chi(n,_) exactly?

There was no e-mail to respond to

edgarcosta:
I believe this is currently fixed on the new version, see:
http://cmfs.lmfdb.xyz/ModularForm/GL2/Q/holomorphic/334/2/a/b/

We have a knowl for 'Character of a modular form' all throughout and links for the Dirichlet character.

edgarcosta:
When  this is in beta we can close it

AndrewVSutherland:
Addressed by #2717.

Issue #2340, opened 2017-12-14
no data  for level = 100, k 7, chi = 99


sanni85:
The only space available, which is the last one, should have dimension 52:

http://www.lmfdb.org/ModularForm/GL2/Q/holomorphic/100/7/?group=1
 
but clicking on the link, it says that it is empty...

edgarcosta:
@JohnCremona, should we close this?
This now sits outside our goal range of Nk^2 <= 4000

AndrewVSutherland:
We should not close any outstanding CMF issues until the updated version is merged.

AndrewVSutherland:
Addressed by #2717, see http://www.lmfdb.org/ModularForm/GL2/Q/holomorphic/100/7/d/.

Issue #2196, opened 2017-07-20
Add Bober's modular form data


jwbober:
I've made some first steps at https://github.com/jwbober/lmfdb/tree/modforms

edgarcosta:
@jwbober, your data is being added. Shall we close this?

edgarcosta:
please reopen if you disagree

AndrewVSutherland:
When the data is actually available (at least on beta.lmfdb.org) we can close this.

AndrewVSutherland:
This has been done and the data is now available on www.lmfdb.org for all newforms satisfying Nk^2 <= 4000.

PR #2717, opened 2019-02-28
Classical modular forms

roed314:
A live version of the code in this PR is available at http://cmfs.lmfdb.xyz/ModularForm/GL2/Q/holomorphic/

# Extent and size

### Boxes

Rather than choosing *N* and *k* in boxes, we organize by *Nk²*, which is asymptotically proportional to the [analytic conductor](http://cmfs.lmfdb.xyz/knowledge/show/mf.elliptic.analytic_conductor) of the newform (a good measure of its complexity).  Our tabulation is [complete](http://cmfs.lmfdb.xyz/ModularForm/GL2/Q/holomorphic/Completeness) within the following ranges

* *Nk²* ≤ 4000
* *Nk²* ≤ 40000, *χ*=1
* *Nk²* ≤ 40000, *N* ≤ 24
* *Nk²* ≤ 14400, *N* ≤ 100, *k* ≤ 12
* *Nk²* ≤ 100000, *N* ≤ 10
* *Nk²* ≤ 40000, *k* > 1, dimension ≤ 50
These boxes were chosen to include everything currently in the LMFDB and other existing tables of modular forms.

### Counts
* 338,551 nonzero newspaces *S<sub>k</sub>(N, χ)* with trace data (52,531 nonzero)
* 281,219 Galois orbits of newforms (48,688 with rational coefficients, 19,306 of weight 1)
* 14,398,359 complex embedded newforms
* 14,396,948 degree 2 L-functions
* 195,163 L-functions for Galois orbits

### Database size
* 325 GiB of embedded modular forms (*q*-expansions in ℂ)
* 123 GiB of trace data (traces of *q*-expansions down to ℚ)
* 45 GiB of portraits (complex graphs of *q*-expansions)
* 2 GiB of exact *q*-expansion data (when dimension is at most 20)

### Computation time
We spent 50-100 CPU years this fall and winter computing the above data.

### Representation
Exact coefficients are stored using either a [sparse cyclotomic representation](http://cmfs.lmfdb.xyz/ModularForm/GL2/Q/holomorphic/1620/1/bp/a/) or in terms of an [LLL-reduced basis for the coefficient ring](http://cmfs.lmfdb.xyz/ModularForm/GL2/Q/holomorphic/6877/2/a/s/).  We have exact coefficients in weight 1 or if the dimension is at most 20.

We compute complex *q*-expansions for all embeddings into the complex numbers, even when the dimension is [large](http://cmfs.lmfdb.xyz/ModularForm/GL2/Q/holomorphic/983/2/c/a/).

### [Reliability and Rigor](http://cmfs.lmfdb.xyz/ModularForm/GL2/Q/holomorphic/Reliability)
Many computations were duplicated in Magma and Pari, providing a check for their correctness.  We also compared with complex analytic data independently computed by Jon Bober, with William Stein's modular form tables and Alan Lauder and Kevin Buzzard's weight 1 tables.  Finally, we have written a verification module for the LMFDB and run various tests on the data.

We have rigorously verified self twists and projective image types in all cases, and inner twists and analytic ranks in most cases.  Any values that are not rigorously computed are explicitly identified.

# User interface
We have rewritten the front end through which a visitor to the LMFDB interacts with classical modular forms, along with many other changes to the LMFDB display code (220 changed files with 14,739 additions and 14,479 deletions).

### Searches
* You can obtain a list of [newforms](http://cmfs.lmfdb.xyz/ModularForm/GL2/Q/holomorphic/?search_type=List) or of [newspaces](http://cmfs.lmfdb.xyz/ModularForm/GL2/Q/holomorphic/?search_type=Spaces) satisfying various constraints (e.g. [rational forms](http://cmfs.lmfdb.xyz/ModularForm/GL2/Q/holomorphic/?dim=1&search_type=List)).
* You can view dimensions of [newforms](http://cmfs.lmfdb.xyz/ModularForm/GL2/Q/holomorphic/?search_type=Dimensions) or of [newspaces](http://cmfs.lmfdb.xyz/ModularForm/GL2/Q/holomorphic/?weight=2-&dim=1-&search_type=SpaceTraces) in a table organized by *N* and *k*.  You can add constraints and change the ranges of *N* and *k* displayed.
* You can browse [newforms](http://cmfs.lmfdb.xyz/ModularForm/GL2/Q/holomorphic/?weight=2-&search_type=Traces) and [newspaces](http://cmfs.lmfdb.xyz/ModularForm/GL2/Q/holomorphic/?weight=2-&dim=1-&search_type=SpaceTraces) using a searchable table of traces for their *q*-expansions, that can be used both integer and mod-p modes.  An application: if you know a curve is modular, you can pin down an associated newform by looking at the trace table at small primes.  Counting points mod p is easy, but computing Euler factors is hard in high genus.
* You can get a random form satisfying specificed constraints (e.g. [CM with level a power of 2](http://cmfs.lmfdb.xyz/ModularForm/GL2/Q/holomorphic/?level_primes=2&cm=yes&search_type=Random)).
* [Search results](http://cmfs.lmfdb.xyz/ModularForm/GL2/Q/holomorphic/?dim=1-&num_forms=10-&char_order=1&search_type=Spaces) for spaces include decomposition into newforms and into Atkin-Lehner eigenspaces.

### Browse
The [front page](http://cmfs.lmfdb.xyz/ModularForm/GL2/Q/holomorphic/) includes a curated lists of interesting newforms and newspaces with hovertext describing why these forms and spaces were chosen

### Statistics
We've added [statistics](http://cmfs.lmfdb.xyz/ModularForm/GL2/Q/holomorphic/stats), as well as a new dynamic statistics feature allowing you to view custom statistics (e.g. statistics on [weight and level for forms with CM](http://cmfs.lmfdb.xyz/ModularForm/GL2/Q/holomorphic/dynamic_stats?cm=yes&col1=level&buckets1=1%2C2-10%2C11-100%2C101-1000%2C1001-2000%2C2001-4000%2C4001-6000%2C6001-8000%2C8001-10000&totals1=yes&proportions=recurse&col2=weight&buckets2=1%2C2%2C3%2C4%2C5-8%2C9-16%2C17-32%2C33-64%2C65-316&totals2=yes&search_type=DynStats))

### Downloads
Downloads are now more broadly available, and you can download various things from homepages (trace forms, *q*-expansions, all stored data) and from search results.

We give linear operators on *S<sub>k</sub>*(*N*, *χ*), expressed as polynomial expressions of Hecke operators, so that the newform is the intersection of their kernels.  This can sometimes speed up construction of the newform in Magma.

### Form homepages
* *q*-expansions are more readable due to changes in representation.  Defaults to showing 10 coefficients, expandable to 100 coefficients on the page, 1000+ coefficients available for download.
* trace expansions are also shown, even when the dimension is [large](http://cmfs.lmfdb.xyz/ModularForm/GL2/Q/holomorphic/983/2/c/a/)
* include projective image type for weight 1 forms, including [A₄](http://cmfs.lmfdb.xyz/ModularForm/GL2/Q/holomorphic/124/1/i/a/), [S₄](http://cmfs.lmfdb.xyz/ModularForm/GL2/Q/holomorphic/148/1/f/a/), [A₅](http://cmfs.lmfdb.xyz/ModularForm/GL2/Q/holomorphic/1763/1/p/b/), [D₂](http://cmfs.lmfdb.xyz/ModularForm/GL2/Q/holomorphic/3600/1/e/a/) and [large dihedral](http://cmfs.lmfdb.xyz/ModularForm/GL2/Q/holomorphic/3997/1/cz/a/).  You can also search for forms with a specified type.
* Matched [weight 1 forms](http://cmfs.lmfdb.xyz/ModularForm/GL2/Q/holomorphic/385/1/q/b/) with [Artin representations](http://cmfs.lmfdb.xyz/ArtinRepresentation/2.5_7_11.6t5.1c1) when present in the LMFDB
* include Sato-Tate groups for all newforms of weight larger than 1
* include provable analytic ranks (e.g. [rational form of weight 42](http://cmfs.lmfdb.xyz/ModularForm/GL2/Q/holomorphic/2/42/a/a/) and [form of weight 14 with analytic rank 1](http://cmfs.lmfdb.xyz/ModularForm/GL2/Q/holomorphic/8/14/b/a/))
* include CM/RM fields for newforms with [self twists](http://cmfs.lmfdb.xyz/ModularForm/GL2/Q/holomorphic/39/1/d/a/)
* include [inner twists](http://cmfs.lmfdb.xyz/ModularForm/GL2/Q/holomorphic/1089/1/r/a/) for twists by characters that yield other forms in the same Galois orbit
* for forms with exact *q*-expansions, we provide the [values of the character](http://cmfs.lmfdb.xyz/ModularForm/GL2/Q/holomorphic/630/2/bv/c/) in the same representation
* We have portraits for both newforms and newspaces, giving plots of the trace form on the Poincare disc.
* change in label scheme but we support [old labels](http://cmfs.lmfdb.xyz/ModularForm/GL2/Q/holomorphic/10/8/1/a)

### Embedded form homepages
* There are now [homepages](http://cmfs.lmfdb.xyz/ModularForm/GL2/Q/holomorphic/983/2/c/a/472/35/) for newforms equipped with an embedding of the coefficient field into the complex numbers.  They are linked to from the tables of embeddings on each newform page.
* q-expansions are shown with complex coefficients, we include links to the dual form (complex conjugate).
* Coefficients (in both arithmetic and analytic normalizations) and Satake angles are viewable up through n=1000, with more sometimes available for download in high level.

### Space homepages
* We give decompositions of[ *S<sub>k</sub>*<sup>new</sup>(*N*, *χ*) into newforms](http://cmfs.lmfdb.xyz/ModularForm/GL2/Q/holomorphic/9075/2/a/), of [*S<sub>k</sub>*<sup>new</sup>(Γ₁(*N*)) into newspaces](http://cmfs.lmfdb.xyz/ModularForm/GL2/Q/holomorphic/9075/2/) and *S<sub>k</sub>*<sup>old</sup>(*N*, *χ*) and *S<sub>k</sub>*<sup>old</sup>(Γ₁(*N*)) into lower level pieces.
* Give dimensions for new and old, Eisenstein and cuspidal subspaces
* Show the trace form
* Give bounds on how many a_n are needed to distinguish newforms within a space.  In addition to the standard Sturm bound, we also provide a computed trace bound based on the forms that actually occur.

### L-functions
* For each complex embedding of a newform with weight at most 200, we have added an L-function.
* both arithmetic and analytic normalizations are available
* We have computed analytic ranks for all of the new L-functions, and proved most of them.
* We compute [Euler factors](http://cmfs.lmfdb.xyz/L/ModularForm/GL2/Q/holomorphic/5/27/c/a/) at all primes up to 100
* zeros on critical line have more precision and are stored in the database rather than computed on the fly, making it easier to find connections to other objects in the LMFDB
* change in label scheme but we support old labels

# Experimental features

We have used classical modular forms as a testing ground for some new features which are applicable to other areas of the LMFDB.

* There is now a verification module to add consistency checks for data in Postgres; you can access it by `from lmfdb.verify import db` and then running either `db.verify` or `db.tablename.verify`.  See the Postgres_FAQ for more details.
* See the description of [dynamic statistics](http://cmfs.lmfdb.xyz/ModularForm/GL2/Q/holomorphic/dynamic_stats?cm=yes&col1=level&buckets1=1%2C2-10%2C11-100%2C101-1000%2C1001-2000%2C2001-4000%2C4001-6000%2C6001-8000%2C8001-10000&totals1=yes&proportions=recurse&col2=weight&buckets2=1%2C2%2C3%2C4%2C5-8%2C9-16%2C17-32%2C33-64%2C65-316&totals2=yes&search_type=DynStats) above.
* A variant on knowls used when displaying very long mathematical content.  You can see the results in the *q*-expansions, basis descriptions, coefficient fields and Hecke kernels for some [high weight newforms](http://cmfs.lmfdb.xyz/ModularForm/GL2/Q/holomorphic/2/218/a/b/) and in various other places on classical modular form pages.
* Some nice features for [2-d tables](http://modularform/GL2/Q/holomorphic/?level=1-150&search_type=Dimensions): sticky headers and rotated labels.  They're currently implemented in a somewhat ad-hoc manner, but can be abstracted if more widely used.
* You can change the sort order on search results
* The ability to get a random object satisfying search constraints
* Multiple search result views (e.g. trace tables, dimension tables)
* There is more modularity in the templates, allowing the same search interface to be used on multiple pages

# Mongo to Postgres
While working on classical modular forms, we've fixed various issues with the transition from Mongo to Postgres, and eliminated the last dependencies on Mongo.  One interface change to be aware of is that you now run `from lmfdb import db` to access the database object, rather than `from lmfdb.db_backend import db`.

One change particularly useful for those of you not at MIT is that we tried to reduce the number of queries to the server when possible.  In particular, at startup and when fetching knowl titles we switch to a single database query and cache the results for 10 seconds.

# Issues resolved

This pull request resolves the following issues.

* Weight 1 cusp forms #437
* L-functions for modular forms should be in database #439
* Data quality for classical modular forms #460
* Dual L-function #617
* Make user adjust Fourier coefficient embeddings #653
* Make pretty coefficients #654
* Create separate page for each embedding of a newform #659
* Table options for old/new/CM #674
* Determine planned extent of classical modular forms #873
* Source and extent of data for classical modular forms #1224
* For imprimitive genus 2 L-functions, give links to factors #1227
* Fix errors in classical modular form data and verify its correctness #1248
* Tests for lcalc files #1257
* Modular form history #1282 (we decided not to link it)
* Search by analytic rank #1362
* Plots of modular forms #1408
* Random modular form #1434
* Product L-functions for modular forms #1443
* Sever error on classical modular form pages #1533
* Resource intensive L-function zeros pages for classical modular forms #1680
* Classical modular forms Gamma1(49) weight 3 #1991
* Add Bober's modular form data #2196
* no data for level = 100, k 7, chi = 99 #2340
* Request for documentation on modular form web pages #2630
* Hecke operators, cosets in subset not subgroup #2631
* Debounce knowl clicking #2687 (this is a PR, that we closed)

EDIT: we were missing the first 3 lines, which I added. Edgar


davidfarmer:
I just loved reading the pull request message!  I had to drop what I was doing and pull that
branch into my local machine.  This is really impressive work.

One thing I noticed is a difference between the posted version and my local copy.
If I browse weight 1 forms, say:
   /ModularForm/GL2/Q/holomorphic/?weight=1
then in my local copy, when I clock on a knowl in the table header (say, Field), then
in the local version the knowl appears in the table header, so it is messed up.

And if I go to another page with a search results table, like,
   /Genus2Curve/Q/?cond=1-499
then when I click on "Equation" the knowl opens in the correct place,
but it is very wide (not constrained by the browser width).

Those both work fine in the posted version.

Okay, I just realized that I had not pulled upstream master before pulling in 
roed314-master.  So I checkout out master, pulled upstream master, and then
checked out roed314-master .

Now when I do  sage -python start-lmfdb.py --debug  it fails with the message:

INFO:LMFDB@2019-02-28 17:56:53,604: Configuration = {'postgresql_options': {'user': 'lmfdb', 'host': 'localhost', 'password': 'lmfdb', 'port': 5432, 'dbname': 'lmfdb'}, 'logging_options': {'slowcutoff': 0.1, 'logfile': 'flasklog', 'slowlogfile': 'slow_queries.log'}, 'flask_options': {'debug': True, 'host': '127.0.0.1', 'port': 41018}} 
    import lmfdb
  File "/scratch/home/davidfarmer/lmfdb/lmfdb/__init__.py", line 5, in <module>
    from website import main
  File "/scratch/home/davidfarmer/lmfdb/lmfdb/website.py", line 52, in <module>
    setup_logging()
  File "/scratch/home/davidfarmer/lmfdb/lmfdb/website.py", line 41, in setup_logging
    formatter = logging.Formatter(utils.LmfdbFormatter.fmtString.split(r'[')[0])
AttributeError: 'module' object has no attribute 'LmfdbFormatter'

Sorry for the long and possibly useless comment!

edgarcosta:
@davidfarmer,  to avoid any merges/surprises, please try: ```git checkout -f roed314/master``` (it doesn't look that you are running the same code as the PR)

edgarcosta:
@davidfarmer  regarding the knowl width we tried to control the knowl width when inside of table.
See: https://github.com/roed314/lmfdb/blob/master/lmfdb/static/lmfdb.js#L165
The guiding principles were
1) no larger than the current row width (for normal tables)
2) no larger than the desired main width = header width - sidebar width (for extra large tables)
3) at least 700px (for small tables)
and then subtract the appropriate margins and borders.

Compare it out:
http://beta.lmfdb.org/Genus2Curve/Q/?cond=1-499 (it just takes the widht of the whole window)
vs
http://cmfs.lmfdb.xyz/Genus2Curve/Q/?cond=1-499

this is extremely relevant for tables for list the embeddings (or the table can be as wide as you want)
or small tables, like the inner twists table, where it would make the table very wide.
Here is an example where both occur: http://cmfs.lmfdb.xyz/ModularForm/GL2/Q/holomorphic/7524/2/l/b/?n=2-100

Another place to compare this behaviour with the current version is:
http://beta.lmfdb.org/L/Genus2Curve/Q/169/a/
vs 
http://cmfs.lmfdb.xyz/L/Genus2Curve/Q/169/a/
and click on any of the invariant knowls, e.g., the conductor

jwj61:
On the whole, it looks great.  So don't take my two comments as indicative of an overall negative impression.

I like the button for a random form at the bottom where it can be subject to search criteria, but think the "random form" link in the browse section (i.e., at the top) should be restored.  The browse area should be for non-experts to explore, and that group should not need to scroll down to the bottom to click for a random form.  This would be easy to change.

I am not a big fan of the big-number-knowls.  I understand that without doing something, the information would get to be too big for the screen.  But, whenever possible, I would like to see the full data.  If I already clicked saying I wanted to see some data, I want to see it even if I don't plan to use all the digits.  Instead, I feel like I am getting a redacted version of the output.  

Aside from that, if someone actually wants to see the information, having the number appear in a knowl below the line becomes a problem if they click on more than one.  They get a stack of knowls containing numbers and have to match them up themselves.

In one case, I opened a knowl to see the defining polynomial for a field, and it a bunch of the coefficients were big-number-knowls.  When I clicked on them, the numbers were not all that big.  The whole polynomial would have reasonably fit in the knowl.  Whatever threshold you use for these, I think it should be much higher when the number is appearing in a knowl.

Here are a few suggestions for dealing with this.  At a minimum, raise the threshold for the number of digits in numbers which use ... so that it happens much less often.  Second, when clicking on a ..., I expected the full number to appear in place, and think that would be better.  It could be a toggle so that one can click on the number again to make it recompress.  

As a bigger change, I think my first choice would be to handle it like pari output.  Pick a threshold for the full display of an item (e.g., of a polynomial or of a field).  Say we pick 1500 characters (of raw latex).  Then, display the first 1500 characters with a symbol at the end to indicate that it continues which one can click on to either get the whole thing, or to get the next 1500 characters.

JohnCremona:
I tend to agree with @jwj61 about hiding digits of numbers.  I went to http://cmfs.lmfdb.xyz/ModularForm/GL2/Q/holomorphic/65/2/o/a/ (a sample for of dimension 20) and to have a tiny number of coefficients partly hidden looks a bit silly.  It is very hard to say what the best heuristic should be, but the idea of showing up to so many characters with an indication that there is more to show and an easy way to show the rest (and hide it again) for a whole formula, not individual coefficients, would probably be better.

On downloads: At the bottom of a search results page there are buttons to down load files to read into the usual 3 systems.  These all define two arrays, of labels and data; and there is a  comment  'To create a list of forms, type "forms:= make_data();' but in none of the systems has a fiction make_data (yet) been implemented.  That will be quite a lot of work; for now perhaps omit that comment.

Overall I think we should get this onto the beta site as soon as possible (and have this branch merged into the master branch).

JohnCremona:
I looked at the Travis failure and it is only a timeout connecting to the Magma website.

davidlowryduda:
Firstly, this is great! And I'm glad that travis doesn't fail after a rerun.

With respect to the big-num-knowls, I looked at a several large weight pages and decided that having some way of not showing all the information is great and very necessary. On the other hand, I can imagine someone deciding that they wanted to see *all the data* without clicking every knowl.

I think a pretty easy-to-implement solution would be to add a button/link towards the top of the page saying `Show all digits`, or perhaps `Show complete coefficients`. Clicking on this would take one to a page with all the same data, but with nothing hidden behind a big-num-knowl.

A slightly different form of this solution which I think is still relatively easy-to-implement (and which might address some of the concerns of @jwj61 and @JohnCremona ) would be to have a button, similar to the analytic/arithmetic normalization buttons on L-function pages (like [this one](http://cmfs.lmfdb.xyz/L/ModularForm/GL2/Q/holomorphic/65/2/o/a/)) that toggles a complete expansion section-by-section. For instance, one might have a pair of buttons for just above the *Basis of the coefficient ring* that alternate between shortened form and complete description. On the other hand, I notice that this will double the size of the payload for these pages, and they're already pretty large.

Aside from that, I like most aspects of the big-num-knowls. It seems that right now, we use knowls if there are more than 8 digits. Some suggest having a higher cutoff, but it seems to me that we probably don't want it much larger. Perhaps 12 digits?

edgarcosta:
I have:
* added a 'random' newform to 'Some interesting newforms:'
* hopefully dealt with the timeout arising from the magma free calculator
* changed the display of polynomials and the use of bigint knowls, in the sense, that now we don't apply the bigint knowls to the coefficients of a polynomial if it wouldn't decrease the number of characters of that polynomial by more than 400 characters.
* changed the default cutoff for bigint knowls from 12/8 to 20.


 
Regarding the knowls for big numers:

- I like @jwj61 suggestion, but that would require substantial javascript coding infrastructure just for that, and that is the reason why we went for knowls. Further, I would suggest to "recompress" we would have a subscript button, otherwise the whole number would be a link.

- Futhermore, I think as a first step one should perhaps try something along the lines of what @davidlowryduda  suggested, where I would even propose that we would disable katex for those pages, as if you want to see the full thing, most likely  it is because you want to copy paste it.

jwj61:
Even if other people like my suggestions with regards to big-int-knowls, I did not intend for them to hold up the general modular forms changes.  This just seemed to be the place for discussion of any changes.

JohnCremona:
@jwj61 Agreed!

AndrewVSutherland:
I think the small changes Edgar made were good ones to make -- they go part way toward addressing some of the points raised by @jwj61 and I think they do address the point @JohnCremona made by avoiding the use of bigint knowls when it is only going to impact a few numbers on the page.

I suggest that once this PR is merged (which I agree would be good to get done soon as there is other work that depends on some of the infrastructure changes included in this PR) we raise an issue to look at the problem more generally and migrate a summary of the suggestions made here to that issue.

Personally,I really like the idea of having a single button that would turn off bigint knowls *and* give you a raw un-katexed text rendering of all the numbers and polynomials on the page -- this would make it easy to cut and paste, which is currently a pain to do on a lot of LMFDB pages, even when the data displayed is quite small (e.g. minus signs don't get copied correctly and exponents often get garbled -- in some cases you can use code snippets and/or download to avoid this, but these are not always available for your intended target and involve a lot more clicking, sometimes you just wish you could do a quick cut and paste).

AndrewVSutherland:
@JohnCremona Regarding downloads, there is a "make_data" function included in the download q-expansion links on each individual newform page that will create the q-expansion as a power series in Sage.  (the download functions for magma also let you create the q-expansion, as well the modular form object and, in cases where it makes sense, the space of modular symbols).

In the search results download you are referring to, the download link just gives you the columns you see on the screen (e.g. the labels, dimensions, traces of a2,a3,a5,a7) for the newforms matching your search.  This is not enough information to reconstruct the newforms (or even produce a meaningful q-expansion), which is why we don't include a make_data function (it would be easy to include a copy of the make_data functions used on the individual newform downloads, but they would not work because the data they need would not be present).

There are two reasons why I don't think it is feasible to provide a search results download that would let you create all the modular forms or even just their q-expansions in Sage/Magma/Pari-GP.  First, the amount of data required is likely to be very large (several hundred KB per form is typical),  Second, for newforms of dimension greater than 20 (which is about a third of them), we don't have exact eigenvalue data and can only give you the q-expansions of a complex embedding, and we would need to choose one (we certainly don't want to download them all for every form, indeed, for the larger forms it is not feasible to do this even  for one form).

In summary, the reason we did not include a make_data function for download search results is not because the code is hard to write, but because it is not feasible to provide the data necessary to do it.  This is very different from the situation with elliptic curves or number fields where the data necessary to quickly reconstruct an individual object is quite small.

Having said all that, there is of course a compact description of each newform, namely, its label. It would be easy to include a "make_data" function that takes a list of labels as input and returns a list of modular forms (indeed, you wrote the Pari/GP version of this function :)).  All it would do is construct the space of newforms S_k(N,chi) using the k,N,chi encoded in the label (the download would need to include the information needed to construct chi, but this is small), decompose the space into eigenforms, sort them according to traces, and return the nth one, where n is encoded in the last component of the label. But this function would take a very long time to run in many cases (possibly years), since it would in effect be reproducing a large part of the computations we performed to construct the database.

JohnCremona:
@AndrewVSutherland I agree with all of that 100% -- and so suggest that removeing the commented line 'To create a list of forms, type "forms = make_data()"'  from the search results download.

JohnCremona:
I nominate @jwj61 to be the one to merge this.  None of the discussion here need delay that.

AndrewVSutherland:
@JohnCremona Agreed, we should remove the comment referencing a make_data function from the search results download.

edgarcosta:
@JohnCremona , the comment has been removed when there is no function.

We didn't mention in the PR, but now we have a class for downloading data in a uniform way, which we hope to spread across LMFDB.

JohnCremona:
@edgarcosta Thanks!  confirmed.  And +1 for the uniform download plans.

JohnCremona:
Hooray!  master-->dev now?

AndrewVSutherland:
Sure, but we need to hold off on master -> prod until we have pushed the new tables.  We can take care of this today or tomorrow.

PR #2719, opened 2019-03-05
master --> dev

AndrewVSutherland:


JohnCremona:
Wow!  We now have the new CMFs on beta.lmfdb.org and finally they are all green not blue!

PR #2720, opened 2019-03-05
Various ec/ecnf script updates (for postgres and more)

JohnCremona:
This only changes import scripts for elliptic curves and ecnf, for postgres.  Originally based on roed/master so now can be merged into master.  As a sanity check I will wait for Travis to pass before merging, but only files in scripts/ have been affected.


JohnCremona:
OK, this needs cleaning up too.

PR #2724, opened 2019-03-06
more ecnf script  updates

JohnCremona:
Just more script updates, as I try to get rid of various branches.  Nothing worth anyone looking at.


JohnCremona:
But some pyflakes fixing is needed as usual...

PR #2727, opened 2019-03-07
Fixing bug in encoding.py

roed314:
This fixes a bug in encoding that prevented @JohnCremona from rewriting ecnf.

One thing to note is that we just use Sage's string representation for encoding real numbers.  At some point it might be worth thinking more about how to deal with precision when encoding/uploading data.

### DIFF
--- a/lmfdb/backend/database.py
+++ b/lmfdb/backend/database.py
@@ -2357,7 +2357,7 @@ class PostgresTable(PostgresBase):
 
         - ``F`` -- an open file handle, at the beginning of the file.
         - ``unordered`` -- a set of the columns in the table.
-        - ``sep`` -- a string giving the column separator.
+        - ``sep`` -- a string giving the column separator.  You should not use comma.
         - ``check`` -- whether to validate the columns.
         - ``adjust_schema`` -- If True, rather than raising an error if the columns
             don't match expectations, will change the schema accordingly.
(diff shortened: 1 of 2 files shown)


JohnCremona:
I will check out this branch and see if the rewrite I attempted now works.

JohnCremona:
With this change the rewrite of ec_nfcurves worked.  I am not qualified to review the code.

JohnCremona:
Well, the rewrite itself produced no errors but I can no longer see any curves on beta.lmfdb.org over feidsl other than quadratic.  e.g. http://beta.lmfdb.org/EllipticCurve/3.3.1957.1/2.1/a/1.   I am not blaming the rewrite code, it was probably something which I did so I will investigate.

JohnCremona:
There is a different encoding/decoding bug in handling string inside jsonb objects.  In ec_nfcurves the local_data column hods a lists of dicts, one key of each of this is 'kod' (Kodaira Symbol) with value a string.  Redundantly these strings start and and with "\(" and "\)" (which would be better inserted by the html template).  Since "\" is the escape character these strings must actually start and end with "\\(" and "\\)". 
During the process of inserting these values into the database the number of backslashes is increased.  Hence every time this table is rewritten, the number is increased.  For example:
`sage: nfcurves.lucky()['local_data'][0]['kod']`
`u'\\\\\\\\\\\\\\\\(I_{1}\\\\\\\\\\\\\\\\)'`

This behaviour should be fixed.  Independently of that I can rewrite the data so this example just stores I_1 and the code (which already has to remove all the extraneous backslashes) can add the "\\(".."\\)" itself.

PR #2728, opened 2019-03-08
master --> dev

JohnCremona:
Apart from irrelevant script updates this pushes the changes in two recent ecnf PRs so that http://beta.lmfdb.org/EllipticCurve/3.3.1957.1/2.1/a/1  (and similar for any curve over a field of degree>2) will no longer cause an error.


JohnCremona:
Thanks.

PR #2729, opened 2019-03-09
Minor formatting change to simplify cmf newform pages

AndrewVSutherland:
Based on feedback, the newform invariants section now omits the symbol column which was used for only a handful of invariants -- A for analytic conductor, ST(f) for Sato-Tate group, Q(\nu) for coefficient field.  These symbols did not add much value and are either not used or defined elsewhere.

Compare:

http://beta.lmfdb.org/ModularForm/GL2/Q/holomorphic/11/2/a/a/
http://127.0.0.1:37777/ModularForm/GL2/Q/holomorphic/11/2/a/a/

http://beta.lmfdb.org/ModularForm/GL2/Q/holomorphic/87/1/d/a/
http://127.0.0.1:37777/ModularForm/GL2/Q/holomorphic/87/1/d/a/

http://beta.lmfdb.org/ModularForm/GL2/Q/holomorphic/41/2/a/a/
http://127.0.0.1:37777/ModularForm/GL2/Q/holomorphic/41/2/a/a/

http://beta.lmfdb.org/ModularForm/GL2/Q/holomorphic/31/2/g/a/
http://127.0.0.1:37777/ModularForm/GL2/Q/holomorphic/31/2/g/a/

Issue #2632, opened 2018-09-26
inconsistencies in truncation vs rounding


edgarcosta:
When displaying floats, should we round or truncate?
Neither is correct, but we should stick to one.

At the moment we do both, and I think we should round.
For example, in Maass forms and L-functions, we truncate almost all the time, with the exception if number rounds to an half-integer, then we then we take that approximation.
To me this doesn't make much sense.
We are happy to display 1 when given 0.99999... and 0.5 when given 0.49999..., but we aren't happy to display 0.1 when given 0.09999...?

ps: this isssue has been brought up before here: #1978, when we dediced to round 0.49999 to 0.5 and here #1293 regarding the zera zeros

davidfarmer:
I think we should truncate.  The text we display is meant for the
eyes of a human.  Humans (me, anyway) want to see correct information.
\pi = 3.1415... .

Note the ... .  The ... and the = are the key.  It is wrong to write
"=" and not have the ... (although maybe the ... is implied).

I am pretty sure that nobody can disagree with this:
    If we write ... after the last digit we show, then we need
    to truncate because the ... stands for digits which have been hidden.

If we are going to round and not use ..., then we need another symbol
than "=", or else the statement is just plain wrong.

I prefer to see

\pi = 3.1415... .

than

\pi \approx 3.1416 .

Does anyone have a 3rd proposal for how to show \pi to 4 decimal
places?  Faced with those two options, I reject the second one because
\approx is not well defined.  We can try to define it to mean "all
digits shown are correct except the last one may be rounded", but I
don't see any real benefit in doing that. Unless I am missing something,
the first is correct and unambiguous.

Two more things:

1) When producing a number for use by a machine, there are good arguments
for rounding instead of truncating.  (But I don't think that is the use
case under discussion.)  And also good arguments for including possibly
incorrect digits after the asserted precision.

2) If the true value is 0.0999999..., then that really equals 0.1, and
it is not a question of rounding.  If the true value is 0.9999996441...,
and we are showing 3 decimal places, then the arguments above say
that we should show 0.999... .

It is true that there are places where we know (or conjecture) that the
actual answer has to be in the set {..., -1/2, 0, 1/2, 1, ...}.
In that case, converting 0.4999.... to 0.5 is not an act of rounding.

edgarcosta:
> If we are going to round and not use ..., then we need another symbol
> than "=", or else the statement is just plain wrong.

I disagree.
[...]

PR #2730, opened 2019-03-09
Change display of EC Invariants and BSD Invariants

JohnCremona:
Similar changes to #2729, for individual elliptic curve pages (over Q only right now).  Some issues on which comments are invited:
 - The heading Invariants is no longer a knowl as that seemed redundant, but I left the heading BSD Invariants as a knowl, linking just "BSD" to   ec.q.bsdconjecture instead.
- I used ec.* knowls rather than ec.q.* knowls unless there was a good reason.  e.g. ec.q.conductor, since by convention the conductor of an elliptic curve over Q is a positive integer while in general it is an ideal.
- Approximate values (Regulator unless r=0, and real period): these used to have an "approximately equals" sign to indicate that the floating point values displayed are (obviously) approximate.  Now that has gone, should I add "..." after the digits instead?
- "Tamagawa product and numbers" is rather long but we do display both, and want to keep that.  The knowl is only for an individual Tamagawa number but that is surely OK.

### DIFF
--- a/lmfdb/elliptic_curves/templates/ec-curve.html
+++ b/lmfdb/elliptic_curves/templates/ec-curve.html
@@ -170,7 +170,7 @@ white-space: pre in order to keep line breaks! #}
 
        </table>
 
-    <h2> {{ KNOWL('ec.q.bsd_invariants', title='BSD invariants') }} </h2>
+    <h2> {{ KNOWL('ec.q.bsdconjecture', title='BSD') }} invariants</h2>
     <div></div>
     <p>
         <table>
--- a/lmfdb/elliptic_curves/templates/ec-curve.html
+++ b/lmfdb/elliptic_curves/templates/ec-curve.html
@@ -96,7 +96,7 @@ white-space: pre in order to keep line breaks! #}
     {%endif %}
 
     {% if data.mw.tor_order!=1 %}
-    <h2> {{ KNOWL('ec.torsion_generators', title='Torsion generators') }}</h2>
+    <h2> {{ KNOWL('ec.torsion_subgroup', title='Torsion generators') }}</h2>
       {{ place_code('tors') }}
     <p> {{ data.mw.tor_gens }} </p>
     {%endif %}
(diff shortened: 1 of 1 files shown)


edgarcosta:
> should I add "..." after the digits instead?

We don't do that anywhere else, and I don't think we should start now.

JohnCremona:
> > should I add "..." after the digits instead?
> 
> We don't do that anywhere else, and I don't think we should start now.

Are you sure?  There have been many discussions about this (involving @davidfarmer and others), about the implication that if you put "..." at the end the implication is that all digits shown are correct, so this is a truncation, while if you put an approximately equals sign and no dots then the implication is that the number shows is correct to however many decimal places are shown, so this is a rounding.  

You may be right that we do not actually use 3 dots in this way.  I tried git grep "\.\.\." but that gets quite a few irrelevant hits...

edgarcosta:
1- next time try `git grep -F "..."`
2- The issue was #2632, which perhaps I should close, as this needs to be solved at the editorial level.
3- Also see https://github.com/LMFDB/lmfdb/blob/master/lmfdb/utils/utilities.py#L290 where I tried to unify all the float printing.

AndrewVSutherland:
@JohnCremona I like this change, definitely an improvement!

I have two small suggestions:

* Link the BSD Invariants heading to the knowl ec.q.bsdconjecture
* Shorten the caption "Tamagawa product and numbers" to just "Tamagawa product" and let the knowl explain that it is the product of the Tamagawa numbers (I actually just updated it to do this and corrected a small type).  I think it will be obvious that the product on the right is the factorization into Tamagawa numbers at bad primes (and this can be seen lower on the page in any case).

JohnCremona:
> @JohnCremona I like this change, definitely an improvement!
> 
> I have two small suggestions:
> 
Thanks

> * Link the BSD Invariants heading to the knowl ec.q.bsdconjecture

I had already done that in the 2nd commit.

> * Shorten the caption "Tamagawa product and numbers" to just "Tamagawa product" and let the knowl explain that it is the product of the Tamagawa numbers (I actually just updated it to do this and corrected a small typo).  I think it will be obvious that the product on the right is the factorization into Tamagawa numbers at bad primes (and this can be seen lower on the page in any case).

Done in 3rd commit.  Thanks for editing the knowl.

AndrewVSutherland:
Looks good.

PR #2731, opened 2019-03-10
Fix bug in encoding, add cleanup_from_reload call to drop_table

roed314:
Fix bug noticed by @JohnCremona in https://github.com/LMFDB/lmfdb/pull/2727

### DIFF
--- a/lmfdb/backend/database.py
+++ b/lmfdb/backend/database.py
@@ -4502,6 +4502,7 @@ SELECT table_name, row_estimate, total_bytes, index_bytes, toast_bytes,
     def drop_table(self, name, commit=True):
         with DelayCommit(self, commit, silence=True):
             table = self[name]
+            table.cleanup_from_reload()
             indexes = list(self._execute(SQL("SELECT index_name FROM meta_indexes WHERE table_name = %s"), [name]))
             if indexes:
                 self._execute(SQL("DELETE FROM meta_indexes WHERE table_name = %s"), [name])
--- a/lmfdb/backend/encoding.py
+++ b/lmfdb/backend/encoding.py
@@ -225,7 +225,7 @@ class Json(pgJson):
                     'data': data}
         elif escape_backslashes and isinstance(obj, basestring):
             # For use in copy_dumps below
-            return obj.replace('\\','\\\\\\\\').replace("\r", r"\r").replace("\n", r"\n").replace("\t", r"\t").replace('"',r'\"')
+            return obj.replace('\\','\\\\').replace("\r", r"\r").replace("\n", r"\n").replace("\t", r"\t").replace('"',r'\"')
         elif obj is None:
             return None
         elif isinstance(obj, datetime.date):


JohnCremona:
I don't know if it is the same bug or a different one but every entry in the ec_nfcurves table has a string column 'equation' which is supposed to start with '\(' and end with '\)' but right now there are three " (double quote) at the beginning and end which makes all pages such as http://beta.lmfdb.org/EllipticCurve/6.6.1528713.1/71.1/a/3
display badly.

I cannot tell whether these are added at each rewrite or if they appeared there at the mongo-to-postgresql conversion.  I don't think it the latter since they would have been noticed before.

I am adding to a current branch code which removes these extraneous " characters before display, but of course they should be removed from the database.  I could do yet another rewrite of ec_nfcurves but they take > 10 minutes.

JohnCremona:
Sorry but I am not really capable of testing this.  Would dumping the ec_nfcurves to a file and looking at that be a test?  Right now the Kodaira symbols have precisely 4 backslashes in from of the ( and ) symbols.  Perhaps it would be worth a rewrite anyway to remove the triple double quotes from front and back of the 'equation' string?

roed314:
One test would be to do a rewrite and remove the quotes and extra backslashes.  If the result has only a single backslash, then the fix should be alright (since dumping to a file is done after calling the rewrite function).

It might be worth then doing another rewrite to remove the \( completely and put them into the code.

This same line could be affecting the quotes (you can see that it replaces " with \").  But I'm not sure why that would be relevant here, since I don't understand how quotes would be getting in in the first place.

JohnCremona:
OK, I will (on master + branch) try such a rewrite.  Right now the code replaces four backslashes with two recursively, but requires the result (for Kodaira symbols) to start '\(' and end '\)'.  So I'll also make a code PR which will correctly display these whether or not '\(', '\)' are there.

JohnCremona:
Sorry but when I do a rewrite of ec_nfcurves now using this branched (merged with master) I get the error I had before:

/scratch/home/jcremona/lmfdb/lmfdb/backend/database.pyc in rewrite(self, func, query, resort, reindex, restat, tostr_func, commit, **kwds)
   2049                     for rec in self.search(query, projection=projection, sort=[]):
   2050                         processed = func(rec)
-> 2051                         searchfile.write(u'\t'.join(tostr_func(processed.get(col), self.col_type[col]) for col in search_cols) + u'\n')
   2052                         if self.extra_table is not None:
   2053                             extrafile.write(u'\t'.join(tostr_func(processed.get(col), self.col_type[col]) for col in extra_cols) + u'\n')

/scratch/home/jcremona/lmfdb/lmfdb/backend/database.pyc in <genexpr>((col,))
   2049                     for rec in self.search(query, projection=projection, sort=[]):
   2050                         processed = func(rec)
-> 2051                         searchfile.write(u'\t'.join(tostr_func(processed.get(col), self.col_type[col]) for col in search_cols) + u'\n')
   2052                         if self.extra_table is not None:
   2053                             extrafile.write(u'\t'.join(tostr_func(processed.get(col), self.col_type[col]) for col in extra_cols) + u'\n')

/scratch/home/jcremona/lmfdb/lmfdb/backend/encoding.pyc in copy_dumps(inp, typ)
    339         return r'\\x' + ''.join(c.encode('hex') for c in inp)
    340     else:
--> 341         raise TypeError("Invalid input %s (%s) for postgres type %s"%(inp, type(inp), typ))

TypeError: Invalid input [0.0458094381868886] (<type 'list'>) for postgres type numeric[]

That's the new list of heights causing the problem it seems.

JohnCremona:
My mistake I merged in #2731 instead of #2727.  I merged that one in too (copy_dumps_array as well as jsonb_backslashes) and if the rewrite works OK I'll merge both the PRs.  The rewrite takes 10 minutes...

JohnCremona:
Backslashes are now fine for me (in the rewrite I removed the offending ones anyway) but the rewrite is adding a double quote character at the front and back of the 'equation' strings in ec_nfcurves:

`sage: db.ec_nfcurves.search(limit=1)[0]['equation']`
`u'" y^2 + a x y + \\left(a + 1\\right) y = x^{3} + \\left(-41 a + 158813\\right) x + 2115 a - 13286543  "'`

My code strips these off but it is assign that the rewrite is not doing everything quite right.  On the other hand there are many columns of type text in this table and this is the only one which gets teh double quotes added:
`sage: [(k,C[k]) for k in nfcurves.col_type if nfcurves.col_type[k]=='text']`
`[(u'ainvs', u'0,1;0,0;1,1;158813,-41;-13286543,2115'),
 (u'field_label', u'2.0.10691.1'),
 (u'class_label', u'2.0.10691.1-1.0.1-a'),
 (u'conductor_ideal', u'[1,1,1]'),
 (u'minD', u'[1,1,1]'),
 (u'label', u'2.0.10691.1-1.0.1-a1'),
 (u'jinv', u'-216,55'),
 (u'conductor_label', u'1.0.1'),
 (u'iso_label', u'a'),
 (u'short_label', u'1.0.1-a1'),
 (u'short_class_label', u'1.0.1-a'),
 (u'equation',
  u'" y^2 + a x y + \\left(a + 1\\right) y = x^{3} + \\left(-41 a + 158813\\right) x + 2115 a - 13286543  "')]`

JohnCremona:
I see no way that teh issue with " is caused by this commit so I will merge it.

roed314:
> Backslashes are now fine for me (in the rewrite I removed the offending ones anyway) but the rewrite is adding a double quote character at the front and back of the 'equation' strings in ec_nfcurves:
> 
> `sage: db.ec_nfcurves.search(limit=1)[0]['equation']`
> `u'" y^2 + a x y + \\left(a + 1\\right) y = x^{3} + \\left(-41 a + 158813\\right) x + 2115 a - 13286543 "'`

Note that an easier way to get an equation example is `sage: db.ec_nfcurves.lucky({}, 'equation')`.


> My code strips these off but it is assign that the rewrite is not doing everything quite right. On the other hand there are many columns of type text in this table and this is the only one which gets teh double quotes added:
> `sage: [(k,C[k]) for k in nfcurves.col_type if nfcurves.col_type[k]=='text']`
> `[(u'ainvs', u'0,1;0,0;1,1;158813,-41;-13286543,2115'), (u'field_label', u'2.0.10691.1'), (u'class_label', u'2.0.10691.1-1.0.1-a'), (u'conductor_ideal', u'[1,1,1]'), (u'minD', u'[1,1,1]'), (u'label', u'2.0.10691.1-1.0.1-a1'), (u'jinv', u'-216,55'), (u'conductor_label', u'1.0.1'), (u'iso_label', u'a'), (u'short_label', u'1.0.1-a1'), (u'short_class_label', u'1.0.1-a'), (u'equation', u'" y^2 + a x y + \\left(a + 1\\right) y = x^{3} + \\left(-41 a + 158813\\right) x + 2115 a - 13286543 "')]`

Strange.  I'm not going to be able to take a look at this today, since I have to focus on teaching stuff, but I'll try to figure out what's going on soon.

PR #2735, opened 2019-03-13
Fix a bug in verify

roed314:
I changed the interface for verify on tables but neglected to make the same change for verify on the database as a whole.


edgarcosta:
Looks good to me!

PR #2744, opened 2019-03-14
ecnf: strip extraneous characters from Kodaira symbols and equations

JohnCremona:
See #2731 .

I am about to run a rewrite of ec_nfcurves (as suggested in #2731) which removes extraneous " and \ characters from the latex strings for equations and Kodaira symbols, also removing the math delimiters \( and \).

This code change ensures that for both, the things will display properly, both by applying the same fix to these strings (just in case) and adding the latex delimiters \( and \) to the template so they do not need to be stored.

This can be merged right away as it will correctly handle the display of these things both before and after the rewrite.  After the rewrite, until this is merged they will not display properly since the latex delimiters will be missing.  Hence it would be good to merge this now (and to dev).


JohnCremona:
I based this on the commit at #2743 (since my script does not work without that) so perhaps that should be merged first.

JohnCremona:
The rewrite of ec_nfcurves has now run successfully and so pages such as
http://beta.lmfdb.org/EllipticCurve/4.4.13625.1/80.3/a/2
will not look right (Weierstrass equations and Kodaira symbols in Local Data) until this PR is merged & pushed to dev.

jwj61:
I tried to look at the curve on beta as suggested and it seems beta is down.

AndrewVSutherland:
beta should be back up in a few minutes, it is being rebooted to get a memory upgrade.

JohnCremona:
Thanks @jwj61 -- as usual there was no easy way to make data and code changes simultaneously, especially now that rewrite always (?) rewrites a table in place rather than creating a new one and allowing you to rename the new table later.

AndrewVSutherland:
Apologies for not warning about the reboot, I only just found out they were going to do the upgrade this morning (it was either now or some time early next week and I didn't want it happening in the middle of the Princeton workshop).

AndrewVSutherland:
@JohnCremona We have a plan that in the future will allow us to make these sort of data changes without breaking either the old or new versions of the code (in fact both would be able to work at the same time)  -- this is something we can discuss in more detail next week.

JohnCremona:
> @JohnCremona We have a plan that in the future will allow us to make these sort of data changes without breaking either the old or new versions of the code (in fact both would be able to work at the same time) -- this is something we can discuss in more detail next week.

It's going to be a busy week :)

AndrewVSutherland:
legendre is back up and beta is available again.  Unfortunately there was a hiccup with the memory upgrade and legendre may need to be rebooted again later today

PR #2750, opened 2019-03-15
removing some static JS that now is served through a CDN

edgarcosta:
In #2736 we started to use CDN to provide standard javascript libraries, like jquery and jquery-ui.

This PR addresses the last places where these libraries were referenced and removes them for our repo.


davidfarmer:
The code looks fine and seems to function as expected. Merging.

PR #2774, opened 2019-03-16
show/hide sidebar with javascript

edgarcosta:
Try it out: http://teal.lmfdb.xyz/

### DIFF
--- a/lmfdb/app.py
+++ b/lmfdb/app.py
@@ -220,13 +220,10 @@ def not_found_503(error):
 
 @app.before_request
 def get_menu_cookie():
-    g.show_menu = request.cookies.get('showmenu') != "False"
-
-@app.after_request
-def set_menu_cookie(response):
-    if hasattr(g, 'show_menu'):
-        response.set_cookie("showmenu", str(g.show_menu))
-    return response
+    """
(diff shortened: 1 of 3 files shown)


AndrewVSutherland:
Nice, this is a lot faster.

JohnCremona:
Yes, that's much faster.  Thanks.

Since the big sidebar rewrite a few years ago I have wanted to redo it (again).  The python code is simple, reading the layout from a yaml file and doing some fairly simple transformations, but the template is horrendously complicated, and hence hard to fix when a small change in sidebar.yaml does not work as it should.  But my point is simply that the contents of the sidebar never changes (while the code is running), yet as far as I can tell whenever every single page is loaded the sidebar html is reconstructed from its template.  (I may be wrong in that.)

This suggests that in the sidebar initialization function (which does run precisely once) that html content could be constructed once and for all -- perhaps stored in some large string? -- and then simply displayed.  One additional advantage would be that the complicated layout would be done in python and not in the templating language.  That is closer to what @davidfarmer first wrote when we started the sidebar.yaml change, which I changed into what we have now, and I am admitting that mine might not have been the best idea.

This could be turned into a new issue.

PR #2776, opened 2019-03-16
Fix bugs in genus 2 search results downloads

AndrewVSutherland:
This fixes two problems with genus 2 search downloads:

(1) There were two sets of download buttons but only the bottom set was using the new download object, I removed the top set of download buttons to be consistent with other search results pages in the LMFDB.

(2) The make_data() function in the download scripts was broken because the curve equations are stored as strings and were being blindly copied into the data array.  To address this, there is now an optional column_wrappers attribute has been added to the Downloader class which allows you to specify a mapping function for any particular column (in this case the mapping function is just string_literal)

I also fixed a formatting issue with the search again button in firefox

To test you can compare downloads on

http://www.lmfdb.org/Genus2Curve/Q/?cond=1-1000
http://localhost:37777/Genus2Curve/Q/?cond=1-1000

You should be able to cut/paste the download scripts for Pari/GP, Sage, or Magma into the relevant system and then run make_data to get the list of curves.

Be aware that if you have visited these pages before you may need to clear your browser cache to see the new behavior.

### DIFF
--- a/lmfdb/genus2_curves/main.py
+++ b/lmfdb/genus2_curves/main.py
@@ -237,6 +237,7 @@ class G2C_download(Downloader):
     table = db.g2c_curves
     title = 'Genus 2 curves'
     columns = 'eqn'
+    column_wrappes = {'eqn':literal_eval}
     data_format = ['[[f coeffs],[h coeffs]]']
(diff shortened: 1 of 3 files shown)


roed314:
Looks good to me!

PR #2784, opened 2019-03-18
Add guiding principles for the LMFDB

davidfarmer:
The development of the LMFDB has been guided by certain principles,
but those have not been officially adopted.

This is a first draft to start the conversation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,14 +15,14 @@ Development
 * [Getting Started](https://github.com/LMFDB/lmfdb/blob/master/GettingStarted.md) -- cheat sheet for setting up lmfdb
 * [Development Guide](https://github.com/LMFDB/lmfdb/blob/master/Development.md) -- organizing development
 
-The design of the LMFDB is guided by the following principles
--------------------------------------------------------------
+Guidelines
+----------
 
-1. Each mathematical object has its own home page.  Home pages
+1. Each mathematical object should have its own home page.  Home pages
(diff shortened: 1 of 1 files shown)


JohnCremona:
See https://github.com/LMFDB/lmfdb/tree/davidfarmer-readme-1 to read the new text as it would appear.

JohnCremona:
On 2: should we say explicitly that there can be more than one valid URL for the same home page, e.g. http://www.lmfdb.org/EllipticCurve/Q/11/a/2 = http://www.lmfdb.org/EllipticCurve/Q/11a1 =http://www.lmfdb.org/EllipticCurve/Q/11.a2  (or perhaps a less contentious example).

On 1: possibly add: (a) Properties box giving very basic facts (all of which should also be on the main part of the page); and Related Objects (Friends), and Downloads etc on the right?  (b) Links to RCS knowls on each home page?

On 4: "entities"?  I suppose you mean software as well as individuals.  It's a pity if listing software means that the authors of the software are not mentioned.  If person X uses a package written by A and B within software S, are all four to be listed?  Note that this sounds as if X only pressed buttons, but X's work can be major (e.g. recent work on CMFs).

On 6: The aim is of course good.  This is the only one which also talks about the process.  I suggest putting the last part "[Home pages and browse-and-search pages] should be designed so that they are useful to a broad mathematical audience.  Normally, this will be achieved by an iterative process involving feedback from both experts and non-experts in the subject." ?

PR #2787, opened 2019-03-18
Null's are converted as None's

edgarcosta:
In MongoDB, if one field (column) is Null, i.e., non-existent, then the dictionary wouldn't have that field as a key.
This doesn't make sense in PostgreSQL, at first, to not break things, we tried to keep this behavior, i.e., if the column is Null, we omitted from the correspondent dictionary.
This pull request tries to fix this behavior but is very likely that we broke a lot of things.
For that reason, we are running this branch on http://olive.lmfdb.xyz for you to try out, and perhaps find broken stuff that relied on the behavior of MongoDB.


JohnCremona:
Looking at all the code changes involved here makes it hard to be confident that everything has been caught.  How are you sure that you have not missed something?

I can check the ecnf and elliptic_curves parts, checking the code changes (and that no more are needed) and the web pages for objects affected by this.  But other people need to be involved too.

JohnCremona:
ecnf: some cases are missed,  The fields 'rank', 'rank_bounds', 'analytic_rank', 'gens', 'ngens' are missing from very many curves.  This absence is caught by AttributeErrors.  Also 'isogeny_matrix' is absent for all curves with 'number'>1 and this needs handling in isog_class.py.  (I'm not 100% sure of that one since in practice we should only construct the class ECNF_isoclass with a class label but the constructor does allow any curve label to be used.

AndrewVSutherland:
As discussed the plan is to handle this in a slightly different way with a new PR

Issue #2789, opened 2019-03-18
Rank 4 elliptic curve L-function


davidfarmer:
Brian Conrey pointed out that the L-fcn of the rank 4 elliptic curve no longer
shows a key feature that appears near t=14.  The domain of the graph needs to
be increased slightly.

http://beta.lmfdb.org/L/EllipticCurve/Q/234446/a/

PR #2793, opened 2019-03-18
extending the plot range

edgarcosta:
This extends the plot range to all available data for the primitive L-functions, and for non-primitive we extend it by 20%.

This is an effort to address #2789


davidfarmer:
I am looking at this.

PR #2794, opened 2019-03-18
missing degree on get_instances_by_trace_hash

edgarcosta:
We were missing the degree argument in `get_instances_by_trace_hash` for Genus 2 Curves.

An example of a broken link is: http://beta.lmfdb.org/L/Genus2Curve/Q/2916/a/

I also added a test to be sure we don't break this page in the future.


AndrewVSutherland:
Looks good, I will merge as soon as travis finishes.

PR #2803, opened 2019-03-20
Fixed capitalization and alpha -> nu (minor display issues).

jvoight:

PR #2804, opened 2019-03-20
Linking to rcs knowls for local fields and Galois groups.

jwj61:
As the title says.  This also constructs the learnmore boxes for these sets of pages in a manner similar to many other sections (with a learnmore_list and the remove function).


AndrewVSutherland:
Test failure is in EC, same as https://github.com/LMFDB/lmfdb/pull/2803#issuecomment-474859112, not related to this PR

JohnCremona:
That test failure fixed in #2807

PR #2810, opened 2019-03-20
Reliability of Data + fix two errors

jenpaulhus:
This pull request adds a "Reliability of the data"  link to pages on the higher genus data main page:
http://localhost:37777/HigherGenus/C/Aut/

And the individual data pages, such as:
http://localhost:37777/HigherGenus/C/Aut/12.120-38.0.2-4-15.1

And the family pages, such as:
http://localhost:37777/HigherGenus/C/Aut/12.120-38.0.2-4-15

It also fixes a small search error which came about after the switch from mongo to Postgres.
If you go to:  http://www.lmfdb.org/HigherGenus/C/Aut/  and type in a single integer into the Group order(s) search box  (like say 60) you will see an error.  That error should be fixed by this pull.  

Finally, it fixes another related small error from the switch from mongo to Postgres on the stats page.  If you go to the stats page  http://www.lmfdb.org/HigherGenus/C/Aut/stats   and click on any of the "maximum group order" links, you will see an error.  Again, this pull request should fix that error.


JohnCremona:
I have checked this out.  When I click on the Reliability link I see a blank page!  Perhaps we can look at it on Thursday.

jenpaulhus:
I just haven't filled in that knowl yet.  I was waiting for the pull request to create the knowl and then was going to write it after the pull request.  I can write it first if that is preferred.

AndrewVSutherland:
I think it would be good to create the reliability knowl now, that way we can see that it shows up correctly in testing, and we know that when we merge this PR there won't be a period of time where the knowl does not exist and users might be led to a blank page.

JohnCremona:
Yes, I think @AndrewVSutherland 's suggestion is good.

jenpaulhus:
The knowl is up now.   

Also I noticed that the titles of the CMF pages for reliability, source, label, completeness were not capitalized so I capitalized those.

PR #2812, opened 2019-03-20
Add root discriminant to the display of a number field

jwj61:
Some bottom knowls point out that a field has small root discriminant.  This adds that value as a basic invariant of the field.  Just look at the home page of any number field.

### DIFF
--- a/lmfdb/number_fields/number_field.py
+++ b/lmfdb/number_fields/number_field.py
@@ -5,13 +5,13 @@ import ast, os, re, StringIO, time
 import flask
 from flask import render_template, request, url_for, redirect, send_file, flash, make_response
 from markupsafe import Markup
-from sage.all import ZZ, QQ, PolynomialRing, NumberField, latex, primes, pari
+from sage.all import ZZ, QQ, RR, PolynomialRing, NumberField, latex, primes, pari
 
 from lmfdb import db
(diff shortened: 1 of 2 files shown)


JohnCremona:
Looks fine to me.  Do you want the r.d. to also appear in the Properties box?

jwj61:
Good suggestion, so now it does it.

I was really thinking of printing rd with fixed precision, which may be LMFDB heresy, but now this pull request does that.

JohnCremona:
Looks good.  For the fixed precision output, I'm not sure why you do not use

>sage: R = RealField(200)
>sage: pi = R.pi()
>sage: pi.str(digits=5)
>'3.1416'
>sage: pi.str(digits=10)
>'3.141592654'
>sage: pi.str(digits=20)
>'3.1415926535897932385'
>sage: pi.str(digits=30)
>'3.14159265358979323846264338328'

but perhaps that does not have the same behaviour?  If we can standardise this the function should go into utils.  But that can wait.

edgarcosta:
Why not `display_float`?

jwj61:
Both display_float and the sage suggestion fixes the total number of digits.  I wanted 2 places after the decimal point regardless of number of digits in the number (so it can produce 3.14 and 10.23, not 3.14 and 10.2).

PR #2814, opened 2019-03-21
Knowl renaming

roed314:
Change the workflow for admins to rename knowls.  The previous workflow would result in broken knowl links on web after the knowl had been renamed but before the code was updated to use the new name.

### DIFF
--- a/lmfdb/knowledge/knowl.py
+++ b/lmfdb/knowledge/knowl.py
@@ -450,10 +450,37 @@ class KnowlBackend(PostgresBase):
         easy_matches = subprocess.check_output(['git', 'grep', knowlid.replace('.',r'\.')]).split('\n')
         return 1 if (len(matches) == len(easy_matches)) else -1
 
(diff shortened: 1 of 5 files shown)


davidfarmer:
I was given a demo of this new workflow, and all seemed fine.

The code looks okay to me, but I do not think it is feasible to comprehensively
check all the use cases of renaming/changing your mind about renaming/
editing while in the midst of renaming/un-changing your mind/etc.  So I suggest
merging and just being aware to check that things work as expected once this
is live.  As I cannot reach the usual threshold of for checking new code,
I am not clicking the 'merge' button.  But if someone else agrees with me, I think
that is good enough (once the usual tests pass).

edgarcosta:
This looks good to me, and given that only admins can rename I will go
ahead and merge it.

Worse case scenario, of David or I not seeing a corner case is that a knowl
won't be displayed, but no content will be lost.

Issue #2816, opened 2019-03-21
Copying KaTeX as TeX


alexjbest:
Currently it can be a little difficult to obtain the TeX representation of a rendered formula quickly. This can be useful to copy a curve equation or q-expansion for instance into a tex document, which I imagine a fair number of people want to do.

Under MathJax one could right click Show Math as-> Tex Commands, but with katex there is no such feature by default, simply selecting the math by drag-clicking and copying does not work well.

I suggest we try out enabling the copy tex extension to katex https://github.com/KaTeX/KaTeX/tree/master/contrib/copy-tex to see if this can be remedied.

One possible hurdle is that some things like q-expansions are broken up into several math blocks IIRC so that they overflow better, so they would possibly not work out of the box (it also might just work!).

Edit:
I have now tried this at #2816 it works quite well for me, simply click a formula to select it then right click copy, (the only downside I saw so far is that it is hard to select a small part of a formula, like a number 1234 in "1234 \cdot 12345") q-expansions as expected are  a little funny, but you can drag to select the whole expansion, and then the result includes a lot of "$ $" between terms.

PR #2817, opened 2019-03-21
Enable katex-copy

alexjbest:
See #2816 hopefully Edgar can make a colour coded demo of this so everyone can try, (may I suggest the colour "slategray").


edgarcosta:
http://purple.lmfdb.xyz/

edgarcosta:
The copy tex works reasonably well, as long as we didn't split the formula in many small blocks.

It even works well on the phone, as I was able to copy a factored discriminant
```
$3\cdot 13\cdot 103\cdot 827\cdot 1401167$
```

PR #2818, opened 2019-03-21
Color settings

roed314:
Add the ability to set the color scheme as a user preference.


AndrewVSutherland:
It looks like pyflakes is not happy and test_css failed?

roed314:
I've fixed the issues and pushed changes again; let's see what travis says.

PR #2819, opened 2019-03-21
as_string  -> mogrify

edgarcosta:
A better error message when there is a `ProgrammingError`

### DIFF
--- a/lmfdb/backend/database.py
+++ b/lmfdb/backend/database.py
@@ -362,16 +362,24 @@ class PostgresBase(object):
                 try:
                     cur.execute(query, values)
                 except (ProgrammingError, NotSupportedError):
-                    print cur.mogrify(query, values)
+                    try:
+                        print cur.mogrify(query, values)
+                    except Exception:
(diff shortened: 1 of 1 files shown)


jwj61:
Need a command to test on

edgarcosta:
This only works if your command is wrong, here is an example:
```
db._execute(SQL("SELECT * from nontable where col = "), ([3.14,2.5],))
```

on the current version you get two lines before raising the error
```
SELECT * from nontable where col = 
([3.14000000000000, 2.50000000000000],)
```
In the new version, you should get the real command in SQL that is indeed tried to run.

roed314:
I updated the PR to replace more instances of `as_string` with `mogrify`.  Here are some examples to try to see the new behavior (the second one might not be slow enough to see the slow query warning:

```
sage: from lmfdb import db
sage: db.nf_fields.analyze({'degree':int(5)},limit=20)
sage: db.nf_fields.search({'ramps':{'$contains':[2,3,11]}}, limit=20)
```

PR #2822, opened 2019-03-21
allowing square brackets for belyi

edgarcosta:


roed314:
There's a problem with flashing the same message twice.  Edgar is fixing it.

PR #2824, opened 2019-03-21
Use sage's pyflakes

davidlowryduda:
We include pyflakes in requirements.txt, so we can expect developers to
have sage's pyflakes available. On systems with a modern python3
development environment, the command `pyflakes` can default to
pyflakes-for-python3.

This (very small) commit has our test.sh use sage's pyflakes by default.

### DIFF
--- a/test.sh
+++ b/test.sh
@@ -39,8 +39,14 @@ fi
 
 WHAT="$@"
 
+SAGE_COMMAND=$SAGE
+if [[ "$SAGE_COMMAND" == "" ]]; then
+  SAGE_COMMAND=sage
+fi
+echo "Using Sage command $SAGE_COMMAND"
+
 echo "Running pyflakes..."
-read PYFLAKES_ERRCNT < <(find . | grep "\.py$" | xargs pyflakes 2>&1 | tee /dev/stderr | grep "py:" -c)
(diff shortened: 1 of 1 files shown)


JohnCremona:
Would using pyflakes2 explicitly also solve this?  I think that pyflakes is a symlink to either pyflakes2 or pyflakes3.

davidlowryduda:
It would solve it if someone had pyflakes2 on their system. But this is different than the pyflakes we installed in `requirements.txt`.

I noticed this as I recently set up a different experimental server-in-the-sky and didn't configure it for python2 development at all --- and I don't have pyflakes2. But I do have sage (and pyflakes2 within sage) set up.

I should say that I don't know anything about the two errors travis caught, except that they're unrelated to this pull request.

roed314:
This would be nice for me.  I don't have pyflakes globally on my system, so I need to use
`sage -python -m pyflakes $LMFDB_ROOT`
when I want to run it.

JohnCremona:
A test is failing since the number of Maass forms in the db has changed. @AndrewVSutherland is suppressing this test until the number has stabilised.

roed314:
Looks good to me.

edgarcosta:
Also looks good to me and most importantly to Travis agrees :)

PR #2825, opened 2019-03-21
Fix bug in record counts

roed314:
This bug is causing us to rarely use the recorded counts tables.  Fixing it should mean many more search results pages will display the number of results, and will prevent duplicated data in counts tables.

### DIFF
--- a/lmfdb/backend/database.py
+++ b/lmfdb/backend/database.py
@@ -3512,17 +3512,17 @@ class PostgresStatsTable(PostgresBase):
         cur = self._execute(selecter, values)
         nres = cur.fetchone()[0]
         if record:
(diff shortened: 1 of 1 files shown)


JohnCremona:
I will first try to see what is making Travis fail since it is messing up our workflow.

JohnCremona:
A test is failing since the number of Maass forms in the db has changed.  @AndrewVSutherland is suppressing this test until the number has stabilised.

JohnCremona:
A test is failing since the number of Maass forms in the db has changed. @AndrewVSutherland is suppressing this test until the number has stabilised.

edgarcosta:
I will go ahead and merge this, as this will significantly improve the performance.

PR #2829, opened 2019-03-21
Nf stats

jwj61:
This updates the number field stats page to stop using the old_stats database.  It just makes queries counting things, which is automatically stored in the counts database.

There have been fields recently added in degrees 4, 6, and 7, so some of the totals for these degrees will be bigger after this pull request compared to what is currently on beta or prod (which are using counts computed before the recent additions).


JohnCremona:
I will look at this and hope to learn a thing or two.

PR #2835, opened 2019-03-21
CRS knowls and placement for HMFs

jvoight:


jenpaulhus:
Looks good except for
/ModularForm/GL2/TotallyReal/Labels
where the labels link is still listed on the "Learn More about" tab.

jvoight:
Good catch Jen!  You definitely earn your associate editor points.  :)  I fixed it to lower case and I think that's still included in this pull request.

jenpaulhus:
That works!

PR #2837, opened 2019-03-21
Fixing typo in inserting into column defines in knowls

edgarcosta:
This is the column where we store the terms defined in the content of the knowl, without this change, we were only inserting the first letter of such term.

You can try it yourself:
```
import  lmfdb.backend.database
db = lmfdb.backend.database.db
SQL = lmfdb.backend.database.SQL
db._execute(SQL("SELECT defines FROM kwl_knowls WHERE id = 'cmf.embedding_label';")).fetchall()
>>> [([u'label'],), ([u'l'],), ([u'label'],)]
```

The first entry was introduced in the conversion, the `l` is what I get when eddited in beta, and `label` is what you get when edditing with this PR

Once this is merged, we will fix all previous entries with a SQL script.

### DIFF
--- a/lmfdb/knowledge/knowl.py
+++ b/lmfdb/knowledge/knowl.py
@@ -108,7 +108,7 @@ def normalize_define(term):
     return ' '.join(term.lower().replace('"', '').replace("'", "").split())
 
 def extract_defines(content):
-    return sorted(set(x[0].strip() for x in defines_finder_re.findall(content)))
+    return sorted(set(x.strip() for x in defines_finder_re.findall(content)))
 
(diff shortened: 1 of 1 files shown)


edgarcosta:
A test will fail 
```
self.check(homepage, "/ModularForm/GL2/Q/Maass/",  'The database contains 16599 Maass forms')
```
 but has nothing to do with this PR

roed314:
Looks good.  Once this is pushed to dev, we'll update the current knowls to fix their `defines` column.

edgarcosta:
Thanks!

PR #2838, opened 2019-03-21
Hack belyi part II

edgarcosta:
I had to remove a couple of extra characters and do it the right way.

To show that it works I created and edited a top knowl:
http://beta.lmfdb.org/Belyi/4T5/%5B4%2C4%2C3%5D/4/4/31/g1/a

This knowl should be deleted, unless @jvoight finds it relevant.


jvoight:
I didn't check that this works (just saw my name), but FWIW I think it's a reasonable knowl to keep.

roed314:
LGTM.

PR #2843, opened 2019-03-22
ecnf extent knowl -- do not always display on browse page!

JohnCremona:
Please merge ASAP since I greatly extended the ecnf extent knowl from a single line, but the entire contents of that knowl is displayed on beta (and www) e.g. www.lmfdb.org/EllipticCurve which we do not want!

I would rather not remove all the new knowl text.  After this PR the one-line summary is displayed as before and the full Extent knowl is available under the "Completeness" link.


AndrewVSutherland:
Looks good, will merge and push to dev+web shortly

PR #2846, opened 2019-03-22
Fix bug in column_counts

roed314:


AndrewVSutherland:
The travis failure occurred the first 30 seconds during setup before any tests were actually run, can you try triggering it again?

edgarcosta:
I hit restart

AndrewVSutherland:
@edgarcosta thanks, travis is happy now, merging.

PR #2853, opened 2019-03-22
RCS knowls for Dirichlet characters

jwj61:
Knowls are added and learnmore list updated.

At some point, the older knowls should be renamed to the rcs convention.


JohnCremona:
Looks good, just waiting for checks.

PR #2855, opened 2019-03-22
RCS knowls for Bianchi forms

JohnCremona:
Instead of displaying the entire content of dq.mf.bianchi.extent at beta.lmfdb.org/ModularForm/GL2/ImaginaryQuadratic/
this just shows the first sentence.  The whole knowl is still accessible via the completeness link.

This is just step 1, further work on RCS knowls for Bianchi will follow.


JohnCremona:
This PR now also adds a Relaibility knowl link to Bianchi pages.  After this, the RCS issue for BMFs is done.

Issue #1293, opened 2016-05-11
Incorrect digits in zeta zeros


jwj61:
From the issue tracker:

"Last digits of every imaginary part of Riemann zeta zeros are incorrect. For example, first zero ends with ....24766, but according to http://oeis.org/A058303, it should end with ...247027. Usually 2-4 last digits are incorrect throughout the whole list. In http://lmfdb.org/zeros/zeta/HowComputed it is said that the zeros are accurate to +-2^-102 ~ 30 digits after the decimal point, so I would suggest to mention this on the main site (http://lmfdb.org/zeros/zeta/), otherwise the list is misleading.

Best wishes,
Marcin"

Please respond to mszynisz@gmail.com

davidfarmer:
When we fix them, we should truncate (not round), and also exlpain
that we are doing that.

For the web page version can we just truncate the strings
as the table is populated?

JohnCremona:
The disclaimer text has been deployed, but the issue stays open until the incorrect digits are not displayed. 
@arbooker

arbooker:
This issue has long been known and there is no perfect resolution. The zeros are stored in binary but displayed in decimal, so there will always be issues with the tablemaker's dilemma, unless we are willing to truncate many places occasionally, which is a needless loss of precision.

My favo(u)red solution is to display round(t*10^30)/10^30. That guarantees that the error is at most 1 in the last place. @jwbober

arbooker:
Another option is to display round(t*10^31)/10^31. Then the approximation is off by no more than 3 in the last place, and it has the added feature that you could recover the exact internal representation from the decimal approximation. I think Odlyzko does something similar in his zeta zero tables.

jwbober:
The zeros should be printed so that when they are corrected rounded to 104 bits of absolute precision, the internal representation is recovered, which will be accurate to +/- 2^(-102). (Perhaps there are some digits in the internal representation that are always zero. I don't know. I'm sure that Dave chose 104 because it is a multiple of 8.) Due to double rounding, this probably means that the displayed decimal number is not necessarily accurate to 2^(-101).

In any case, I'm printing too many digits. An offending line (there is more than one) is at https://github.com/LMFDB/lmfdb/blob/master/lmfdb/zeros/zeta/platt_zeros.py#L78
```python
    mpmath.mp.prec = log(t1, 2) + 10 + 101  # We make sure that the working precision
                                            # is large enough. Note that we are adding
                                            # a little too much here, so when these
[...]

PR #2856, opened 2019-03-22
Adjust precision of zeta zero display

AndrewVSutherland:
This addresses issue #1293


AndrewVSutherland:
There is no easy way to test this locally due to the dependence on the sqllite datafiles stored in GCP buckets.  Andy Booker and I have tested using a local workaround.  As soon as Travis finishes I am going to merge and push to beta

PR #2858, opened 2019-03-22
Fix display precision for zeta zeros

AndrewVSutherland:
Addresses issue #1293


AndrewVSutherland:
There is no easy way to test this locally due to the need to access sqllite files stored in GCP buckets.  Andy Booker and I have tested this using a local hack/workaround -- the changes are very small and easy to undo or modify if there is a problem.  I am going to merge this as soon as travis is happy and merge it and push to beta.

PR #2865, opened 2019-03-22
RCS knowls for ec/Q

JohnCremona:
There is a new Reliability knowl.  The other two have only had minor edits.  Please also review the knowl content!


JohnCremona:
Checking up on the Travis failure now...

Issue #2873, opened 2019-03-23
Genus 2 curve missing HMF friend


AndrewVSutherland:
The genus 2 curve L-function

http://www.lmfdb.org/L/Genus2Curve/Q/3125/a/

and the HMF L-function

http://www.lmfdb.org/L/ModularForm/GL2/TotallyReal/2.2.5.1/holomorphic/2.2.5.1-125.1-a/0/0/ 

are the same, but the genus 2 curve and the HMF do not know they are friends.

We should fix this and search for any other missing HMF or BMF friends of genus 2 curves.

jvoight:
As we discussed at IAS, I think the time to be systematic is when we're finishing up our many author paper.  If I were to do this, I would go over each row of the giant table and make sure all matches are there--but it will be more efficient if we do this while writing, especially if any mathematical issues come up.  

It would be more reliable if there was a source-independent way (e.g., hash) on the L-functions in the database--or perhaps equivalently the relevant CMFs, HMFs, and BMFs--and these friends were generated automatically.  Then the humans could check the computations and hopefully we agree.  

I'm not sure this is worth holding up v1.1: the links to CMFs seem to be there, we never promised that all friends will be there, and there's no way we will forget to do this.

AndrewVSutherland:
I agree this is not worth holding up v1.1, but I believe we already have enough data to completely automate this (which I agree is absolutely necessary).  I'll take a stab at this today.  I don't think it will be hard to generate a list of potential friends that is provably complete, and I think this will be a short list (tens not hundreds).

AndrewVSutherland:
There is a related and perhaps more pressing issue.  The genus 2 curve

http://www.lmfdb.org/Genus2Curve/Q/12500/a/12500/1

knows its HMF friend, and the L-function lists both the genus 2 curve and the HMF as related objects

http://www.lmfdb.org/L/Genus2Curve/Q/12500/a/

because both are listed in lfunc_instances.  But If you click the HMF link on the L-function page, and then click the L-function link on the HMF page, you return to a different version of the same L-function

http://www.lmfdb.org/L/ModularForm/GL2/TotallyReal/2.2.5.1/holomorphic/2.2.5.1-500.1-a/0/0/

[...]

PR #2874, opened 2019-03-23
Updates to degree 2 and 4 L-function tables

AndrewVSutherland:
I added some missing links to weight 1 modular forms on the degree 2 L-function table, and added an example of a non-self-dual weight 1 modular form, as well as cleaning up some of the formatting.

To see the changes compare:

http://beta.lmfdb.org/L/degree2/
http://127.0.0.1:37777/L/degree2/

On the degree 4 L-function table the genus 2 curve that was listed has an imprimitive L-function (conductor 169), which seemed strange because we don't list any other imprimitive L-functions on this table  I changed it to the genus 2 curve of conductor 277 and also listed the corresponding paramodular form (which of course is not in the LMFDB yet, but the same applies to other objects on this page such as K3 surfaces and hecke characters).  I also added genus 2 curves with CM, RM, and QM and corresponding Hilbert and Bianchi modular forms (we list elliptic curves with and without CM on the degree 2 page, so this seemed reasonable).

To see the changes compare:

http://beta.lmfdb.org/L/degree4/
http://127.0.0.1:37777/L/degree4/

@davidfarmer Was the genus 2 curve of conductor 169 listed here intentionally?  I left it commented out in the file so it can be easily restored, but if we do that there are a lot of other imprimitive L-function we could add (that are different in interesting ways).


davidfarmer:
On the degree 4 page, looks good, except that the link from
      CM Hilbert cusp form over $\Q(\sqrt{5})$  (8th row, 2nd column of degree 4 page)
goes here:
     /ModularForm/GL2/TotallyReal/2.2.5.1/holomorphic/2.2.5.1-125.1-a/0/0/
but that link is broken.  I tried to find the correct link, but when I go to the
genus 2 curve that is in the same box, it does not have a Hilbert modular form friend.

On the degree 2, bottom row, the link to the weight 1 cusp form goes to a
Galois orbit of forms, not to the specific embedding that corresponds to the linked
L-function.  The specific form is:
    /ModularForm/GL2/Q/holomorphic/3997/1/cz/a/13/1/
Also on the bottom row, the Gamma factors are listed as "0,1", meaning two Gamma_R s.
But when you follow that link, the functional equation shows Gamma_C(s).  I do not
suggest changing it (but I am okay if you do).  But at some point we have to decide
how to deal with the ambiguity.   (It is a function of the underlying object.
For this particular example there is an associated Artin representation.  Once we
have it, its L-function naturally wants to be written as two Gamma_R s, right?)

Issues I saw that were not strictly related to this pull request:

There is a link to:
   http://localhost:37777/L/ModularForm/GL2/TotallyReal/2.2.5.1/holomorphic/2.2.5.1-31.1-a/0/0/
(that is the link labeled 3.67899).  When I follow that link, it does not show the E/Q(sqrt(5))
as a friend, even though it is a friend.

The terms CR, RM , and QM are used, but when you follow those links, the home page of
the curve does not indicate what is meant by that terminology (not in a way that a
nonexpert can understand).

AndrewVSutherland:
@davidfarmer I fixed the 2 specific issues you noticed, thanks!  I think the others should probably be addressed in a separate PR, but let me know if you disagree.

The fact that the genus 2 curve and HMF are not friends is issue #2873

davidfarmer:
I will merge after checks finish, and also submit those issues (and another I just noticed).

Issue #2451, opened 2018-05-16
Weil polynomial issues


jwj61:
From the feedback page from Stefano Marseglia (stefanom@math.su.se):

I think there is a problem with the search function when setting the initial coefficient to 0 (and eventually something else). It seems that the 0's are ignored.

I believe that the Weil polynomials with real roots are missing from the database.
For example the polynomial (p*x^2-1) does not correspond to an isogeny class of elliptic curves over F_p, but it's square defines an isogeny class of surfaces (over F_p), see (1971, Milne, Waterhouse - Abelian Varieties over Finite Fields, Theorem 8).
I think that polynomials of this form and their multiples are missing.

kedlaya:
For p = 2, I'm not seeing (1-2x^2)^2 in dimension 2 nor (1-2x^2)^4 in dimension 4.

I just checked my original code that generates Weil polynomials and it does pick up this polynomial; so the issue is probably with the code that filters for the Milne-Waterhouse condition.

PR #2875, opened 2019-03-23
Allow specifying file during rewrite

roed314:
We've recently run into some problems on legendre.mit.edu where the `rewrite` method fails due to lack of disk space in `/tmp`.  This PR makes it possible to pass a filename to `rewrite` and thus adds the ability to use a different disk.

### DIFF
--- a/lmfdb/backend/database.py
+++ b/lmfdb/backend/database.py
@@ -2331,7 +2331,7 @@ class PostgresTable(PostgresBase):
         # Sort and set self._out_of_order
         pass
 
-    def rewrite(self, func, query={}, resort=True, reindex=True, restat=True, tostr_func=None, commit=True, **kwds):
+    def rewrite(self, func, query={}, resort=True, reindex=True, restat=True, tostr_func=None, commit=True, searchfile=None, extrafile=None, **kwds):
         """
         This function can be used to edit some or all records in the table.
 
@@ -2359,8 +2359,20 @@ class PostgresTable(PostgresBase):
         # operate on the results, but then func would have to process the strings
         if tostr_func is None:
             tostr_func = copy_dumps
-        searchfile = tempfile.NamedTemporaryFile('w', delete=False)
-        extrafile = EmptyContext() if self.extra_table is None else tempfile.NamedTemporaryFile('w', delete=False)
+        if searchfile is None:
+            searchfile = tempfile.NamedTemporaryFile('w', delete=False)
+        elif os.path.exists(searchfile):
+            raise ValueError("Search file %s already exists" % searchfile)
+        else:
+            searchfile = open(searchfile, 'w')
+        if self.extra_table is None:
+            extrafile = EmptyContext()
+        elif extrafile is None:
+            extrafile = tempfile.NamedTemporaryFile('w', delete=False)
+        elif os.path.exists(extrafile):
+            raise ValueError("Extra file %s already exists" % extrafile)
+        else:
+            extrafile = open(extrafile, 'w')
         try:
             with searchfile:
                 with extrafile:
--- a/lmfdb/backend/database.py
+++ b/lmfdb/backend/database.py
@@ -4613,7 +4613,7 @@ SELECT table_name, row_estimate, total_bytes, index_bytes, toast_bytes,
 
         INPUT:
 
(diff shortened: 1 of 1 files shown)


AndrewVSutherland:
Testing this now.

AndrewVSutherland:
There seems to be a problem.  I created a table test with text search columns col1, col2 and text extra columns col3, col4, populated it with a million records and then wrote a function that sets col2 to col3 and col4 to col1.  I created a writable directory /scratch/temp and then tried to do a rewrite specifying /scratch/temp/a.txt as searchfile and /scratch/temp/b.txt as extrasfile.  I hit the following error:

```
sage: db.test.rewrite(myfunct,searchfile="/scratch/temp/a.txt",extrafile="/scratch/temp/b.txt")
INFO:test@2019-03-23 12:14:57,302: SELECT "id", "col1", "col2" FROM "test" ran in  1.49226498604s  [database.py:383]
INFO:test@2019-03-23 12:14:57,303: Replicate with db.test.analyze({}, ['id', u'col1', u'col2', u'col3', u'col4'], None, 0) [database.py:385]
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
<ipython-input-29-4b0383857e83> in <module>()
----> 1 db.test.rewrite(myfunct,searchfile="/scratch/temp/a.txt",extrafile="/scratch/temp/b.txt")

/home/drew/github/lmfdb/lmfdb/backend/database.pyc in rewrite(self, func, query, resort, reindex, restat, tostr_func, commit, searchfile, extrafile, **kwds)
   2392             self.log_db_change("rewrite", query=query, projection=projection)
   2393         finally:
-> 2394             searchfile.unlink(searchfile.name)
   2395             if self.extra_table is not None:
   2396                 extrafile.unlink(extrafile.name)

AttributeError: 'file' object has no attribute 'unlink'

```

roed314:
Oops, sorry about that.  I've updated the `finally` clause in a way that should fix the problem.

AndrewVSutherland:
Thanks, I'll retest as soon as I finish fixing Dirichlet characters

roed314:
Creating a `test` table caused the API to fail (every table needs an underscore for the api to work).  Should we add the underscore requirement to `create_table` or change the api to support table names without underscores?

AndrewVSutherland:
For the moment, let's add an underscore requirement (seems safer)

roed314:
I added an underscore requirement to `create_table`, but I didn't delete `test` so Travis will still fail.

AndrewVSutherland:
I deleted test, so travis should be good.

AndrewVSutherland:
So this seems to work, but I'm concerned about the performance of rewrite when there is an extras file.  I ran a test where I created a table with two short text columns, populated it with 100,000 rows and did a rewrite using a function that copies one column to the other, it took very little time (half a second), all as expected.

I then repeated the test, this time with an extras file with two additional columns, and the same rewrite is taking many minutes (you can see the files f.txt and g.txt in /scratch/temp being written to if you look now it is only about 1/4 of the way through).

AndrewVSutherland:
I did verify that create_table now requires underscores

roed314:
The reason that an extras table causes a dramatic slowdown is that the `rewrite` method uses `search` to get the input for the rewriting function, and search wasn't intended to return large numbers of rows with columns from the extras table (the whole point of the extras table was that these columns aren't used in search results).  The right solution is to change the `search` and `_search_iterator` methods to use a join instead of doing things in an ad-hoc way, but that will require some work.  I think it's better for me to focus on #2451 for the moment.

I've created #2881 to track the issues with slow rewrites, but I'm not going to be able to deal with it immediately.  In the mean time, I would suggest using copy_to and copy_from, and making the changes to the files outside of postgres.

AndrewVSutherland:
OK, I will merge this PR.

JohnCremona:
I was too late to vote but had a comment on the underscore requirement:  I think it would be useful to have a way to easily distinguish test tables from real data tables, and might have suggested "no underscore" as a way to identify the former.  But other conventions are possible, e.g. any table with "test" in the name (anywhere).  If agreed this convention should be documented somewhere.
(Note that the inventory system relies on data tables having names with underscores and uses the parts before and after the underscore to collect them into what used to be collections within one database.  It can / should / will ignore test tables.)

PR #2878, opened 2019-03-23
Fix to #2877

### LINKED ISSUE #2877: Dirchlet character search lists incorrect parity
As reported by Jörg Czeranski, the Dirichlet character search results page does not list the correct parities of the characters matching the search (it appears to always list odd, even for the trivial character!)

> Dear LMFDB developers,
> 
> I searched for Dirichlet characters with modulus 35, but all 24
> characters are shown with parity = odd:
> http://www.lmfdb.org/Character/Dirichlet/?start=0&count=12&modulus=35&parity=All&primitive=All&limit=25&refine=1
> 
> If I restrict the search to even characters, I get 12 characters as
> expected, but the result table still shows parity = odd:
> http://www.lmfdb.org/Character/Dirichlet/?start=0&count=24&modulus=35&parity=Even&primitive=All&limit=25&refine=1
> 
> Best regards
> Jörg Czeranski

AndrewVSutherland:
This fixes #2877, which was caused by doing an inappropriate boolean test on a string.  It also adds selectors for parity and primitivity to the front search page (they were already present on the refine results page), and fixes a number of minor issues in the search/browse pages.  To test, first compare

http://127.0.0.1:37777/Character/Dirichlet/?modulus=35
http://www.lmfdb.org/Character/Dirichlet/?modulus=35

Then goto the main search page and try searching on various combinations of inputs, including parity/primitive selections.  I recommend specifying some constrains on conductor/order/modulus in your searches -- unconstrained searches work but they are quite slow (this is perhaps a separate issue we should address, but not in this PR)


AndrewVSutherland:
We should expect travis to fail on the api test (which will go away as soon as I delete the table "test").  this has nothing to do with this PR, the character tests all pass.

AndrewVSutherland:
As expected, the only error travis hit was the expected one on the api test.

AndrewVSutherland:
Added fix to #2880, the trivial character now has an L-function.

JohnCremona:
Looks good.  One might add "(Riemann \eta-function)" after "L-function" in the Friends box, but I'll not wait for that.

Issue #2652, opened 2018-10-11
Duplicated related objects in Genus 2 curves


edgarcosta:
At the moment in related objects, we display the curves and the isogeny classes.
We should pick if we want to list the isogeny class or the curve in related objects.

http://prod.lmfdb.xyz/Genus2Curve/Q/2916/b/11664/1
vs
http://beta.lmfdb.org/Genus2Curve/Q/2916/b/11664/1

<strike>Picking the two to stay is related to #2651</strike>

AndrewVSutherland:
We should continue to show the curves (there are rational maps from the Jacobian to each of these elliptic curves, but this is not true of other elliptic curves in the isogeny class).  I don't think we need to list the isogeny classes in addition.

AndrewVSutherland:
To test, compare:

http://beta.lmfdb.org/Genus2Curve/Q/2916/b/11664/1
http://127.0.0.1:37777/Genus2Curve/Q/2916/b/11664/1

and note that with the update only the elliptic curve friends are listed in related objects (rather than both the elliptic curve and its isogeny class) as was the case pre-postgres:

http://prod.lmfdb.xyz/Genus2Curve/Q/2916/b/11664/1

edgarcosta:
Closed via #2879

PR #2879, opened 2019-03-24
Fix to #2652

AndrewVSutherland:
Fixes issue #2652 with more careful filtering of duplicate related objects, and being careful to add them in the correct order (when relevant, we should prefer links to specific elliptic curves over links to their isogeny classes since the former conveys more information).

### DIFF
--- a/lmfdb/genus2_curves/web_g2c.py
+++ b/lmfdb/genus2_curves/web_g2c.py
@@ -412,7 +412,7 @@ def lfunction_friend_from_url(url):
 def add_friend(friends,friend):
     for oldfriend in friends:
         if oldfriend[0] == friend[0] or oldfriend[1] in friend[1] or friend[1] in oldfriend[1]:
-            print "not adding friend", friend,  "already friends with "oldfriend
+            print "not adding friend", friend,  "already friends with ",oldfriend
             return
     print "adding friend", friend
     friends.append(friend)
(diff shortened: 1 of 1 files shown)


JohnCremona:
While I agree that we need not link to both elliptic curves and their isogeny class I am puzzled as to how any one curve is preferred or carries more information.  Is there a claim that the curves being linked to are the ones which minimize the degree of the isogeny from the Jacobian to the product of curves?

AndrewVSutherland:
@JohnCremona In order to get a principally polarized abelian surface the elliptic curves need to be compatible (e.g. have isomorphic N-torsion Galois modules) in a way that is not invariant under isogenies of elliptic curves.  There is actually more information we could give here (e.g. the value of N, which gives rise to an N-N isogeny from the Jacobian to the product whose kernel is isomorphic to the N-torsion of both the elliptic curves (as a Galois module)).

JohnCremona:
Thanks for the additional explanation.  I see now that "isogenous" is being used to mean "isogenous with maximal isotropic kernel" (if that is equivalent, which I think it might be).  I suggest that the knowl g2c.decomposition includes this, or that there is a new knowl linked from "up to isogeny",   to explain that only certain kinds of isogeny are considered.

JohnCremona:
I'll leave this open a while until you respond in case you want to add a new knowl to the template; otherwise it is merge-worthy.

AndrewVSutherland:
I won't have time to work on this today and the change would be in a 
different file in any case, so I would go ahead and merge.  In addition 
to adding a knowl I'd like to see if we can add more information (e.g. 
the degree of the isogeny), but I need to check with Jeroen, so this 
should really be a separate PR (I'll add an issue for this later).

Issue #1820, opened 2016-07-18
Fix input validation and error handling on Maass form search pages


AndrewVSutherland:
On the main Maass form page http://www.lmfdb.org/ModularForm/GL2/Q/Maass/ all of the following lead to internal server errors:
- Enter anything into the level input box other than a single integer or a pair of integers separated by two periods (e.g. 1-10 does not work, even though it works in almost any other place where we ask for a range of positive integers)
- Enter anything into R in range other than a single float or pair of floats separated by two periods.
- Enter anything for the weight other than a single integer (which is allowed to be negative).
- Enter anything for the character other than a single integer (which is allowed to be negative).  In particular, you cannot enter the label of a Dirichlet character (you must enter its Conrey index).

The first three fields should be changed to our standard search parsing functions (e.g. parse_ints) and display meaningful errors to the user when the input is invalid (we may need to add a parse_reals function to handle ranges of reals).

The "Character" input box is confusing in that it is really asking for a Conrey index, you cannot type in a character label.  If you leave the level blank and type in 3 for the character, the search results include forms both for the character chi_8(3,.) and chi_13(3,.).  I cannot imagine a scenario in which that is a search one would want to do (but I'd be interested if there is one!).

I think either the label of the "Character" input should be changed to "Conrey index" (or "Character index"), or it should also accept a character label (e.g. 13.3) and check compatibility with the level entered (if any).

jenpaulhus:
No Maass!

AndrewVSutherland:
@jenpaulhus Maass!  FYI, I just threw all the input validation we discussed into mwf_main.py right at the entry point and flash errors there (a total hack, but it works).

jenpaulhus:
Thanks for the update, good to know we weren't so far off!

PR #2882, opened 2019-03-24
Fix to #1820

AndrewVSutherland:
Adds some input validation to the Maass forms search page in an effort to address #1820.  Not perfect, but a lot better than what was there (it is now much harder to generate a server error).  To test go to 

http://127.0.0.1:37777/ModularForm/GL2/Q/Maass/

and try various combination of input (especially invalid input)


davidfarmer:
I tried various invalid values, and it worked as expected.

The only valid weight for now is 0, and it will only ever be 0 or 1,
but it is fine to leave as-is.  Once we have more extensive data, we can
rethink this interface.

Merging now.

PR #2890, opened 2019-03-26
New contributor, and weight 1 correction

davidfarmer:
Added a new contributor from the IAS workshop, adjusted some contributor info,
and omitted an incorrect statement from
/L/degree2/
The last phrase on that page was incorrect:  only in the weight 1 case is a Maass form
with eigenvalue 1/4 holomorphic.  But it is arithmetic for either weight 0 or weight 1.


AndrewVSutherland:
Looks good, will merge as soon as Travis finishes.

PR #2893, opened 2019-03-27
Create table like

roed314:
This utility function was a feature request last week.  To test:
```
sage: db.create_table_like('av_test', 'av_fqisog')
sage: db.drop_table('av_test')
```

I also fixed a bug in `db.create_table` resulting from the change of default from jsonb to arrays.

### DIFF
--- a/lmfdb/backend/database.py
+++ b/lmfdb/backend/database.py
@@ -901,6 +901,7 @@ class PostgresTable(PostgresBase):
         """
         Initialize the sorting attributes from a list of columns or pairs (col, direction)
         """
+        self._sort_orig = sort
         self._sort_keys = set([])
         if sort:
             for col in sort:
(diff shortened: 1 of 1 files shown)


roed314:
Test failures are due to data changes in abelian varieties.  Once #2895 is merged I'll pull the changes into here and the tests can be rerun.

roed314:

roed314:
The test failure seems to be us being blocked by `magma_free` again....

AndrewVSutherland:
I like it.  One thing I noticed while testing which is not directly relevant to this PR is that after you drop a table the db object still knows about it until you restart sage -- dtopped tables should be removed from the ephemeral state of the db object as well being removed from the database unless there is a good reason to keep the state around?

AndrewVSutherland:
I'll leave this open for the moment in case you want to add a commit (but its fine if you don't).

roed314:
I've added a `delattr` at the end of `drop_table`.

AndrewVSutherland:
@roed314 thanks, merging.

PR #2894, opened 2019-03-27
Copy to select header

roed314:
These changes make it easier to use the `_copy_to_select` function to produce files that can be read in by the `copy_from` and `reload` methods.  They are not used in rendering the website.

To test, try something like
```
sage: from lmfdb.backend.database import db, SQL
sage: db._copy_to_select(SQL("SELECT label, p_rank FROM av_fqisog WHERE g=2 AND q=2"), "/scratch/importing/select_test.txt", header = "label:p_rank\ntext:smallint\n\n", sep=":")
```


roed314:
Test failures are due to data changes in abelian varieties. Once #2895 is merged I'll pull the changes into here and the tests can be rerun.

roed314:

AndrewVSutherland:
Looks good, but it would be nice if it gave you confirmation that something actually happened (e.g. "wrote x records to file blah", or even "created file blah").

roed314:
I added a print statement, together with the option to silence it (since this method is used twice where the calling function prints something).

PR #2895, opened 2019-03-27
Fixed abvar to no longer use oldstats, add reliability knowl

### LINKED ISSUE #2451: Weil polynomial issues

roed314:
To see the difference visit http://localhost:37777/Variety/Abelian/Fq/ and note that there is a Reliability link in the Learnmore box, and that there are 35 isogeny classes for g=q=2 rather than 34.

Once this is merged, I think we can push to web and resolve #2451.

### DIFF
--- a/lmfdb/abvar/fq/main.py
+++ b/lmfdb/abvar/fq/main.py
@@ -4,6 +4,7 @@ import ast, re, StringIO, time
 
 from flask import flash, render_template, url_for, request, redirect, send_file
 from markupsafe import Markup
+from collections import defaultdict
 from sage.rings.all import PolynomialRing, ZZ
 
 from lmfdb import db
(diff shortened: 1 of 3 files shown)


roed314:
The test failures are due to new and improved principal polarization and Jacobian data.  I'll update the tests....

roed314:
Tests pass!  Once this is merged into master I'll merge it into the other open PRs which have tests failing.

PR #2892, opened 2019-03-26
Fixes for finishing renaming

roed314:
I've fixed a few bugs in the renaming process, and made the code changes required for the two renames currently in progress.

To test, wait until #2891 is pushed to web and then try to finish the rename process for [dq.character.dirichlet.source](http://localhost:37777/knowledge/edit/dq.character.dirichlet.source) and [content.how-to-cite](http://localhost:37777/knowledge/edit/content.how-to-cite).

PR #2898, opened 2019-03-27
Rename cmf.hecke-orbit to cmf.hecke_orbit

roed314:
Mainly for testing #2892.

### DIFF
--- a/lmfdb/classical_modular_forms/templates/cmf_space.html
+++ b/lmfdb/classical_modular_forms/templates/cmf_space.html
@@ -171,7 +171,7 @@ function show_qexp(qstyle) {
 </div>
 {% endif %}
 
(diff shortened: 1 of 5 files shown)


roed314:

roed314:
I merged in master, so this should be less scary to review now.

AndrewVSutherland:
Looks good, but I'm slightly confused by the fact that the knowl being renamed still has knowl references to it (even though it has no code references to it).  Is the assumption that knowls won't be updated until after the code has changed?  I guess this makes sense in that it delays all changes until something irrevocable happens (updated code gets merged), but I just want to confirm that this is intentional.

In any case, the answer to my question is irrelavent to this PR, so I will merge it.

roed314:
Yeah, the current workflow is that the Finish Renaming step updates the other knowls in the database.

Issue #2883, opened 2019-03-24
Colors on Dir Char and EC Isogeny Class table


davidfarmer:
There are three colors in the cells of Dirichlet character tables, such as

http://beta.lmfdb.org/Character/Dirichlet/?modbrowse=21-40

The dark color is too dark for the numbers in those cells.  (I realize it is
the same as occurs in the main headings on the sidebar, but in the sidebar
the letters are bold.)

I suggest replacing the dark background by what is currently the light
background, and making the light background white.  The very slight
parity difference indicated by the light and white background is of
minimal value (to me), so just two background colors in that table
should be fine.

Another shortcoming with that table is that if you show all characters
(meaning, both the primitive and imprimitive boxes are checked)
then it is not easy to tell which are which.  When you hover over the
word imprimitive, then that background is a reasonable way to do it.
But that disappears when you stop hovering.  I would be okay with that
background appearing always.  The hovering over Primitive is not
useful, and it can be eliminated if the imprimitive always have a background.

It may seem odd to highlight the imprimitive, because those are less
interesting.  But they are hidden by default and only appear when the
user asks for them, so it makes sense to highlight what the user asked for
 (which is what I just did, which led me to notice these issues).

davidfarmer:
The dark color also occurs on individual character pages:

http://beta.lmfdb.org/Character/Dirichlet/27/4

AndrewVSutherland:
Given that we are changing the color scheme in 1.1, it makes sense to address this now (and it should not be hard to do).  I agree the colors are too dark.  We should use the lighter background color, as on http://beta.lmfdb.xyz/ModularForm/GL2/Q/holomorphic/?search_type=SpaceDimensions, for example.

JohnCremona:
To me all those colours on the character pages look fine.  What definitely *is* too dark is the shading (background colour) of on line in a table such as http://beta.lmfdb.org/EllipticCurve/Q/11/a/ (to "highlight" the optimal curve).  The colour here is 555555 and is much too dark, making the text unreadable. To me.

AndrewVSutherland:
Wow, that is a bad.  Let's include fixing it in this issue (even if there are two PRs, it's really the same issue, with the color scheme change we need to take a close look at all the pages and fix any problems we find).

JohnCremona:
Agreed -- the issue really is "some of the components for this colour scheme needs adjusting".
[...]

PR #2899, opened 2019-03-27
Tweaking some colors

roed314:
This PR resolves the color issues pointed out in #2883.

### DIFF
--- a/lmfdb/abvar/fq/main.py
+++ b/lmfdb/abvar/fq/main.py
@@ -4,6 +4,7 @@ import ast, re, StringIO, time
 
 from flask import flash, render_template, url_for, request, redirect, send_file
 from markupsafe import Markup
+from collections import defaultdict
 from sage.rings.all import PolynomialRing, ZZ
 
 from lmfdb import db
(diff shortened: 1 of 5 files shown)


roed314:

JohnCremona:
I am happy with the change to the isogeny class display, and the characters look OK to me but @davidfarmer should look too.

davidfarmer:
The checkerboard looks better, and I can see the difference between the black and other
color when primitive and imprimitive are both shown.  However, the cost is that we typically
do not use black when something is clickable.

I'll let someone else merge after the pull request this depends on is merged.

We still need to rethink the way primitive/ and imprimitive are highlighted.  The current
interface is misguided.  To highlight/imprimitive, you have to hover while your eyes are
looking elsewhere.  And then the highlighting disappears if you then move to click
on something that was highlighted.

I don't know of other cases where such toggling is an option.  Do we need that elsewhere?
Or should it be eliminated here?  Or are Dirichlet characters special?

JohnCremona:
Since #2895 has already been merged I am merging this.

PR #2901, opened 2019-03-28
Fix jumps in link suggestion mode

### LINKED ISSUE #2832: knowl editing link suggestions
When using the Link Suggestions in a large knowl with many suggestions (and on a laptop screen), hovering over the word "insert" jumps the focus back to the edit frame in order to highlight the suggested insertion point -- which makes it impossible to click on "insert" since by then it has disappeared off the bottom of the screen.
A possible alternative to highlighting the insertion point would be to show a few words before and after on the same line as each suggestion, to show the context.  That should be enough for people to be able to decide whether to insert the suggested link.

roed314:
Resolves #2832.  To test, edit a knowl with lots of link suggestions.  The layout of the suggestions has changed, and there is now a separate link to highlight the text rather than have it be on the hover.

### DIFF
--- a/lmfdb/knowledge/templates/knowl-edit.html
+++ b/lmfdb/knowledge/templates/knowl-edit.html
@@ -190,7 +190,7 @@ var all_defines = {
 };
 
 $("body").on("click", ".insert_klink", insert_klink);
-$("body").on("mouseenter", ".insert_klink", hover_klink);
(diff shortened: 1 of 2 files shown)


JohnCremona:
I tried this and it works well.  There are some cases (e.g. the word "isogeny") where the context manages to extend to several lines but that is OK.  Good!

PR #2902, opened 2019-03-28
Knowl bread

### LINKED ISSUE #2831: Knowl breadcrumbs
The breadcrumbs for a knowl page such as http://beta.lmfdb.org/knowledge/show/g2c.discriminant  looks like 
Knowledge -> Discriminant...
It would be nice to have an intermediate breadcrumb, in this case 'g2c',  which linked to the same page as "up".

roed314:
Resolves #2831.  See any knowl homepage to test.

### DIFF
--- a/lmfdb/abvar/fq/main.py
+++ b/lmfdb/abvar/fq/main.py
@@ -4,6 +4,7 @@ import ast, re, StringIO, time
 
 from flask import flash, render_template, url_for, request, redirect, send_file
 from markupsafe import Markup
+from collections import defaultdict
 from sage.rings.all import PolynomialRing, ZZ
 
 from lmfdb import db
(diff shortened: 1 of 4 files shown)


JohnCremona:
One of the changed files implements this, the other two changed files do different things.  This makes it harder to review, since I am not sure what the side-effects of the other two changes will be.
Meanwhile, would it be possible also to have a view such as http://localhost:37777/knowledge/?category=ec  to have the ec in the breadcrumbs?  That seems logical, and currently the category name is highlighted somehow to I expect that this will not be hard.

roed314:
I've added the bread to the index page when relevant.

I'm confused about the other two files.  That change was already merged in #2891, so I don't know why it's still showing up here when comparing against master....

JohnCremona:
Thanks, I'll look at the new version.  Don't worry about the other two files -- as you say, github should probably not be showing them.  Perhaps you did a forced push so that the most recent common ancestor is not what it logically is.  I'll ignore that.

JohnCremona:
Looks good: to be merged once tests have passed.

roed314:
I do remember doing a forced push since I pushed the wrong branch somewhere.  That seems like a likely explanation.

PR #2903, opened 2019-03-28
Backup annotations

### LINKED ISSUE #2802: Annotations for EC isogeny classes
Elkies asked about implemented top/bottom annotations for isogeny classes. This is twofold: they should display on the isogeny class web page, but ideally also on the page for each curve in the isogeny class.

roed314:
Resolves #2802.  To test, create an annotation for an elliptic curve isogeny class and make sure that it displays on the isogeny class and on every curve in the class.

Note that if you're logged in, clicking "edit" on a curve homepage should take you to the edit page for the curve knowl, not the isogeny class knowl which is being displayed.  This might be a bit confusing, but I think it's worse to not be able to directly edit the curve annotation if there's an isogeny class annotation present.

### DIFF
--- a/lmfdb/abvar/fq/main.py
+++ b/lmfdb/abvar/fq/main.py
@@ -4,6 +4,7 @@ import ast, re, StringIO, time
 
 from flask import flash, render_template, url_for, request, redirect, send_file
 from markupsafe import Markup
+from collections import defaultdict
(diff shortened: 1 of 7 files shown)


AndrewVSutherland:
Nice!  I like that the curve knowl supersedes the isogeny class knowl and then reverts to the isogeny class knowl if you delete the curve knowl.  Merging.

PR #2905, opened 2019-03-28
Add barrier to dropping production tables

roed314:
Make it harder to drop important tables.  I made 46 tables important (basically, whether they're used in production).


roed314:
I've also add confirmation messages for dropping columns and tables.

AndrewVSutherland:
Looks good, I like the interface.  Merging.

PR #2906, opened 2019-03-28
EC/Q labels 

JohnCremona:
1. allows Cremona labels (of curves or isogeny classes) to be entered into the search box without any prefix.  Try entering 11a1 or 11a into http://localhost:37777/EllipticCurve/Q/
2. adjusts title of an elliptic curve's home page depending on the sort of label used to get there.  Compare http://localhost:37777/EllipticCurve/Q/11a1  (with no rewriting of the URL) and http://localhost:37777/EllipticCurve/Q/11.a2 (whose URL is renamed to http://localhost:37777/EllipticCurve/Q/11/a/2)
3. Same as 2 but for isogeny classes -- to be added in an extra commit, so hold on.

### DIFF
--- a/lmfdb/elliptic_curves/elliptic_curve.py
+++ b/lmfdb/elliptic_curves/elliptic_curve.py
@@ -264,8 +264,10 @@ def elliptic_curve_search(info, query):
         # fails on 990h3
         query['number'] = 1
 
-    info['curve_url'] = lambda dbc: url_for(".by_triple_label", conductor=dbc['conductor'], iso_label=split_lmfdb_label(dbc['lmfdb_iso'])[1], number=dbc['lmfdb_number'])
-    info['iso_url'] = lambda dbc: url_for(".by_double_iso_label", conductor=dbc['conductor'], iso_label=split_lmfdb_label(dbc['lmfdb_iso'])[1])
+    info['curve_url_LMFDB'] = lambda dbc: url_for(".by_triple_label", conductor=dbc['conductor'], iso_label=split_lmfdb_label(dbc['lmfdb_iso'])[1], number=dbc['lmfdb_number'])
+    info['iso_url_LMFDB'] = lambda dbc: url_for(".by_double_iso_label", conductor=dbc['conductor'], iso_label=split_lmfdb_label(dbc['lmfdb_iso'])[1])
+    info['curve_url_Cremona'] = lambda dbc: url_for(".by_ec_label", label=dbc['label'])
+    info['iso_url_Cremona'] = lambda dbc: url_for(".by_ec_label", label=dbc['iso'])
 
 ##########################
 #  Specific curve pages
--- a/lmfdb/elliptic_curves/elliptic_curve.py
+++ b/lmfdb/elliptic_curves/elliptic_curve.py
@@ -84,7 +84,6 @@ def rational_elliptic_curves(err_args=None):
         'counts': counts,
         'stats_url': url_for(".statistics")
     }
-    #credit = 'John Cremona and Andrew Sutherland'
     t = 'Elliptic Curves over $\Q$'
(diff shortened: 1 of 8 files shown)


JohnCremona:
OK so 3 above is now also implemented.  
Queries for reviewer: 
(1) I left the breadcrumbs alone so that going to http://localhost:37777/EllipticCurve/Q/2340c/ where the LMFDB label is 2340.g, the breadcrumbs are --> Q --> 2340 --> g; similarly for curves.  I can change that too if it will look better.
(2) In a search result page such as http://localhost:37777/EllipticCurve/Q/?conductor=101-200 the LMFDB curve labels are links but the Cremona labels are not; should they be?  And the C-labels for isogeny classes are no in the table at all: should they be?

JohnCremona:
NB ignore the X next to commit f4f6262: Travis cancelled the pending request when the new commit came in.  The latest one 5a208f1 is waiting to run.

AndrewVSutherland:
@JohnCremona Please do add the Cremona labels of the isogeny labels to the table -- this cause some confusion for us here at IAS just yesterday!  And I like the idea of making the Cremona labels links.

JohnCremona:
> @JohnCremona Please do add the Cremona labels of the isogeny labels to the table -- this cause some confusion for us here at IAS just yesterday! And I like the idea of making the Cremona labels links.

Will do both -- coming up.

JohnCremona:
OK: to see the latest changes:  http://localhost:37777/EllipticCurve/Q/?conductor=101-200
Note the layout of the table header, with both curve columns followed by both isogeny class columns.  I think this is the right order for the columns.  I also tweaked the label knowls slightly.

AndrewVSutherland:
I like this a lot!  But I'm going to resist the urge to merge it immediately so I can ask for one more feature:
it would be nice to have Cremona label curve links on isogeny class pages like 

http://localhost:37777/EllipticCurve/Q/102c

and also on

http://localhost:37777/EllipticCurve/Q/102/b

Regarding the breadcrumbs, I think it's fine to leave them as is, but if you were motivated to change them, that would also be fine by me.

JohnCremona:
> I like this a lot! But I'm going to resist the urge to merge it immediately so I can ask for one more feature:
> it would be nice to have Cremona label curve links on isogeny class pages like
> 
> http://localhost:37777/EllipticCurve/Q/102c
> 
> and also on
> 
> http://localhost:37777/EllipticCurve/Q/102/b
> 

Will do.  Maybe even the isogeny graph labels, but the graphs are going to be enhanced anyway.

> Regarding the breadcrumbs, I think it's fine to leave them as is, but if you were motivated to change them, that would also be fine by me.

JohnCremona:
Here's what the latest commit adds:  
- all links to related objects now use the same labelling scheme as the page 
- graph vertices are lager and now contain the letter part of the label as well as the number, using whichever labelling scheme is in operation.  The isogeny matrix is permuted if necessary accordingly.
- the knowls ec.q.cremona_label and ec.q.lmfdb_label should be cut after this, by deleting the second paragrph in each case.
- fixed a bug whereby class 990h showed the wrong curve as optimal (but the curve pages were correct).
- small changes to main search page examples

Test cases to try to compare isogeny graph labels:  http://localhost:37778/EllipticCurve/Q/235200/bbo/ (8 curves and 3-letter class code), and http://localhost:37777/EllipticCurve/Q/15a/  and http://localhost:37777/EllipticCurve/Q/14a/.  Please check the labelling of the graph and compatibility with the matrix entries.

We should perhaps add a test for the 990h special case since it was being handled OK by the curve code but not by the isogeny class code (where it was being applied to 990.h which is not the same as 990h!!!)

AndrewVSutherland:
Looks good!

JohnCremona:
Thanks

PR #2907, opened 2019-03-28
removed use of oldstats from elliptic_curves (*not* ecnf yet)

JohnCremona:
To test this go to http://localhost:37777/EllipticCurve/Q/stats and http://beta.lmfdb.org/EllipticCurve/Q/stats in adjacent tabs so you can switch between them easily and note that the nothing changes.

Also do "git grep -p oldstat | grep elliptic" and see nothing.

The update_stats() script in scripts/elliptic_curves/import_ec_data.py can be run with recount=False to see what is stored, and with recount=True to force several recounts (which only takes a few seconds).

Question for someone who understands stats and counts (e.g. @roed314 ) :  the update_stats() function, when run  with recount=True, uses db.ec_curves.stats._slow_count() to force a new count which is then stored.  That is the point: we need to be able to update the counts when data changes.  But it also uses db.ec_curves.stats._slow_max() to find the max of 4 quantities, which are retrieved in the website code using db.ec_curves.max(); but I think that _slow_max() does *not* store the max value, and the reason why using plain max() in the main code works is that something else has happened, perhaps an add_stats()?   However I could not use add_stats() in the update_stats() script since it (apparently) does not have a forced recompute option.  I think that makes two feature-requests, and without them it may be true that after a call to  db.ec_curves.stats._clear_stats_counts(), the stats page will crash.

### DIFF
--- a/lmfdb/elliptic_curves/ec_stats.py
+++ b/lmfdb/elliptic_curves/ec_stats.py
@@ -46,18 +46,19 @@ class ECstats(object):
         if self._counts:
             return
         logger.debug("Computing elliptic curve counts...")
-        ecdbstats = db.ec_curves.stats
+        ecdb = db.ec_curves
         counts = {}
-        rankstats = ecdbstats.get_oldstat('rank')
-        ncurves = rankstats['total']
+        ncurves = ecdb.count()
+        nclasses = ecdb.count({'number':1})
         counts['ncurves']  = ncurves
         counts['ncurves_c'] = comma(ncurves)
-        nclasses = ecdbstats.get_oldstat('class/rank')['total']
         counts['nclasses'] = nclasses
         counts['nclasses_c'] = comma(nclasses)
-        max_N = ecdbstats.get_oldstat('conductor')['max']
+
+        max_N = ecdb.max('conductor')
+
         # round up to nearest multiple of 1000
-        max_N = 1000*((max_N/1000)+1)
+        max_N = 1000*int((max_N/1000)+1)
         # NB while we only have the Cremona database, the upper bound
         # will always be a multiple of 1000, but it looks funny to
         # show the maximum condictor as something like 399998; there
@@ -66,7 +67,7 @@ class ECstats(object):
 
         counts['max_N'] = max_N
         counts['max_N_c'] = comma(max_N)
-        counts['max_rank'] = int(rankstats['max'])
+        counts['max_rank'] = ecdb.max('rank')
         self._counts  = counts
         logger.debug("... finished computing elliptic curve counts.")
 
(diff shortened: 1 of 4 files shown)


roed314:
I will take a look shortly.  Note that I started switching ecnf stats last night; I should have a PR today for that.

JohnCremona:
> I will take a look shortly. Note that I started switching ecnf stats last night; I should have a PR today for that.

Thanks.  If anything is not clear please ask.  I have notes on what stats are stored.  There is a list of torsion orders and structures (easy) and triples (#curves, #classes, max(conductor norm)) for (i) each individual field, (ii) subtotal over all fields of a given signature, (iii) subtotal over all fields of a given degree, (iv) grand total.  For the display one also needs a list of degrees, a list of signatures per degree and a list of fields per signature.  The update_stats() script in scripts/ecnf/import_ecnf_data stores the stuff now in oldstats, and I would like to see that script updated at the same time  so that it can easily be rerun when data changes.

The HMF stats will be very similar to ecnf.

AndrewVSutherland:
@roed314 Did you review this?  Is it ready to merge?  I'm happy to review the UI functionality but it would be good for you to take a look at the code changes.

roed314:
I'm working on stats now.  Overall, things look fine.  I'm adding some features that will speed up some of what @JohnCremona has done, so I'm planning an addendum.  I can either do that as a separate PR or as a pull request to this one.

AndrewVSutherland:
Either is fine by me, I'm not in a rush to merge I just wanted to make sure there weren't PR's sitting around waiting for someone to look at them.

JohnCremona:
@roed314 do you mean adding new backend features (as I suggested somewhere else)? Or making changes to what I wrote which use existing features I could have used but did not?  Just for info.

roed314:
Adding new backend features; I don't think that there were any existing features that you should have used instead.

roed314:
I should go to sleep soon, but just to answer your question about max:

You're right that `_slow_max` does not store the result, but `max` does.  So as long as you're not calling `_slow_max` directly, that's why you're seeing statistics being stored.

It's true that `add_stats` doesn't have a forced recompute.  Instead, when the data is changed by a copy_from or reload, it calls `refresh_stats` which recomputes all of the statistics that have been added.  So there is no need to manually run an update script to redo the statistics computation.  I added some documentation on this in #2910; let me know if it's still not clear.

roed314:
Some comments on the code:

 * `1000*int((max_N/1000)+1)`: if the maximum conductor is ever a multiple of 1000 after some future update, won't this be off?  I think it should be `1000*int(((max_N-1)/1000)+1)`
 * You write `ncu = ecdb.count({'rank':r})`, which will work, but you can also use `ecdb.stats.column_counts('rank')` (and `ecdb.stats.column_counts('rank', {'number':1})`) to get the results into a dictionary with one database lookup.  Similarly for the torsion structure and sha.
 * As I mentioned in the previous comment, your changes to `import_ec_data` aren't necessary.  The updating stats should happen automatically with `reload` or `copy_from`.

roed314:
Anyway, I'm happy with these changes and Drew should feel free to merge them.  The tweaks I mentioned will just affect performance slightly, and have no bearing on correctness.

JohnCremona:
> Some comments on the code:
> 
> * `1000*int((max_N/1000)+1)`: if the maximum conductor is ever a multiple of 1000 after some future update, won't this be off?  I think it should be `1000*int(((max_N-1)/1000)+1)`

Multiples of 1000 are never conductors, so this does not arise.  I thought the comment said that.

> * You write `ncu = ecdb.count({'rank':r})`, which will work, but you can also use `ecdb.stats.column_counts('rank')` (and `ecdb.stats.column_counts('rank', {'number':1})`) to get the results into a dictionary with one database lookup.  Similarly for the torsion structure and sha.

Ah yes, that is something you implemented last week and I forgot.

> * As I mentioned in the previous comment, your changes to `import_ec_data` aren't necessary.  The updating stats should happen automatically with `reload` or `copy_from`.

But not when new data is uploaded!  If I were to add (say) curves of conductor 400000-401000 (as until 2016 I was regularly adding, in batches of 10000 usually) then after adding all those rows surely something needs to be done to trigger the stats update?

AndrewVSutherland:
Looks good!  One thing that would be nice to add in the future is to make the counts in the stats tables clickable (taking you to search results), as on genus 2 curves and cmfs.  But the current PR seems to fully replicate the functionality of the old stats, which was the immediate goal.  Merging.

PR #2910, opened 2019-03-29
Statistics documentation and new features

roed314:
In addition to adding documentation, I added a couple new methods to the statistics objects:

 * `numstats` and `add_numstats`, used in `ec_nfcurves` to find the maximum conductor norm, grouped by degree, signature or field label.
* `status`, used to display the contents of the statistics.  Try calling it on `ec_nfcurves` and on `mf_newforms` (the newform output isn't ideal, but improving it will be tricky).

### DIFF
--- a/lmfdb/backend/database.py
+++ b/lmfdb/backend/database.py
@@ -3513,9 +3513,9 @@ class PostgresStatsTable(PostgresBase):
 
         sage: db.ec_nfcurves.stats.quick_count({'torsion_structure': [2,4]})
         5100
-        sage: torsion_structures = db.ec_nfcurves.stats.column_counts(['torsion_structure'])
+        sage: torsion_structures = db.ec_nfcurves.stats.column_counts('torsion_structure')
         sage: torsion_structures[4,4]
-        14
+        14L
 
     Adding statistics on norm_conductor, grouped by signature::
 
--- a/lmfdb/backend/database.py
+++ b/lmfdb/backend/database.py
@@ -36,7 +36,7 @@ from psycopg2.extras import execute_values
 from sage.all import cartesian_product_iterator, binomial
 
 from lmfdb.backend.encoding import setup_connection, Json, copy_dumps, numeric_converter
-from lmfdb.utils import KeyedDefaultDict
+from lmfdb.utils import KeyedDefaultDict, make_tuple
 from lmfdb.logger import make_logger
 from lmfdb.typed_data.artin_types import Dokchitser_ArtinRepresentation, Dokchitser_NumberFieldGaloisGroup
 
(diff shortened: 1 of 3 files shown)


roed314:
I think I understand where the test failure is coming from (I think it's due to changing ccols and cvals from None to []); I'll fix it when I wake up.

JohnCremona:
These functions look extremely useful -- and thanks @roed314 for all the documentation!

lmfdb/backend/database.py
```diff
@@ -3432,13 +3436,178 @@ def show_check(name, check, typ):
...
+    Adding statistics on torsion structure::
+
+        sage: db.ec_nfcurves.stats.add_stats(['torsion_structure'])
+
+    This make counts available::
+
+        sage: db.ec_nfcurves.stats.quick_count({'torsion_structure': [2,4]})
+        5100
+        sage: torsion_structures = db.ec_nfcurves.stats.column_counts(['torsion_structure'])
+        sage: torsion_structures[4,4]
```
JohnCremona: Should be
torsion_structures[((4,4),)]
(and the result is 14L not 14)

JohnCremona:
I vote to merge this when the test failure is sorted, but someone who understands the backend better should also look at the new code.

JohnCremona:
Something weird: the torsion_structures example you just fixed (thanks) shows only 4 curves with trivial torsion.  And these two
>sage: db.ec_nfcurves.count({'torsion_order':1})
>193336
>sage: db.ec_nfcurves.count({'torsion_structure':[]})
>193332
are inconsistent (searching on the beta website gives the same, different numbers) and differ by 4 which is suspicious.  

Interesting that this data problem was shown up by testing this PR -- it is not actually relevant to the PR of course, but I want to sort it out.

roed314:
Yeah, there are four curves where the torsion structure is an empty dictionary rather than an empty list:
```
list(db.ec_nfcurves.search({'torsion_order':1,'torsion_structure':{'$ne':[]}},['label','torsion_structure']))
[{'label': u'2.0.3.1-113556.2-a1', 'torsion_structure': {}},
 {'label': u'2.0.3.1-116364.1-a1', 'torsion_structure': {}},
 {'label': u'2.0.3.1-146716.4-a1', 'torsion_structure': {}},
 {'label': u'2.0.3.1-146716.4-b1', 'torsion_structure': {}}]
```
This could be a jsonb vs array problem, since an empty array is indicated by `{}`.  I assume that these are the four new curves; what function did you use to upload them?

JohnCremona:
These are the 4 I added recently (after postgres).  See lines 336 and 409 of scripts/ecnf/import_ecnf_data.  There, torstruct is list of ints and the dict used in the insert_many (line 672) uses that as the value for key 'torsion_structure'.  It's chance that these 4 new curves all had trivial torsion so the list is just [].
What is needed now is (1) a one-time fix for this column on four rows; (2) a change to that upload script so that it will not happen again next time.

roed314:
I just manually updated them so that the torsion structure is [].

I just glaced at the insert script and `insert_many` and didn't see an obvious problem....

JohnCremona:
> I just manually updated them so that the torsion structure is [].

Thanks.  I then did db.ec_nfcurves.stats.refresh_stats() and now the torsion_structures example correctly shows 193336 counts for torsion_structure=(), same as torsion_order=1.  It does also show a count of 0 for torsion_structure (1,) which is invalid but that is not serious.

> 
> I just glaced at the insert script and `insert_many` and didn't see an obvious problem....

I did a test using a table like ec_nfcurves, after creating a similar table (using the new creta_table_like() function), and uploaded to that.  With the upload script as it is the empty lists were stored as {}.  But after wrapping the lists with Json() they are stored correctly as ().  I think that sorts that one out.

PR #2911, opened 2019-03-29
Ecnf stats

roed314:
Switches elliptic curves over number fields to the new statistics functionality.  Based on #2910, which should probably be reviewed first.

To test, check that there are no more references to oldstats in the `ecnf` folder, and that the browse pages such as http://localhost:37777/EllipticCurve/browse/2/ and http://localhost:37777/EllipticCurve/browse/3/3/ look correct and included the 4 curves @JohnCremona recently added (to 2.0.3.1).

### DIFF
--- a/lmfdb/backend/database.py
+++ b/lmfdb/backend/database.py
@@ -3168,7 +3168,11 @@ class PostgresTable(PostgresBase):
                 self._search_cols.append(name)
             self.log_db_change("add_column", name=name, datatype=datatype)
 
-    def drop_column(self, name, commit=True):
+    def drop_column(self, name, commit=True, force=False):
+        if not force:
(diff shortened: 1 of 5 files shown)


JohnCremona:
Testing this -- the code looks a lot simpler, good.

http://localhost:37777/EllipticCurve/browse  fails -- this is a link from the page above so that's a problem.

Otherwise all is perfect.

roed314:
Thanks; I fixed the top level browse page.

JohnCremona:
To make sure I understand  how things now work:
* all necessary counts and stats are now in place for the various browse/stats pages
* even after dropping all counts & stats everything would still work but there would be a small delay the first time each count/stat was used as it would be (re)computed and stored
* after any new data upload, calling refresh_stats will recompute all the currently stored stats
* hence there is no need for a function like the old update_stats()

PR #2913, opened 2019-03-31
Switch to using joins for extra columns

### LINKED ISSUE #2881: Rewrite is slow for tables with extras
It's not feasible to use `rewrite` for elliptic curves or number fields, since the `rewrite` method uses `search` to get the input for the rewriting function, and search wasn't intended to return large numbers of rows with columns from the extras table (the whole point of the extras table was that these columns aren't used in search results).  The right solution is to change the `search` and `_search_iterator` methods to use a join instead of doing things in an ad-hoc way.

roed314:
This PR includes one trivial change, one minor change and one substantial change.

* #2905 introduced an importance column to `meta_tables`, but the schema at the top of `database.py` wasn't updated.  We add the new column.
* The `projection` keyword for `search` and `lucky` supported `projection=1.1` if you wanted to include the `id` (1 indicates all search columns, and 2 all search and extra columns).  Using a floating point value to indicate this option is silly.  This option was used in the api, where the only reason we didn't return all fields (both search and extras) was a limitation of the search functionality.  Since it's now feasible to do so, we switch to an option 3 which includes all search and extra columns, plus `id`.
* The main change in this PR is that we use a join between search and extra columns (on `id`, which was specifically designed for this purpose) when returning search results, rather than doing a new SELECT.  This change simplifies the code and will greatly speed up searches when columns from the extra table are needed.

This PR should resolve #2881.

### DIFF
--- a/lmfdb/api/api.py
+++ b/lmfdb/api/api.py
@@ -169,7 +169,7 @@ def api_query(table, id = None):
     if fields:
         fields = ['id'] + fields.split(DELIM)
     else:
-        fields = 1.1
+        fields = 3
 
     if sortby:
         sortby = sortby.split(DELIM)
--- a/lmfdb/api/test_api.py
+++ b/lmfdb/api/test_api.py
@@ -77,5 +77,5 @@ class ApiTest(LmfdbTest):
             if '8T3' in query:
                 assert '"name": "E(8)=2[x]2[x]2"' in data
             if '11a1' in query:
-                assert '"lmfdb_label": "11.a2",' in data
-                assert '"jinv": "-122023936/161051",' in data
+                assert '"lmfdb_label": "11.a2"' in data
+                assert '"jinv": "-122023936/161051"' in data
--- a/lmfdb/backend/database.py
+++ b/lmfdb/backend/database.py
@@ -115,11 +115,11 @@ def jsonb_idx(cols, cols_type):
     return tuple(i for i, elt in enumerate(cols) if cols_type[elt] == "jsonb")
 
 _meta_tables_cols = ("name", "sort", "count_cutoff", "id_ordered",
-        "out_of_order", "has_extras", "stats_valid", "label_col", "total")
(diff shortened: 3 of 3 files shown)


JohnCremona:
One of the tests in test_api is failing.

roed314:
The failure was due to a changed order in the returned json object (so that the lmfdb label was last and didn't have a comma).

Do you agree that including more data in the api search results is an improvement?  I think that the reason we only included the columns from the search table was just a technical one, but perhaps I'm missing something.

jwj61:
I pulled this and am trying a rewrite on the number field database.

JohnCremona:
> The failure was due to a changed order in the returned json object (so that the lmfdb label was last and didn't have a comma).

Well spotted.  I am making a separate trivial PR to change that since it is causing other PR's to fail which is distracting.

> 
> Do you agree that including more data in the api search results is an improvement? I think that the reason we only included the columns from the search table was just a technical one, but perhaps I'm missing something.

Yes I agree, we should include everything.  Certainly the new api will do this.

JohnCremona:
I think this is fine but I am not really competent to judge the correctness.

jwj61:
By the way, this is still running.  Based on the sizes of the two files it creates, it is progressing more slowly than it did before.  If they are a good indication, we are only 1% done after 2 days.

roed314:
@jwj61  That's too bad.  I assume that your rewrite function isn't doing anything complicated?

It sounds like I'll need to profile this and figure out why it's so slow.  I'm probably not going to be able to get to that this week.

jwj61:
It should all be fast.  It uses the stored discriminant and computes the root discriminant, it does a string replace of backslashes on units, and it calls polsturm from gp for polynomials defining index 2 subfields as part of determining if a field is cm.  (A previous attempt used sage's cm test, but it had to initialize the number field which took too long.)

edgarcosta:
The code looks good to me, this indeed improves the rewrite.
However, doesn't fully fix everything.
David and I found another way to improve the rewrite (and the search in general), by doing a buffered search. We will do another PR with those changes, but this one should be merged.

Issue #2582, opened 2018-08-19
Grayed out input on search pages


jvoight:
On search pages, we often put example input into search fields in gray.  We have been given feedback that this is dark enough of a gray that it looks like it's part of the search, and anyway this information is listed to the side of a search box anyway.  

<img width="528" alt="screen shot 2018-08-18 at 9 48 21 pm" src="https://user-images.githubusercontent.com/4120797/44304665-787cc280-a330-11e8-9a57-fc13b4da1189.png">

I happen to like the gray, but I am passing along this issue for discussion.

davidfarmer:
I think this can be made light enough so that it does not distract,
while still being useful.  Assigning myself to look into this.
Note: placeholder, example, and maybe other things, are used as attributes

roed314:
We lightened the gray at some point in the last six months, so I think this issue can be closed.

AndrewVSutherland:
Agreed, I think this issue has been addressed, but let's give @davidfarmer  a chance to comment before closing.

Issue #2915, opened 2019-03-31
Remove browser-specific selectors


davidfarmer:
The CSS selectors which start `-moz`,  `-webkit`, and `-o` are mostly,
and possibly completely, archaic.  Those can be replaced by the standard CSS
selectors.

I can do this, but will wait for comments in case this may cause problems I am not
aware of.

This issue came to my attention when I was investigating why issue #2582 was
not fixed, even though some people said it was.  The problem is that the
selector was specific to webkit, and I was viewing in Firefox.  (I am about to send
a small pull request fixing that other issue.)

It is good to check CSS changes in several browsers.

edgarcosta:
I agree with this, but not all of them are deprecated for example, ```sticky``` is not yet supported, and the hack to make it work is```-webkit-sticky```

davidfarmer:
Addressed by  #2947

PR #2916, opened 2019-03-31
Placeholder css

davidfarmer:
Fixes issue #2582 in all browsers.  See any page with placeholders in the input field of a form,
such as
   /EllipticCurve/Q/

See issue #2915 for an explanation why this was needed.


JohnCremona:
The api test failure -- which is already present in master -- is fixed in #2913.

JohnCremona:
The new version looks fine to me on chrome and firefox on ubuntu -- in fact it looks the same before and after.  So, no objection to merging from these eyes!

PR #2917, opened 2019-04-01
fix failing api test by removing commas so order is irrelevant

JohnCremona:
See #2913.  The only reason the test was failing was because it was looking for some commas in the output which made the test sensitive to the order of columns.

This fix was borrowed from #2913 which is still under review (as it has more substantial changes) so that other PR's do not fail spuriously.  This one can be merged as soon as tests have passed.


roed314:
Looks good to me!

JohnCremona:
OK, I triggered a retest for #2916

roed314:
Don't you need to merge in master?

Issue #500, opened 2015-10-28
Tensor products of elliptic curves not working and/or extremely resource intensive.


swisherh:
If you choose to tensor product two elliptic curves, then select a random elliptic curve in each box and click on “tensor these”, you get a server error

http://beta.lmfdb.org/TensorProducts/show/?obj1=EllipticCurve%2FQ%2F272080%2Fl%2F2&obj2=EllipticCurve%2FQ%2F132800%2Fc%2F1

JohnCremona:
On 28 October 2015 at 13:40, Holly Swisher <notifications@github.com> wrote:

> If you choose to tensor product two elliptic curves, then select a random
> elliptic curve in each box and click on “tensor these”, you get a server
> error
>
>
> http://beta.lmfdb.org/TensorProducts/show/?obj1=EllipticCurve%2FQ%2F272080%2Fl%2F2&obj2=EllipticCurve%2FQ%2F132800%2Fc%2F1
>

The tensor product code is nowhere near powerful enough for this, which is
why we took it out of beta (or something).

AndrewVSutherland:
While in the process of cleaning up some of the L-function code I came across this and also cleaned up the tensor_products code and modified it to use the new error handling code introduced in PR #1667 so these internal server errors no longer occur and a more reasonable error message is displayed (in this case the issue is that current implementation requires both curves to be semistable).

It will at least attempt to compute tensor products of semistable elliptic curves, e.g. http://beta.lmfdb.org/TensorProducts/show/?obj1=EllipticCurve%2FQ%2F11%2Fa%2F2&obj2=EllipticCurve%2FQ%2F14%2Fa%2F1, but the computation is too expensive and it gets aborted after 30 seconds.

I'm not sure how feasible it is to compute these on the fly.  It would be a very cool feature to have but will require work to support efficiently.  I don't think a python computation on the fly is going to be fast enough, it will likely require some supporting pre-computation and/or C code.

edgarcosta:
Could we put a more reasonable time limit on it?
like in the case of the zeros or graph of an L-function?

AndrewVSutherland:
@edgarcosta How do we change the timeout for a specific computation?  I didn't realize one could do this.

edgarcosta:
I don't know at the moment.
I would need to look it up. My first approach would be to check the code of L-functions and zeros, to see how they do it.

davidfarmer:
Presumably we should precompute those tensor products, which will
[...]

PR #2921, opened 2019-04-02
Unreachable code in lfunction.py (LGTM)

davidlowryduda:
I've continued to use LGTM for static code checking. It recognized several instances of unreachable and unused code in lfunction.py.

This pull request has 3 commits. The first two commits each correspond to removing unreachable code. 

The third commit is larger, but only because I made the code compatible for both python2/3 and did some basic linting (in part because my py2to3 conversion utility does some basic linting as well). It's full of whitespace correction, maps-->list(maps), and things of that nature.

I also note that I commented out the tensor product code at the end, as it's not being used anywhere. But I didn't remove it, since the code suggests that it should be reconsidered when #500 is handled. I notice that #500 is currently closed, so I was tempted to fully remove the code --- but I didn't.

### DIFF
--- a/lmfdb/lfunctions/Lfunction.py
+++ b/lmfdb/lfunctions/Lfunction.py
@@ -1317,7 +1317,6 @@ class Lfunction_HMF(Lfunction):
 
         # Compute Dirichlet coefficients ########################
         R = QQ['x']
-        (x,) = R._first_ngens(1)
         K = NumberField(R(str(f['hecke_polynomial']).replace('^', '**')), 'e')
         iota = K.complex_embeddings()[self.number]
 
--- a/lmfdb/lfunctions/Lfunction.py
+++ b/lmfdb/lfunctions/Lfunction.py
@@ -6,7 +6,8 @@
 # DedekindZeta, ArtinLfunction, SymmetricPowerLfunction,
 # Lfunction_genus2_Q
 
-import math, re
+import math
+import re
 
 from flask import url_for, request
 from sage.all import (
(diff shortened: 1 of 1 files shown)


AndrewVSutherland:
I don't think these are the kind of changes we want to be making right now with Release 1.1 imminent.  In cases where LGTM identifies an actual bug I'm all in favor of fixing it, but making blanket code changes that are mostly (but not all) formatting changes introduces risks that I don't think we should be taking -- it is very hard for anyone reviewing a PR like this to be sure that all the changes are truly harmless and our test coverage is too low to have confidence that we will catch everything.

I'm fine with commenting out the unused tensor product code and commenting out (or adding a comment immediately above) the identified unreachable code if you want to submit a new PR that does just this.

JohnCremona:
(written before reading @AndrewVSutherland 's comment)

LGTM (I find it helps to use the facility of github to hide whitespace diffs when viewing the changes).

If we are going to start to make changes based on LGTM suggestions, which I think is a good thing, then developers need to know *why* some of these changes have been made (not the whitespace ones).  Perhaps a good practice would be to include with such PRs a link to the appropriate LGTM page.

And I really do not care about whitespace / layout changes.

JohnCremona:
To be fair to @davidlowryduda (who had some encouragement from me), I'm note sure how people who were not at the recent Princeton meeting are supposed to know that "release 1.1 is imminent".

jwj61:
Also, the unreachable code in the first commit has a comment saying that the part currently being used is a temporary fix, which implies someone intends to come back later and use the part you deleted.

AndrewVSutherland:
@JohnCremona, @davidlowryduda Apologies if my comment came across the wrong way, I think it's a great idea to be using LGTM to try to find bugs and I would encourage this to continue! 

Independent of our release schedule, making apparently innocuous linting/formatting changes that touch lots of code is something that is best done either in conjunction with work on a specific section of the LMFDB when we know there will be a lot of testing done anyway, or during a planned interval when we make a conscious decision to hold off pushing code to web for an extended period of time (if/when we decide to switch to Python 3 we will definitely want to do this).

davidlowryduda:
@AndrewVSutherland  I understand. I don't know anything about the upcoming 1.1 release, but I'll keep that in mind.

What I think I'll do is submit a new PR without any linting. But I think I might still include python2/3 compatibility, since it's on my mind. What would you think about that @AndrewVSutherland ?

JohnCremona:
I think that a reasonable way forward is this.  Let's try and avoid *any* automatic reformatting.  LGTM is about a lot more than that.
* LGTM lists (currently) 611 Alerts.  This includes 132 Errors, 203 Warnings and the rest are Recommendations.
* We should attend to these in order of severity, fixing the Errors first.  Even some of these do not seem serious to me (e.g. some has wrapped a block of code inside "if False:" as a quick way of commenting it out.)
* We can allocate subsets of these to different people (e.g. someone other than me should handle the javascript ones)
* If there are some common errors being picked up we should think about preparing some kind of style guide for the future.

It would be good if we ever got to the point where our code was 100% LGTM-clean, as then we could maintain that in the same way that we use pyflakes.  But that might be too much to hope for.

davidfarmer:
Small and possibly obvious comment:

A pull request that primarily addresses code layout (i.e., mostly white space adjustments)
should only contain such changes.  This current pull request also includes commenting
out some tensor product code, which I assume was some of the unreachable code.
I find it hard to evaluate the changes because my brain has to be alert for actual content
changes amidst a bunch of straightorward changes.

I am okay with these layout changes, but I'd like to see the tensor products code left as-is
until someone is ready to actually to actually tackle tensor products as a separate project.

PR #2922, opened 2019-04-02
Added comments to "download to" for HMF. 

jvoight:


jvoight:
This is addresses a comment made by Akshay Venkatesh, to make sure the user knows that there is code at the bottom to recreate the HMF.  (I'd like to rewrite much of this in light of our CMF experience, but that's all for later.)

He also suggested that we rename "Download to magma" with something else to indicate that there is more there.  I made up something on the spot that I immediately disliked (but seemed fine to everyone else).  Now I'm thinking "Download/recreate in magma"?  Whatever we decide, I think it should look the same across other pages.

jvoight:
I'm going to modify CMFs and HMFs to follow the script "Download <blank> to <blank>", and if we're happy we'll do this more broadly.

jvoight:
OK, as systematically as possible I replaced all downloads links with "download X to Y", as we discussed today.  I think this is clearer.  Only names changed, no functionality should be affected.  

The only thing that I wasn't completely sure about was renaming "Download Magma code" to "Download code to Magma".  I prefer consistency and chose the latter, but I'm happy if the former is preferred.

AndrewVSutherland:
You could use "Download code for Magma", but I'm OK with "Download code to Magma" if you prefer (I think both are better than "Download Magma code").

Issue #2923, opened 2019-04-02
Katex not rendering inside properties box


edgarcosta:
See for example http://www.lmfdb.org/ModularForm/GL2/Q/holomorphic/805/2/p/a/

alexjbest:
As I recall this was sort-of intentional, katex works fine on the properties box, just not stuff that is added dynamically by javascript after katex has run when the page first loads. I.e. friends for L-functions.

Any content that is dynamically added by javascript should be passed to KaTeX by calling `renderMathInElement(<element-id>, katexOpts);`.


Why was this semi-intentional? For the "monster form" 983.2.c.a with a lot a friends we didn't want katex to waste time parsing a bunch of links that didn't contain any math. At that time there was no Sato-Tate link there so it wasn't a problem.


#2925 should be a minimal fix this in this case, there I make katex render before the L-function links but after Sato-Tate, best of both worlds?

davidfarmer:
Fixed by #2925

PR #2925, opened 2019-04-02
Force katex render on L-function friends in properties box

alexjbest:
Addresses #2923


davidfarmer:
Looks good.  Will merge.

PR #2926, opened 2019-04-03
Fix bmf_dims

### LINKED ISSUE #2854: Fix dimension table display for Bianchi modular forms
This did not survive migration to postgres.  For example, 
http://localhost:37777/ModularForm/GL2/ImaginaryQuadratic/gl2dims/2.0.4.1
displays blank data and it should display a lot.

roed314:
This resolves #2854.  To test, visit http://localhost:37777/ModularForm/GL2/ImaginaryQuadratic/gl2dims/2.0.4.1, etc.

The problem was that postgres only returns the first 50 results, which were then filtered away by the BMF code.  I added four columns to the `bmf_dims` table: `sl2_new_totaldim` (the total of the `new_dim` entries of `sl2_dims` for all weights), `sl2_cusp_totaldim`, `gl2_new_totaldim`, and `gl2_cusp_totaldim`.  For our search purposes we only needed whether these were zero or nonzero, but I thought having an integral column might be useful.

### DIFF
--- a/lmfdb/bianchi_modular_forms/bianchi_modular_form.py
+++ b/lmfdb/bianchi_modular_forms/bianchi_modular_form.py
@@ -190,7 +190,6 @@ def bmf_field_dim_table(**args):
         query[totaldim] = {'$gt': 0}
     t = ' '.join(['Dimensions of Spaces of {} Bianchi Modular Forms over'.format(info['group']), pretty_field_label])
     data = list(db.bmf_dims.search(query, limit=count, offset=start, info=info))
-    print info
     nres = info['number']
     if not info['exact_count']:
         info['number'] = nres = db.bmf_dims.count(query)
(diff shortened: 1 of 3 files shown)


JohnCremona:
OK thanks.  I don't have time to test this now.  It's a pity that the data in the table needs changing as now I will have to also change upload scripts to store the extra stuff, and I'll almost certainly forget.  But I have no plans to extend this data in the near future so that's OK.

AndrewVSutherland:
The next and previous buttons don't seem to work (on linux, using both chrome and firefox).  You can switch pages by manually updating the URL, but if you go to 

http://127.0.0.1:37777/ModularForm/GL2/ImaginaryQuadratic/gl2dims/2.0.4.1?start=50&count=50&level_flag=cusp

clicking on the prev or next button doesn't seem to do anythying.

AndrewVSutherland:
Also, it displays the total number of levels as 1000 when it really means at least 1000 (I assume this is the usual issue of not wanting to count the search results, but for these dim tables the totals are small enough that I would just count them rather than giving the user the option to ask for it).  Compare the lines "Displaying 50 of ... levels" in:

http://127.0.0.1:37777/ModularForm/GL2/ImaginaryQuadratic/gl2dims/2.0.11.1
http://prod.lmfdb.xyz/ModularForm/GL2/ImaginaryQuadratic/gl2dims/2.0.11.1

roed314:
Test failures should be fixed by #2935.  Once that's merged into master, I'll merge it into this PR and make sure tests pass.

PR #2928, opened 2019-04-04
Lfunction rcs knowls

jwj61:
Try 2 (I don't know why travis failed before):

This adds rcs knowls to L-function pages.

It for source and rigor, it reads from the database to see which knowls to use. Note, the database seems to have a typo in source knowl identifier for load_type=Costa.

Unlike other areas, the knowl information is typically specific to an L-function, so we add on to the url and bread. It is possible we should not do this for Completeness since there is only one such knowl?

To test, look at various L-function pages and click on the rcs links in the learnmore about area. Most are stubs, but you can look at the identifier of the knowl to check that it is correct.

### DIFF
--- a/lmfdb/lfunctions/main.py
+++ b/lmfdb/lfunctions/main.py
@@ -489,34 +489,27 @@ def render_lfunction_exception(err):
     info = {'explain': errmsg, 'title': 'Error displaying L-function', 'bread': bread }
     return render_template('problem.html', **info)
 
+def learnmore_list(path, remove=None):
+    prepath = re.sub(r'^/L/', '', path)
+    prepath = re.sub(r'/$', '', prepath)
+    learnmore = [
+        ('Completeness of the data', url_for('.completeness', prepath=prepath)),
+        ('Source of the data', url_for('.source', prepath=prepath)),
+        ('Reliability of the data', url_for('.reliability', prepath=prepath))]
+    if remove:
+        return filter(lambda t:t[0].find(remove) <0, learnmore)
+    return learnmore
 
 def initLfunction(L, args, request):
     ''' Sets the properties to show on the homepage of an L-function page.
     '''
     info = L.info
     info['args'] = args
-    prepath = request.path
-    prepath = re.sub(r'^/L/','',prepath)
-    prepath = re.sub(r'/$','',prepath)
     info['properties2'] = set_gaga_properties(L)
 
     set_bread_and_friends(info, L, request)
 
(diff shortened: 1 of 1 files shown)


jwj61:
I decided Completeness should be a top-level url (L/Completeness) like other areas since one knowl will cover all L-functions.  I then added it as a learn more about item to all of the L-function pages.

As something to consider, we have a front page for L-functions of elliptic curves (and half a dozen similar pages).  The completeness knowl there will be for all L-functions, but it might make sense for it to be just for L-functions of elliptic curves.  It will be simple to change if we decide to write such a knowl.

AndrewVSutherland:
One issue I noticed: on dynamically compute L-function pages like 

http://127.0.0.1:37777/L/ModularForm/GL2/TotallyReal/2.2.8.1/holomorphic/2.2.8.1-34.2-a/0/0/

clicking on either the source or reliability links triggers a recomputation of the Z-plot and zeros (you can see this happening if you run in debog mode, and the delay is quite noticeable).  This is especially problematic if you try to click on either knowl link before it has finished the first computation of the zeros and Z-plot (they get displayed twice, then you go to the source/reliability knowl).

jwj61:
I don't think it is Z-plot and zeros.  I found those functions, put in print statements.  When the main page lots, those print statements printed, but when I went to the Source page, they did not.

The logger warnings are from computing special values (easy to grep for the message in the source).  I think the situation seems worse than it is when running a test version because the website is multithreaded, but the when testing, I think it is really single threaded.  It seems that sometimes code from the last click has to finish before the new click is processed, but that wouldn't hold for the main site.

There is still the fact that the special values are computed for the Source/Reliability pages.  Before proceeding, my leading options are (1) ignore it since you should wait for the current page to finish loading to better simulate the main site, and then it is not too bad; or (2) make it so that those are only computed when needed.  For (2), I could make that spot produce a function which produces the values and put return that instead.  Then change places which use the values so that they first call the function.

Opinions?

AndrewVSutherland:
@jwj61 I expect your are right that this will be less noticeable in production.  Let's go ahead and merge this and see how it looks on beta and then we can take further action if needed (it will also be easier for people to fill in the stubbed knowls once it is on beta).

If you are OK with this I will go ahead and merge (or free to do so yourself).

jwj61:
Merging.  Before it goes to production, we should have all the knowls written.  BTW, I just fixed the data error in the lfunc_rs_knowls table mentioned above (source knowl for Costa).

Issue #2754, opened 2019-03-15
Related object links from Artin representations to weight 1


roed314:
Add related objects links from Artin representations back to weight 1 forms.  For example, http://www.lmfdb.org/ModularForm/GL2/Q/holomorphic/23/1/b/a/ has a link to http://www.lmfdb.org/ArtinRepresentation/2.23.3t2.1c1 but there is no link back.

edgarcosta:
Perhaps we could achieve this by adding L-functions instances for Artin representations and use that table to figure out additional friends for Artin reps.

AndrewVSutherland:
Given that we just addressed #433, we should address this for release 1.1.  Friendships really should be mutual.

Additionally, the Artin rep has its own (dynamically computed) L-function.  We are planning to address this in the July Bristol workshop, but for Artin reps that correspond to weight 1 modular forms in the database, we could link their L-functions to the CMF L-functions (if the Artin rep knows its friend, it also knows its friend's L-function).

edgarcosta:
Does #2930 address this issue completely?

AndrewVSutherland:
Yes, I think this can be closed (but we still want to change the embedded CMF pages so they link only to the artin rep with the same L-function).

PR #2929, opened 2019-04-04
Load L-functions from database for HMF and Artin when available

AndrewVSutherland:
This PR partially addresses #2754 and #2873 by having the L-function code check the instances database when displaying HMF and Artin L-functions.  If the L-function is stored in the database it will be used, otherwise it will revert to computing them on the fly as it does now.  To keep the code changes minimal and isolated I created a new L-function class Lfunction_HMFDB that loads the L-function from the database and will fail if it is not present -- this is called from the routing function l_function_hmf_page inside a try/except and will catch the error and ignore it (this is a small hack that will go away once all HMF functions are stored in the database.  The case of Artin L-functions is handled similarly.

To test, compare:

http://127.0.0.1:37777/L/ModularForm/GL2/TotallyReal/2.2.5.1/holomorphic/2.2.5.1-500.1-a/0/0/
http://beta.lmfdb.org/L/ModularForm/GL2/TotallyReal/2.2.5.1/holomorphic/2.2.5.1-500.1-a/0/0/

and also try clicking from the L-function page to the HMF and back, test the bread, click through to different origins and then back via the origin page L-function links etc. (other than the URL the L-function page should look the same now matter which origin you came from).  In particular, the L-function page should list all the related objects now matter how the page was reached.

Note that this change does not only impact HMFs whose L-functions arise as genus 2 L-functions that are in the database, it also catches L-functions of elliptic curves over number fields, e.g. compare

http://127.0.0.1:37777/L/ModularForm/GL2/TotallyReal/2.2.5.1/holomorphic/2.2.5.1-31.1-a/0/0/
http://beta.lmfdb.org/L/ModularForm/GL2/TotallyReal/2.2.5.1/holomorphic/2.2.5.1-31.1-a/0/0/

For an Artin L-function example, compare

http://127.0.0.1:37777/L/ArtinRepresentation/2.23.3t2.1c1/
http://beta.lmfdb.org/L/ArtinRepresentation/2.23.3t2.1c1/

(again clicking through to the Artin rep and back should work).  Note that this is currently the only Artin L-function for which I have added an instance record, but I will be adding more later today (this doesn't involve any code changes so you can test the PR now).

Examples of L-function pages that should not change (i.e. they should still be rendered on the fly) because we have not precomputed them via another origin:

http://127.0.0.1:37777/L/ModularForm/GL2/TotallyReal/2.2.5.1/holomorphic/2.2.5.1-101.2-a/0/0/
http://beta.lmfdb.org/L/ModularForm/GL2/TotallyReal/2.2.5.1/holomorphic/2.2.5.1-101.2-a/0/0/

http://127.0.0.1:37777/L/ArtinRepresentation/2.2e8_3e2.8t5.2c1
http://beta.lmfdb.org/L/ArtinRepresentation/3.229.4t5.1c1/

### DIFF
--- a/lmfdb/lfunctions/Lfunction.py
+++ b/lmfdb/lfunctions/Lfunction.py
@@ -42,6 +42,12 @@ from LfunctionDatabase import (
     get_instance_by_url, getHmfData, getHgmData,
     getEllipticCurveData, get_multiples_by_Lhash)
 
+def artin_url(label):
+    return "ArtinRepresentation/" + label
+
+def hmf_url(label):
+    "ModularForm/GL2/TotallyReal/" + label.split("-")[0] + "/holomorphic/" + label
+
 def validate_required_args(errmsg, args, *keys):
     missing_keys = [key for key in keys if not key in args]
     if len(missing_keys):
@@ -1260,7 +1266,7 @@ class Lfunction_HMFDB(Lfunction_from_db):
         self.label = kwargs['label']
         self.origin_label = self.label
         self._Ltype = "hilbertmodularform"
-        self.url = "ModularForm/GL2/TotallyReal/" + self.label.split("-")[0] + "/holomorphic/" + self.label
+        self.url = hmf_url(self.label)
         Lfunction_from_db.__init__(self, url = self.url)
 
 class Lfunction_HMF(Lfunction):
@@ -1672,7 +1678,7 @@ class ArtinLfunctionDB(Lfunction_from_db):
         self.label = kwargs['label']
         self.origin_label = self.label
         self._Ltype = "artin"
-        self.url = "ArtinRepresentation/" + self.label
+        self.url = artin_url(self.label)
         Lfunction_from_db.__init__(self, url = self.url)
 
     @lazy_attribute
--- a/lmfdb/lfunctions/Lfunction.py
+++ b/lmfdb/lfunctions/Lfunction.py
@@ -1265,11 +1265,15 @@ class Lfunction_HMFDB(Lfunction_from_db):
     def __init__(self, **kwargs):
         constructor_logger(self, kwargs)
 
-        validate_required_args ('Unable to construct L-function.', kwargs, 'label')
+        validate_required_args ('Unable to construct Hilbert modular form ' +
+                                'L-function.', args, 'label', 'number', 'character')
+        validate_integer_args ('Unable to construct Hilbert modular form L-function.',
+                               args, 'character','number')
+        
         self.label = kwargs['label']
         self.origin_label = self.label
         self._Ltype = "hilbertmodularform"
-        self.url = hmf_url(self.label)
+        self.url = hmf_url(args['label'],args['character'],args['number'])
         Lfunction_from_db.__init__(self, url = self.url)
 
 class Lfunction_HMF(Lfunction):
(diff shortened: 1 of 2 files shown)


edgarcosta:
@AndrewVSutherland, I must have misunderstood something, I got the idea that I was the one taking care of this, and I was planning to submit my PR tomorrow for HMF and BMF.
Would it be possible for me to selectively merge some of your changes? Alternatively, I can hijack your PR.

AndrewVSutherland:
@edgarcosta Don't worry about, I can add BMF as well, I need to fix the Travis errors anyway  (some L-function tests need to be tweaked).

edgarcosta:
@AndrewVSutherland, thanks.
I might have something to add when you are done.

Should we deprecate the 0/0 for HMFs?

AndrewVSutherland:
@edgarcosta either we should, or the instance URL should include it.

edgarcosta:
@AndrewVSutherland The 0/0 stands for trivial character and first embedding of the Hecke field.
This in CMFs would correspond to 1/1
For the HMF with Hecke field QQ having 0/0 or not is irrelevant, and perhaps we should have instances for both, as we do for CMFs.

However, for the instances arising from genus 2 curves, things are a bit more complicated.
In this case, the eigenvalues are Galois conjugates, and thus either embedding will match the genus 2 curve, but the L-function of the orbit (i.e. without the 0/0) is a degree 8 L-function

AndrewVSutherland:
@edgarcosta The HMFs should really have Galois orbit and embedded instance  like CMFs and lfunc_instances should use whichever is correct.  But I don't think this is something we want to do right now and I don't t think it is relevant to this PR (none of the instance URL's have embedding suffixes and I am simply matching L-funciton URLs that end in /0/0 against these.  In any case, we can talk about this when you get back.

edgarcosta:
@AndrewVSutherland Okidoki
ps: I'm in my office

AndrewVSutherland:
OK, Travis is happy, and I have added support for BMF L-functions.  The BMF pages don't have active L-function links on them yet, but you can manually insert an L into the URL to get to the L-function page:

http://127.0.0.1:37777/L/ModularForm/GL2/ImaginaryQuadratic/2.0.11.1/24964.1/a/

Updating the BMF pages to add links to their L-functions will happen in a separate PR.

This PR should be ready to review now.

edgarcosta:
I will review it.

AndrewVSutherland:
I have added L-function instance records for the 6479 Artin L-functions that correspond to (embedded) modular forms of weight 1, for compare:

http://127.0.0.1:37777/L/ArtinRepresentation/2.2e2_641.7t2.1c3/
http://beta.lmfdb.org/L/ArtinRepresentation/2.2e2_641.7t2.1c3/

Note that for CMFs of dimension greater than 1 there will be multiple L-functions (one for each embedding), each of which corresponds to an Artin representation.  Note that the ordering of embbeddings of CMFs os not necessarily the same as the ordering of conjugate Artin representations, as in this example where the CMF corresponding to the L-function above (for the third conjugate Artin rep) is the first embedding:

http://127.0.0.1:37777/L/ModularForm/GL2/Q/holomorphic/2564/1/d/b/2563/1/

Currently the embedded modular form pages list all the conjugate Artin reps as related objects, but this will be changed in a separate PR.

PR #2940, opened 2019-04-08
Server sider cursors

edgarcosta:
We now provide buffered searches, when the user doesn't specify a limit constraint.
This also speeds up significantly rewrite, and hopefully fully addresses #2881, rewriting `nf_fields` should take about a day based on some measurements on the "writing" to disk part:
We averaged   0.0006s on the first 1.6 million records, 0.003s for several degree 16, and 0.008s for the last couple thousand records.

For example, now 
```
for rec in db.nf_fields.search():
    print  rec
    break
```
is instantaneous, while in the past it first fetched all the records into memory.

Compare against with the current code (this will take about 1min):
```
for rec in db.gps_small.search():
    print  rec
    break
```

### DIFF
--- a/lmfdb/backend/database.py
+++ b/lmfdb/backend/database.py
@@ -5019,12 +5019,12 @@ class PostgresDatabase(PostgresBase):
     def cursor(self, buffered=False):
         """
         Returns a new cursor.
-        If buffered, then it creates a server side cursor that should be manually
+        If buffered, then it creates a server side cursor that must be manually
         closed after done using it.
         """
         if buffered:
             self.server_side_counter += 1
-            return self.conn.cursor(str(self.server_side_counter))
+            return self.conn.cursor(str(self.server_side_counter), withhold=True)
         else:
             return self.conn.cursor()
 
--- a/lmfdb/backend/database.py
+++ b/lmfdb/backend/database.py
@@ -1550,14 +1550,17 @@ class PostgresTable(PostgresBase):
             if info is not None:
                 # caller is requesting count data
                 info['number'] = self.count(query)
-            return self._search_iterator(cur, search_cols, extra_cols, projection)
+            return self._search_iterator(cur, search_cols,
+                                         extra_cols, projection)
         if nres is None:
             exact_count = (cur.rowcount < prelimit)
             nres = offset + cur.rowcount
         else:
             exact_count = True
         res = cur.fetchmany(limit)
-        res = list(self._search_iterator(res, search_cols, extra_cols, projection))
+        res = list(
+                self._search_iterator(res, search_cols, extra_cols, projection)
+                )
         if info is not None:
             if offset >= nres:
                 offset -= (1 + (offset - nres) / limit) * limit
--- a/lmfdb/backend/database.py
+++ b/lmfdb/backend/database.py
@@ -4921,7 +4921,11 @@ class PostgresDatabase(PostgresBase):
         return connection
 
     def reset_connection(self):
-        logging.info("Connection broken (status %s); resetting..."%self.conn.closed)
+        """
+        Resets the connection
+        """
+        logging.info("Connection broken (status %s); resetting...",
+                     self.conn.closed)
         conn = self._new_connection()
         # Note that self is the first entry in self._objects
         for obj in self._objects:
@@ -4932,6 +4936,7 @@ class PostgresDatabase(PostgresBase):
         self._objects.append(obj)
 
     def __init__(self, **kwargs):
+        self.server_side_counter = 0
         self._nocommit_stack = 0
         self._silenced = False
         self._objects = []
@@ -4958,11 +4963,11 @@ class PostgresDatabase(PostgresBase):
             cur = sorted(list(self._execute(SQL("SELECT privilege_type FROM information_schema.role_table_grants WHERE grantee = %s AND table_schema = %s AND table_name=%s AND privilege_type IN (" + ",".join(['%s']*len(privileges)) + ")"), [self._user,  'userdb', 'users'] + privileges)))
             self._read_and_write_userdb = cur == sorted([(priv,) for priv in privileges])
 
-        logging.info("User: %s" % self._user)
-        logging.info("Read only: %s" % self._read_only)
-        logging.info("Super user: %s" % self._super_user)
-        logging.info("Read/write to userdb: %s" % self._read_and_write_userdb)
-        logging.info("Read/write to knowls: %s" % self._read_and_write_knowls)
+        logging.info("User: %s", self._user)
+        logging.info("Read only: %s", self._read_only)
+        logging.info("Super user: %s", self._super_user)
+        logging.info("Read/write to userdb: %s", self._read_and_write_userdb)
+        logging.info("Read/write to knowls: %s", self._read_and_write_knowls)
         # Stores the name of the person making changes to the database
         from lmfdb.utils.config import Configuration
         self.__editor = Configuration().get_logging().get('editor')
@@ -4996,6 +5001,12 @@ class PostgresDatabase(PostgresBase):
         return "Interface to Postgres database"
 
 
+    def cursor(self):
+        """
+        Returns a new server side cursor, with automatic name generation
+        """
+        self.server_side_counter += 1
+        return self.conn.cursor(int(self.server_side_counter))
 
 
     def login(self):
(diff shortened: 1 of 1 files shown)


jwj61:
Just started a nf rewrite using this (and the scare quotes fix).

roed314:
I worked on this with Edgar today and the changes are all fine with me.

AndrewVSutherland:
The example you gave is far from instantaneous for me (running on legendre).  I started it 2-3 minutes ago and it is still running

`~/github/lmfdb$ git checkout edgarcosta/server_sider_cursors 
M	lmfdb/genus2_curves/main.py
Previous HEAD position was ee09eaf... Fix typo in insert_many
HEAD is now at 6b75ecd... formatting for David
~/github/lmfdb$ sage
┌────────────────────────────────────────────────────────────────────┐
│ SageMath version 8.3, Release Date: 2018-08-03                     │
│ Type "notebook()" for the browser-based notebook interface.        │
│ Type "help()" for help.                                            │
└────────────────────────────────────────────────────────────────────┘
sage: from lmfdb import db
INFO:LMFDB@2019-04-09 06:58:21,649: Configuration = {'postgresql_options': {'user': 'editor', 'host': 'localhost', 'password': '****', 'port': 5432, 'dbname': 'lmfdb'}, 'logging_options': {'slowcutoff': 0.1, 'logfile': 'flasklog', 'editor': 'drew', 'slowlogfile': 'slow_queries.log'}, 'flask_options': {'debug': False, 'host': '127.0.0.1', 'port': 41001}} 
INFO:LMFDB@2019-04-09 06:58:22,654: Connecting to PostgresSQL server as: user=editor host=localhost port=5432 dbname=lmfdb... 
INFO:LMFDB@2019-04-09 06:58:22,671: Done!
 connection = <connection object at 0x7fba8fabea10; dsn: 'user=editor host=localhost password=xxx port=5432 dbname=lmfdb', closed: 0> 
INFO:LMFDB@2019-04-09 06:58:22,696: User: editor 
INFO:LMFDB@2019-04-09 06:58:22,697: Read only: False 
INFO:LMFDB@2019-04-09 06:58:22,697: Super user: False 
INFO:LMFDB@2019-04-09 06:58:22,698: Read/write to userdb: True 
INFO:LMFDB@2019-04-09 06:58:22,698: Read/write to knowls: True 
sage: for rec in db.nf_fields.search():
....:     print rec
....:     break
....: 
`

AndrewVSutherland:
@edgarcosta FYI, on proddb it took 90 seconds to print the first record of nf_fields (using the old code, not this PR), but on legendre (using your PR) it actually looks like it is stuck waiting on something (it is not using any CPU time).

If I try ec_curves instead it is instantaneous (versus 15 seconds on proddb).

I don't think the problem with nf_fields is not related to your PR, but there is definitely a problem, the number field database appears to be completely locked up right now, number field queries on beta are throwing server errors, as are pages that try to access the number fields table like http://beta.lmfdb.org/ModularForm/GL2/Q/holomorphic/124/1/i/a/

AndrewVSutherland:
I restarted postgres and beta is now functioning again (and I can confirm that the first record is returned by db.nf_fields.search instantaneously.  I think this PR is good to go but let's hold off merging until we understand what happened (possibly related to #2913 ?).

@jwj61 if you had a rewrite running I'm sure it got nuked (but it was clearly stuck in any case).  @edgarcosta @roed314 we should try to understand exactly what happened here -- I;m guessing it was triggered by some combination of the rewrite @jwj61 has been trying to get done on nf_fields, the recent changes to db backend, and/or this PR (and the testing therreof).

jwj61:
Looking at the dumped files, they do seem to be sorted by id, the integer which links them.  But, they are not matched.  Some data will need to be recomputed.  Even for fields where the id is present on both sides, we can't be sure they actually go together if the numbering got off track at some point.

edgarcosta:
I agree with Drew's, but the key thing to realize is that during all that time that you waited for the first record, you were loading all of it into memory.
I did it twice, and the second time was about 10s faster:
```
sage: print datetime.datetime.now()
2019-04-09 20:44:03.527926
sage: for rec in db.nf_fields.search():
....:     print rec
....:     print datetime.datetime.now()
....:     break
....:
INFO:nf_fields@2019-04-09 20:45:14,250: SELECT "disc_rad", "iso_number", "disc_abs", "class_number", "label", "coeffs", "class_group", "degree", "used_grh", "cm", "r2", "galt", "ramps", "disc_sign" FROM "nf_fields" ORDER BY "degree", "disc_abs", "disc_sign", "iso_number" ran in  69.051432848s  [database.py:383]
INFO:nf_fields@2019-04-09 20:45:14,251: Replicate with db.nf_fields.analyze({}, 1, None, 0) [database.py:385]
{u'galt': 1, u'cm': False, u'degree': 1, u'used_grh': False, u'ramps': [], u'label': u'1.1.1.1', u'disc_sign': 1, u'r2': 0, u'iso_number': 1, u'disc_rad': 1, u'disc_abs': 1, u'coeffs': [0, 1], u'class_number': 1, u'class_group': []}
2019-04-09 20:45:14.252298
```
but you can also see in the log:
http://legendre.mit.edu/munin/static/dynazoom.html?plugin_name=PostgreSQL%2Fproddb%2Fmemory&start_iso8601=2019-04-09T15%3A49%3A28-0400&stop_iso8601=2019-04-09T17%3A17%3A13-0400&start_epoch=1554839368&stop_epoch=1554844633&lower_limit=&upper_limit=&size_x=800&size_y=400&cgiurl_graph=%2Fmunin-cgi%2Fmunin-cgi-graph

the memory use of that sage process taking ~6GB

edgarcosta:
I don't understand why fetching the whole table in `legendre` is so much slower than in `proddb`, in either case, if you are not connected locally the bottleneck will be bandwidth, and this PR will balance that by buffering the incoming data.

Regarding the locks yesterday, I have no clue who created them, as the server was restarted.
Anyhow, would you mind testing the PR again?

edgarcosta:
@jwj61, were you editing nf_fields in any other way? or were you just running the rewrite?

jwj61:
No, the last attempt was last night which was killed when it stalled and the postgres server was restarted.  Should I try restarting it with this?

edgarcosta:
@jwj61,  can you tell which script are you using, and can I be the one running it? to see if I understand what goes wrong, and it this plays any role in it

jwj61:
/scratch/home/jj/nf-lmfdb/modify.py which I run with sage -python

edgarcosta:
I will start to run that now and monitor it, and if it gets grid locked we will have a clue of what happened

edgarcosta:
Rewrite has been successful, and it took about 7h.

So I think we can go along and merge this PR.

For what went wrong earlier on, I only have somewhat tinfoil hat explanation.

First, from the logs, we know that the rewrite was done with reloading the data and creating all the indexes by 3 am, and since then I assume that the script was waiting for its access exclusive lock to rename the tables.

Now lock attribution works in queues, and since that the access exclusive lock is the possible strongest lock, no one else can do anything with nf_fields until the rewrite is done with its lock, even if it never gets it.

Now, why was there some lock on the server?
Most likely us while testing this PR did something like 
```
for rec in db.nf_fields.search():
    print  rec
    break
```
and did keyboard interrupt and never closed the connection, and thus the lock was left on the Postgres server.

I was able to recreate such behavior using the table mf_hecke_lpolys.

jwj61:
Great.  I will test it right away since I commented out one of the modifications (to be 100% sure it wasn't slowing things down).

edgarcosta:
ps: This PR doesn't fix the issue of leaving locks around while the connection has not yet been closed

edgarcosta:
@jwj61, I have just pushed one more commit that fixes a bug that I found today that is extremely relevant for tables with extras, please make sure that your rewrite is using this latest commit, otherwise you will need to add primary keys by hand

jwj61:
OK, I just pulled this and started it.

edgarcosta:
@jwj61, it has been 48h since you started the rewrite, and from the DB side, everything seems to have worked out alright, at least you are spending  99% of your time on doing math instead of waiting for data.

jwj61:
It seems that pari in sage gets bogged down after too many calls, so I killed it and restarted the computation a different way, and it finished in less than a day.  

The counts table gets wiped out by a rewrite.  I don't know if it should be rebuilt automatically.

PR #2947, opened 2019-04-10
CSS cleanup

davidfarmer:
Remove some un-necessary browser-specific css.

Also slightly cleaned up the main css file, eliminating tabs and leading spaces
where there is more than 6.

The check is that no page should look different!  That is unrealistic, but a specific
place to check is the word "Weight" on the left edge of the table on
   /ModularForm/GL2/Q/holomorphic/?search_type=Dimensions&dim=1

I checked in Firefox, Safari, and Chrome and did not see anything.

Fixes issue #2915 to the extent possible (apparently "sticky" still is browser specific).

### DIFF
--- a/lmfdb/templates/style.css
+++ b/lmfdb/templates/style.css
@@ -2,8 +2,6 @@
 
 {%- macro BORDER_RADIUS(R) -%}
   border-radius: {{R}}px;
-  -webkit-border-radius: {{R}}px;
-  -moz-border-radius: {{R}}px;
 {%- endmacro -%}
 
 {# end defs #}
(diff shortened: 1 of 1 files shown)


edgarcosta:
The changes all look good to me.
Compare:
http://blue.lmfdb.xyz/ModularForm/GL2/Q/holomorphic/?search_type=Dimensions&dim=1
vs 
http://beta.lmfdb.xyz/ModularForm/GL2/Q/holomorphic/?search_type=Dimensions&dim=1
they look identical to me in Safari, Chrome, and Firefox.
ps: they look weird in Safari, but this not a bug introduced by this PR

I will merge this by the end of the day if no one else raises an issue

edgarcosta:
I compared beta vs blue for a bit longer and couldn't spot any differences. Mergin

PR #2949, opened 2019-04-11
Displaying ST as a related object and making use of embedded_related_objects

edgarcosta:
This PR makes sure that in CMFs:
- we use `sato_tate_group` column to display ST as a related object
 - avoid duplicates if such an entry is also present in `related_objects`
 - using the column `embedded_related_objects` for embedded new forms

There will be a data change to make use of this PR, and so far everything should look the same

### DIFF
--- a/lmfdb/classical_modular_forms/web_newform.py
+++ b/lmfdb/classical_modular_forms/web_newform.py
@@ -271,7 +271,7 @@ class WebNewform(object):
         nf_url = ns_url + '/' + self.hecke_orbit_label
         if self.sato_tate_group:
             res.append(('Sato-Tate group \({}\)'.format(get_name(self.sato_tate_group)[0]),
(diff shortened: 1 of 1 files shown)


AndrewVSutherland:
The links are not quite right, on 11.2.a.a the ST group link should be

http://beta.lmfdb.org/SatoTateGroup/1.2.3.c1

but on http://127.0.0.1:37777/ModularForm/GL2/Q/holomorphic/11/2/a/a/ the link is

http://127.0.0.1:37777/ModularForm/GL2/Q/holomorphic/11/2/a/a/SatoTateGroup/1.2.3.c1

you need to omit the ModularForm/... prefix.

(sorry for the temporary mislabelling, I had the maass form issue open in an adjacent tab)

edgarcosta:
I fixed the typo :)

PR #2952, opened 2019-04-11
Fixing the usage of embedded_related_objects

edgarcosta:
A good test case is
citing Drew:

> ```
> sage: db.mf_newforms.lookup("71.1.b.a")["embedded_related_objects"]
> [[u'ArtinRepresentation/2.71.7t2.1c3'],
>   [u'ArtinRepresentation/2.71.7t2.1c1'],
>   [u'ArtinRepresentation/2.71.7t2.1c2']]
> ```
> 
> So on the first embbedded newform home page
> 
>     http://localhost:37777/ModularForm/GL2/Q/holomorphic/71/1/b/a/70/1/
> 
> out of the three Artin representations only 
> ArtinRepresentation/2.71.7t2.1c3
> should appear.

and on http://localhost:37777/ModularForm/GL2/Q/holomorphic/71/1/b/a/ all three should appear

### DIFF
--- a/lmfdb/classical_modular_forms/test_cmf.py
+++ b/lmfdb/classical_modular_forms/test_cmf.py
@@ -18,7 +18,7 @@ class CmfTest(LmfdbTest):
         assert '?search_type=Dimensions&dim=1' in data
         assert '?search_type=SpaceDimensions&char_order=1' in data
         assert "./stats" in data
-        data = self.tc.get("/ModularForm/GL2/Q/holomorphic/?search_type=SpaceDimensions",follow_redirects=True).data
+        data = self.tc.get("/ModularForm/GL2/Q/holomorph/ic/?search_type=SpaceDimensions",follow_redirects=True).data
         assert r'<a href="/ModularForm/GL2/Q/holomorphic/23/12/">229</a>' in data
         data = self.tc.get("/ModularForm/GL2/Q/holomorphic/?search_type=SpaceDimensions&char_order=1", follow_redirects=True).data
         assert r'<a href="/ModularForm/GL2/Q/holomorphic/18/4/a/">1</a>' in data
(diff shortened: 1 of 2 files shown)


edgarcosta:
Ups, I had forgotten to push...

AndrewVSutherland:
I'm looking at this now to see if the CMF test failure is a data problem or a code problem.

edgarcosta:
Seems a data problem to me, as I would expect 
```len(embedded_related_objects) = dim ```

and that is not the case:
http://beta.lmfdb.org/api/mf_newforms/?label=11.2.a.a

AndrewVSutherland:
The problem is the nulls are empty arrays rather than \N.  I can fix this, but if your code did

if self.embedded_related_objects:
    ....

all would be well (note that python treats an empty list as None, but a list of empty lists is not None)

edgarcosta:
I will do that then.

AndrewVSutherland:
I'll test again once Travis is done.  @edgarcosta I have modified the upload script to set embedded_related_objects to use \N rather than {} (for next time).

AndrewVSutherland:
test_embedded_invariants needs to be updated to not look for the genus 2 isogeny class 169.a on the embedded newform page http://127.0.0.1:37777/ModularForm/GL2/Q/holomorphic/13/2/e/a/, this is really an invariant of the Galois orbit, not the embedded newform.

One could argue that for embedded newforms of dimension 1, e.g.http://127.0.0.1:37777/ModularForm/GL2/Q/holomorphic/11/2/a/a/1/1/ we should  always show the related objects for the Galois orbit (in this case the elliptic curve isogeny class 11a now appears on the main newform page but not on the embedded newform page).  I could of course do this in the data but it would be better to do it in the code in case we change our minds (this is more a question of presentation than a mathematical one -- I don't have a strong preference).

edgarcosta:
I fixed the test and added an if statement for embedded forms that come from 1 dim galois orbit

Issue #2953, opened 2019-04-11
not-equal sign rendering as equal sign on Safari


AndrewVSutherland:
As reported by Mike Bennett, using any math mode command for not equal, including "\ne", "\neq", "\not =" will render as "=" on Safari.

To see an example run Safari (on a mac or ipad), go to any elliptic curve page (e.g. http://www.lmfdb.org/EllipticCurve/Q/11/a/2) and click on the discriminant knowl, where you will $\Delta \ne 0$ rendered as $\Delta = 0$.

alexjbest:
@AndrewVSutherland I assume this is #2798 again, see the links there.
It seems this is a recurring issue for them, but I think https://github.com/KaTeX/KaTeX/commit/ff5ebea681c0710805c32f62e7efe96c1b432938 fixes this. Note that this has been added to master since the last official katex release. So we can wait for release 0.10.2 (or whatever) and upgrade when that comes out to fix this.

Alternatively if we want to get these sorts of issues seen to faster we could start building our own KaTeX from master and serve that rather than loading from a CDN?

AndrewVSutherland:
@alexjbest In general I think we can live with minor annoyances (e.g a slash not appearing in the right place on some browsers), but a rendering problem that produces perfectly rendered but false mathematical output is not something we can live with, at least not without warning users about it, which is why I tagged this issue for Release 1.1 (which effectively means May 1).  If we can't fix it I think we will need to add a warning message that appears when the website is accessed via Safari.

alexjbest:
@AndrewVSutherland I completely agree, its hard to think of a worse bug for a math renderer to have (that still renders anything). I am amazed they haven't done a new hotfix release for this.

I have just verified that building the latest KaTeX from source (on my laptop for now). And hosting it from the static folder fixes this bug on Safari.

I have made a PR for this #2954, up to you if/when you want to merge it, but its there if the 1.1 deadline is looming over you.

edgarcosta:
@AndrewVSutherland, could we merge `master` to `dev` and posteriorly, after checking that we are all happy, to `web` so we can close this issue?

AndrewVSutherland:
Done -- let's hold off merging other PRs to master until we get this on web (or decide we don't want to).

AndrewVSutherland:
This is now fixed on www.lmfdb.org, closing.

PR #2954, opened 2019-04-12
Host our own copy of katex

alexjbest:
Currently at katex master  https://github.com/KaTeX/KaTeX/commit/b9deec098f35919b9ad77baa6b11b7593473d488

This fixes the Safari rendering bugs from #2953 , we may wish to revert to a hosted version of katex when they move to 0.10.2.


edgarcosta:
This indeed resolves  #2953 and we should merge it as a temporary fix.

Note, the only relevant diff is lmfdb/templates/base.html where we now load the katex from our repo.

@AndrewVSutherland, are you fine with this fix?

AndrewVSutherland:
@edgarcosta Yes I'm okay with this, and I agree that it makes sense to merge it so we have plenty of time to test it.  It will only take a small code change to undo it.  If you are comfortable that it is ready to merge go ahead (but note that Travis is not going to notice any rendering problems so aside from obvious errors like files missing, the only way to really test it is to look at pages).

Issue #2886, opened 2019-03-25
CM, RM, and QM 


davidfarmer:
There are genus 2 curves which (elsewhere in the LMFDB, such as
/L/degree4/)
are referred to as RM, CM, or QM.  But if you go to the home pages of those
curves, a non-expert would find it difficult to understand what RM, etc,
means, or that those curves have those properties. 

If those properties relate to the Sato-Tate group, then that would be
good to point out somewhere.

edgarcosta:
Honestly, we should also be able to search on them.
At the moment you need to use ```Sato-Tate identity component``` to search for those examples, which for me is almost as hard as doing mental arithmetic.

AndrewVSutherland:
Actually, you need to specify both the Sato-Tate identity component and that the Jacobian is Qbar-simple to get an exact correspondence, but it would be trivial to add a search for surfaces that have potential CM, RM, QM over Qbar using this correspondence, in fact you should be able to search on the geometric endomorphism algebra (which would also let you search for Jacobians Qbar-isogenous to products of elliptic curves with/without CM isogenous/non-isogeneous, 9 possbilities in total).

This would be easy to add, and the knowl for this could explain what RM/CM/QM mean and be linked to from the degree 4 L-function page.

edgarcosta:
>  and that the Jacobian is Qbar-simple to get an exact correspondence
I disagree somewhat.
I know what you have in mind, but this perhaps reveals that we first should agree on what we want to search. We want to search on "is" a RM/CM/QM abelian surface or "has"?

For me, an abelian surface "has" RM/CM/QM if we can embed a real field/a CM field/a quaternion algebra in End(A), and for the "is" we expect an isomorphism.

For example, 
* A = (non CM)^2, then is a QM abelian surface with QM by M_2(Q)
* A = (CM)^2, then A has CM
* If End(A) = B, a quaternion algebra where disc B != 1, then A also has CM and RM
* If End(A) = K and K is a quartic CM field, then A is a CM abelian surface, but also has RM.
* etc...

AndrewVSutherland:
I have added the column geom_end_alg to the table g2c_curves, which can take 9 values and I plan to add this as a search field.  Of the 9 values, the 4 that correspond to geometrically simple varieties are Q, RM, CM, QM.  The others are Q x Q, CM x CM, Q X CM, M_2(Q), M_2(CM).

[...]

PR #2956, opened 2019-04-13
Genus 2 curve search by geometric endomorphism algebra

AndrewVSutherland:
As discussed in #2886, this PR adds the ability to search on the geometric endomorphism algebra of the Jacobian of a genus 2 curve by type.  As explained in the newly created knowl http://beta.lmfdb.org/knowledge/show/g2c.geom_end_alg there are nine types, four of which arise for geometrically simple Jacobians (Q, RM, CM, QM).  This knowl could also be linked to from the degree 4 L-function browse page (but this change is not part of this PR, which only touches the genus 2 code).

The type of geometric endomorphism algebra is also shown in the properties box (the exact value of the geometric endomorphism algebra can still be found in the endomorphisms section of the curves home page as before).

While making this change I noticed an inconsistency in the ordering of the search options on the main search and refine search pages which I fixed.

A new column has been added to g2c_curves to support this search; the table needs to copied to proddb before the code is pushed to web (I will take care of this once this PR has been merged, the new data won't have any impact on the old code).

### DIFF
--- a/lmfdb/genus2_curves/main.py
+++ b/lmfdb/genus2_curves/main.py
@@ -44,6 +44,10 @@ real_geom_end_alg_to_ST0_dict = {
         'R':'USp(4)'
         }
 
+# End_QQbar tensored with QQ
+geom_end_alg_list = ['M_2(CM)', 'M_2(Q)', 'QM', 'CM x CM', 'CM', 'CM x Q', 'Q x Q', 'RM', 'Q']
+geom_end_alg_dict = { x:x for x in geom_end_alg_list }
+
 aut_grp_list = ['[2,1]', '[4,1]', '[4,2]', '[6,2]', '[8,3]', '[12,4]']
 aut_grp_dict = {
         '[2,1]':'C_2',
@@ -101,6 +105,8 @@ def index_Q():
     info["aut_grp_dict"] = aut_grp_dict
     info["geom_aut_grp_list"] = geom_aut_grp_list
     info["geom_aut_grp_dict"] = geom_aut_grp_dict
+    info["geom_end_alg_list"] = geom_end_alg_list
+    info["geom_end_alg_dict"] = geom_end_alg_dict
     title = 'Genus 2 Curves over $\Q$'
     bread = (('Genus 2 Curves', url_for(".index")), ('$\Q$', ' '))
     return render_template("g2c_browse.html", info=info, credit=credit_string, title=title, learnmore=learnmore_list(), bread=bread)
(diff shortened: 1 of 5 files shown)


davidfarmer:
This is good, but it still takes a bit of puzzling for a non-expert to put the pieces together.

If the properties box had "RM by Qsqrt5" instead of just "RM", that would help.
(and also explain why the same item has the value "RM" in the properties box
and "Qsqrt5" at the bottom of the page).  May want to change the E/Q case to be
consistent with this.  (I personally would prefer the field and not the "D" value to
be listed in the elliptic curve/Q case, but I am happy for experts to tell me that D
is the right way to do it.)

It is more convenient for the reader if the Qsqrt5 at the bottom of /Genus2Curve/Q/961/a/961/3 
(and at the bottom of /Genus2Curve/Q/12500/a/12500/1 twice)
was the number field knowl and not a link to its home page.

There is a knowl for "Endomorphism ring" which can be used in the heading "Endomorphism ring over Q".  

For the knowl link to the new knowl "Geometric endomorphism algebra", could the final
subheading be "Endomorphism algebra over Qbar"?  Or, could after the second one
there be a parenthetic knowl after it, similar to the (Complex multiplication) knowl
in the Invariants section of an elliptic curve /EllipticCurve/Q/27/a/3  ?

There could also be a Learn more about:  Endomorphism algebras.  Or what would be
even better, a learn more about that talks about Sato-Tate groups and endomorphism
algebras and their relationship, since that is interesting and I still don't have it straight in my head.
(In particular, there is a clear relationship between S-T groups and L-functions, but I am
fuzzy on what the endomorphism algebra says about L-functions.)

I don't think anything needs to be knowled on the degree 4 L-function page.  It is enough
that when you follow the link "RM genus 2 curve" on that page, then you find out
what it means.  That is the point of that issue and this pull request.

AndrewVSutherland:
@davidfarmer Thanks for the feedback.   I will update the PR as follows:

1. Add RM/CM field(s) to the properties box (I will also try to give more detail in the QM case).
2. Convert the number field links to knowls (this isn't really part of this PR, but it's easy to do).
3. Add knowl links for the endomorphism ring headings.
4. Try to make the endomorphism section and headings more consistent/readable.
5. Add a learn more about link for endomorphisms that will give an overview of the information we provide about endomorphism rings/algebras (including the connection between the real endomorphism algebra and the Sato-Tate group).

Regarding the CM discriminant for elliptic curves (over Q and other number fields), the value of D is the discriminant of the endomorphism ring (an order in an imaginary quadratic field), and this gives more information than the field discriminant (e.g.http://www.lmfdb.org/EllipticCurve/Q/27/a/1 has CM by the order of discriminant -27 in Q(sqrt(-3)), not the ring of integers).  The integer D uniquely identifies the endomorphism ring for elliptic curves over number fields (D=1 in the non-CM case), so I think it does belong in the properties box.

AndrewVSutherland:
I should add that I may not get to this for a few days.  Feel free to leave this PR open, or merge it and I will submit a new PR addressing 1-5 above.

edgarcosta:
I have just added some tests to test the new feature.
I will merge this once the tests pass.

PR #2961, opened 2019-04-18
blue favicon

### LINKED ISSUE #2945: Favicon color should be updated
If we are changing the color scheme, the background color of the favicon (small image that appears on a tab) should be updated to match.

edgarcosta:
This resolves #2945

However, should not be merged until we are ready to commit to the blue on www.lmfdb.org

Some side by side comparisons:

In my browser:
Screen Shot 2019-04-18 at 08 19 44

When they are zoomed in:
Screen Shot 2019-04-18 at 08 26 20

ps: I was lazy and used http://www.xiconeditor.com/ to create the new favicon.
If you want to go fancy you can use GIMP and perhaps you take advantage of favicon.ico.xcf (that i just deleted)
pps: the `db` is centered, but the`m` is not...


AndrewVSutherland:
Looks good, but we want to hold off merging this until Release 1.1 (this will PR change the icon on www.lmfdb.org as soon as it it merged to web and there are other changed to master that we do want to push to web before Release 1.1).  I'm going to close this PR temporarily and ask @edgarcosta to resubmit it when we are ready.

AndrewVSutherland:
FYI, Travis is likely going to fail on an ecnf test (due to a data issue that should be fixed within the hour).

AndrewVSutherland:
The ecnf data has been restored, so Travis should be happy the next time it runs.

PR #2965, opened 2019-04-20
Replace st_group knowls with more generic group knowls

kedlaya:
While reviewing Sato-Tate knowls, I noticed that some of them are redundant with knowls for arbitrary groups. Without judging the question of whether Sato-Tate groups should be filed under groups, I would propose that we use generic group knowls to refer to properties of generic groups (order, abelian).

### DIFF
--- a/lmfdb/galois_groups/templates/gg-search.html
+++ b/lmfdb/galois_groups/templates/gg-search.html
@@ -150,7 +150,7 @@ table.ntdata a {
 <th>{{KNOWL('gg.simple_name', title='Name')}}</th>
 <th>{{KNOWL('group.order', title='Order')}}</th>
 <th>{{ KNOWL('gg.parity', 'Parity') }} </th>
-<th>{{ KNOWL('gg.solvable', 'Solvable') }}</th>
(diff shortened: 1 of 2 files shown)


AndrewVSutherland:
@kedlaya Agreed.  Some of the more generic knowls weren't there when I originally added the st_group knowls, but I will go through the knowl links and make changes where appropriate.

AndrewVSutherland:
@kedlaya Nevermind, it looks like you are doing (or have done) what I was going to do (I misread your PR as an issue).  Just let me know when its ready to review.

kedlaya:
I made a couple of changes along these lines, which you can go ahead and merge; but you may want to take a look yourself to see whether there are other places that need similar treatment.

AndrewVSutherland:
@kedlaya Now I know why I was confused by this PR.  You need to submit it to LMFDB:master, not dev (all PRs go to master, we then push code to dev and web from master when appropriate -- this lets us control the timing of when changes appear on beta.lmfdb.org and web.lmfdb.org which are running code directly from github and will update in quasi-real-time (about 5 minutes).

Of if you prefer, I can take care of this in a new PR, just let me know

PR #2966, opened 2019-04-20
Replace st_group knowls with more generic group knowls 

kedlaya:
OK, hopefully this is the right target this time.


AndrewVSutherland:
Looks good, I'll merge as soon as Travis is done.

Issue #2697, opened 2019-01-19
search option to filter CM number fields fields


tscholl2:
It would be nice if there was an option to filter global number fields by whether they are CM.

jwj61:
Sure.  We are not displaying it now because the CM field was introduced in the postgres conversion.  If it is used, then we need to be sure to not push it to the main server until that uses postgres.

edgarcosta:
@jwj61 Okidoki, we have not yet started to use prod.lmfdb.xyz, before we do, most likely we will copy the nfields table.

@tscholl2, Given your shown interest in doing it yourself.
Here is how I would do it:
1- Display if a field is CM or under the invariants and properties box 
Under invariants the knowl to use is`ag.cm_field`
This would be done by editing:
- the template: https://github.com/LMFDB/lmfdb/blob/master/lmfdb/number_fields/templates/number_field.html
- and the function that calls the template https://github.com/LMFDB/lmfdb/blob/master/lmfdb/number_fields/number_field.py#L335

2- Add the search by CM
This amounts to edit the 3 files:
- https://github.com/LMFDB/lmfdb/blob/master/lmfdb/number_fields/templates/number_field_search.html
- https://github.com/LMFDB/lmfdb/blob/master/lmfdb/number_fields/templates/number_field_all.html
- https://github.com/LMFDB/lmfdb/blob/master/lmfdb/number_fields/number_field.py#L693

Let me know if you have any questions.
I'm happy to give more details and answer questions.

tscholl2:
Ok I think I am getting pieces of it. My plan for the search is to use a `<select>` with three options: `""`, `"true"`, and `"false"`.

Does the [`WebNumberField`](https://github.com/LMFDB/lmfdb/blob/fa39b0e46b1a5e16c2145641708649141bd892d9/lmfdb/WebNumberField.py) class also need a `is_cm` method?

jwj61:
Yes, add the method to WebNumberField and then use it.

I don't like blank as an option for the select.  See the CM select for elliptic curves.

AndrewVSutherland:
Agreed, blank should be avoided on selects.

edgarcosta:
To be clear, the default value of the select should always be "", what you put on the text is up to you.
Otherwise, clean_submit that cleans up the url by removing superfluous constraints doesn't remove that constraint.

For example, Elliptic curves does that wrong with `surj_quantifier` and even if you don't specify anything your URL always looks like:
[...]

Issue #482, opened 2015-10-28
Search number fields by root discriminant


AurelPage:
Feature request: it is annoying (and prone to error) to type lots of zeros when searching for large degree number fields by discriminant. It would be nice if we could search by root discriminant.

jwj61:
Done by #2970

PR #2970, opened 2019-04-21
Number fields: adds search/display for CM and root discriminant

jwj61:
From the search page for number fields, you can pick CM, not-CM, or leave blank, and can put in a range for the root discriminant of a field.

On the page of a number field, we say if the field is CM or not.  The root discriminant is listed on the page and in the properties box.  If it is larger than 9999, we put commas in the root discriminant.  For example

http://127.0.0.1:37777/NumberField/11.11.779296860357698534095483320599059601917194890273797415048589282424956242368505315786699569648072067627971942731199225.1

On this page on beta

http://beta.lmfdb.org/NumberField/11.11.779296860357698534095483320599059601917194890273797415048589282424956242368505315786699569648072067627971942731199225.1

notice that in the list of ramified primes, the big prime has an L at the end (courtesy of python).  That has been removed in this pull request.

Also on this page, the properties box is quite wide because the label is so long.  One solution would be to put an ellipses in the middle if the label has more than 30 characters.  It is already at the top of the page, so it would not be a big loss.

### DIFF
--- a/lmfdb/number_fields/code.yaml
+++ b/lmfdb/number_fields/code.yaml
@@ -41,6 +41,11 @@ discriminant:
   pari:  K.disc
   magma: Discriminant(K);
 
+rd:
+  sage: (K.disc().abs())^(1./K.degree())
+  pari: abs(K.disc)^(1/poldegree(K.pol))
+  magma: Abs(Discriminant(K))^(1/Degree(K));
+
 ramified_primes:
   sage:  K.disc().support()
   pari:  factor(abs(K.disc))[,1]~
--- a/lmfdb/number_fields/number_field.py
+++ b/lmfdb/number_fields/number_field.py
@@ -5,13 +5,13 @@ import ast, os, re, StringIO, time
 import flask
 from flask import render_template, request, url_for, redirect, send_file, flash, make_response
 from markupsafe import Markup
-from sage.all import ZZ, QQ, RealField, PolynomialRing, NumberField, latex, primes, pari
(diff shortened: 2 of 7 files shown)


jwj61:
This would address issues #482 and #2697

AndrewVSutherland:
Looks good, just a few comments:

- I assume you are going to create the nf.cm_field knowl once this is merged? (I see the link is there).  
- Searching on root discriminant ranges feels slow (but not horrible), do we need to add an index?  Perhaps we should wait and see what performance is like once this is live on beta.
- The property box in your example above is certainly wide :)  I would vote for shortening the label (with dots in the middle) if it is more than 30 digits.

None of the comments above are reasons not to merge, so I will do so.

jwj61:
I think we do need an index.

jwj61:
Knowl for cm field is now in place

AndrewVSutherland:
@jwj61 We have added an index for root discriminant on both the beta and production databases.  Queries with a narrow range on root discriminant are now quite quick, e.g.

http://www.lmfdb.org/NumberField/?start=0&rd=10-11

however queries with a larger range (where large means a range that postgres believes will produce a lot of results) are still slow, e.g.

http://www.lmfdb.org/NumberField/?start=0&rd=1-10

This is due to the fact that the query planner in postgres will prefer a linear index scan when it expects many results, because it assumes it can more quickly find the first 1000 results using a linear scan, but this may wind up causing it to scan all the quadratic fields before it gets the first 1000 results (or even the first 50).  This is a general issue that arises elsewhere and is not trivial to resolve (and not something we are going to be able to address for 1.1)

PR #2972, opened 2019-04-21
Extend local field browse table to degree 15

jwj61:
The change is the browse table at

  http://beta.lmfdb.org/LocalNumberField/

where the degree direction has been extended thanks to new additions


roed314:
Looks good to me!

PR #2984, opened 2019-04-25
Genus 2 friend finder

### LINKED ISSUE #2873: Genus 2 curve missing HMF friend

AndrewVSutherland:
This PR addresses issue #2873 by making genus 2 curves and isogeny classes look harder for friends, now any object with the same rational L-function stored in the database will appear as a friend; in particular this means that HMF and BMF friends that were missing are now found (but note that the HMF "friend" example given in the description of #2873 does not actually fit this description; as noted by @edgarcosta its rational L-function has degree 8 and is the square of a genus 2 L-function, and for the moment this means it does not meet our current bar for friendship).  Once this PR is merged I will close #2873 -- we will have an opportunity to revisit L-function friends more broadly when we migrate to an origin and Lhash/trace_hash independent L-function labeling system.

In the course of implementing this I converted the elliptic curve over imaginary field labels stored in g2c_endomorphisms (as factors of the base change of the Jacobian) to the new ecnf labelling system (which uses our newer bracket-free conductor labels).  I did not modify any of the ecnf labels in lfunc_instances, instead the genus 2 code uses the ecnf utility function convert_IQF_label to convert them, but eventually we should update these also.

To see the difference, look at an L-function like

http://beta.lmfdb.org/L/Genus2Curve/Q/20736/i/

which has 4 origins: an HMF, a corresponding ECNF isogeny class, a dimension 2 CMF, and a genus 2 isogeny class.  The HMF and CMF both all their friends, but the genus 2 isogeny class (and the curve in it) does not know the HMF, but with this PR it now does

http://beta.lmfdb.org/Genus2Curve/Q/576/a/
http://127.0.0.1:37777/Genus2Curve/Q/576/a/

with the change it know does.  Note that the L-function pages 

http://beta.lmfdb.org/L/Genus2Curve/Q/576/a/
http://beta.lmfdb.org/L/ModularForm/GL2/TotallyReal/2.2.8.1/holomorphic/2.2.8.1-9.1-a/0/0/

are different versions of the same L-function (one has zeros computed to higher precision, and in general the Z-plots may get flipped).  Dealing with duplications like this is a separate issue that cannot be properly dealt with until we complete the switch to origin-independent L-function labels.

Also note that the ECNF only knows its HMF friend, this will be fixed in a separate (Release 1.1) PR

For a more complicated example, try

http://beta.lmfdb.org/L/Genus2Curve/Q/20736/i/

where you will see that there are 5 origins, a BMF, and HMF, an imaginary quadratic EC isogeny class, a real quadratic EC isogeny class, and a genus 2 isogeny class (the two elliptic curves are both base changes of a pair of Qbar-isogenous elliptic curves over Q).  Then compare

http://beta.lmfdb.org/Genus2Curve/Q/20736/i/
http://beta.lmfdb.org/Genus2Curve/Q/20736/i/373248/1

to 

http://127.0.0.1:37777/L/EllipticCurve/2.2.8.1/9.1/a/
http://127.0.0.1:37777/Genus2Curve/Q/20736/i/373248/1

Before this PR neither the genus 2 curve or its isogeny class knew about either the HMF or BMF, now they both do.

### DIFF
--- a/lmfdb/genus2_curves/test_genus2_curves.py
+++ b/lmfdb/genus2_curves/test_genus2_curves.py
@@ -169,3 +169,37 @@ class Genus2Test(LmfdbTest):
                 ('M_2(CM)', '2916.b.11664.1')]:
             L = self.tc.get('/Genus2Curve/Q/?geom_end_alg={}'.format(endo))
             assert text in L.data
+
+    def test_related_objects(self):
+        for url, friends in [
+                ('/Genus2Curve/Q/20736/i/373248/1',
+                    ('L-function',
+                        'Isogeny class 20736.i',
+                        'Elliptic curve 576.f3',
+                        'Elliptic curve 36.a4',
+                        'EC isogeny class 2.0.8.1-324.3-a',
+                        'Bianchi MF 324.3.a',
+                        'Hilbert MF 2.2.24.1-36.1-a',
+                        'EC isogeny class 2.2.24.1-36.1-a',
+                        'Twists',)
+                    ),
+                ('/Genus2Curve/Q/20736/i/',
+                    ('L-function',
+                        'EC isogeny class 576.f',
+                        'EC isogeny class 36.a',
+                        'EC isogeny class 2.0.8.1-324.3-a',
+                        'Bianchi MF 324.3.a',
+                        'Hilbert MF 2.2.24.1-36.1-a',
+                        'EC isogeny class 2.2.24.1-36.1-a',)
+                    ),
+                ('/Genus2Curve/Q/576/a/',
+                    ('L-function',
+                        'EC isogeny class 2.2.8.1-9.1-a',
+                        'Modular form 24.2.d.a',
+                        'Hilbert MF 2.2.8.1-9.1-a',)
+                    )
+                ]:
+            data = self.tc.get(url).data
+            for friend in friends:
+                assert friend in data
+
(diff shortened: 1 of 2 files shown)


AndrewVSutherland:
Once this PR is merged I will update g2c_endomorphisms on proddb -- it is not strictly necessary for this to happen before the code is pushed to web, but it will look better if it does (otherwise we will have the same ecnf referred to by two different looking labels on the same page).

edgarcosta:
Given that we are doing this in some many other places, I will try to write a fcn that does this generically.

edgarcosta:
@AndrewVSutherland, are you okay with having just "Isogeny Class" for Elliptic curves, instead of "EC Isogeny Class", as it says when you look at the L-fcn?

AndrewVSutherland:
@edgarcosta Hmm, I think this is OK, but it might be nicer to change the L-function origin to say EC Isogeny Class.

edgarcosta:
@AndrewVSutherland I just added the tests, and won't touch the G2C codebase, as at the moment we have a mixture of things that match the L-function exactly and the factors of the L-function, and you only want a subset of those.
In other words, if you look at the L-function, e.g. http://127.0.0.1:37777/L/Genus2Curve/Q/20736/i/
in the G2C isogeny class, we list Origins and then only a subset of Origins of factors.

I will do a PR after this one is merged that rewrites some of the friend finders for ECNF, BMF, and HMF.

I will merge this as soon as the tests pass.

Issue #2983, opened 2019-04-25
Analytic conductor appears twice


davidfarmer:
In the search form for classical modular forms, bottom of
   http://beta.lmfdb.org/ModularForm/GL2/Q/holomorphic/
Analytic conductor is one of the fields, and so is Nk^2.  But both
of those are knowls, with the same content.

I see that they function differently: you can put an integer for Nk^2,
such as 400, but the Analytic conductor has to be given as a range.
But when I put "400" for Nk^2, the returned list has "A" values that
are not close to 400.

Either those should point to different knowls, or it should be noted that Nk^2 is not a good approximation when k is small.

AndrewVSutherland:
Good catch, these should be different knowl -- they are not the same thing.

AndrewVSutherland:
Addressed by #2985.

PR #2985, opened 2019-04-25
Fix for #2983

AndrewVSutherland:
This PR addresses #2983 by giving Nk^2 its own knowl, distinct from that of the analytic conductor.


AndrewVSutherland:
This is a trivial change that I will merge as soon as Travis finishes.

PR #2987, opened 2019-04-26
Galois names

jwj61:
When a Galois group appears with a "pretty" name X, we now (often) show it as "X (as nTj)".  This is used the properties box for an individual local or global field, on the main page for either type of field (using knowls, if possible, for both the abstract group and the nTj) and on search results pages for these two areas.  It is also used in dynamic field knowls for local or global number fields.

I did not touch the Artin section since the abstract group takes precedence -- we may produce as a permutation group acting on roots of a low degree polynomial, but we are really less interested in the particular permutation representation.

I also did not use it on Galois group pages since it seems that it would be more cumbersome there (e.g., when listing resolvents and other representations).  Sample pages below.

Galois group of a local field (group is in small group database so both pretty name and nTj are knowls):
  http://beta.lmfdb.org/LocalNumberField/5.8.0.1
  http://127.0.0.1:37777/LocalNumberField/5.8.0.1
Galois group of a global field (but the group is not in the small group database):
  http://beta.lmfdb.org/NumberField/9.1.32344469.1
  http://127.0.0.1:37777/NumberField/9.1.32344469.1
Global field search page (all S_4, but with 2 different nTj showing for it):
  http://beta.lmfdb.org/NumberField/?degree=6&galois_group=S4&ram_quantifier=exactly
  http://127.0.0.1:37777/NumberField/?degree=6&galois_group=S4&ram_quantifier=exactly
Global field where the Galois group does not have a pretty name (so just like before):
  http://beta.lmfdb.org/NumberField/21.3.3960879820606443846053527.1
  http://127.0.0.1:37777/NumberField/21.3.3960879820606443846053527.1

Also in this pull request are two invisible changes.  I removed a function which was no longer needed for displaying transitive groups, and fixed a potential bug when displaying a small group (if the function is fed a name to use, it sometimes did not use it -- I don't know of an instance where this currently comes up, but better to get it right).

This does not deal with error messages when searching.  It is a bit different, so will leave that for a separate pull request.

### DIFF
--- a/lmfdb/artin_representations/math_classes.py
+++ b/lmfdb/artin_representations/math_classes.py
@@ -4,7 +4,7 @@ from lmfdb import db
 from lmfdb.utils import url_for, pol_to_html
 from lmfdb.typed_data.standard_types import PolynomialAsSequenceTooLargeInt
 from sage.all import PolynomialRing, QQ, ComplexField, exp, pi, Integer, valuation, CyclotomicField, RealField, log, I, factor, crt, euler_phi, primitive_root, mod, next_prime
-from lmfdb.galois_groups.transitive_group import group_display_knowl, group_display_short, tryknowl
+from lmfdb.galois_groups.transitive_group import group_display_knowl, group_display_short
 from lmfdb.number_fields.web_number_field import WebNumberField
 from lmfdb.characters.web_character import WebSmallDirichletCharacter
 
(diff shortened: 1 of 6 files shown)


AndrewVSutherland:
I like it!  Especially on search results pages where you can see the same abstract group showing up as different transitive groups, e.g. http://127.0.0.1:37777/NumberField/?start=0&galois_group=S3&ram_quantifier=contained&ram_primes=2%2C3

PR #2989, opened 2019-04-27
Added a few more workshops.

jvoight:


jwj61:
I think the failure is my fault.  I am making data changes and got caught at a bad time.

jvoight:
This only changes a template, so maybe we should just ignore the failed checks and merge?

PR #2991, opened 2019-04-27
Remove word "download" from download link

jvoight:
Thought I caught all of these!


jwj61:
Same reason for the test failure.  The needed rewrite of the number field database is underway (1.8 million fields so far), but it will be hours before it finishes.

jwj61:
rewrite is complete.  The number field test should pass now.

Issue #2994, opened 2019-04-29
Field of Kronecker character


davidfarmer:
In the table at the bottom of this page:
    http://beta.lmfdb.org/ModularForm/GL2/Q/holomorphic/17/3/e/b/
there is a heading "Type".

1) "Type" is not a good name.  Is "Field" an acceptable heading?
(I do not think it is confusing to see "CM by Qsqrt-3" under the "Field" heading.)

2)  I deleted "this column shows" from the knowl for that column heading,
since the knowl was just a definition of what that column contains.

3) It took me a while to understand that the heading of the table was
"Inner twists" but the column heading mentioned above was only
relevant to those rows that are self-twists.   (Which is why there are
blank entries in that table.)  I don't have a suggested fix,
but somehow it should be more clear that the self twist (which does not
always exist) is special and more information is provided about it.

edgarcosta:
Here is a more illustrative example of the type column: http://beta.lmfdb.org/ModularForm/GL2/Q/holomorphic/120/1/i/a/

AndrewVSutherland:
I think changing for the "Type" column to "Self Twist" might address both 1 and 3 (and is more informative than "Field").  I would then edit the knowl to explain that inner twists which are self twists are either trivial or the Kronecker character of a quadratic file (which may be imaginary (CM) or real (RM)).

Does that sound reasonable to you @davidfarmer?

We could also put "no" in the entries that are currently blank, but personally I think blank is fine (and less busy).

davidfarmer:
I agree that Self twist is better than Field and addresses my concern.

AndrewVSutherland:
Addressed by #2998, closing.

PR #2998, opened 2019-04-30
Change self twist field column heading in inner twist table for CMFs

AndrewVSutherland:
This PR addresses #2994.


edgarcosta:
The single line change looks good to me. But I won't review the content.

AndrewVSutherland:
It would be good for @davidfarmer to review and close the issue if he is satisfied.  I also changed the knowl content.

davidfarmer:
I think this is clearer, and less confusing when not all rows have an entry.  Will merge.

Now the knowl cmf.self_twist_col should be renamed to cmf.self_twist .
The knowl should also mention that it is talking about holomorphic GL(2) modular forms.

AndrewVSutherland:
@davidfarmer I renamed the knowl to cmf.self_twist_field (the knowl cmf.self_twist defines a self twist and is already referenced by this knowl).  The knowl does refer to newforms, but I will move this to to the first sentence to make it clearer (after the rename finishes).

PR #3009, opened 2019-04-30
Typo

sehlen:


edgarcosta:
Then change Elliptic -> elliptic in midsentence looks good to me

Issue #3014, opened 2019-04-30
Specific knowls to replace


davidfarmer:
This is related to issue #2992 , but here I am listing pairs of knowls
which both exist but one of them should be deleted and replaced by
another.  Se that issue to understand the possible pitfalls.  In particular,
the content of the knowl which is to be deleted will actually survive,
so you have to first copy the good material into the bad knowl.

I am assigning myself to do this, but will not object if someone else does it first!

Feel free to add to this list.

Delete:   mf.maass.mwf.fricke
Replace by:   cmf.fricke

Delete:  mf.maass.mwf.atkinlehner
Replace by:  cmf.atkin-lehner

Delete:  lfunction.level
Replace by:  lfunction.conductor

pascalmolin:
Added to this list:
Delete character.dirichlet.induced

I have already removed all references and redirected them to the dupplicate character.dirichlet.induce

In my opinion, the content of character.dirichlet.induced can be discarded (I do not think the extra details are important in any context), but someone may disagree.

AndrewVSutherland:
We have automated tools that make this very easy, I just made the changes, see PR #3015

AndrewVSutherland:
@pascalmolin I just deleted character.dirichlet.induced (safe since it is not referenced anywhere, and I agree the additional content was superfluous).

@davidfarmer if you could just sanity check and then merge PR #3015, I will then push to dev/web, delete the knowls, and then close this issue.

davidfarmer:
I will check and merge.  I want to make sure that all the correct
content is in the knowls that will remain.

AndrewVSutherland:
The code with the updated knowl links is now in production and the knowls mf.maass.mwf.fricke, mf.maass.mwf.atkinlehner, lfunction.level, character.dirichlet.induced have been deleted.

Closing.

PR #3015, opened 2019-04-30
Knowl renaming to address #3014

AndrewVSutherland:
This PR makes the following knowl id changes to all source code and templates:

mf.maass.mwf.fricke -> cmf.fricke
mf.maass.mwf.atkinlehner -> cmf.atkin-lehner
lfunction.level -> lfunction.conductor

Once this is merged and pushed to web, the old knowls will be deleted.


davidfarmer:
I have organized the material in the knowls, so it is safe to delete the ones which are no longer needed.

AndrewVSutherland:
@davidfarmer Thanks.  We don't want to delete the knowls until the code is pushed to web -- I will take care of both.

PR #3016, opened 2019-05-01
removing print closing cursor

edgarcosta:
Removes 
```
print "closing cursor"
```

I also changed some formatting and white space around that print statement


roed314:
LGTM

PR #3018, opened 2019-05-01
pari code for kronecker symbol

pascalmolin:
provide parigp code to convert a real character to kronecker symbol representation


edgarcosta:
Here is an example: http://localhost:37777/Character/Dirichlet/13/12
and this looks good to me

PR #3021, opened 2019-05-01
Don't include deleted knowls when getting categories

roed314:
To test, look at http://localhost:37777/knowledge/ and check that some categories with only deleted knowls (e.g. [hecke_algebras](http://beta.lmfdb.org/knowledge/?category=hecke-algebra)) have disappeared and the counts have been updated to not include deleted knowls.


edgarcosta:
Looks good to me, I just added one word to the doc string.

PR #3030, opened 2019-05-02
Add try/except on local algebras

jwj61:
On beta, defining polynomials for some local fields are being changed.  A global number field tries to display local algebras at ramified primes, but it looks up the local field by defining polynomial.  Getting the data rewritten is taking longer than expected, so ...

This puts try/except around that code which looks up the local fields in the algebra of a global number field.  If something fails, it catches the error and the output to the user is that the data is not computed.  It is probably a good idea to make the change anyway so it doesn't rely on certain data being in the database.

A sample page where this comes up:
  http://127.0.0.1:37777/NumberField/22.0.178400853291024459894627.1
Currently on beta gives a server error.


edgarcosta:
LGTM

PR #3031, opened 2019-05-02
Short label

jwj61:
When the label of a number field is very long, this replaces part of it with an ellipsis in the properties box.  Otherwise, the properties box gets excessively wide.  The label is still there in its entirety on the page title.

An example:
  http://beta.lmfdb.org/NumberField/22.0.178400853291024459894627.1
  http://127.0.0.1:37777/NumberField/22.0.178400853291024459894627.1

An example which does not change
  http://127.0.0.1:37777/NumberField/3.3.49.1


edgarcosta:
LGTM

edgarcosta:
LGTM

PR #3048, opened 2019-05-06
Knowl rename

jwj61:
This just renames knowls

Prefix is removed from dq.X.extent for X in [artin, nf, lf, gg]

Changed name of the know explaining the Gap small group numbering


AndrewVSutherland:
The only test failure is a CMF test that needs to be updated to accomodate a data change (it is looking for a specific coefficient in a download that is now represented used a different basis).  I will fix the test shortly, but in the mean time I will merge this.

PR #3052, opened 2019-05-08
No bananas

jwj61:
This covers local fields, global fields, and Artin representations, with a small tweak to Galois groups to make them more uniform.

In any of these areas, enter something faulty in the box for a label.  The error message depends on whether it is the right format or not.

Second, go to the index page for an area, and append banana to the url (as if trying to directly access something by its label).  This should be caught as well.


JohnCremona:
testing now...

JohnCremona:
http://localhost:37777/ArtinRepresentation/3.2e3_7e5.7t3.1c1  (the example given next to the search-by-label box) now raises an error.  It's OK on beta and www.

Otherwise all is well -- I tried various different fruits on all four sets of pages in both the ways suggested.

Issue #3033, opened 2019-05-02
Change how multiplicities are shown for transitive groups


jwj61:
When looking at a Galois group page or knowl (or search results), other transitive groups appear as subfields, resolvent fields, and other representations.  In each case, we currently show multiplicity by repeating the item.  We should show multiplicities more compactly.  For example, 8T4 (which is D_4 as an octic) has 3 quadratic subfields which are 2T1, and we should show "C_2 x 3" instead of "C_2 as 2T1 x 3" (also adding the adjustment for abstract group vs permutation representation).  For that group, it is not too bad, but for 16T67, it gets bad, and there are much worse examples.  In expanding the degree, there will be cases where the same group appears over 100,000 times!

On the plus side, this will make the information easier to digest, take less space, and speed up loading of the relevant pages.  Using "x 3" for multiplicty would match how multiplicity is shown for subfields of a number field.

jwj61:
Done in #3063

PR #3063, opened 2019-05-10
Multiplicities in Galois group displays

jwj61:
This can wait for 1.2, but it is better to submit it now in case there is feedback.  It addresses issue #3033 

When listing things like resolvents, "subfields", or siblings for a Galois group, we want to explicitly give multiplicities rather than just repeat the item.  This is good when there is a modest number of repeated items, and essential when there is a large number.  It is used on the page displaying a group, on the search result page, and in the knowl for a transitive group.  Some of the examples below time out on beta.  Some probably work on the main site because it does not have the data for permutation representations in degrees bigger than 23.  One way to see the difference is to run with --debug, and then switch branches redisplaying the page.

Other changes: display of actual permutations has better spacing (e.g., list of generators on a group page), and we don't display character tables when the number of conjugacy classes is too large (they are currently computed on the fly, and it takes a long time, plus the display from gap cuts up the table).  This also consolidates some computations, like conjugacy classes, so they don't have to be done several times.

Here are some pages

Search results in degree 16:
  http://127.0.0.1:37777/GaloisGroup/?n=16
The page has multiplicities and contains knowls also showing multiplicities (e.g. C_2^3 which is a subfield of 16T3)

Search results in degree 22:
 http://127.0.0.1:37777/GaloisGroup/?n=22&count=100
probably doesn't load on beta.

A particularly hard group:
  http://127.0.0.1:37777/GaloisGroup/22T28
It is still slow, but not horrendous anymore.  It will speed up if we precompute conjugacy classes and store them in the database.

If/when this is merged, the data will need to be copied to the cloud just before merging it there.  It uses several new columns in the database.

### DIFF
--- a/lmfdb/galois_groups/main.py
+++ b/lmfdb/galois_groups/main.py
@@ -182,8 +182,8 @@ def render_group_webpage(args):
         data['chartable'] = ctable
         data['parity'] = "$%s$" % data['parity']
         data['cclasses'] = conjclasses(G, n)
-        data['subinfo'] = subfield_display(n, data['subs'])
-        data['resolve'] = resolve_display(data['resolve'])
+        data['subinfo'] = subfield_display(n, data['subfields'])
+        data['resolve'] = resolve_display(data['quotients'])
         if data['gapid'] == 0:
             data['gapid'] = "Data not available"
         else:
--- a/lmfdb/galois_groups/main.py
+++ b/lmfdb/galois_groups/main.py
@@ -172,10 +172,11 @@ def render_group_webpage(args):
             G = gap.SmallGroup(n, t)
         else:
             G = gap.TransitiveGroup(n, t)
-        if ZZ(order) < ZZ('10000000'):
+        cclasses = conjclasses(G, n)
+        if ZZ(order) < ZZ(10000000) and len(cclasses) < 21:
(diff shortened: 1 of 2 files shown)


AndrewVSutherland:
The single test failure is due to the ecnf data issue and unrelated to this PR.  Reviewing now.

AndrewVSutherland:
@jwj61 This is a huge improvement!  I vote we include it in 1.1 (and we really can't merge this if we don't include it because there are other 1.1 PRs in process that need to be merged and pushed to web for 1.1).

Is it safe to copy gps_transitive to the production database now? 




  Is it safe to copy gps_transitive to production now (and then I will merge this)

jwj61:
Yes, it is safe, modulo the fact that some group pages will time out until the code is merged as well.  But the odds of someone looking at one of those pages in the interim is pretty small.

AndrewVSutherland:
I just copied the file over and pushed to web.  Should be up within the next 5 minutes.

AndrewVSutherland:
Code and data are now live on www.lmfdb.xyz, can you take a quick look @jwj61 and make sure everything is Kosher?

jwj61:
Looks good.  Thanks.

Issue #3047, opened 2019-05-05
Inconsistent/incorrect error handling on bad input


AndrewVSutherland:
If you type an invalid label or identifier (e.g. "banana") you see a range of behviours

BMFs, HMFs, Maass Forms, genus 2, higher genus families, AV/FF, Dirichlet characters, Galois groups, and Sato-Tate groups all exhibit what I would consider the most desirable behavior: they flash an error and leave you on the main browse page where you can immediately type a corrected label (maybe your fingers just slipped).  For example

http://www.lmfdb.org/Genus2Curve/Q/banana

Elliptic curves over Q, number fields, local fields take you to an empty search results page and display an error message there (and you then have to navigate back to the main browse page if you want to reenter the label)

http://www.lmfdb.org/ModularForm/GL2/Q/holomorphic/banana/
http://www.lmfdb.org/EllipticCurve/Q/?label=banana&jump=curve+or+isogeny+class+label
http://www.lmfdb.org/NumberField/?natural=banana&search=Go
http://www.lmfdb.org/LocalNumberField/banana

ECNF and Artin Reps flash an error and then throw a server error,  These should definitely be fixed for 1.1:

http://www.lmfdb.org/ArtinRepresentation/?natural=banana&search=Go
http://www.lmfdb.org/EllipticCurve/?label=banana&jump=label

CMFs do something really bizarre, which is to transfer the bad label you typed into the level field and then tell you the level is invalid (but a valid label works fine).  This one should also definitely be fixed for 1.1:

http://www.lmfdb.org/ModularForm/GL2/Q/holomorphic/banana/

AndrewVSutherland:
Closed via #3064

PR #3064, opened 2019-05-10
No more bananas!

AndrewVSutherland:
This PR completes the work to address #3047.  You should now be able to enter "banana" or any other invalid label name into the search by label box on any of the main browse pages and see an error message flashed and no navigation (you should be left right where you were so you can retype the label you mistyped rather than being taken to a search results page where you cannot type a label).

There is one exception to the above statement.  Typing in "random" will yield "random" results: on some pages it will take you to a random object, on others it will complain that random is not a valid label, and on ECNF (for reasons I do not understand) it will do nothing (but any other bad label will flash an error).  I don't see this as a serious issue.

In addition to typing in banana (or other bogus labels) reviewers should also be sure to type in a valid label to make sure that still works (I have tested this but another set of eyes would not hurt).

### DIFF
--- a/lmfdb/abvar/fq/main.py
+++ b/lmfdb/abvar/fq/main.py
@@ -263,7 +263,7 @@ def by_label(label):
         validate_label(label)
     except ValueError as err:
         flash(Markup("Error: <span style='color:black'>%s</span> is not a valid label: %s." % (label, str(err))), "error")
-        return redirect(url_for(".index"))
+        return redirect(url_for(".abelian_varieties"))
     g, q, iso = split_label(label)
     return redirect(url_for(".abelian_varieties_by_gqi", g = g, q = q, iso = iso))
 
--- a/lmfdb/abvar/fq/main.py
+++ b/lmfdb/abvar/fq/main.py
@@ -263,7 +263,7 @@ def by_label(label):
         validate_label(label)
     except ValueError as err:
         flash(Markup("Error: <span style='color:black'>%s</span> is not a valid label: %s." % (label, str(err))), "error")
-        return search_input_error()
+        return redirect(url_for(".index"))
     g, q, iso = split_label(label)
     return redirect(url_for(".abelian_varieties_by_gqi", g = g, q = q, iso = iso))
 
--- a/lmfdb/artin_representations/main.py
+++ b/lmfdb/artin_representations/main.py
@@ -45,7 +45,7 @@ def parse_artin_label(label):
     if LABEL_RE.match(label):
         return label
     else:
-        raise ValueError("input %s is not in a valid form for an Artin representation label, such as 9.2e12_587e3.10t32.1c1"% label)
+        raise ValueError("input %s is not in a valid form for an Artin representation label, such as 9.2e12_587e3.10t32.1c1")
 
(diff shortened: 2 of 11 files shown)


AndrewVSutherland:
FYI, I made a couple of tweaks to error message formatting in places where the error message was already being displayed correctly to make sure the bad label is "highlighted" (meaning displayed in black to contrast with the red text of the error message).  This is the only reason artin_representations/main.py changed, for example.

AndrewVSutherland:
There are some test failures in CMF and ECNF due to changes in the error handling behavior, I'll fix them tomorrow morning.

AndrewVSutherland:
OK, I believe this PR should now be ready.  Travis is still going to fail on one ecnf test due to a temporary data problem in ec_nfcurves (currently all the curves over 2.0.3.1 have the same record, see http://beta.lmfdb.org/EllipticCurve/?field=2.0.3.1) -- this caused me a fair bit of confusion which is why you see a ridiculous number of commits above.  In the course of fixing the tests I also fixed another existing production bug, which is a server error on labels that are well-formed but do not correspond to an elliptic curve, e.g. http://www.lmfdb.org/EllipticCurve/Q/11/a/7

The data problem can be quickly fixed by copying the records back from production, I'm just waiting for confirmation from @JohnCremona that it is safe to do this (I don't want to interfere with any other work he may currently be doing on the table).

JohnCremona:
Indeed this was some fallout from mistakes I made yesterday which corrupted ecnf data for 3 fields, 2.0.3.1 and two cubics.  The cubics are already fixed and the quafratic one is having its data reuploaded right now.  I'll post here when that is finished.  It will take a while since there are 77000 curves to put back (the old records are already deleted) and they are taking 25s/100.  Sorry for this delay.

jwj61:
Are you loading them one at a time?  For number fields I just do a single insert_many with all of them.  The insert does not take long, and then you only rebuild indexes once.

JohnCremona:
@jwj61 no there will be a single insert_many.  But there is work done on each curve, which I usually hardly notice since I do not usually upload so many at a time (this is more than 10% of all the ecnfs).
For each curve I create is as a Sage elliptic curve, compute torsion, test for cm, test for base-change and compute local data.  Now you will suggest that I should compute all that offline and store it (as I do for some of the other data for each curve), and you would be right.
If people prefer we could also copy the records *for this field only* from prod; the only glitch in the base_change data for this field was that 4 newly add curves which were not b.c. had that field set to {} instead of [].

jwj61:
I see, I didn't realize you were doing computations as part of the process.  It seems that what you are doing is equivalent to doing it offline.

AndrewVSutherland:
I can confirm that the only test that failed is the expected failure in the ECNF tests where it is looking for a curve over 2.0.3.1 that isn't there (this test will pass after John Cremona's update finishes and is unrelated to the changes I made).

JohnCremona:
If you don't want to wait you could change that test....

@jwj61 yes, equivalent except that re-uploading would be a lot faster if this extra data was kept somewhere.  I'll look into that since I really like to be able to re-upload data.

JohnCremona:
The upload of curves over 2.0.3.1 will have finished about an hour from now.

PR #3068, opened 2019-05-11
auto add numstats if missing

edgarcosta:
If you browse to a page that requires numstats, and these have not been added, it tries to add them.
However, one shouldn't rely on this to make sure that all numstats are there.
Furthermore, while browsing on the live versions, most likely this won't do anything as gunicorn will timeout.


AndrewVSutherland:
This is awesome, thank you for making our lives just a little bit easier!

PR #3069, opened 2019-05-11
Beta sidebar changes for Release 1.1

AndrewVSutherland:
This PR makes some minor changes to the beta sidebar for Release 1.1 (there are no changes to the production sidebar).

The only functional changes are that Hecke characters and L-function operations have been reverted to future status -- neither of these pages are currently functional so there is no reason to provide links to them.  Both pages are still accessible via URL at http://beta.lmfdb.org/Character/Hecke/ and http://beta.lmfdb.org/TensorProducts/) and they can easily be restored to beta status once they are in a working state.

### DIFF
--- a/lmfdb/homepage/sidebar.yaml
+++ b/lmfdb/homepage/sidebar.yaml
@@ -50,7 +50,7 @@
           url_for: "zeta zeros.zetazeros"
         - title: Operations
           url_for: "tensor_products.index"
-          status: beta
+          status: future
         - title: dummy
           status: future
         - title: Local L-functions
--- a/lmfdb/homepage/sidebar.yaml
+++ b/lmfdb/homepage/sidebar.yaml
@@ -65,6 +65,7 @@
    parts:
      - title: GL(2)
        style: rotate
+       status:
        part2:
          - title: Classical
            url_for: "cmf.index"
@@ -84,13 +85,13 @@
            url_for: "modlmf.modlmf_render_webpage"
            status: future
            colspan: onecolumn
-         - title: Hecke Algebra 
+         - title: Hecke algebra 
            url_for: "hecke_algebras.hecke_algebras_render_webpage"
            status: future
            colspan: onecolumn
      - title: GL(3)
        style: rotate
-       status: beta
+       status: future_rotated
        part2:
          - title: Maass
            status: future
@@ -100,14 +101,12 @@
            colspan: onecolumn
      - title: Other
        style: rotate
-       status: beta
+       status: future_rotated
        part2:
          - title: Siegel
            status: beta
            url_for: "smf.index"
          - title: U(2,1)
-           status: future
-         - title: Other
            status: future
 
 4Var:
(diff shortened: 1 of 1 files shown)


AndrewVSutherland:
Travis failure looks like a side effect of PR #3068, although the fact that there were stats missing that were hit by a number field test is interesting in and of itself.  I'm running tests now on legendre with edit privileges (so the missing stats should be created), but @edgarcosta and @jwj61 could you take a look and let me know if you think my diagnosis is correct?

Of course this has nothing to do with the sidebar changes in this PR.

AndrewVSutherland:
Number field stats are currently broken on beta http://beta.lmfdb.org/NumberField/stats (but OK on production), which explains the error.  But it looks like is it failing because an expected column count is missing, not because of a numstats error:

```
    @nf_page.route("/stats")
    def statistics():
        # FIXME use StatsDisplay
        fields = db.nf_fields
        title = 'Global Number Field Statistics'
        bread = [('Global Number Fields', url_for(".number_field_render_webpage")),
                 ('Number Field Statistics', '')]
        init_nf_count()
        ntrans = [0, 1, 1, 2, 5, 5, 16, 7, 50, 34, 45, 8, 301, 9, 63, 104, 1954,
                  10, 983, 8, 1117, 164, 59, 7, 25000, 211, 96, 2392, 1854, 8, 5712]
        degree_stats = fields.stats.column_counts('degree')
        n = [degree_stats[elt + 1] for elt in range(23)]
    
        degree_r2_stats = fields.stats.column_counts(['degree', 'r2'])
        nsig = [[degree_r2_stats[(deg+1, s)] for s in range((deg+3)/2)]
               for deg in range(23)]
E       KeyError: (23, 1)

lmfdb/number_fields/number_field.py:221: KeyError
```

AndrewVSutherland:
It looks like the code expects there to be a column count for degree=23, r2=1, but in fact there are no such fields in the database, so I think in fact the stats are correct and perhaps the statistics code needs to be changed to handle this case.

AndrewVSutherland:
number field stats are now fixed (just asking for the missing column count sufficed to create it).

jwj61:
I rewrote the database from legendre to populate a searchable column for regulator.  Currently, a necessary step after a rewrite is to regenerate the counts, which I had not done yet.

AndrewVSutherland:
@jwj61 Thanks for the explanation!  Now I understand why things broke.  Just to be clear, the regulator search option won't be included in Release 1.1 (the number fields data on proddb won't have a searchable regulator column in it until we copy the updated tables over and we definitely don't want to do today).

jwj61:
I take it back.  @edgarcosta change which sped it up considerably uses the stats database, and that is the where it is failing (i.e., it pulls from db.nf_fields.stats and the array does not have all of the expected values).

jwj61:
I was not expecting the regulator search to be in 1.1.

JohnCremona:
Since this isn't merged yet, is there any objection to my adding an additional commit which deletes the greyed-out Q-curves entry?

jwj61:
I thought it was already decided to do this.

John

AndrewVSutherland:
@jwj61 It was and it has been done (@JohnCremona  was looking at the wrong branch, which he realized shortly after posting that comment and then deleted it).

PR #3070, opened 2019-05-11
LMFDB Release 1.1

AndrewVSutherland:
This PR adds an LMFDB version number to the footer (now set to Release 1.1).  This PR should be merged after #2961 and #3069.

### DIFF
--- a/lmfdb/app.py
+++ b/lmfdb/app.py
@@ -85,7 +85,7 @@ def ctx_proc_userdata():
     vars['title'] = r'LMFDB'
 
     # LMFDB version number displayed in footer
-    var['version'] = LMFDB_VERSION
+    vars['version'] = LMFDB_VERSION
 
     # meta_description appears in the meta tag "description"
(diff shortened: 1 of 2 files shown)


AndrewVSutherland:
Travis failure is due to a numberfields stats problem unrelated to this PR (contrary to what I wrote earlier, number field stats are broken on beta again, although they were working briefly).

JohnCremona:
If nf stats work like ecnf stats then when any are missing (e.g. if you have just cleared all stats and counts) the only way to recreate them is to run locally with editing rights and visit every stats or count related page.  This appears to be  a design decision though not mine: I would prefer to have one function to run to recreate all stats and counts at once. (Using the refresh stats function is not useful here as it only recounts the counts which are there. )

AndrewVSutherland:
@JohnCremona Agreed, this is something we will look into next week.  For today I just want to get number field stats working on beta and be sure that the the recent numstats code change was not responsible for the problem before pushing Release 1.1 to web.  @edgarcosta and I are looking at this now.

PR #3071, opened 2019-05-11
fix stats

edgarcosta:
A patch to fix the issue mentioned in #3070 

Again, this is not the right way to do stats! We should be using StatsDisplay


AndrewVSutherland:
@edgarcosta Thanks, this fixes the immediate problem, but before I merge it I would like to understand how the code worked before and why it broke.  What changed?

edgarcosta:
The reason that the zero counts were there before, is that the old page that did 5k+ counts put them there.
While now we are relying on stats which just loops over distinct values.

Issue #3072, opened 2019-05-11
Next button on number field browse page broken


AndrewVSutherland:
If you do any number field search with more than 50 results there is no way to see results past the first 50 because the next button simply redisplays the first 50 (on both beta and web).  I'm not sure when this happened, but it should be easy to fix.

edgarcosta:
start is being converted to None

@roed314 will look into this later in the afternoon

AndrewVSutherland:
Now fixed on beta and production via #3075.

edgarcosta:
everything has been fixed.

I also fixed the data issue in other instances.

PR #3075, opened 2019-05-11
Add count hidden input to number fields

roed314:
To test, try clicking the "Next" button on a number field search results page.  Addresses #3072.

### DIFF
--- a/lmfdb/number_fields/templates/number_field_search.html
+++ b/lmfdb/number_fields/templates/number_field_search.html
@@ -6,6 +6,7 @@
 
 <form id="re-search" onsubmit="cleanSubmit(this.id)">
 <input type="hidden" name="start" value="{{info.start}}"/>
+<input type="hidden" name="count" value="{{info.count}}"/>
 <table border="0">
 
 <tr>


AndrewVSutherland:
Looks good, thanks @roed314.  I'll merge as soon as Travis is finished.

AndrewVSutherland:
pyflakes is unhappy about an unused variable in the ecnf script @JohnCremona updated a few hours ago, but all other tests passed.  Merging.

AndrewVSutherland:
@JohnCremona Actually, in addition to the unused Json import, pyflakes is complaining that curve_data_filename is never used (is that intended?) and it looks like there might be a bug in check_database_consistency, the variable number_1_only_keys is not defined.

JohnCremona:
Sorry about the pyflakes issues in that script file.
Recently, wrapping arrays in Jason() was required before insertion, not it is not allowed, so that import can go.
The selection of filenames in the upload function is regularly changed by commenting stuff out.
The check_database_consistency function is hardly needed now that schemas exist, but in fact updating this function for postgres was quite helpful since it is not easy to find out what the current schema actually is, so running this alters me to changes.  It used to be true that some fields (columns) only existed for one curve in each class ("number 1"), but now that is not allowed, though instead some columns are allowed to be null and non-existent when inserting while others are not.  I don't know how to determine which are which, but now the columns which used to be allowed to be absent for number>1 curves are set to None instead, and the list of number_1_keys was deleted -- but not one reference to it so that was a bug.
I have fixed these and a trivial PR is imminent.

Issue #3086, opened 2019-05-14
EC/Q stats broken on beta


AndrewVSutherland:
See http://beta.lmfdb.org/EllipticCurve/Q/stats

But http://www.lmfdb.org/EllipticCurve/Q/stats wortks fine.

JohnCremona:
I will fix this, it is easy:
       max_sqrt_sha = max_sha.sqrt() # exact!
Here max_sha is now an int where it used to be a Sage integer, and python ints have no sqrt() method.

AndrewVSutherland:
@JohnCremona It seems to be working now, but I don't think the code changed, how did you fix it?

JohnCremona:
Fixed -- PR in a minute -- but I am puzzled since the code I changed works OK on www and also there is a test which loads this page so why did the test not fail?  The code extract the maximum value of the 'sha' column of ec_curves, like this:

sage: db.ec_curves.max('sha')
5625
sage: type(_)
<type 'sage.rings.integer.Integer'>

and it appears that sometimes the result delivered is an int (hence crash) and sometimes it is in Integer (no crash).  So while my PR will stop this happening again, it does not really answer the question Why?

JohnCremona:
I didn't fix it!  At least, only with PR #3088 which is not yet no beta.  As I said above, there is a random feature here which is very mysterious:   the max('sha') is sometimes an int and sometimes not.

The type of ec_curves.max('sha') is sometimes int and sometimes Integer, while the type of all the 'sha' column entries is int.

JohnCremona:
Also, if I look at ec_curves.stats there is a row which seems to be saying that the max of the sha column is 49, which it is not.  I have no idea what is going on here.

AndrewVSutherland:
@JohnCremona We're looking into it now.

roed314:
There is a column "threshold" which was set to 1000 on that row.  The threshold is intended for the following situation: if you ask for statistics on a column with a few common values but many uncommon values, you might want to only include values where there are at least N rows.  That was done here, and indeed 49 is the maximum size of sha where there are least 1000.

I don't know why this threshold was getting set though.  I've deleted those rows since I was confused what what the problem was.

PR #3088, opened 2019-05-14
fix for ec/Q stats

JohnCremona:
Fix for #3086 : ensure that the max of the sha column is converted to Integer before having its sqrt extracted.


AndrewVSutherland:
Looks good, I'll merge as soon as Travis finishes.

AndrewVSutherland:
Travis is failing because a number field test is calling galois_sib_data which expects the repns column to exist in gps_transitive, but it apparently no longer exists on beta, did you remove it @jwj61 ?  If so, it looks like some code changes are needed.

In any case this test failure has nothing to do with this PR.

jwj61:
Yes I did.  Working on it ...

PR #3089, opened 2019-05-14
nf_fields_new, coeffs -> nf_fields, coeffs_array

edgarcosta:
I created the column coeffs_array where coeffs is stored in array format, we can now use that column to run the verification script.


jwj61:
The test failure is fixed by #3090

Issue #3054, opened 2019-05-08
Add index by default to counts tables


edgarcosta:
By default, we should do
```
CREATE INDEX tablename_counts_cols_vals_split ON tablename_counts USING
btree (cols, values, split);
```
when creating a table and then rename them accordingly when reloading table.

See #2971

edgarcosta:
Also, 
```
CREATE INDEX tablename_counts_cols_split ON tablename_counts USING
btree (cols, split);
```

PR #3093, opened 2019-05-14
Pulling number field stats from the statistics table.

jwj61:
This should only affect the number field statistics page.  It should be unchanged except for the last table, which was still displaying information from the old stats table, so totals of numbers of fields has increased in some places.

I attempted to add the needed statistics to the nf_fields.stats table, but every time the page is rendered, there is logger output like:

INFO:nf_fields@2019-05-14 16:33:47,349: SELECT values, count FROM "nf_fields_counts" WHERE cols = '["class_number", "degree"]' AND split = false ran in  0.112394094467s  [database.py:394]

I don't know why, but maybe someone else can figure out how to add the needed data to the nf_fields.stats database.

### DIFF
--- a/lmfdb/backend/database.py
+++ b/lmfdb/backend/database.py
@@ -3848,7 +3848,7 @@ class PostgresStatsTable(PostgresBase):
             one_col = False
             cols = sorted(cols)
         if constraint is None:
-            ccols, cvals, allcols = [], [], cols
+            ccols, cvals, allcols = Json([]), Json([]), cols
         else:
             ccols, cvals = self._split_dict(constraint)
             allcols = sorted(list(set(cols + constraint.keys())))
--- a/lmfdb/number_fields/number_field.py
+++ b/lmfdb/number_fields/number_field.py
@@ -243,7 +243,6 @@ def statistics():
     has_hdeg_stats = nfstatdb.column_counts('degree', {'class_number': {'$exists': True}})
     has_hdeg = [has_hdeg_stats[deg+1] for deg in range(23)]
     has_h = sum(has_hdeg[j] for j in range(len(has_hdeg)))
-    # old has_h = fields.count({'class_number': {'$exists': True}})
     hdeg = [[{'cnt': comma(hdeg[nn][j]),
               'prop': format_percentage(hdeg[nn][j], has_hdeg[nn]),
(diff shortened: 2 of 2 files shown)


AndrewVSutherland:
@jwj61 I think this is just an informational message that is prineted for all queries that take more than 0.1s, but @edgarcosta can confirm.  In any case, it is harmless (hence the INFO) and should not have any impact on the functionality, other than possibly identifying a performance issue (but 0.11s seems fine).

edgarcosta:
In any case I will add an index on cols and split, see #3054

jwj61:
I thought the messages meant that counts were taking place since there are also messages which say 

INFO:nf_fields@2019-05-14 16:33:41,170: Statistics already exist [database.py:4415]

Some calls to the stats table produce the "already exist" and others produce messages like the one above.  So, I assumed that that meant that some counts are not being found.  Also, the earlier message references nf_fields_counts, and I didn't think the new code called that at all.

edgarcosta:
With my last commit you should no longer see:
```
INFO:nf_fields@2019-05-14 16:33:41,170: Statistics already exist [database.py:4415]
```

Regarding 
```
INFO:nf_fields@2019-05-14 16:33:47,349: SELECT values, count FROM "nf_fields_counts" WHERE cols = '["class_number", "degree"]' AND split = false ran in 0.112394094467s [database.py:394]
```
this was because there were 45k+ rows matching that criteria.
It seems that it had an entry for every possible tuple for the values of ["class_number", "degree"]. maybe because someone ran
```db.nf_fields.add_stats(["class_number", "degree"])```
anyhow, we deleted everything and repopulated it again by loading it locally

AndrewVSutherland:
I just test it and it looks good (no more extraneous INFO messages and the page reloads pretty quickly).  I'm happy to merge this once travis finishes.

jwj61:
I probably did the erroneous add stat command while trying different combinations.  Thanks for cleaning it up.

AndrewVSutherland:
FYI, we cleared all counts using  db.nf_fields.stats._clear_stats_counts() and then running start-lmfdb.py locally (with editor privileges) and refreshing the stats page to recreate the counts that are actually being used (this only took a few minutes).

Be aware that if/when you do this stats will be unavailable until you recreate them (so someone trying to access stats on beta may get a server error).

PR #3094, opened 2019-05-14
Regulators

jwj61:
This adds the ability to include regulators in number field searches.  Just go to the number field main page and include regulators in a search.

In the process, the field was copied from the number field extras table to the search table, and was renamed regulator (from reg) to avoid a conflict.  So, there will need to be a data upload to go with this if/when it is ultimately pushed to the web.


jvoight:
Excited for this, but waiting for the data upload before merging.  JV

jwj61:
Probably the sequence of events is merge, then data to cloud version, then merged to cloud.

AndrewVSutherland:
@jwj61 Agreed, but there is a PR from @edgarcosta in the works to address some issues with the db_backend functions that we use to move data from beta to production that I would like to get merged and pushed to web before we move the updated nf_fields table to production (it will ensure that the process goes more smoothly, and given the time it takes to copy the nf_fields table and rebuild all the indexes, I'd like it to go smoothly).  So I'd like to hold off merging this for the moment.

AndrewVSutherland:
Actually, now that I think about it I'm not sure I do agree, is there any reason not to move the data first?  The old code obviously still works with the new data format.

jwj61:
There is no problem in moving it.  The only difference in the data is that a column is duplicated (reg and regulator) as a step toward moving it from extra to searchable.

AndrewVSutherland:
Great, then let's plan to move it first, which I'll do as soon as Edgar has the PR I mentioned ready to go.

AndrewVSutherland:
nf_fields has now been copied to the production database, merging.

PR #3097, opened 2019-05-17
improved wording on bmf space pages

JohnCremona:
The improves the wording of the preamble to the table showing the Bianchi newforms at a single level.  Almost all the newforms in the database at present have rational coefficients (dim=1) but some over 2.0.4.1 have quadratic coefficients (dim=2).

Examples:

1 rational only:
http://localhost:37777/ModularForm/GL2/ImaginaryQuadratic/2.0.4.1/200.2

2 rational:
http://localhost:37777/ModularForm/GL2/ImaginaryQuadratic/2.0.4.1/194.2

1 of each:
http://localhost:37777/ModularForm/GL2/ImaginaryQuadratic/2.0.4.1/377.1

1 quadratic only:
http://localhost:37777/ModularForm/GL2/ImaginaryQuadratic/2.0.4.1/485.2

1 rational, 2 quadratic:
http://localhost:37777/ModularForm/GL2/ImaginaryQuadratic/2.0.4.1/1521.2

### DIFF
--- a/lmfdb/bianchi_modular_forms/bianchi_modular_form.py
+++ b/lmfdb/bianchi_modular_forms/bianchi_modular_form.py
@@ -293,6 +293,9 @@ def render_bmf_space_webpage(field_label, level_label):
                     'cm': cm_info(f.get('CM','?')),
                     } for f in newforms]
                 info['nnewforms'] = len(info['nfdata'])
+                # currently we have newforms of dimension 1 and 2 only (mostly dimension 1)
+                info['nnf1'] = sum(1 for f in info['nfdata'] if f['dim']==1)
+                info['nnf2'] = sum(1 for f in info['nfdata'] if f['dim']==2)
                 properties2 = [('Base field', pretty_field_label), ('Level',info['level_label']), ('Norm',str(info['level_norm'])), ('New dimension',str(newdim))]
                 friends = [('Newform {}'.format(f['label']), f['url']) for f in info['nfdata'] ]
 
--- a/lmfdb/bianchi_modular_forms/bianchi_modular_form.py
+++ b/lmfdb/bianchi_modular_forms/bianchi_modular_form.py
@@ -117,6 +117,7 @@ def bianchi_modular_form_postprocess(res, info, query):
                            ('Search Results', '.')],
              learnmore=learnmore_list,
              properties=lambda: [])
+
 def bianchi_modular_form_search(info, query):
     """Function to handle requests from the top page, either jump to one
     newform or do a search.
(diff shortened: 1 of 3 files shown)


AndrewVSutherland:
In looking at these pages the question that immediately comes to mind when you see the words "in the database" is whether the results are complete or not (e.g. in case 1, are there any other newforms with rational coefficients at this level).  I think the answer is yes, in which case I would drop the words "in the database", but it was actually slightly hard for me to tell for 2 reasons:

- The learnmore box is missing from the pages in question, and I notice it is also missing from the home pages of Bianchi newforms.  I think it should appear on all BMF pages (as it does for CMFs, HMFs, and elsewhere in the LMFDB), not just on the top level browse page.

- The completeness knowl doesn't make it clear whether, in the cases where we have newforms with quadratic coefficients (which I think is just for newforms over Q(i)), whether we have all of them or not.

If it is indeed the case that whenever we list newforms on these pages with rational/quadratic coefficients we are listing all of them, I would suggest dropping the words "in the database".

Finally, two minor wording nitpciks:

- I would write "newforms" rather than "cuspidal newforms", since according to http://www.lmfdb.org/knowledge/show/mf.bianchi.newform, newforms are cuspidal by definition, and this is consistent with the usage on the main browse page (which just talks about newforms, not cuspidal newforms), and with the CMF pages.

- I of course know exactly what you mean when you write "quadratic coefficients" but I can imagine some (or at least some pedants) might not.  It might be better to write "quadratic coefficient field".

JohnCremona:
Thanks for the detailed feedback.  I will follow your suggestions, including adding more links to the rcs knowls.  The whole point of this to to make things clearer!  (You are right that "in the database" is redundant.  I could easily also add code to check whether the dim 1, or dim 1 & dim2 together for Q(i), exhaust the newspace, and when they do not, add a sentence like "newforms with coefficient fields of degree greater than {1,2} are not yet in the database."

JohnCremona:
I'll update the PR shortly.  I have edited the completeness knowl quite a bit and hope it is clearer.  The Learnmore stuff does now appear on individual newform pages and space pages.

The logic behind trying to say what is/is not there on the space pages got too much so I have simplified it (old template code just commented out).  We do *sometimes* have to say "in the database" when we know that there are newforms in the space which we do not have (which the code does now test for).  Usually this is whenever the number of rational newforms is less than the total new dimension, but the code allows for the dimension 2 newforms we have (only over Q(i) and for level norm up to 5000) to say the right thing.

I may have missed some cases in checking.  In addition to the spaces listed above, try these (after the updated PR appears).

new dimension 2, no rational newforms but the dim 2 newform is not present: http://localhost:37777/ModularForm/GL2/ImaginaryQuadratic/2.0.7.1/226.2

new dim 6, we have one dim 2 but the other 4 are not present:
http://localhost:37777/ModularForm/GL2/ImaginaryQuadratic/2.0.4.1/2209.1

JohnCremona:
I'll fix the failing test caused by the new wording.

AndrewVSutherland:
Looks good!  As a side question (unrelated to this PR), are you planning to determine which of the dimension 2 newforms has CM?  I would naively suppose that at least for the ones that do not have CM, it should not be too hard to prove this.

JohnCremona:
> Looks good! As a side question (unrelated to this PR), are you planning to determine which of the dimension 2 newforms has CM? I would naively suppose that at least for the ones that do not have CM, it should not be too hard to prove this.

For the record -- I replied to this offline.

Issue #2848, opened 2019-03-22
minor db backend issues in reload_all


AndrewVSutherland:
1) When db.reload_all is given a folder name that does not exist it should raise an error rather than reporting success (and uploading no files).

2) Any counts/stats cached by the db object should be cleared, since they may have been changed if a counts/stats file was reloaded

AndrewVSutherland:
@edgarcosta has this been addressed, or will it addressed by PR #3098?

edgarcosta:
Now yes

edgarcosta:
Closed via #3098

Issue #2622, opened 2018-08-31
Better error messages when uploading data


roed314:
When creating new tables, schemas usually change frequently.  This can generate mismatches between the header row and the data stored below (for the format used in `reload` and `copy_from`).  The error messages generated by Postgres are often quite unhelpful.  Since we have the header columns, we can try to print something more useful.

roed314:
This has been fixed.

PR #3098, opened 2019-05-17
Various backend things

### LINKED ISSUE #3054: Add index by default to counts tables

edgarcosta:
- We no longer use to_regclass, to test if a relation exists, as it behaves in a weird fashion with upper case names.
- The bug above was creating name clashing of the indexes on reload, when these had upper case characters
- We now avoid name clashing when restoring or renaming indexes, by renaming them
eg
```
WARNING:LMFDB@2019-05-16 19:04:22,895: Index with name bmf_forms_field_label_level_label_CM_bc_tmp already exists. It has been renamed to bmf_forms_field_label_level_label_CM_bc_dep0_tmp and it can be deleted with the following SQL command:
DROP INDEX "bmf_forms_field_label_level_label_CM_bc_dep0_tmp" 
```
- Added indexes on the counts tables on reload, and thus resolves #3054
- Added a warning message when more than 10k rows are inserted into the counts tables, eg
```
WARNING:LMFDB@2019-05-17 10:26:18,285: 45083 rows, with with cols = '["degree", "class_number"]', were added into nf_fields_counts.This might decrease the counts table performance significantly! Consider clearing all the stats db.nf_fields_clear_stats_counts() and rebuilding them more carefully.
```
as we saw in #3093

- Better error message when a SQL command fails

### DIFF
--- a/lmfdb/backend/database.py
+++ b/lmfdb/backend/database.py
@@ -2067,16 +2067,17 @@ class PostgresTable(PostgresBase):
                 raise ValueError("Index %s does not exist in meta_indexes"%(name,))
             type, columns, modifiers, storage_params = cur.fetchone()
             creator = self._create_index_statement(name + suffix, self.search_table + suffix, type, columns, modifiers, storage_params)
-            try:
+            if self._index_exists(name + suffix):
                 # Reloading indexes can cause crashes in a way we don't understand.
                 # Postgres complains that the index already exists, even though we haven't
                 # created it yet on this table.
-                # As a workaround, we ignore errors in the index creation code and print a big warning
+                # As a workaround, we avoid errors in the index creation code and print a big warning
+                print "*"*80
+                print "Index restoration failed for {}".format(name + suffix)
+                print "*"*80
+                return
+            else:
                 self._execute(creator, storage_params.values())
-            except Exception as err:
-                print "*"*80
-                print "Index creation failed: %s" % (err,)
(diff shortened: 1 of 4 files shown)


AndrewVSutherland:
These are all great changes that should solve a lot of annoying problems, but it is tricky to really test it properly.  @roed314 have you looked at these changes?

edgarcosta:
David is traveling. Might only have time to look into this on Monday.

AndrewVSutherland:
@edgarcosta any of the 13 issues labelled "postgres" other than #3054 adressed by this PR?

edgarcosta:
This now also addresses #2848

edgarcosta:
I also believe that this addresses #2622 in two ways:
- Now when there is a db error, we reraise the error the SQL command, thus it's clearer what is going on
- By default adjust the schema of a table on reload, i.e., we use the headers from the files to adjust the schema (or even create the table with `reload_all`) automatically

roed314:
* `begin_renamer = SQL("ALTER TABLE {0} RENAME CONSTRAINT").format(tablename)` and the needs to change `tablename` to `SQL(tablename)` (and the same for two lines below).

Otherwise I don't see any problems.

PR #3099, opened 2019-05-17
Count -> self.class size

edgarcosta:
Replacing a count, which was then being recorded in the counts table, by using an attribute that we already have.

@AndrewVSutherland and @roed314, this also solves the mystery why ec_curves_counts have so many records


JohnCremona:
OK, I will review this.

JohnCremona:
In fact that line can simply be deleted since it just copies self.class_size (which is in the database) to self.ncurves (which is never referred to in code or template).  I think this is a left-over from when the class_size was *not* stored in the database, and this count was the only way to get at it.  That was bad, so I stored the sizes, but missed this.

edgarcosta:
Yup, you are right, not used anywhere. I have just deleted it.

JohnCremona:
I'll merge after Travis...

PR #3101, opened 2019-05-17
test stats

edgarcosta:
Adding a route for `statshealth` that quickly tests if all the stats page load accordingly.
We will, later on, use some automated service to check this route periodically.


AndrewVSutherland:
I'm currently travelling, but I will find time to review this (and some of the other PRs) later today.

PR #3104, opened 2019-05-17
Tweaks to statshealth

AndrewVSutherland:
Adds stats for genus 2 curves and more ecnf stats (response time is still under 2 seconds).  Also changes confirmation message.


AndrewVSutherland:
A few small additions to PR #3101 that occurred to me after I merged it.  I'll merge this as soon as Travis finishes.

Issue #2088, opened 2017-06-13
L-function of Trivial Artin representation


SallyKoutsoliotas:
Linking to the L-function friend from the Artin representation:
/ArtinRepresentation/1.1.1t1.1c1
gives
/L/Riemann/
However, adding a /L in front gives a different homepage:
/L/ArtinRepresentation/1.1.1t1.1c1/

Shouldn't this redirect to the /L/Riemann/ ?

AndrewVSutherland:
Closed via #3115.

PR #3115, opened 2019-05-19
Trivial Artin L-function = Reimann zeta function

AndrewVSutherland:
This addresses #2088 by redirecting the L-function page for the trivial Artin L-function to the Riemann zeta function page (and adding the trivial Artin representation as an origin).

### DIFF
--- a/lmfdb/lfunctions/main.py
+++ b/lmfdb/lfunctions/main.py
@@ -210,6 +210,7 @@ def l_function_riemann_page():
     return render_single_Lfunction(RiemannZeta, args, request)
 
 
+@l_function_page.route("/ArtinRepresentation/1.1.1t1.1c1")
 @l_function_page.route("/Character/Dirichlet/1/1/")
 @l_function_page.route("/NumberField/1.1.1.1/")
 def l_function_riemann_redirect():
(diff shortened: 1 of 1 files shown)


AndrewVSutherland:
This one-line change simply redirects

http://www.lmfdb.org/L/ArtinRepresentation/1.1.1t1.1c1

to 

http://www.lmfdb.org/L/Riemann/

as is already done for the trivial Dirichlet L-function

http://www.lmfdb.org/L/Character/Dirichlet/1/1

To test simply check http://127.0.0.1:37777/L/ArtinRepresentation/1.1.1t1.1c1 and then check that the trivial Artin L-function shows up in the origins box of the Riemann zeta function page and takes you where it should.

jwj61:
It seems to be the number field stats page.  There was a rewrite last night, but I thought when using the stats table, one didn't have to regenerate the statistics.

I just loaded the table, which took a while, and then loaded it again, which was fast.  So, I am guessing that it has now recomputed what it needed.

AndrewVSutherland:
Number field stats look fine to me, I just restarted Travis, let's see how he does this time.

PR #3118, opened 2019-05-21
Fixing computations for bigger bigger groups

jwj61:
Some Galois group pages were not working now that the range of degrees has been extended.

For some computations, we would create the transitive group in gap, but we are now beyond gap's knowledge.  So, we have precomputed generators for the groups and stored them in the database.  This means we don't have to compute them on they fly, and they can be used to initialize the group in gap.  For example,

  http://127.0.0.1:37777/GaloisGroup/47T6

gives a server error on beta, but now loads.

This group has over 100,000 conjugacy classes.  So, with this change, we do not try to compute/display conjugacy classes if there are more than 49 of them.

On a page like

  http://127.0.0.1:37777/NumberField/22.0.3388109619085565148335762714038503428163.1

the knowl for conjugacy classes and the character table is now just a "not computed" message.

On a page like
  http://127.0.0.1:37777/NumberField/7.3.2007889.2
the knowls for conjugacy classes and character table work as before (because the number of classes is manageable).

### DIFF
--- a/lmfdb/galois_groups/main.py
+++ b/lmfdb/galois_groups/main.py
@@ -17,7 +17,7 @@ from lmfdb.number_fields.web_number_field import modules2string
 from lmfdb.galois_groups import galois_groups_page, logger
 from .transitive_group import (
     group_display_pretty, small_group_display_knowl, galois_module_knowl_guts,
-    subfield_display, resolve_display, generators, chartable,
+    subfield_display, resolve_display, chartable,
     group_alias_table, WebGaloisGroup)
 
 # Test to see if this gap installation knows about transitive groups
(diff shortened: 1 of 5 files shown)


JohnCremona:
Looks pretty good to me.  Some questions about these examples, actually only about 47T6:
- any chance of saying something like "(there are over *** conjugacy classes)" after "data not computed"?   No problem if impractical, but the code does know something about the number when it decides whether or not to display.
- the group order makes the Properties box rather wide.  I don't know how that could be made to look better -- the factorization is about the same width, unlike the case of number fields with very large discriminant where you have only the factored expression in that box as it is always (?) shorter.
- in the Group Invariants section the line for Character Table the text "data not available" is in a different font, it is wrapped with html "pre" tags.  It would look better without them.

jwj61:
Good points.  I will say how many conjugacy classes there are.  That is now being stored in the database.

I don't know a good solution for the group order.  It is wide, but I think I have seen worse in other parts of the lmfdb.

I will see about the font switch for the non-existent character table.  It should not be hard.

JohnCremona:
@jwj61 OK, I will re-review when you have seen to this (ignoring the second point).

jwj61:
Hmm, I just pushed changes but Travis hasn't restarted

JohnCremona:
Tried to merge but failed.  If someone else can sort this, please do as I'm about to catch a train.  
Meanwhile one could fix "...representatives...is not computed" by removing "is" but I was not going to hold this up for that.

JohnCremona:
OK the merge worked OK second time.

jwj61:
I changed the phrasing for a conjugacy class knowl in pull request #3121

PR #3121, opened 2019-05-22
Galois show bounds

jwj61:
As we expand the range of data on Galois groups, we cannot always compute things to the same extent.  On a page such as 

  http://127.0.0.1:37777/GaloisGroup/24T18003

we explicitly say below the resolvents and below the siblings the extent to which we have computed each of these.  For resolvents, there are now also column headers to clarify what one is looking at, and the knowl for the header "Low degree resolvents" tries to explain things more fully.

We also have entries where we have not (yet) been able to compute arithmetically equivalent fields (i.e., H such that (G,G_1, H) is a Gassmann triple with G_1 the stabilizer of 1).  If we haven't computed it we give a message saying that, as on

  http://127.0.0.1:37777/GaloisGroup/28T86

### DIFF
--- a/lmfdb/galois_groups/main.py
+++ b/lmfdb/galois_groups/main.py
@@ -186,7 +186,8 @@ def render_group_webpage(args):
                                                       int(data['gapid']),
                                                       str([int(data['order']), int(data['gapid'])]))
         data['otherreps'] = wgg.otherrep_list()
-        ae = wgg.arith_equivalent()
+        ae = data['arith_equiv']
+        del data['arith_equiv']
         if ae>0:
             if ae>1:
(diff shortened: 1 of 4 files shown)


jwj61:
Also fixed the phrasing on a number field page for the spot where the knowl for conjugacy classes would be.  Three possible options are

  http://127.0.0.1:37777/NumberField/11.1.129891985607.1

where we show the knowl

  http://127.0.0.1:37777/NumberField/1.1.1.1

where we show the knowl but the number of classes is 1, and

  http://127.0.0.1:37777/NumberField/12.0.41223887921.1

where we don't shown the knowl

JohnCremona:
This looks great.  I really like those knowls containing the conjugacy class data and character tables.

Issue #3111, opened 2019-05-18
Server error for a higher auto page


jwj61:
From the feedback page:

> server error for this page:
> 
> http://www.lmfdb.org/HigherGenus/C/Aut/3.2-1.2.0.1
> 
> I arrived at it from this search result.
> 
> http://www.lmfdb.org/HigherGenus/C/Aut/?start=0&paging=0&genus=3&count=200
> 
> It's the search result with signature [2;2]
> 
> Thanks,
> Justin Lanier
> 

Reply to jlanier8@gatech.edu

jenpaulhus:
I think this can now be closed.  Notice:
http://beta.lmfdb.org/HigherGenus/C/Aut/3.2-1.2.0.1
is fixed by last pull.

AndrewVSutherland:
Agreed, but let's wait until the server error is actually fixed in production (http://www.lmfdb.org/HigherGenus/C/Aut/3.2-1.2.0.1 still fails).  I will push the data and the last PR to production momentarily.

AndrewVSutherland:
The fix is now live, thanks Jen!

Issue #3109, opened 2019-05-18
Ability to change sort order of search results


jwj61:
For CMF searches one can choose the sort order, see 

    http://www.lmfdb.org/ModularForm/GL2/Q/holomorphic/?search_type=List

It would be nice if the same option were available on other search results pages.  We received a feedback page request to do this specifically for higher genus curves with automorphisms, but there are plenty of other candidates.  Specific examples that have come up before

- Genus 2 curves (sorting by Sato-Tate group would be particularly useful)
- Elliptic curves over Q (rank, torsion order)
- Number fields (Galois group, class number)

Note that unless the search results all fit on the first page, this is not a matter of  simply sorting the results that appear on the screen (which could be done using javascript), it is reissuing the query with a different sort order -- this is both more useful and much easier to implement.

One other point worth noting, while it is standard UI practice to allow users to sort a column by clicking on it, we don't want to do that because it would make the knowls attached to the column headings inaccessible (and these are particularly important because column headings are often abbreviations or even just symbols whose meaning is not immediately obvious).  This is why we used a drop down option list in the CMF implementation.

From the feedback page:

> This is a feature request, not a problem. In the searchable database of curve automorphism groups, results are returned sorted by ascending dimension. Or maybe more accurately (I think, and depending on what is being searched for), results are returned sorted first by genus, then by dimension, then by group id, and then by signature.
> 
> It would be a great and very useful feature if search results could be toggled to be sorted by any one of the columns.
> 
> Thanks!
> Justin Lanier

Reply to jlanier8@gatech.edu

AndrewVSutherland:
This is a duplicate of #2765, but given that there is more content here (and a request from the feedback page) I will close #2765.

jenpaulhus:
This is implemented in Higher genus so could we get rid of that label on this issue, please?

AndrewVSutherland:
Done!

AndrewVSutherland:
Like the output column selection, this should be done generically across the entire LMFDB

roed314:
I've made progress on this, which is visible on [purple](https://purple.lmfdb.xyz).  Suggestions on specific sort orders to include are welcome; I'm just making things up for areas of the LMFDB I'm less familiar with.  I should be able to finish this today.

AndrewVSutherland:
Can you give  examples of some pages where the progress is visible?

roed314:
p-adic fields and above on the sidebar have a controllable sort order.

[...]

Issue #3108, opened 2019-05-18
Unable to search high genus curves for signatures with semicolons, e.g. [1;2]


jwj61:
From the issue tracker:

> There is a genus 3 curve whose automorphism group has a subgroup with signature [1;2]
> 
> http://www.lmfdb.org/HigherGenus/C/Aut/?start=0&paging=0&genus=3&group_order=8
> 
> But when I try to search for a signature of [1;2] with genus 3 (or other genera), the search result turns up empty
> 
> http://www.lmfdb.org/HigherGenus/C/Aut/?start=0&paging=0&genus=3&signature=%5B1%3B2%5D
> 
> In fact, the page changes the semicolon to a comma on the results page. It appears that in general there is a problem with searching for signatures that involve a semicolon.
> 
> Looking forward to an update. The site is very useful, thanks!
> Justin Lanier

Reply to jlanier8@gatech.edu

jenpaulhus:
I think this can be closed as well.  Notice:
http://beta.lmfdb.org/HigherGenus/C/Aut/?start=0&paging=0&genus=3&signature=%5B1%3B2%5D

AndrewVSutherland:
And it is now fixed on the production site

http://www.lmfdb.org/HigherGenus/C/Aut/?start=0&paging=0&genus=3&signature=%5B1%3B2%5D

Issue #2577, opened 2018-08-18
Downloads for higher genus curves not working


davidfarmer:
On dev.lmfdb.xyz, click on Higher genus, then click on genus 5.

If you then click on "Magma" to download the results in Magma code,
all it seems to do is reload the page.

AndrewVSutherland:
This works correctly on prod.lmfdb.xyz

AndrewVSutherland:
None of the download links (for either magma or gap) on the search results pages for higher genus curves appear to be working properly on beta.lmfdb.org or www.lmfdb.org (but as noted above, they used to work in the mongo db version, as can sill be seen on prod.lmfdb.xyz, so something changed with the switch to postgres).  This is not specific to genus 5.

I also note that the download links at the top and bottom behave differently; neither work but the ones at the top refresh the current page, whereas the ones at the bottom take you to the search results page for an empty query.

@edgarcosta I notice that you self-assigned this PR back in August, any thoughts on what is going on here?

jenpaulhus:
I am on airport wireless atm but I might have a fix for this in my code.  I'll check tomorrow and report back.

jenpaulhus:
I do not have a fix for this.  I'm going to make a pull request shortly which fixes some magma/gap download issues on the higher genus pages, but not from the search result pages.  I don't actually know what's happened with the switch from Mongo to Postgres.  Did you have any thoughts, @edgarcosta?

edgarcosta:
I'm looking into it

PR #3126, opened 2019-05-23
Add quotient genus >0 to higher genus

### LINKED ISSUE #3110: Confusion about notation for signatures, and seemingly missing entries for higher genus curves
From the feedback page:

> Actually, I'm confused about the signature of the entry I just submitted about.
> 
> It's an automorphism of a genus 3 surface of order 2. 
> 
> http://www.lmfdb.org/HigherGenus/C/Aut/?start=0&paging=0&genus=3&count=200
> 
> Or to see it in a shorter list:
> 
> http://www.lmfdb.org/HigherGenus/C/Aut/?start=0&paging=0&genus=3&group_order=2&count=200
> 
> The signature says [2;2]. But shouldn't it be [2; -], or some other notation for no branch points? This should be a free action on a genus 3 curve, with quotient a genus 2 curve.
> 
> I feel like this may be an issue throughout the database. For instance here:
> 
> http://www.lmfdb.org/HigherGenus/C/Aut/?start=0&paging=0&genus=4&group_order=3&count=200
> 
> The [2;2] entry should be [2; -]
> 
> And I was going to say here, next:
> 
> http://www.lmfdb.org/HigherGenus/C/Aut/?start=0&paging=0&genus=7&group_order=3&count=200
> 
> There should be a [3; -] entry. But it doesn't even appear, despite the fact that there are genus 7 curves with an order 3 automorphisms acting freely and with a genus 3 quotient. 
> 
> If I'm mistaken or if some other convention I'm unfamiliar with is being used, please let me know.
> 
> Thanks!
> Justin Lanier
> 

Reply to jlanier8@gatech.edu

jenpaulhus:
This code fixes a number of small issues on the higher genus data pages.

I have updated the hgcwa_passports table in a few places to fix some small errors with the data,  so that will need to be updated on the production version eventually.  This fixed, for example, issue  Issue #3108  as you can see if you run the search on beta now:
http://beta.lmfdb.org/HigherGenus/C/Aut/?start=0&paging=0&genus=3&signature=%5B1%3B2%5D

The primary issue is that some data I thought did not make the switch to Postgres last August actually did make it.  This "new" data is for quotient genus >0  (only  for genus of the original curve up to 4 for the moment) and this data required a series of small changes.  But the pages for that data were works in progress back then, and so aren't displaying correctly on lmfdb.org.  This pull request fixes them. It might be worth considering these as "new" pages even though there are only a handful of entries currently in the database which display this way (instead of just as a pull request fixing errors). If you search for quotient genus >0 you will find one of the "new" pages.

We've fixed some pages returned errors:
http://www.lmfdb.org/HigherGenus/C/Aut/3.2-1.2.0.1
(See #3111  too)

Or a page like the following which had () instead of Id(G) for the identity element of the group.
http://www.lmfdb.org/HigherGenus/C/Aut/2.2-1.1.2-2.1

We also made searching by quotient genus a part of the main search functionality

Magma and Gap downloads had some issues on the page and I fixed most of them (except see #2577)

Notice that [2;2] now correctly shows as [2;-] which fixes #3110
http://localhost:37777/HigherGenus/C/Aut/?start=0&paging=0&genus=3&signature=%5B2%5D&count=200

Also in #3110 they ask about genus 7, but I only have data for genus 4 and quotient genus >0 in the database yet---Note the completeness knowl specifically says only quotient genus =0 but I will change that to include the quotient genus >0 once this pull request is accepted. I don't want to add higher genus data until I am sure these pages are stable. Similarly I  will need to update a couple knowls like signature and generating vectors to explain the quotient genus >0.  

There was an error if you searched for certain labels.  It was a small regex error.

Sort order like #3109 is coming soon...


jenpaulhus:
Just noticed there were two files from maass_form which register changes with this pull request. You will notice they are just whitespace changes.   I likely forgot to commit or stash changes when I was trying to fix an old error message with maass_forms (#1820) and that corrupted this branch.   Those changes should have 0 impact on any pages.

AndrewVSutherland:
Thanks for this Jen!  I took a look and I think there are still a few issues that need to be addressed.

(1) Searching on group order 2 gives different results than searching on group [2,1], compare:

http://localhost:37777/HigherGenus/C/Aut/?group_order=2
http://localhost:37777/HigherGenus/C/Aut/?start=0&paging=0&group=%5B2%2C1%5D

I expect this is because there are four entries in the search results where the group id still has an extra space in it.  This is not the only place where this happens, the database search

    db.hgcwa_passports.search({"group":{"$like":"[%, %]"}})

returns 128 results.  Presumably these also need to be fixed?  When they are you should let me know so I can copy them to the production database.

(2) There is a problem in how group order searches are being processed.  The value entered into the group order field on the main browse page is not present on the search results page, compare:

http://localhost:37777/HigherGenus/C/Aut/?group_order=2
http://beta.lmfdb.org/HigherGenus/C/Aut/?group_order=2

also, if you enter the group order on the search results page (rather than from the main browse page) it does not work correctly, compare

http://localhost:37777/HigherGenus/C/Aut/?start=0&paging=0&groupsize=2
http://beta.lmfdb.org/HigherGenus/C/Aut/?start=0&paging=0&group_order=2

I presume this is due to the discrepancy between "groupsize" and "group_order".  This change was made in the commit

     https://github.com/jenpaulhus/lmfdb/commit/8f971e7bad79577f030cf9f9f6e0cba768076c4f

labelled "these aren't ready, just saving" on March 22.  This commit (along with the inadvertent changes to Maass forms) makes me nervous; it includes a lot of changes that are not obviously related to this PR and makes a lot more changes to the search results template than are needed.

Would it make sense to create a completely new branch that contains only the changes intended for this PR?  It seems dangerous to be including code changes from two months ago that were obviously not made in response to the issues this PR is trying to address and look like they may be based on an older version of the code.

(3) I don't understand why the search results for quotient genus 2:

    http://localhost:37777/HigherGenus/C/Aut/?g0=2

include the signature [1;2,2,2].  According to

    http://localhost:37777/knowledge/show/curve.highergenus.aut.signature

the first entry of the signature (1 in this case) is the quotient genus.  Have I misunderstood something?

I also think the signature knowl should reference the quotient genus knowl.

(4) Just a small suggestion for improvement: it seems like the quotient genus would be a natural thing to add to the properties box and the list of invariants on the home page of the passport.

(5) I think it would be a good idea to add some additional tests to test_hwcga.py that specifically check the problems that this PR is intended to fix so that they don't recur, and to check that the new quotient genus search/display works as desired.

jenpaulhus:
Thanks, Drew.   I will work on the changes you suggested (and/or start a different, clean pull request if needed).  I'm a bit swamped for the next week, so give me a bit of time to implement the changes.

jenpaulhus:
Closing this because it is superseded by #3143

PR #3130, opened 2019-05-26
fixing downloads for HGCWA

edgarcosta:
The aim of this PR is to fix downloads for HGCWA #2577 

I made something to work, however, I must confess I don't understand the previous old structure of `hgcwa_code_download_search` or the old code before MongoDB to PostgreSQL transition, as both were giving me way too many results when I downloaded something.
I tried to do something that makes sense from my perspective, but it would be nice to make sure that this is what we want.

I have also removed the download buttons on top next to search again, and the bottom ones were replaced by the template `download_search_results.html`.

### DIFF
--- a/lmfdb/higher_genus_w_automorphisms/main.py
+++ b/lmfdb/higher_genus_w_automorphisms/main.py
@@ -42,7 +42,7 @@ def split_passport_label(lab):
     return passport_label_regex.match(lab).groups()
 
 
-credit ='Jen Paulhus, using group and signature data originally computed by Thomas Breuer'
+credit = 'Jen Paulhus, using group and signature data originally computed by Thomas Breuer'
 
 
 def get_bread(breads=[]):
@@ -64,9 +64,9 @@ def sign_display(L):
         signL = "[ " + str(L[0]) + "; -]"
     else:
         signL = "[ " + str(L[0]) + "; "
-        for i in range(1,sizeL-1):
-            signL= signL + str(L[i]) + ", "
-        signL=signL + str(L[sizeL-1]) + " ]"    
+        for i in range(1, sizeL-1):
+            signL = signL + str(L[i]) + ", "
+        signL = signL + str(L[sizeL-1]) + " ]"
     return signL
 
 def cc_display(L):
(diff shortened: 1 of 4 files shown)


jenpaulhus:
Thanks, Edgar!  I see this and will look it over, but it might not be until next week.

edgarcosta:
I have fixed the merge conflict and did some cosmetic changes in the last commit.

edgarcosta:
For the friends line maybe you want that block to be replaced by:
```
friends = [("Full automorphism " + Lf, Lf) for LF in friends]
friends += [("Family containing this refined passport",  urlstrng)]
```

jenpaulhus:
Assuming I understand git merges correctly, it is probably best if #3147 gets merged before this PR as otherwise I think that PR will overwrite Edgar's cleanup.

I also would rather not have any additional deleting of code from these pages for the moment, unless something is actually not working.  As you can see from #3148 there is some specialized code we wrote a while ago that is no longer present/working and I'd like to return that functionality to the pages before removing or rewriting current code.

edgarcosta:
To be clear, no function was deleted, I rewrote
`hgcwa_code_download_search` and had to move it before `search_wrap`.

jenpaulhus:
I must have misunderstood yesterday. I thought you said that in your cleanup you had deleted a function that was no longer in use.   

I also didn't mean to imply that this pull request had created the problem with searching with a variable, I suspect that happened way back when the switch to Postgres was happening, and it very likely was something I did.  But, the error messages you flagged to fix here were part of that old code, and so I don't want to touch that until I have a chance to go back and see what happened to the rest of the code.

edgarcosta:
My fault, I confused myself about what I had done, and that is why I went
through the code again to be sure that nothing was broken.

edgarcosta:
The test was broken by upgrading sage on Legendre to 8.7, however, Travis doesn't yet run 8.7 and I don't have time to upgrade Travis right now, so I have reverted the commit.

jenpaulhus:
I've checked this PR again and it appears to do exactly what I expect it to.  I think it's ready to be merged.  I would still like to see the download buttons at the top, but since the rest of the site doesn't have that, I agree it is reasonable to be consistent across the site for now.

edgarcosta:
@jenpaulhus I removed the download buttons on top next to search again as I couldn't easily use the template `download_search_results.html`.
The best option, so we don't forget, is to open a ticket as a feature request, and we can work from there.

jenpaulhus:
Ticket opened!

Issue #3131, opened 2019-05-28
Use of commit flag in rewrite not working or not clear


jwj61:
As a trial run, I tried to rewrite the transitive group database with commit=False thinking it would go through the motions, but not actually change the database.  But, it did change the database.  I don't think the commit flag is documented so maybe I guessed wrong as to what it means.  But then, it at least needs to be documented.

jwj61:
In trying to fix this, it looked like
 db.gps_transitive.reload_revert()
would revert to the most recent backup.  It gives:

TypeError                                 Traceback (most recent call last)
<ipython-input-2-e09bc7397b18> in <module>()
----> 1 db.gps_transitive.reload_revert()

/scratch/home/jj/lmfdb/lmfdb/backend/database.pyc in reload_revert(self, backup_number, commit)
   3257                 if self._table_exists(tablename + old):
   3258                     tables.append(tablename)
-> 3259             self._swap(tables, indexes, constraints, '', '_tmp')
   3260             self._swap(tables, indexes, constraints, old, '')
   3261             self._swap(tables, indexes, constraints, '_tmp', old)

TypeError: _swap() takes exactly 4 arguments (6 given)

roed314:
The `reload_revert` error should be fixed by #3133.  Can you check to see if that gets you back to the state you were in before the problem?

You're correct about the intention of `commit=False`.  Can you tell us exactly what commands you tried to execute?  I'm curious whether `db.conn.rollback()` would have gotten back to the desired state after calling reload.

jwj61:
The command with commit=False was
  gg.rewrite(mod,searchfile='/scratch/home/jj/nf-lmfdb/DELETE.ME', commit=False)

where mod is a function which changes values in a given row and gg= db.gps_transitive.

I am already part way through fixing the errors moving forward, so I am not excited about going back now.  In fact, I am stalled because a rewrite is just sitting after building indexes.  My guess is that it is redoing things in the counts table, but can be sure.  I ran a lot of commands with count yesterday on this database.  How can I see what is in there, and potentially wipe out its contents?

edgarcosta:
I'm looking into it.

There are no counts:
```
lmfdb=> select count(*) from gps_transitive_counts;
 count
-------
     0
(1 row)

Time: 91.211 ms
```

edgarcosta:
I was the hold-up, the rewrite should have been cleared.

edgarcosta:
[...]

PR #3133, opened 2019-05-28
Update _swap calls in reload_revert to use new signature

roed314:
Addresses part of #3131


edgarcosta:
This looks good to me.

roed314:
The test failure looks to be associated with the broken `gps_transitive` table, so I'm going to merge this.  @jwj61, let me know if you're now able to revert the reload.

PR #3135, opened 2019-05-28
better error message on knowldb.search

### LINKED ISSUE #3132: Server error on knowl regex search
Regex knowl searches that being with an asterix throw a server error:

http://www.lmfdb.org/knowledge/?search=*&regex=on

edgarcosta:
Resolves #3132

### DIFF
--- a/lmfdb/backend/database.py
+++ b/lmfdb/backend/database.py
@@ -31,7 +31,7 @@ from glob import glob
 import csv
 import sys
 
-from psycopg2 import connect, DatabaseError, InterfaceError, OperationalError, ProgrammingError, NotSupportedError
+from psycopg2 import connect, DatabaseError, InterfaceError, OperationalError, ProgrammingError, NotSupportedError, DataError
 from psycopg2.sql import SQL, Identifier, Placeholder, Literal, Composable
 from psycopg2.extras import execute_values
 from psycopg2.extensions import cursor as pg_cursor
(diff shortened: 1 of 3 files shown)


AndrewVSutherland:
This is still not very user friendly, why not just say "Error: *banana is not a valid regular expression."  as we do elsewhere, e.g.

http://www.lmfdb.org/Genus2Curve/Q/banana

edgarcosta:
The crux of it is that I don't really have a way to test if it is a valid
regular expression for postgres without trying to run it, and if it fails
the type DataError is too generic to claim that is not a valid regular
expression

AndrewVSutherland:
The error message contains the string "invalid regular expression", so you could just search for that, and if it is not there just display a generic "Search on regular expression failed." (but my guess is the only reason this ever fails is because the regular expression is invalid).

```
"Error: Unexpected error while performing the search:
knowldb.search(category='', filters=[], types=['normal'], keywords=*, regex=True)
DataError: invalid regular expression: quantifier operand invalid happens while executing SELECT "id", "authors", "cat", "content", "last_author", "timestamp", "title", "status", "type", "links", "defines", "source", "source_name" FROM (SELECT DISTINCT ON (id) "id", "authors", "cat", "content", "last_author", "timestamp", "title", "status", "type", "links", "defines", "source", "source_name" FROM kwl_knowls WHERE status != -1 AND content ~ '*' OR title ~ '*' OR id ~ '*' ORDER BY id, timestamp DESC) knowls WHERE status >= 0 AND type = 0
```

AndrewVSutherland:
@edgarcosta Don't worry about it, I went ahead and made the change I suggested and incorporated it in #3134 (sorry you got sidetracked by this).

PR #3147, opened 2019-06-07
Update the sort orders for searches in HGCWA

jenpaulhus:
For these first three, you can experiment with search order using this search:
/HigherGenus/C/Aut/?start=0&paging=0&genus=3-5&g0=0-1&group_order=2-6&count=91
* Designated a fixed sort order for search results.
* Created drop down menu to give options for search result priority (completing #3109 for HGCWA)
* Made groups pretty print on search instead of just gap/magma number, and rearranged order of columns in search a bit

Three small unrelated fixes:
* Got rid of "name" in parse calls (see PR #3130 comments)
* Fixed incorrect "friends" loop  (see PR #3130 comments)
* top matter on index page: /HigherGenus/C/Aut/   For the moment I just hard coded in max and min of higher quotient genus data, because redoing the statistics on HGCWA is upcoming, and so this will just be a temporary fix.)

I think this pull request should get merged before #3130, otherwise I think Edgar's "cleanup" will be overwritten with this pull.


AndrewVSutherland:
Looks good!

PR #3149, opened 2019-06-07
Add two new cmf search options

alexjbest:
Searching on hecke ring index and minimal coeffs needed to generate it, first one was requested by a user, added two for symmetry.


AndrewVSutherland:
Looks good, but I would change the captions "Hecke generators needed" to "Coefficient ring generators" (abbreviate on search results page).  I don't think you need to include the word needed, the knowl will explain exactly what this number means (and the word "needed" is potentially confusing, since this quantity is not the minimal number of generators, it is the least n such that a_1,...,a_n generate the coefficient ring, which is not necessarily the same thing -- the know will explain this).

alexjbest:
@AndrewVSutherland Ok, I have to abbreviate gens on the browse page too though to get it into 160px width.

PR #3151, opened 2019-06-11
fixing upsert, when used for insertion

### LINKED ISSUE #3136: Odd error from upsert in backend
I'm trying to use the upsert to change records that contain Jsonb fields, but SQL query construction is broken. Example:

```
sage: rec_find = json.loads('{"db_id":14, "name":"fields_human"}')
sage: rec_set = json.loads('{"nice_name":"fields_human", "NOTES":null, "INFO":nu
....: ll}')
sage: lmfdb.db['inv_tables'].upsert(rec_find, rec_set)
```
```
ProgrammingError: syntax error at or near "{"
LINE 1: INSERT INTO {0} ({1}) VALUES ({2}) RETURNING *
                    ^
 happens while executing INSERT INTO {0} ({1}) VALUES ({2}) RETURNING *
```

Any ideas?

edgarcosta:
This resolves #3136 and fixes another minor bug where the data object submitted for upsert could be unintentionally modified by the upsert call.

you can try it out
```
db.create_table_like('test_inv_tables', 'inv_tables')

import json

rec_find = json.loads('{"db_id":14, "name":"fields_human"}')
rec_set = json.loads('{"nice_name":"fields_human", "NOTES":null, "INFO":null}')
print db.test_inv_tables.upsert(rec_find, rec_set)
db.drop_table('test_inv_tables')
```

### DIFF
--- a/lmfdb/backend/database.py
+++ b/lmfdb/backend/database.py
@@ -2823,7 +2823,8 @@ class PostgresTable(PostgresBase):
                     self._execute(updater, dvalues)
                 if not self._out_of_order and any(key in self._sort_keys for key in data):
                     self._break_order()
-            else: # insertion
+
+            else:  # insertion
                 if "id" in data or "id" in query:
                     raise ValueError("Cannot specify an id for insertion")
                 new_row = True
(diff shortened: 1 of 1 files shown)


JohnCremona:
@hratcliffe Can you test this?  If you're happy we'll merge it.

hratcliffe:
That seems to fix the bug. I'm not sure about using the dict constructor rather than a deepcopy for the other bug - @edgarcosta are you sure a shallow copy like that is always enough?

edgarcosta:
@hratcliffe, the only issue I'm aware is this line:
```
search_data["id"] = row_id = self.max_id() + 1
```
where we add the `id` entry, thus a shallow copy should suffice.

Alternatively, instead of doing a copy we could just pop that item after the insertion.

Issue #3155, opened 2019-06-12
Isogeny matrix of EC


edgarcosta:
I believe that the Isogeny matrix is using the wrong labels, i.e., using the Cremona labels when should use LMFDB label and vice versa.

Compare:
http://beta.lmfdb.org/EllipticCurve/Q/15/a/
vs
http://beta.lmfdb.org/EllipticCurve/Q/15a/

And note that 15.a5 = 15a1 is the unique curve without an isogeny of degree 8.

side note, it also took me a while to figure out what meant search by isogeny degree for a curve in http://beta.lmfdb.org/EllipticCurve/Q/, as the knowl only defines isogeny degree.

JohnCremona:
@edgarcosta I cannot see any inconsistency.  The curve [1,1,1,-10,-10] has no cyclic 8-isogeny.  This is 15a1 = 15.a5.  In the page http://beta.lmfdb.org/EllipticCurve/Q/15/a/ using LMFDB labels a5 is in the middle of the graph and row 5 of the matrix has no 8's.  In the page http://beta.lmfdb.org/EllipticCurve/Q/15a/ using C labels, a1 is in the middle and row 1 has no 8's.

edgarcosta:
I agree with what you read from the matrix, and that is what lead me to this inconsistency, as the page of that curve (http://beta.lmfdb.org/EllipticCurve/Q/15/a/5 or http://beta.lmfdb.org/EllipticCurve/Q/15a1/)
says
```
This curve has non-trivial cyclic isogenies of degree $d$ for $d=$
2, 4, 8 and 16.
```
and that curve also shows here:
http://beta.lmfdb.org/EllipticCurve/Q/?conductor=15&isodeg=8&surj_quantifier=include

JohnCremona:
OK I see -- there is a bug, possibly two, but not on the isogeny graph/matrix page.  On the curve's own home page it gets that list of isogeny degrees directly from the database.  It appears that when I added the list of degrees to the db I messed up:
sage: db.ec_curves.lucky({'label':"15a1"})['isogeny_degrees']
[1, 2, 4, 8, 16]
Best solution is is obviously to fix that column in the database.  Meanwhile, a trick like
sage: db.ec_curves.lucky({'iso':"15a", 'lmfdb_number':1})['isogeny_degrees']
[1, 2, 4]
might succeed in getting the correct number into the home page but will not correct the wrong search result.

[...]

PR #3156, opened 2019-06-13
correct list of isogeny degrees on ec/Q home pages

JohnCremona:
See #3155.   Previously the list of isogeny degrees shown on a curve's home page was sometimes incorrect, having been taken from the wrong row of the isogeny matrix.  This was caused by having different numbersing of curves in the isogeny class for Cremona labels and LMFDB labels.
Compare http://beta.lmfdb.org/EllipticCurve/Q/15/a/5 which incorrectly says that the curve has 16-isogenies, with http://localhost:37777/EllipticCurve/Q/15/a/5 which correctly says that the degrees are at most 4.  This curve is '15a1' and to justify the above statement:
sage: E = EllipticCurve('15a1'); E.ainvs()
(1, 1, 1, -10, -10)
sage: [E.isogeny_degree(E2) for E2 in E.isogeny_class()]
[1, 2, 2, 2, 4, 4, 4, 4]

NB We will also have to change the isogeny_degrees column in db.ec_curves, as this fix does not correct searching by isogeny degrees (this curve is still found by searching for isogeny degree 16), but half a loaf is better than none.

### DIFF
--- a/lmfdb/elliptic_curves/web_ec.py
+++ b/lmfdb/elliptic_curves/web_ec.py
@@ -249,7 +249,10 @@ class WebEC(object):
 
         cond, iso, num = split_lmfdb_label(self.lmfdb_label)
         self.one_deg = ZZ(self.class_deg).is_prime()
-        isodegs = [str(d) for d in self.isogeny_degrees if d>1]
+        # NB self.isogeny_degrees is currently incorrect (as of
+        # 2019-06-13) but we can get the degrees from the correct row
+        # of the isogeny matrix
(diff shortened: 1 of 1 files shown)


edgarcosta:
@JohnCremona given that we can extract the right data from data already in the database (as you show on this PR), could we instead run a rewrite to fix the data? and hence not need this PR?

JohnCremona:
I have been persuaded to make the data change (to the isogeny_degrees column) which will make this PR redundant, so I am going to close it.

PR #3168, opened 2019-06-26
Fix sibling count

jwj61:
This bug is currently only visible for one field:

  http://127.0.0.1:37777/NumberField/40.4.818380493827663732613543259594667183634328825351814767249590812771172429611764531578535936000000000000000000000000.1
vs
http://beta.lmfdb.org/NumberField/40.4.818380493827663732613543259594667183634328825351814767249590812771172429611764531578535936000000000000000000000000.1

The one on beta has a section indicating that the S_40 field has a degree 40 sibling.  It doesn't.  Ultimately, it is because for larger degree transitive groups, we put -1 for the count of arithmetically equivalent fields if we don't know it (yet).

### DIFF
--- a/lmfdb/number_fields/web_number_field.py
+++ b/lmfdb/number_fields/web_number_field.py
@@ -407,11 +407,12 @@ class WebNumberField:
             for s in sibcnts:
                 repcounts[s[0][0]] += s[1]
             gc = 0
-            if galord<24:
+            if galord<48:
                 del repcounts[galord]
                 if self.degree() < galord:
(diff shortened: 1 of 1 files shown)


AndrewVSutherland:
Looks fine.  In the course of testing I noticed that there are some high degree fields (e.g. 
http://127.0.0.1:37777/NumberField/30.0.17761887753093897979823770061456102763834271.1) whose pages won't load (they timeout with a server error).  But I'll raise an issue for this.

Issue #3160, opened 2019-06-19
Specify number of ramified primes in number field search


jwj61:
From the issue tracker:

The online Jones-Robert tables of number fields gave the option of bounding the number of ramified primes. I almost always used this feature, since it's nice to have examples of fields ramified at exactly one prime which is unknown in advance (or possibly ramified at the prime 2 and a single other prime not known in advance). Could that functionality be included here? 

Responded to the reporter by e-mail that I will do this.

edgarcosta:
Addressed by #3170

PR #3170, opened 2019-06-27
Number of ramified primes

jwj61:
Address issue #3160 allowing the user to specify the number of ramifying primes.  To test, go to the number field index page and search.

Also, adjusted some capitalization, laid out the search refinement to a style used elsewhere, and simplified the template code on the search results page.

### DIFF
--- a/lmfdb/number_fields/number_field.py
+++ b/lmfdb/number_fields/number_field.py
@@ -110,7 +110,7 @@ def NF_redirect():
 @nf_page.route("/Source")
 def source():
     learnmore = learnmore_list_remove('Source')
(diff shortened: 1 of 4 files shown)


edgarcosta:
Everything works fine.
I added an index on `num_ram`.

However, regarding the browse page, I would prefer to keep `Unramified primes` and `Set of ramified primes` together, hopefully, in the same line.

jwj61:
I think number of ramified primes is more closely connected to the set of ramified primes.  If we get another search field, then unramified bumps over to be directly below the set of ramified primes.

edgarcosta:
Okidoki

PR #3183, opened 2019-07-01
Adding more cyclotomic fields

jwj61:
We have a dictionary of field labels for Q(zeta_n) and Q(zeta_n)^+ which are used to display the fields and to allow users to specify the former by nickname.  Now that we have fields up to degree 47, this pull request expands those dictionaries accordingly.

As a test, enter Qzeta37 or Qzeta132 in the label box on the number field index page.  Both will take you to the corresponding field, and down on the page where subfields are listed, their real quadratic subfields should be there.


JohnCremona:
Looks good.  I would suggest (perhaps for a new PR) adding Qzeta5 (or similar) in the e.g. list under the "go to a specific field..." search box, as users might not know this facility exists.  Also, might it be nice to have similar nicknames for input for the real subfield?  I could remember whether this was possible so I tried putting Qzeta37+ and Qzeta37plus into the box and this raised a server error; so there should be some more robust error-catching there.

None of those suggestions requires delaying this though so I'll merge.

PR #3185, opened 2019-07-02
Serve static files from inventory sub-dir

hratcliffe:
Backup routing for static files, so that they are correctly handled if webserver does not intercept them.

### DIFF
--- a/lmfdb/inventory_app/inventory_app.py
+++ b/lmfdb/inventory_app/inventory_app.py
@@ -1,4 +1,4 @@
-from flask import render_template, request, url_for, make_response, jsonify, Blueprint
+from flask import render_template, request, url_for, make_response, jsonify, Blueprint, send_from_directory
 from flask_login import login_required
 from lmfdb.users import admin_required
 import inventory_viewer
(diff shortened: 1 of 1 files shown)


JohnCremona:
How could we test this, given that in a local instance of the webserver the code would not run?  I see that Edgar has just merged it so perhaps the plan is to push to beta and see.

edgarcosta:
@JohnCremona, I just checked that you can load the static files in `lmfdb/inventory_app/static/`, e.g.:
http://localhost:37777/inventory/static//inventory_general.js
or 
http://localhost:37777/inventory/static/index_viewer.js

and now that these files load, you should be able to test inventory in a local instance

JohnCremona:
@edgarcosta I am puzzled.  I could already test the inventory locally and did so before merging the big PR.  I just ran into problems on beta.  As for the two URLs above, the first shows the text file which I think is expected, but the second returns a Bad Request page saying "No collection static.index_viewer.js exists".

edgarcosta:
This PR should fix that.

JohnCremona:
> This PR should fix that.

Yes it does, thanks.  Any objection to merging to beta now?

edgarcosta:
Nope!

PR #3194, opened 2019-07-11
Added full API2 implementation

csbrady-warwick:
This should be an interface complete and mostly backend complete implementation of the version 2 API. Will bring the documentation repository fully up to sync with the implementation once merge completed.


csbrady-warwick:
Missing file. Will add and resubmit ASAP

csbrady-warwick:
Now fixed missing files.

davidlowryduda:
I'm looking into this.

davidlowryduda:
There is a lot here to examine. For those who want to check out this api, the best way to that is to navigate to the entry point http://localhost:37777/api2/ and read. This is to be compared to the current api: http://localhost:37777/api/

In its current version, this api gives access to three collections: elliptic curves over Q, abelian varieties, and Sato-Tate groups. These are the `searchers` listed here: http://localhost:37777/api2/description/searchers

Each `searcher` can describe its own results. For example, see http://localhost:37777/api2/description/elliptic_curves_q

This is a nice feature over the current version of the API.


Not explicitly described is how one adds a collection to the API. This seems to be very easy, but it's not automatic. One can see how a collection is added by looking at `lmfdb/elliptic_curves/__init__.py`, and examining the two new lines `register_search_function(...)` and `register_singleton(...)`.

I haven't yet wrapped my head around the behavior of singletons yet and how adding them might work for generic collections. In particular, I'm thinking about L-functions.


Overall, I like the amount of documentation and that the API is largely self documenting. It seems easy to use and easy to parse, and the range-functionality seems to work pretty well.

But I also have a few questions as I continue to test this out.

1. I notice that every response includes the key-item pair `has_api2: false`, and this is hard-coded in `build_api_wrapper` in api2/utils.py. What is the idea here? This doesn't seem to be referenced anywhere else in the code or the documentation.

2. In the documentation on the /api2 page, under "Interpreting the records", I see a link `/api2/inventory/:searcher`. I think that this is probably supposed to be `/api2/inventory/<searcher>`, in line with the optional arguments in the rest of the document. Is that right?

3. Could you add an example singleton page?

4. In http://localhost:37777/api2/data/elliptic_curves_q?label=11a2 (for example), the first key in data:records is `"degree"`, and its value is `"5"` (with quotes). On the other hand, there is also the pair `"rank"`:`0` (no quotes). After reading the description page http://localhost:37777/api2/description/elliptic_curves_q I would expect both results to be unquoted integers. But I notice also that the description of "rank" within the description page doesn't have a key for `postgres_type`, so perhaps something a bit odd is going on. Can you help me figure out my confusion?

JohnCremona:
A few additional comments, all rather superficial ones about the content of the /api2 page itself.

1. Some items such as /api2/description/<searcher>  are rendered as links even though the (literal) link does not of course work since <searcher> is a wildcard.  I suggest not making that a link; instead put in an example right there which is a link, e.g./api2/description/elliptic_curves_q .
2. Similarly for the other paths shown with wildcards, in particular examples for <other_path> and <path> would be welcome.  I thought that http://localhost:37777/api2/pretty/description/searchers would be an example, but it is not valid.
3. I thought http://localhost:37777/api2/livepg/ec_curves would be a valid example but is gives an error  (TypeError: create_search_dict() got an unexpected keyword argument 'database')
4. An example of the % for "Test if a value is contained in a list"?  Neither http://localhost:37777/api2/data/elliptic_curves_q?rank=0&conductor=%11,14 nor ...%[11,14] work.
5. How do we edit the "human names"?  e.g. I would capitalise Q in "Elliptic curves over q"
6. In the "Get results" example section there's a typo "elliptical_curve_q" should be "elliptic_curve_q", but even with that charge the URL does not work for me with the sample _fields query part since it does not like the field "2adic_index" for some reason.  (Possibly since that column is in the extras table?)

I also note that we'll need to think up and decide on a large number of "Canonical names".  I think that the way to fill these in is by inventory editing, right?

csbrady-warwick:
Hi,
To answer David's points first

1) Sorry about that. That used to do something but now doesn't. I'll remove it from the response

2) Yes, that should be <searcher>. I'll swap that too

3) No worries. I'll put one in and create some more documentation on how to set up all of the searchers and the singletons

4) I'll have to look at that one more carefully. I suspect that the Inventory has gotten out of sync somehow

And now John's

1-2) Good catch. I'll mark them to prevent them becoming links

3) Ah, OK. There must be an error in the "live" section. I'll check that

4) Oh dear. I'll get that working then.

5) The "Human Names" are specified where the searchers are, so in the __init__ file for the relevant section

6) OK. I'll check out what's causing that. It might well be connected to being in the extras table. How the back end deals with extras has changed a bit I think at times so something might have broken

And yes, the canonical names should now be editable through the inventory

csbrady-warwick:
Per David's point 4, everything is OK with the inventory and my code but for some reason the backend is returning "rank" from the main table as a Python 'int' but "degree" from the extras table as 'sage.rings.integer.Integer'. I'll raise an issue on it.

csbrady-warwick:
OK. I've managed to address all of the other elements except John's point  4 which is a misdocumentation. The % operator is intended to look up a single supplied value in a list in the database, not look up multiple possible values for a single value in the database. I'll look into adding that since it would be useful. I'll push fixes for the rest later today

JohnCremona:
In case it is relevant: the rank will never be large (the record is 28 but
we currently have none >4) but the degree gets large. For consistency
perhaps bith shoukd be returned as Sage integers.  Not sure.

JohnCremona:
Thanks for fixing the merge conflict.  There's just a pyflakes failure:

./lmfdb/api2/utils.py:407: local variable 'f_e_c' is assigned to but never used
./lmfdb/api2/utils.py:408: local variable 'count_only' is assigned to but never used

PR #3196, opened 2019-07-17
fixing term order

### LINKED ISSUE #3195: Euler factor terms displayed in the wrong order on rational L-function CMF
On the L-function page

    http://www.lmfdb.com/L/ModularForm/GL2/Q/holomorphic/2401/2/a/f/

The 675*p^3T^12 term of the Euler factor at 2 appears before the 147p^5T^11 term, and a similar problems appear in the Euler factor at 3 for the T^5 and T^6 terms.

It appears that the terms are being ordered base on the total degree in p and T, rather than the degree in T (this should be an easy fix).

edgarcosta:
This resolves #3195, by having the sort based on degree, where deg(p) = 0 and deg(T) = -1.


AndrewVSutherland:
Thanks for the quick fix!

PR #3202, opened 2019-07-23
Nitpick: spaces of modular forms include Eisenstein series and old forms

jvoight:
so they shouldn't be called "Space of Newforms" in the title.


edgarcosta:
The tests failed because one of the tests checks for the title. Could you please change the tests accordingly? You only need to change those lines 103 and 106 in  `lmfdb/classical_modular_forms/test_cmf.py`

PR #3208, opened 2019-07-25
Next nitpick: the Hecke cutters work on the Galois orbit space

jvoight:


jvoight:
Hmm, this requires a choice of character to be true; I'm going to think about how best to say this.

PR #3211, opened 2019-07-31
bfH to bbH

jvoight:
We use blackboard bold for RR, and CC; we should do the same for HH, the next term in the sequence.  :)


jwj61:
We have a macro for this, \H.  It is better to use the macros (including for Z, Q, R , C, etc) in case we change styles at some point in the future.

jvoight:
Want to revert and change it to \H then?  (Sorry, I didn't know about the macros.)

jwj61:
There are various places they are not used.  In any case, I wouldn't revert, but just change again on top of this change.

PR #3213, opened 2019-08-01
avoiding deprecation warning from Markdown, converting add to register in markdown

edgarcosta:
We recently started to get
```
DeprecationWarning: 
Using the add method to register a processor or pattern is deprecated. 
Use the `register` method instead.
```
This PR fixes this.
My changes were based on https://github.com/Python-Markdown/markdown/blob/master/markdown/util.py
No page should behave differently.


AndrewVSutherland:
Thanks for fixing this!

Issue #3216, opened 2019-08-01
Isogeny degree queries on elliptic curves over number fields return no results


AndrewVSutherland:
Searching for elliptic curves over number fields by isogeny degree does not appear to be working properly; entering in any number (or range of numbers) gives no results.  Entering 1 should find all the curves in the database, and entering 2 should find all the curves with a 2-torsion point, of which there are many, so something is wrong.

These searches work correctly for elliptic curves over Q.

AndrewVSutherland:
Now fixed on www.lmfdb.org

PR #3217, opened 2019-08-01
Isodeg

roed314:
Addresses the bug pointed out in #3216.


AndrewVSutherland:
Thanks for the quick fix!

PR #3220, opened 2019-08-02
More artin script

jwj61:
Just a script update


roed314:
Why is `nottest` set consecutively to `False` then `True`?  Otherwise it looks okay.

jwj61:
It is so that I can change the value by inserting a comment character on the second line which is several keystrokes less than just changing the value.

PR #3224, opened 2019-08-07
Add update_from_file method to db object

roed314:
The current methods for updating a table don't handle the case of updating a single column for every row very well: `upsert` is slow when run on many rows, `update` can't handle data that varies by row, `rewrite` requires that the new values be computable from existing ones, and using a raw `reload` is cumbersome.

This PR adds a new method that allows a developer to provide a file of the same format used in `copy_from` and `reload`: a header and then one line per record in the table to be updated.  Unlike those methods, only columns being change need to be included: other data will remain unchanged.  By default, the change is made to a copy of the table so that it can be undone using `reload_revert`; an `inplace` keyword is also available which may be useful for large tables like L-functions.

### DIFF
--- a/Postgres_FAQ.md
+++ b/Postgres_FAQ.md
@@ -297,7 +297,10 @@ Note that you need editor priviledges to add, delete or modify data.
 
 1. How do I update data?
 
-   There are a number of methods for updating data.  One option is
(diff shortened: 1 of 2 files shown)


edgarcosta:
I will review this

Issue #3225, opened 2019-08-09
elliptic curve link broken


JohnCremona:
From the issue tracker: "When looking at the elliptic curve http://www.lmfdb.org/EllipticCurve/Q/147b1/, at the bottom of the page you are able to see that over another field it has torsion Z/13Z. The 'base change curve' link there redirects to nothing, as in the broken link URL above. I am not sure if it is supposed to link to anything or not." <cgmcwhor@syr.edu>

I can confirm that the link is to the URL http://www.lmfdb.org/EllipticCurve/3.3.49.1/27.1/a/2 which gives an error.  The isogeny class page http://www.lmfdb.org/EllipticCurve/3.3.49.1/27.1/a/ is fine but the links to the curves in the class are broken.

The error is in Lfun['Lhash'] on line 606 of WebEllipticCurve.py

JohnCremona:
http://www.lmfdb.org/EllipticCurve/3.3.49.1/27.1/a/ now works fine.

PR #3226, opened 2019-08-09
L-function missing

roed314:
See #3225.  There was a code path in the construction of an ecnf where it expected to have an L-function computed on the fly (totally real with `self.abs_disc ** 2 * self.conductor_norm < 70000`), but then the L-function lookup failed because it wasn't in the database.  I'm not sure that this is the right solution, but it at least prevents a server error.


edgarcosta:
I will review this

PR #3230, opened 2019-08-12
Updating Travis to use Sage to 8.8

edgarcosta:
I also fixed a test, so it passes with 8.3 and 8.8


edgarcosta:
Given that there no code changes, except a test, I will go ahead and merge this.

Issue #2771, opened 2019-03-15
Remove input boxes for changing count from refine search results


roed314:
There not be an input box for changing count when refining search results, since this screws up the Prev and Next buttons.  For example, if you go to http://www.lmfdb.org/NumberField/?start=50, change the `number of fields to display` box to 5, then click next, you'll see results 56-60.

jwj61:
I thought this used to work.  Is it unfixable?

edgarcosta:
@roed314 and @jwj61 perhaps you should talk about this in person

AndrewVSutherland:
It has been agreed that the number of results to display input box will be removed from all refine search results pages (it is not really necessary and it takes up valuable screen real estate).

This means we don't have to fix this bug (which really is a bug, it did work at one point), but we do have to fix the somewhat related problem #2641.

jwj61:
This was done for number fields, so removing the nf label

roed314:
Fixed by #3233.

Issue #2641, opened 2018-10-02
Reset offset when refining a search.


roed314:
When re-searching after clicking the "next" button a few times, the new search will start at the same offset rather than 0.

alexjbest:
The somewhat related ticket I hinted at earlier was #2621

AndrewVSutherland:
Just to bump this issue, this really is a bug, and it can lead to some seriously confusing results e.g.

http://www.lmfdb.org/EllipticCurve/Q/?start=50&conductor=11&surj_quantifier=include

which to the unwary makes it look as if either there are no elliptic curves of conductor 11 in the LMFDB, or there are 50 of them but they are not being displayed (note the the download links will all give the correct answer, it is only the screen display that is broken).

(to reproduce, go to the http://www.lmfdb.org/EllipticCurve/Q/, click search, then click next, then enter 11 into the conductor input box, then click search again)

JohnCremona:
I note that if you click on "search again" a second time then start is reset and you see the 3 curves you should.  I agree (of course) that that should be what you see the first time, but this this might help debug.

roed314:
This has been resolved by #3233.

PR #3233, opened 2019-08-14
Search again

roed314:
This PR makes a number of changes related to search results pages.

* It addresses #2771 by removing input boxes for changing the number of results per page (on results pages).
* It addresses #2641 by setting `start=0` whenever the Search Again button is pressed (actually, it's set to the empty string so that it's removed by the cleaner)
* It changes the behavior when a large value of start is entered into the URL: now the last page will be shown instead of an empty list of results.  To observe this behavior, go to a search result page and add `&start=1000000000`
* It encapsulates the hidden input boxes into a separate template so that they can be adjusted in the future.
* It cleans up malformatted HTML on various search result pages.
* It removes a `paging` mechanism that wasn't functional.

### DIFF
--- a/lmfdb/abvar/fq/templates/abvarfq-search-results.html
+++ b/lmfdb/abvar/fq/templates/abvarfq-search-results.html
@@ -11,8 +11,7 @@
 <h2> Refine search </h2>
 
 <form id='re-search' onsubmit="cleanSubmit(this.id)">
-    <input type="hidden" name="start" value="{{info.start}}"/>
-    <input type="hidden" name="count" value="{{info.count}}"/>
+{% include 'hidden_search_inputs.html' %}
 
     <table border="0">
 
@@ -120,7 +119,7 @@
 
         <tr>
             <td class="button">
-                <button type='submit' value='refine'>Search again</button>
+                <button type='submit' value='refine' onclick='resetStart()'>Search again</button>
             </td> </tr>
     </table>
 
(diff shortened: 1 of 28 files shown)


jwj61:
In html, closing tags for table rows and table columns are optional, so leaving them out is not malformed html.  If you put one of our pages through an html validator, you will get tons of errors, but missing closing tags for td and tr are among them. I may be alone on this, but I think the reduced clutter makes templates easier to read.

roed314:
> In html, closing tags for table rows and table columns are optional, so leaving them out is not malformed html. If you put one of our pages through an html validator, you will get tons of errors, but missing closing tags for td and tr are among them. I may be alone on this, but I think the reduced clutter makes templates easier to read.

Oh, that's interesting: I hadn't realized that they were optional.  Now that I know, I think there's a slight argument for changing them for consistency (I think most places in the lmfdb use closing tags in tables), but I don't have a strong preference.  Since John likes them better the way they were, I'll go ahead and change them back unless someone else weighs in.

edgarcosta:
I'm also quite surprised about the fact that the closing tags td and tr are optional. 
I prefer the consistency of having them closed, as they are the exception to the rule that tags should be closed.

jwj61:
For the record, they are not the only exceptions.  Some other tags which don't have to be closed are html, head, body, p, dt, dd, li, thead, tbody, etc, not to mention tags where closing tags are forbidden like input, hr, br, meta, etc.

Issue #3227, opened 2019-08-11
Number field search, set of ramified primes superset of is slow


alexjbest:
On http://www.lmfdb.org/NumberField/ we can search fields by ramififed primes containing a given set, for instance http://www.lmfdb.org/NumberField/?ram_quantifier=include&ram_primes=2 but every such query I've tried has timed out, adding more restrictions on degree seems to help, but not always. Maybe some of these queries are cached so be sure to try this with different primes to see.

I appreciate there are a lot of number fields matching such a  query, but I wonder if this is a situation where adding indexes or whatever could help?

edgarcosta:
The root of the problem is the data type.
It should be an array of numerics instead of a jsonb, as indexes on jsonb do not support the superset operator, see https://www.postgresql.org/docs/current/gin-builtin-opclasses.html.

However, the fix isn't as easy as changing the data type, we already have a column `ramps_array`  with the right data type, but our current backend doesn't support searches on `numeric[]`, as PostgreSQL doesn't auto-cast integers to numeric.

```
 SELECT * FROM nf_fields WHERE ramps_array <@ ARRAY[5] LIMIT 1;
ERROR:  operator does not exist: numeric[] <@ integer[]
LINE 1: SELECT * FROM nf_fields WHERE ramps_array <@ ARRAY[5] LIMIT ...
                                                  ^
HINT:  No operator matches the given name and argument type(s). You might need to add explicit type casts.
```

However,
```
SELECT * FROM nf_fields WHERE ramps_array <@ ARRAY[5]::numeric[] LIMIT 1;
```
or
```
SELECT * FROM nf_fields WHERE ramps_array <@ '{5}' LIMIT 1;
``` 
works fine.


@roed314 and I are considering typecast every array, but we haven't gotten around to do it.

roed314:
I'm not sure that the data type is the true issue.  The following queries are executed on grace.mit.edu; the only difference is the LIMIT of 1 versus 100.

```
lmfdb=> EXPLAIN ANALYZE SELECT * FROM nf_fields WHERE ramps_array <@ ARRAY[5]::numeric[] LIMIT 1;
                                                     QUERY PLAN                                                      
---------------------------------------------------------------------------------------------------------------------
 Limit  (cost=0.00..19.56 rows=1 width=363) (actual time=0.014..0.015 rows=1 loops=1)
[...]

PR #3236, opened 2019-08-15
Add support for explicit typecasts on array columns

roed314:
Relevant for #3227, but I'm not sure if it actually addresses the full issue there.


edgarcosta:
I reverted the change on the number field search to use arrrays, as I will change the table instead

PR #3244, opened 2019-08-20
Split ors

roed314:
Add the ability to split the `$or` clause of a query into multiple queries which are then combined and sorted in Python.  This speeds up searches for Galois groups because it causes Postgres to use the correct indexes.

### DIFF
--- a/lmfdb/backend/database.py
+++ b/lmfdb/backend/database.py
@@ -27,6 +27,7 @@ You can search using the methods ``search``, ``lucky`` and ``lookup``::
 
 import datetime, inspect, logging, os, random, re, shutil, signal, subprocess, tempfile, time, traceback
 from collections import defaultdict, Counter
+from itertools import islice
 from glob import glob
 import csv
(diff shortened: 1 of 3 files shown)


AndrewVSutherland:
There are two CMF Travis errors -- I don't know why/how this change would have caused them but we need to figure this out before we can merge this.

AndrewVSutherland:
On the plus side, I just tested this and Galois group searches are now very snappy, well done!

AndrewVSutherland:
@roed314 I just confirmed that ./test.sh lmfdb/classical_modular_forms succeeds on master but fails on this branch, so there is definitely something here that needs to be investigated.

roed314:
I agree; I'm trying to figure out what's causing the problem now.

roed314:
I found the problem and pushed a fix; it was an issue with searching with `limit=None`.

roed314:
This travis failure is real too.  I'll fix it later this evening.

PR #3245, opened 2019-08-20
Adjusting search options for p-part and prime-to-p-parts of A,B data for HGMs

jwj61:
It replaces the A_2 and A^2, ... lines with just one set where the prime is picked from a drop down.


AndrewVSutherland:
I like it!   But I'm wondering if A_{p^\perp} would be easier to read if we instead used A_{\perp p}, which I think communicates the same thing (possibly in a better way, since it matches the notation n \perp p used to indicate that n is coprime to p).

jwj61:
The perp symbol is now a superscript to A_p and B_p to make it more visible

PR #3264, opened 2019-08-22
Fix issue 3234 - NF search fails on range starting with 0

jwj61:
To test, search for quadratic fields with discriminant range 0..30, and variations of that.  Currently it produces no results, but with this change, it gives the same result as 1..30, as it should.

The source was that the sign of the sage integer 0 is 0, so we work around that.


roed314:
Looks good to me once tests pass.

PR #3273, opened 2019-08-23
Rename "imprimitivity index" to "rotation number"

jvoight:
I didn't rename the column, it should probably also be renamed to avoid confusion.


edgarcosta:
to avoid merge conflicts I redid these changes in my hgm branch

PR #3275, opened 2019-08-23
Renaming and moving (to make room for knowls)

jvoight:
The header of the HGM page contained static text that I think should be in a knowl.


roed314:
I agree that this looks like it would be better in a knowl, but you just delete it without replacing it with a `KNOWL_INC`.  Was that intentional?

PR #3278, opened 2019-08-27
correctly setting _read_and_write_knowls

edgarcosta:
When running with `webserver`, the code was always reporting 
```
Read/write to knowls: False
```
as when we changed the knowl infrastructure, we didn't adapt this part of the code.

This PR fixes that.


roed314:
Looks good once tests pass.

PR #3282, opened 2019-08-29
Artin efficiency

jwj61:
There were several places where Artin representations would compute the discriminant of the underlying number field which can hang inside sage as it tries to factor a big polynomial discriminant.  This removes all that problem.  In some places, we don't need the discriminant and can easily make a list of "bad" (i.e. ramified) primes or "hard" primes from data in the database.  Otherwise, we call nfinit from pari inside sage with a list of hard primes and pari will not need to factor the polynomial discriminant.

There was then a place where we did something like this, once with pari(...) and once with gp(...).  The results differ in how sage sees the indexing of list entries.  Now we just use pari(...) which simplified some code.

An example where the above problem happens is

  http://beta.lmfdb.org/ArtinRepresentation/2.2777e2.24t138.2c1

which times out (I think it would take a minimum of several days to succeed), but 

  http://beta.lmfdb.org/ArtinRepresentation/2.2777e2.24t138.2c1

will load.  (Note, this page also suffers from katex won't typeset long strings, but that's another fish to fry.)

### DIFF
--- a/lmfdb/artin_representations/math_classes.py
+++ b/lmfdb/artin_representations/math_classes.py
@@ -885,14 +885,8 @@ class NumberFieldGaloisGroup(object):
         try:
             return self._residue_field_degrees(p)
         except AttributeError:
-            # Try to make WebNumberField, but only helps if the field is in our database
-            wnf = self.wnf()
-            if wnf._data is None:
-                from lmfdb.number_fields.number_field import sage_residue_field_degrees_function
-                fn_with_pari_output = sage_residue_field_degrees_function(self.nfinit())
-            else:
-                from lmfdb.number_fields.number_field import residue_field_degrees_function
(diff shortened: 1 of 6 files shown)


JohnCremona:
Yes: in Sage as in python and C, arrays are indexed from 0.  Same in gp, but the pari library is a C library so indexes from 0.  Using pari rather than gp is much more efficient since data is passed at the C level rather than via strings.

jwj61:
I didn't realize there was a gain by using pari(...) as opposed to gp(...).  I was just simplifying things by being consistent.

I realize the link above which is supposed to work is a repeat of the one on beta.  It is supposed to be the same thing, but for a version running this branch.

I just updated this pull request so that katex will be happy with Artin pages, such as the one referenced above.

Issue #3181, opened 2019-07-01
Optimality of elliptic curves /Q


JohnCremona:
The web pages for elliptic curves and isogeny classes over Q claim show which curve in each class is Gamma_0(N)-optimal but this has not been proved in all cases.  A more complete story is told in [these notes](https://github.com/JohnCremona/ecdata/blob/master/doc/manin.txt).  Note that I have in all *optimal* cases shown that the Manin constant is 1, but curve home pages do not say this.

- [x] I will right away add something to both isogeny class and curve home pages to alert users that we do not always know which curve in the class is optimal.  This is only necessary when the class has more than one curve and the conductor is >60000.  At the same time each optimal curve's home page should state that the Manin constant is known rigorously to equal 1.  I have not yet computed all the Manin constants for non-optimal curves -- something else to do. 20190809: optimality verified for all conductors to 250000, with 2 exceptions (see below).

- [ ] I have [data files](https://github.com/JohnCremona/ecdata/tree/master/optimality) listing all the currently ambiguous cases (which isogeny classes, and which are the possible optimal curves).  This data could be uploaded into a new optimality column in the ec_curves table in order to display better information, pending the next action begin completed.

- [ ] For any of the 245299 currently ambiguous classes I can carry out extra computations to remove the ambiguity.  Each one does not take too long but it would take several weeks at least to process all of them. 20190729: optimality verified for all conductors to 250000:  no change in data required in the range 60k-250k except for two classes (235470bb and 235746u) requiring the two curves with Cremona numbers 1 and 2 to be swapped (not yet done).  There are 8 further such classes between 250000 and 400000 which will be switched.

JohnCremona:
I am in the process of computing all the Manin constants --given that the optimal curve has c=1 that is quick, and almost all are 1 anyway (I have yet to see c>3 except for 11a3, 15a8 and 17a4 which have c=5,4,4).  The computations proving that curve#1 is optimal in all cases will take much longer (at least a month, probably longer).

JohnCremona:
I have added new columns 'optimality' and 'manin_constant' to the ec_curves table, currently filled with default values.  The plan is to encode what we know as follows:

[...]

PR #3293, opened 2019-09-02
EC/Q: improved display of optimality and Manin constant data 

JohnCremona:
After new data upload, we can now display better information about optimality and Manin constants for elliptic curves over Q.  See Issue #3181 .

Cases to look at:

Class of size 2, either curve may be optimal:

http://localhost:37777/EllipticCurve/Q/300006b1/
http://localhost:37777/EllipticCurve/Q/300006b2/

3 curves in class, any of the first 3 may be optimal, the 4th is not.
All Manin constants are conditional on the first curve being optimal:

http://localhost:37777/EllipticCurve/Q/499992a1/
http://localhost:37777/EllipticCurve/Q/499992a2/
http://localhost:37777/EllipticCurve/Q/499992a3/
http://localhost:37777/EllipticCurve/Q/499992a4/

2 curves in class, N>260000 but we know which is optimal:

http://localhost:37777/EllipticCurve/Q/300003a1/
http://localhost:37777/EllipticCurve/Q/300003a2/

2 curves in class, the second has Manin constant 2:

http://localhost:37777/EllipticCurve/Q/104393a1/
http://localhost:37777/EllipticCurve/Q/104393a2/

The rogue class 990 where curve 99h3 is optial (LMFDB class label 990.i with 990.i3 optimal):

http://localhost:37777/EllipticCurve/Q/990h/

NB When I was testing this using test.sh I sometimes saw failures in the test of random curves suggesting that the 'min_quad_twist' column is sometimes invalid (it is a dictionary and the key 'label' is invalid).  I still would like this PR to be reviewed for its logic and how the display looks, but I also try to track down the data issue -- which has nothing to do with this PR.

Also, ignore the script code change.

### DIFF
--- a/lmfdb/elliptic_curves/elliptic_curve.py
+++ b/lmfdb/elliptic_curves/elliptic_curve.py
@@ -270,11 +270,20 @@ def elliptic_curve_search(info, query):
         mode = 'append'
     parse_primes(info, query, 'nonsurj_primes', name='non-maximal primes',
                  qfield='nonmax_primes',mode=mode, radical='nonmax_rad')
+    # The button which used to be labelled Optimal only no/yes"
+    # (default no) has been renamed "Curves per isogeny class all/one"
+    # (default one) but the only change in behavious is that we no
+    # longer treat class 990h (where the optial curve is #3 not #1) as
+    # special: the "one" option just restricts to curves whose
+    # 'number' is 1.
     if 'optimal' in info and info['optimal'] == 'on':
+        query.update({'number':1})
+
+        # Old behaviour was as follows:
         # For all isogeny classes except 990h the optimal curve is number 1, while for class 990h it is number 3.
         # So setting query['number'] = 1 is nearly correct, but fails on 990h3.
         # Instead, we use this more complicated query:
-        query.update({"$or":[{'iso':'990h', 'number':3}, {'iso':{'$ne':'990h'},'number':1}]})
+        # query.update({"$or":[{'iso':'990h', 'number':3}, {'iso':{'$ne':'990h'},'number':1}]})
 
     info['curve_url_LMFDB'] = lambda dbc: url_for(".by_triple_label", conductor=dbc['conductor'], iso_label=split_lmfdb_label(dbc['lmfdb_iso'])[1], number=dbc['lmfdb_number'])
     info['iso_url_LMFDB'] = lambda dbc: url_for(".by_double_iso_label", conductor=dbc['conductor'], iso_label=split_lmfdb_label(dbc['lmfdb_iso'])[1])
--- a/lmfdb/elliptic_curves/isog_class.py
+++ b/lmfdb/elliptic_curves/isog_class.py
@@ -60,12 +60,16 @@ class ECisog_class(object):
         self.curves = [db.ec_curves.lucky({'iso':self.iso, number_key: i+1})
                           for i in range(ncurves)]
 
-        # Set optimality flags.  The optimal curve is number 1 except
-        # in one case which is labeled differently in the Cremona tables
-        self.optimality_known = (self.ncurves==1) or (self.conductor<OPTIMALITY_BOUND)
+        # Set optimality flags.  The optimal curve is conditionally
+        # number 1 except in one case which is labeled differently in
+        # the Cremona tables.  We know which curve is optimal iff the
+        # optimality code for curve #1 is 1 (except for class 990h).
+        self.optimality_known = (self.optimality==1) or (self.iso=='990h')
         self.optimality_bound = OPTIMALITY_BOUND
+        self.optimal_label = self.label if self.label_type == 'Cremona' else self.lmfdb_label
         for c in self.curves:
-            c['optimal'] = (c['number']==(3 if self.iso == '990h' else 1))
+            c['optimal'] = (c['optimality']!=0)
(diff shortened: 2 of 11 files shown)


JohnCremona:
The data error in min_quad_twist is currently being fixed with a rewrite.  This does *not* affect this PR.

JohnCremona:
> The data error in min_quad_twist is currently being fixed with a rewrite. This does _not_ affect this PR.

Fixed, just waiting for the ec_curves table to be copied from dev to web.  This PR need not wait for that, it is independent.

AndrewVSutherland:
Things look good but I do have a few comments/question:

(1) The search results when "optimal" is chosen currently include both conditionally optimal curves, and in cases where we don't know which is optimal, one of the curves (e.g. 300006b1 but not 300006b2, even though the curve pages now say that their optimality is unknown.  I think we should probably not change the search (i.e. still always return exactly one curve in each isogeny class when optimal is chosen), but I think we should put an asterisk on the "Optimal" caption in the browse and refine search pages and update the knowl ec.q.optimal to explain that this the optimal curve in an isogeny class is not always known (or may only be conditionally known).  This would be analagous to what we do with analytic rank and analytic order of Sha in the genus 2 pages.

(2) I'm curious about isogeny classes like 499992a where the Manin constant is conditionally 1, depending on a particular curve being optimal.  Would the Manin constant change if a different curve were optimal, or is this simply not known?  If we know it would still be 1 no matter which curve is optimal then I don't think we need to include the "conditional" parenthetical.

JohnCremona:
Excellent points.

   1. This is an issue I considered but had not solved yet.  In fact the search does not currently use the 'optimality' column at all but just relies (as it always did) on the optimal curve being #1 (except 990h).

I mostly agree with your proposal: the reason for choosing "optimal only" is normally to get one curve per class, not because the searcher particularly cares about optimal.  So would it be better just to rename that button from "optimal only" yes/no to something like "all curves"/"one curve per isogeny class"?

   2. Some where I documented the fact that even when we do not know which curve is optimal, we do know that the optimal curve has m.c.=1.  The reasons are given in the paper cited in the Manin Constant knowl.  More information is in https://github.com/JohnCremona/ecdata/blob/master/doc/manin.txt
which I rewrote a couple of days ago (and is not otherwise published).
Hence it would be possible to give a complete list of what the Manin constants would be for every curve in the class for each possibility (of which is the optimal one), but I thought that more work than really useful.  Here's the story for 49992a:

Let x=0.118954 and y=0.09233926.  The the four period lattices (of the four global minimal models of the curves a1, a2, a3, a4) are

L1 = [4x, 2x+2yi]
L2 = [2x, 2yi]
L3 = [2x, yi]
L4 = [x,2yi]

with L1 subset L2 \subset L3 and also L2 \subset L4.  (It's a theorem of Stevens that one of these is contained in all the others, and his conjecture is that this one is always the Gamma_1(N)-optimal curve.  Add to that the fact that the Gamma_1(N)- and Gamma_0(N)-optimal curves are almost always the same and one would conjecture that curve 1 is optimal.)

The modular symbol computations in the plus space tell you that either the lattice is rectangular with least real period 2x (compatible with L2 and L3) or is non-rectangular with least real period 4x (compatible with L1 only).  This is because with the plus space only we only determine the projection of the lattice onto the real line.

If a1 is optimal then all Manin constants are 1.  (Proof:  Let c be the constant for a1.  By definition, comparing the period lattices of curve a1 and the newform,  they are homothetic with factor c (that is the definitino of c);  but the least real periods of both the newform and the curve are 4x (to several decimal places) and c is an integer (result of Edixhoven) so c=1.  Note that this does not require much precision, as we compute two real numbers whose ratio is known to be an integer and want to be sure that the ratio=1.   Now since the lattice L1 is contained in the others it follows that all the constants are 1.

If a2 were optimal then the constants would be 2, 1, 1, 1 while if a3 were optimal they would be 4, 2, 1, 2.  So the only curve whose M,c. is certainly 1 is a3!

I hope this detailed example was illuminating!  It makes it very hard to display concisely the maximum amount of information.

Meanwhile my runs to eliminate the ambiguities by computing the minus modular symbols have reached 268000 and are continuing.

JohnCremona:
On thing I realised while writing the above:  at the end of https://github.com/JohnCremona/ecdata/blob/master/doc/manin.txt is a list of 48 isogeny classes where plus symbols were not enough to determine that optimal curve has m.c.=1.  For each of these I had already computed the full modular symbol space to eliminate the ambiguity.  Most of these have N<260000 anyway, but the last 18 have N>260000, and I should change the data for those 18 cases to show that the optimal curves is known (and is #1). I'll do that now before I forget.

AndrewVSutherland:
@JohnCremona Thanks for the detailed explanation!  I don't have a strong opinion about how the "optimal only" button is labelled.  It might be safer (and less aggravation) to label it differently and say that it just returns the curve whose Cremona label ends in 1 (and you could even get rid of the hardwired code to handle 990h3 if you wanted to)

JohnCremona:
The latest commit renames the "Optimal only" button on the top search page (and search results page) to "Curves per isogeny class" with options "all" (default) and "one".  Behaviour is exactly the same as before *except* that for class 990h (as for all other classes) when "one" is selected the curve returned is 990h1 and not 990h3.  This required a test change too.

I also increased OPTIMALITY_BOUND to 270000 since I have passed that point in my verification.

PR #3297, opened 2019-09-03
EC/Q: small efficiency gain for minimal quadratic twists

JohnCremona:
Until now each elliptic curve (over Q) stored its minimal quadratic twists's Cremona label only (together with the twisting parameter) since the upload script did not have access to LMFDB labels.  I have now added LMFDB labels to the min_quad_twist column's dictionary, and also changed the upload script so new curves uploaded will have this set too.  This avoids the need to do a database lookup in creating a WebEC instance to find out the LMFDB label of the MQT from its C-label.  At the same time I adapt the code so that ec-curve pages show the C-label or LMFDB label  of the MQT as appropriate.

Compare http://localhost:37777/EllipticCurve/Q/7744y1/ (which shows the MQT as 11a3) with http://localhost:37777/EllipticCurve/Q/7744/k/3 (showing the MQT as 11.a3), noting that 7744y1=7744.k3 and 11a3=11.a3.

The branch in this PR was based on the one in PR #3293 so this should wait until that one is merged.  This is safe to be pushed to web since the amended curve table has already been copied.


AndrewVSutherland:
Looks good but I'll hold off merging until #3293 is merged.

AndrewVSutherland:
Looks like there is a minor merge conflict that needs to be resolved before this can be merged.

JohnCremona:
OK, I will fix that.

Issue #3283, opened 2019-08-30
EC/Q search combinations involving "Optimal only" fail


JohnCremona:
(From the feedback spreadsheet)

If I try to search for elliptic curves of conductors 32 *or* 64 (other variants possible) by entering "32,64" in the field conductor and then select "Optimal only: Yes," I get all curves of conductor 32 or 64, not only optimal ones. If I later click "Search again" I lose my results and get a ton of curves instead of only of conductors 32 or 64.	

I confirm that http://www.lmfdb.org/EllipticCurve/Q/?conductor=32%2C64&surj_quantifier=include&optimal=on returns over 1.7 million curves while without the last part http://www.lmfdb.org/EllipticCurve/Q/?conductor=32%2C64&surj_quantifier=include gives only 8 (the correct 8).

JohnCremona:
Another similar report, should be dealt with at the same time:

"I searched for curves with low conductor (using the default 1-99) and rank 1, but as you can see from the URL, it is entering the search term as the torsion order.  When I directly change the URL, I receive (what I assume is) the correct list of curves, but now the "1" is entered in the CM box instead of the rank box.  I have screen shots of the searches that I would be happy to send if for some reason you are unable to reproduce it on your end."

JohnCremona:
Fixed on beta by the PRs mentioned above.

PR #3301, opened 2019-09-04
Fix part of issue 3283.

jwj61:
This addresses the second problem mentioned in issue #3283 where giving the conductor and rank causes the search to fail.  The problem was just that input boxes and labels were not aligned properly.


AndrewVSutherland:
@jwj61 Thanks for fixing this, I'm sure this was a mistake I made when implementing #3286.

PR #3304, opened 2019-09-05
Finish fixing issue 3283

jwj61:
See issue #3283 for samples.  The problem was that the optimal flag would overwrite any "or" in the search, and the example there was producing an or for conductor (equals 32 or equals 64).  The fix is to use update_ors so it combines them properly.


jwj61:
Pull request #3293 fixes the same issue in a different way, and will conflict with this one, so closing this one.

JohnCremona:
Just so I understand, the problem was from including this in the query 
query.update({"$or":[{'iso':'990h', 'number':3}, {'iso':{'$ne':'990h'},'number':1}]})
(now removed by #3293) ?

jwj61:
Yes, there was already an "or" and that line replaced it.

PR #3317, opened 2019-09-06
Fix local field knowls

jwj61:
In the knowl for a local field, one of the pieces of data is the degree.  Currently, it shows the T-number of the Galois group and this fixes it so it shows the degree.  As an example:

  http://127.0.0.1:37777/LocalNumberField/2.8.8.7
  http://www.lmfdb.org/LocalNumberField/2.8.8.7

Look in the intermediate fields section for local field knowls, and then inside a knowl, for the line which gives the degree

### DIFF
--- a/lmfdb/local_fields/main.py
+++ b/lmfdb/local_fields/main.py
@@ -73,7 +73,7 @@ def local_field_data(label):
     ans += 'Extension of $\Q_{%s}$ defined by %s<br>'%(str(f['p']),web_latex(coeff_to_poly(f['coeffs'])))
     gt = f['gal']
     gn = f['n']
-    ans += 'Degree: %s<br>' % str(gt)
+    ans += 'Degree: %s<br>' % str(gn)
     ans += 'Ramification index $e$: %s<br>' % str(f['e'])
     ans += 'Residue field degree $f$: %s<br>' % str(f['f'])
     ans += 'Discriminant ideal:  $(p^{%s})$ <br>' % str(f['c'])


edgarcosta:
Your changes look good to me.

The test failure comes from `HigherGenusWithAutomorphismsTest.test_random`. I'm not sure why. I will trigger Travis to rerun.

JohnCremona:
From Travis:

   def test_random(self):
        L = self.tc.get('/HigherGenus/C/Aut/random',follow_redirects=True)
>       assert 'Dimension of the corresponding' in L.data
E       AssertionError: assert 'Dimension of the corresponding' in '<!DOCTYPE html>\n<html lang="en">\n  <head>\n    <meta http-equiv="Content-Type" content="text/html; charset=UTF-8">\...middot;\n      SageMath version 8.8\n      &middot;\n      LMFDB Release 1.1\n    </div>\n</div>      </body>\n</html>'
E        +  where '<!DOCTYPE html>\n<html lang="en">\n  <head>\n    <meta http-equiv="Content-Type" content="text/html; charset=UTF-8">\...middot;\n      SageMath version 8.8\n      &middot;\n      LMFDB Release 1.1\n    </div>\n</div>      </body>\n</html>' = <Response 16732 bytes [200 OK]>.data
lmfdb/higher_genus_w_automorphisms/test_hgcwa.py:29: AssertionError

Hard to test these random pages.  Perhaps random tests like this should output a label or URL so we can test which pages gave the error.  I will restart Travis while I take a look at the PR since obviously the failure has nothing to do with these changes.

PR #3320, opened 2019-09-06
Nilpotency

jwj61:
This fulfills a request to be able to search Galois groups based on whether or not they are nilpotent.  Rather than yes/no, we have computed the nilpotency class as an integer.  You can include it in searches (1 for abelian, -1 for not nilpotent, ...).

To see the display the value on a home page of a group, from 

  http://127.0.0.1:37777/GaloisGroup/?solv=True&n=8

you can click on Q_8 to see a group of class 2, and 8T37 for a group which is not nilpotent, in which case there is a little added message on the page.

There is an issue for copying data to the cloud before this goes to the main server.


JohnCremona:
Please can you make "Nilpotency class" on an individual group page a knowl?

Otherwise good to go.

jwj61:
Sure -- an oversight.  I am in an airport, but will get to it tomorrow.

jwj61:
Fixed the knowl id on the page showing a group.

Issue #3319, opened 2019-09-06
Searching for curves with torsion


davidfarmer:
On the main page for Elliptic Curves ( /Elliptic ),
if I put 16 in the search box for Torsion order,
the resulting page has the 16 in the Isogeny degree field.

Also, I did not see any E/Q among the 117 curves it found.
(The first such example has conductor 210.)
Also, I could not guess the order in which the results were returned.

If I do the same search on the E/Q page ( /Elliptic/Q ),
then on the resulting page the 16 is in the Rank field.
(And searching again produces the same results, even though
it looks like I am now searching for Rank 16).

AndrewVSutherland:
This EC/Q torsion order 16 search works correctly on www.lmfdb.org so that part of the issue must be due to  very recent change (I'm pretty sure I know where the problem is, I will look at it shortly).

JohnCremona:
First quick point: the databases of curves over Q and over other fields are separate.  If you search from http://www.lmfdb.org/EllipticCurve/  you are not looking in the same table as if you search from http://www.lmfdb.org/EllipticCurve/Q/.  The only warning that this is the case is in the top line of http://www.lmfdb.org/EllipticCurve/ where it says (not including Q) somewhere.  Perhaps this should be made clearer.

The second thing is  fixable, as we just fixed a similar issue for EC/Q -- the 3rd oint above is OK on beta after that fix.

AndrewVSutherland:
Thanks @JohnCremona, you are right, the 3rd problem @davidfarmer noted is fixed on beta (via #3283) now that I have pushed the fix.

JohnCremona:
I would like to close this since I have no intention that searching on the top Elliptic Curves page should include curves over Q.  But it can be kept open pending a clearer indication that for results over Q a different search is needed.

davidfarmer:
It is a mixed signal that the first thing to browse on the top
Elliptic Curves page is curves over Q.  Probably that is why I
expected to see results from the search.

Maybe removing that is all that is needed?

JohnCremona:
Yes. I suggest deleting Q from the top of that browse list and instead add a sentence above that list saying something appropriate like 

This page is only for elliptic curves defined over number fields other than Q. Elliptic curves over Q have a dedicated set of pages.

With a link to the /Q page somewhere in there.

davidfarmer:
Good.  Then let's interpret that as the point of this issue.

rbommel:
I will change this.

PR #3321, opened 2019-09-06
Fix ecnf searching

jwj61:
Fix issue #3319, at least the part where torsion order ends up in the wrong box on the search results page.


JohnCremona:
Looks good to me, can ber merged once tests pass.

AndrewVSutherland:
Thanks for fixing this @jwj61, this bug was also surely my fault (introduced in #3286).

PR #3335, opened 2019-09-06
Gsm link

jwj61:
On the bottom of the home page of a local field is a Galois splitting model (if we have one).  This defines a global field, so it should be a number field knowl if we have the field.  This pull request does that.  One

http://127.0.0.1:37777/LocalNumberField/13.6.5.5
http://beta.lmfdb.org/LocalNumberField/13.6.5.5

the polynomial is such a knowl, but 

http://127.0.0.1:37777/LocalNumberField/5.15.15.6
http://beta.lmfdb.org/LocalNumberField/5.15.15.6

is a case where we do not have the field in the database, so it just prints the polynomial.

### DIFF
--- a/lmfdb/local_fields/main.py
+++ b/lmfdb/local_fields/main.py
@@ -15,6 +15,7 @@ from lmfdb.local_fields import local_fields_page, logger
 from lmfdb.galois_groups.transitive_group import (
     group_display_knowl, group_display_inertia,
     group_pretty_and_nTj, small_group_data, WebGaloisGroup)
+from lmfdb.number_fields.web_number_field import formatfield
 
 import re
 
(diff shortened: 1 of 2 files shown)


JohnCremona:
When I look at 13.6.5.5 the polynomial is a knowl but it has no content (appears to open up to a tiny box with nothing in it).  But the website (running on legendre) keep crashing on me so I'll try again later.  By crashing I mean that entire python process dies either with no message on the command line or a huge python traceback.

JohnCremona:
OK, it is working fine now.

edgarcosta:
@JohnCremona, I honestly don't understand what is going on with Legendre.
@jwj61 reported the same thing, and I have also observed it, but I don't understand what is going on

jwj61:
Sometimes I got an error message which I think was similar to the one we saw in Bristol.  It was some internal sage thing involving gap libraries.  When running on legendre, I found that running without debug seemed to avoid the weird crashes -- it is just less useful when debugging something.

edgarcosta:
This might be the issue:
https://trac.sagemath.org/ticket/28250

Issue #3341, opened 2019-09-06
Searching number fields by Galois group and signature simultaneously returns a server error.


alisonbethmiller:
I tried to search for S3 cubic fields of signature [1, 1] with the query:

http://www.lmfdb.org/NumberField/?signature=%5B1%2C+1%5D&galois_group=S3

and it returns a server error.

jwj61:
The error is in _split_ors in the search parser.

jwj61:
Looking more, the choice of Galois group produces an or of two pairs of (degree, t-number).  The signature restriction forces the degree, and it claims there is an error when it hits the conflicting combination.  Really, it should not raise an error and accept that branch of the or should be dropped.

AndrewVSutherland:
This is still broken, do you have any thoughts on this @roed314 or @edgarcosta

edgarcosta:
We will take care of it on Friday

AndrewVSutherland:
Fix is now on beta, and will be pushed to web shortly, closing.

PR #3364, opened 2019-09-16
Add option for search parser

jwj61:
This fixes issue #3341 .  In general, when the user gives input which doesn't parse correctly, an error message is generated from the input and a message which is customized to the problem.  If a user entered an illegal Galois group nickname, then the message tried to have a knowl with the nicknames, but the html was being escaped, so the result looks like garbage.

This change adds an option for search parsers.  The option is only turned on for the Galois group parser.  It makes it so that the main text for the error (which is hardwired in the parser code) does not get escaped.  The user's input is still put through a cleaner and escaped, so it is still safe.

To test,

  http://127.0.0.1:37777/NumberField/?degree=3&galois_group=3s3&ram_quantifier=exactly
  http://beta.lmfdb.org/NumberField/?degree=3&galois_group=3s3&ram_quantifier=exactly

should show the difference in the error messages.  You can try putting raw html (or other bad things) in the input box(es), to see that it still sanitizes it.

### DIFF
--- a/lmfdb/utils/search_parsing.py
+++ b/lmfdb/utils/search_parsing.py
@@ -29,7 +29,7 @@ BRACKETING_RE = re.compile(r'(\[[^\]]*\])') # won't work for iterated brackets [
 
 
 class SearchParser(object):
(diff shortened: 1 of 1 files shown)


edgarcosta:
Thanks!

PR #3367, opened 2019-09-18
Add progress statements to rewrite

jwj61:
The statements are just for the dumping phase, and are eerily similar to the ones from the mongo version of rewrite.  You can give an optional argument of progress_count to say how many lines should be dumped between each progress statement (defaults to 10000).  For testing, you can always add the argument commit=False so that it will dump, and then go back to what you had before.


edgarcosta:
Pyflakes is failing:
./lmfdb/backend/database.py:28: redefinition of unused 'time' from line 28

apart from that looks good to me

jwj61:
Learn something every day.  I didn't know using time.time() meant I did not need to import it first.

edgarcosta:
@jwj61, you need to import time, but not two times :P

PR #3371, opened 2019-09-19
Add searchers for API2

davidlowryduda:
This pull request essentially adds the on-main searchers of the LMFDB to API2.

## Example comparison between API1 and API2

Let me suggest a comparison between the old API and API2 to see what this pull request helps implement. 

### API2

Going to `/api2/description/searchers` returns the available API2 endpoints. Suppose we are interested in `char_dirichlet_orbits`. One can go to `/api2/description/char_dirichlet_orbits` to see the available keys and their meaning (as taken from the Inventory `/inventory`). I see that modulus is a key. Suppose we want to look at the orbits of modulus 7. One can go to `/api2/data/char_dirichlet_orbits?modulus=7`.

In practice the idea is for this to be mostly machine-readable as opposed to human-readable --- but the discoverability is a nice bonus.

### API1

One goes to `/api/` and sees **dir_orbits**, taking them to `http://localhost:37777/api/char_dir_orbits/`. After reading a few of the first 100 entries in the database, one can get a sense for the available keys. To use API1 to do the same modulus examination as before, one can go to`http://localhost:37777/api/char_dir_orbits/?modulus=i7`, where the `i` in the address means that the modulus is an integer. This sort of type-adding is necessary for the old API.

## Notes

In the original API, database collections that aren't on production, or are maybe even grayed on beta, are exposed. I did not match this.

Chris and Heather also added the ability to have "singleton" pages that reflect the data on the webpage for any individual object in the LMFDB. This is implemented for elliptic curves over Q, but no other collection. This takes a bit more work, and is also something that I haven't done.

### DIFF
--- a/lmfdb/artin_representations/__init__.py
+++ b/lmfdb/artin_representations/__init__.py
@@ -2,6 +2,7 @@
 from lmfdb.app import app
 from lmfdb.logger import make_logger
 from flask import Blueprint
+from lmfdb.api2.searchers import register_search_function
 
 artin_representations_page = Blueprint(
     "artin_representations", __name__, template_folder='templates', static_folder="static")
@@ -18,3 +19,10 @@ import main
 assert main # silence pyflakes
 
 app.register_blueprint(artin_representations_page, url_prefix="/ArtinRepresentation")
+
+register_search_function(
+    "artin_representations",
+    "Artin representations",
+    "Search over Artin representations",
+    auto_search = 'artin_reps'
+)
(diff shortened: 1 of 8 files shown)


JohnCremona:
As a comment to those not following the API2 story:  the first PR for API2 only implemented 2 or 3 of these and David has extended it to all the collection on the main website.  Documentation is as in http://www.lmfdb.org/api2/ (which is unchanged).  This PR touches 8 init files, in each one registering one or more searcher functions.  

Compare http://www.lmfdb.org/api2/description/searchers which lists 3 collections as searchable to http://localhost:37777/api2/description/searchers which lists many.  This is just boiler-plate code (but thanks to @davidlowryduda for doing it), and the only thing which had to be decided was the "human name" of each collection and the "description" of each searcher (which are rather similar).

Issue #3313, opened 2019-09-05
AbVarFq Isogeny Classes - Gaga box


tdupu:
-The angle rank picture was unclear to people
-ordinary/supersingular; simple not displayed

roed314:
Can you elaborate?  I don't understand what you're suggesting to change here.

AndrewVSutherland:
I think part of the request is to display the simple and ordinary attributes in the properties box (of course not ordinary does not imply supersingular).  The p-rank is already there, but it wouldn't hurt to also list ordinary true/false, and I certainly agree that the property box should say whether the variety is simple or not (this seems like the must fundamental invariant after dimension).

I'm not sure what is meant but the "angle rank" picture, but I assume this refers to the plot showing the roots of the Weil polynomial on the circle.  I'm not sure it makes sense to add anything to the properties box explaining this, but we could link the picture to a knowl (in fact the same issue "what does the picture mean" arises elsewhere, so this part of the issue isn't really about abvar per se and should perhaps be a separate issue).

tdupu:
Drew is right. 

Also the way it is displayed it looks like p-rank has something to do with the picture. 

I guess this fits into a general discussion that was had regarding regarding images in the LMFDB, the lack of explanations for them, and where the most appropriate place to out this information is. 

Also, to clarify, all these posts came from the discussions during the September AIM workshop when we were showing a group of new users the database. David Farmer encourages us to post these comments.

AndrewVSutherland:
It looks like ordinary is now in the properties box, but primitive and supersingular are not.  At a minimum I think primitive should be added.

I think the pictures are fine (whether we should be explaining the pictures in the properties boxes or not is a much broader issue that I don't think should be addressed as part of this one).

roed314:
See #3957

PR #3378, opened 2019-09-20
Abelian variety gaga box changes

alexjbest:
Addresses some of the issues in #3313.

Move the picture just below the label, consistent with other pages, hopefully emphasising that is a general picture, a knowl would still be good

Add is ordinary as a yes/no

Make the principally polarized and is jacobian boxes yes no questions. 

Add a general option to encode pictures with a transparent background, which can look nicer.


AndrewVSutherland:
The abvar property box tests is failing.

roed314:
Some issues:

* There are examples where it is unknown whether or not the isogeny class contains a Jacobian or is principally polarized (e.g. http://localhost:37777/Variety/Abelian/Fq/5/2/ac_b_c_ag_i and http://localhost:37777/Variety/Abelian/Fq/4/4/b_af_a_bc).  The Gaga box are displaying these as "No," which is incorrect.  Note that the corresponding fields in the database encode No as `-1`, Yes as `1` and Unknown as `0`.
* Personally, I don't like the changes to the image (the location change is fine).  I understand that the white box is a bit glaring and it's nice to have that absent.  But the blue points have more contrast against the white than against the transparent.  Moreover, I like the thinner circle since it's just meant to orient the points rather than be a part of the picture of the polynomial.  Finally, the bounding box is now too small: if there are points on the real axis they are truncated (e.g. http://localhost:37777/Variety/Abelian/Fq/1/4/ae).
* Finally, a minor style gripe.  Can you add spaces after your commas in function calls?  For example, `encode_plot(P,pad=0,pad_inches=0,transparent = True)` and `fig.savefig(virtual_file, format='png', pad_inches=pad_inches, bbox_inches=bbox_inches,transparent=transparent)`.

Issue #3326, opened 2019-09-06
Elliptic curves isogenies -- knowl


grundman7:
On individual elliptic curves' web pages (for example, http://www.lmfdb.org/EllipticCurve/Q/40/a/2), it would be helpful if the "Isogenies" heading were a knowl.

JohnCremona:
Good point.  This would be a very good first job for a new developer -- it only involves a small edit to one template.

PR #3379, opened 2019-09-20
Isogenies knowl, second try

SamSchiavone:
Added link to Isogenies knowl in ec-curve.html, as suggested in https://github.com/LMFDB/lmfdb/issues/3326


AndrewVSutherland:
Looks good!

PR #3384, opened 2019-09-23
parsing list of rationals, not ranges

edgarcosta:
Just adding the functionality of not allowing ranges when parsing rationals, and exposing `parse_range2rat`


roed314:
I talked to Edgar and we came up with an alternate solution to his original problem which makes this PR unnecessary.

Issue #1958, opened 2016-12-22
Genus 2 curves: search by geometric invariants


sanni85:
After a talk I was discussing with someone about genus 2 and 3 curves and we wanted to look up data.
Adding searching by Igusa-Clebsch invariants would help users. This is connected to Issue #1930

rbommel:
Sam and I are working on this.

PR #3385, opened 2019-09-23
Geometric invariant search for genus 2 curves

SamSchiavone:
@rbommel and I added the ability to search genus 2 curves using Igusa-Clebsch, Igusa, or G2-invariants as requested in https://github.com/LMFDB/lmfdb/issues/1958.


roed314:
Looks good.

PR #3386, opened 2019-09-23
Hgm

edgarcosta:
Rewrote great part of how the pages of HGM families are rendered.

Now we display L-functions for certain p or t as desired.

The display for this L-functions isn't perfect but it is a step in the right direction.

I moved some code from L-functions to other parts, e.g., so we can use the same infrastructure in HGM.

ps: There is a line change in number fields `@cached_function` which should have been part of #3382


jvoight:
Doesn't work for me (probably my fault).  I get

Error: p must be an integer, range of integers or comma separated list of integers

jvoight:
OK, latest patch fixed some issues.  The symbol "x" for ordinary could mean either "yes" or "no", I suggest writing "no" when not ordinary.

PR #3393, opened 2019-10-01
Allow keeping some backups when cleaning up from reload

roed314:
Add the ability to keep a few tables when cleaning up from reloading.


edgarcosta:
We have tested this on nf_fields

Issue #3266, opened 2019-08-23
Some degree 47 number field pages timeout


AndrewVSutherland:
If you click on any of the last 3 search results for degree 47 number fields:

    http://www.lmfdb.com/NumberField/?degree=47

the request will timeout before the page is displayed.  My understanding is that this is due to time spent computing some invariants on the fly that are trivial to compute for lower degree fields but happen to be more difficult in these cases (all of which have Galois group C_47.   There is one other C_47 page, the fourth from the bottom on the list above that will load but takes a long time).

We should either store the invariants that are hard to compute on the fly, or not display them in cases where we know they are going to be tough to compute (the former is clearly the best solution, but the later might be easier to implement in the short term).

jwj61:
I thought we could get away with precomputing only some of the invariants.  In fact, in some cases initializing the number field takes to long, even if we provide the ramifiying primes or an integral basis.  By the time the next rewrite is done, we should have everything precomputed.

AndrewVSutherland:
This is fixed via #3395 (there is still a katex rendering problem, but this is a separate issue.

jwj61:
I submitted a bug report to katex.  They said that the browser was chopping up long bits of text into shorter nodes, so katex was not seeing the matched start/end for latex expressions.  I have not had a chance to look into this further.  If someone else is inclined to tackle this, here are quotes from the two responses on their issue tracker.

----------------------------------
It seems Chrome optimizes by splitting a long Text node into several Text nodes.

You can call normalize() on the element to merge them before passing to KaTeX. I'm not sure it'll be worth adding it to auto-render, as it may increase processing time, but we should definitely add it to the documentation.
----------------------------------
Instead of normalizing, we could do some work in auto-render.js's renderElem/renderMathInText to detect consecutive text fields where the first one has an open delimiter but no closing delimiter, and treat them as if they were merged... though it definitely would make the code messier. ☹️
----------------------------------

Back to me.  The issue is not specific to chrome, but maybe calling normalize at the right time will fix all browsers.

PR #3395, opened 2019-10-02
Use precomuted data for number fields

jwj61:
Several pieces of data have now been precomputed for number fields and put in the database: conductor and Dirichlet character groups for abelian fields, roots of unity information and Frobenius splittings for all number fields.  This changes the code to just read that data from the database.  Some pages will load significantly faster fixing issue #3266 .  Most pages should look the same as before, such as

http://127.0.0.1:37777/NumberField/16.0.675821419082285056.2
http://beta.lmfdb.org/NumberField/16.0.675821419082285056.2

which has root of unity of order bigger than 2 (although the two pages give different generators for the torsion part of the unit group, they are just different choices of generators).

Now

http://127.0.0.1:37777/NumberField/47.47.6807739517051283990370106856968124192474229146184900948675688481754810058334971205734548194436392534997735758198660848138706901604387352879006731328411841.1

loads quickly instead of timing out.  (On the downside, katex cannot handle its integral basis, but that is a different issue.)

### DIFF
--- a/lmfdb/number_fields/number_field.py
+++ b/lmfdb/number_fields/number_field.py
@@ -4,7 +4,7 @@ import ast, os, re, StringIO, time
 
 import flask
 from flask import render_template, request, url_for, redirect, send_file, make_response
-from sage.all import ZZ, QQ, PolynomialRing, NumberField, latex, primes, pari, RealField
+from sage.all import ZZ, QQ, PolynomialRing, NumberField, latex, primes, RealField
 
 from lmfdb import db
 from lmfdb.app import app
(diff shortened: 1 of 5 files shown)


JohnCremona:
@jwj61 You can manually get Travis to rerun tests!   Follow the links t the Travis report page (e.g. https://travis-ci.org/LMFDB/lmfdb/builds/592682462?utm_source=github_status&utm_medium=notification is a couple of clickes away from https://github.com/LMFDB/lmfdb/pull/3395) and in the top right there's a button "restart build".

jwj61:
That's helpful.

Issue #3410, opened 2019-10-11
Cleaning up search urls


edgarcosta:
Some search pages are missing the url clearner on submit
```
onsubmit="cleanSubmit(this.id)
```

For example, in HMF's if one searches the default thing, we get a very long url:
http://beta.lmfdb.org/ModularForm/GL2/TotallyReal/?field_label=&deg=&disc=&weight=&cm=include&level_norm=&bc=include&dimension=&count=50

The cleaner takes away empty strings, therefore, we should make the default option on the drop-down options to have value an empty string.

Here are a couple examples that should be fixed:
http://beta.lmfdb.org/NumberField/?ram_quantifier=exactly
http://beta.lmfdb.org/Genus2Curve/Q/?geometric_invariants_type=igusa_clebsch_inv&bad_quantifier=include
http://beta.lmfdb.org/Genus2Curve/Q/?geometric_invariants_type=igusa_clebsch_inv&bad_quantifier=include
http://beta.lmfdb.org/EllipticCurve/Q/?surj_quantifier=include&bad_quantifier=include

This is somewhat related to #2277

edgarcosta:
PR #3411 took care of this issue for every search with the exception of Lattices.

PR #3411, opened 2019-10-11
Fixing ugly search URLs

rbommel:
Together with @SamSchiavone, I fixed the issue #3410 with the search URLs raised by @edgarcosta. Hopefully we got all and everything still works.


edgarcosta:
@rbommel and @SamSchiavone, this PR broke your own tests for genus 2 curves invariants

SamSchiavone:
Thanks for the information. I'll take a look at this. Presumably we deleted something we shouldn't have since the Igusa-Clebsch test is still passed.

Issue #3407, opened 2019-10-07
Adding tests for the new genus 2 curves


edgarcosta:
We recently implemented two new search features:
- searching for bad primes
- searching by invariants

However, there are no tests covering these new features.

CC: @rbommel, @SamSchiavone

PR #3414, opened 2019-10-11
Tests for searching by geometric invariants for genus 2 curves

SamSchiavone:
As requested in https://github.com/LMFDB/lmfdb/issues/3407


edgarcosta:
The test that failed was a timeout out "http://www.msri.org/programs/262"

PR #3420, opened 2019-10-18
Fix to #3412

### LINKED ISSUE #3412: Repeated search yields fewer results for Sato-Tate
When searching for Sato-Tate groups, the search counts drops by 2 every time you search. To reproduce this, start at http://beta.lmfdb.org/SatoTateGroup/?rational_only=yes&count=25 and click on 'Search again'.

AndrewVSutherland:
Fixes #3412 which was caused by a change in how the info dictionary is filled in on searches when we switched from mongo db to postgres (this fix simply changes the indentation of two lines so the the count and number get reset).


roed314:
The main problem is fixed.  Did you also want to change the default count from 25 to 50?  Should we add a test for this behavior?

AndrewVSutherland:
Done.

roed314:
Looks good.

PR #3434, opened 2019-11-01
Fix _split_ors

roed314:
See #3341.


AndrewVSutherland:
The travis failure is in Dirichlet characters and has nothing to do with this PR (which works as desired).  Merging.

PR #3440, opened 2019-11-01
enforce lmfdb.com -> lmfdb.org

edgarcosta:


roed314:
Looks good; test failures are not due to this PR.

Issue #3166, opened 2019-06-23
Download button at top of search result


jenpaulhus:
As per conversation in #3130,  I suggest we include download buttons near the top of searches on the site as well as at the bottom.

AndrewVSutherland:
As noted in https://github.com/LMFDB/lmfdb/pull/3442#issuecomment-552515811, we need to think a bit about exactly how we want to address this issue before implementing a solution -- I think we should revisit this after we have finished uniformizing the search results display (this work is in progress).

AndrewVSutherland:
We discussed this today at our editorial meeting and decided to go ahead and include this in 1.1.2.  In order not to lose any vertical space we would like to have the download links in the same line as the next/prev buttons (perhaps floated to the right). If we shorten "Results (displaying matches 1-50 of ...)" to "Results 1-5o of ..." there should be plenty of space.

tomgrubbmath:
I have attached a crude first pass at what this would look like. At this point I would like some feedback before implementing fully. 

- In what order do people prefer the "Results," "Forward/Back," and "Download" bits? 
- Should there be roughly even spacing between the three, or justified some other way? 
- Is there a central list somewhere telling which objects are "downloadable" and which are not? I can do this just by brute force search through each potential page, but it'd be easier if I could just look it up somewhere. 

![Screenshot from 2020-09-08 15-05-59](https://user-images.githubusercontent.com/53013402/92532534-5723f300-f1e5-11ea-84c8-1285e34b4e28.png)

roed314:
What you've done looks pretty good to me.  I think the "Results" and "Forward/Back" are conceptually linked to each other, so I think it makes sense to have them closer together rather all three equally spaced.

I don't think we need to repeat the number of results now that this has been moved up to the top.  But we do need to think about the case of what it looks like when we haven't computed the number of results. I think the current message, "In order to download results, [determine the number of results](url)" works, moved up to the same place.

I think all of the search pages allow for downloading results (except ones like Siegel modular forms that are non-standard). Is that what you're looking for or do you mean something else?  Probably the ideal thing to do would be to update one of the global templates (e.g. `lmfdb/templates/matches.html`).

tomgrubbmath:
@roed314 Thanks! I will continue along these lines. 

[...]

PR #3442, opened 2019-11-01
Added download button on the top of search results page

rbommel:
Also added a message to the dummy download section saying that downloads haven't been implemented to not confuse the user. This can be seen on all search results pages (e.g. /EllipticCurve/Q/?all=1) except a few that are implemented in another way. This concerns issue #3166.

### DIFF
--- a/lmfdb/abvar/fq/templates/abvarfq-search-results.html
+++ b/lmfdb/abvar/fq/templates/abvarfq-search-results.html
@@ -10,7 +10,7 @@
 
 <h2> Refine search </h2>
 
-<form id='re-search' onsubmit="cleanSubmit(this.id)">
+<form id='re-search' onsubmit="cleanSubmit(this.id)" action="?">
 {% include 'hidden_search_inputs.html' %}
 
     <table border="0">
--- a/lmfdb/artin_representations/templates/artin-representation-search.html
+++ b/lmfdb/artin_representations/templates/artin-representation-search.html
@@ -2,7 +2,7 @@
 {% block content %}
 {# Refine search page #}
 
-<form id="re-search" onsubmit="cleanSubmit(this.id)">
+<form id="re-search" onsubmit="cleanSubmit(this.id)" action="?">
   {% include 'hidden_search_inputs.html' %}
   <table>
     <tr>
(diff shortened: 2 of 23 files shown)


roed314:
What's the purpose of `action="?"`?

roed314:
I think the intention of #3166 was to add duplicate links at the top that actually download data, rather than just link to the bottom.  But @edgarcosta and @jenpaulhus should correct me if I'm wrong.

AndrewVSutherland:
@roed314 Is correct, the intention is to have download links in both places.

rbommel:
The purpose of the ? Is to make sure you go to the top when you press the next button.

I was experimenting with download buttons on the top, but I had two problems. It breaks the JavaScript in case the number is not computed yet. Also, it became a bit ugly on the top, unless I was taking more vertical space which people didn't seem to like.

jenpaulhus:
Chaotic 2 weeks here so I don't have time right now to check this PR but yes, what David said is what I was suggesting.

edgarcosta:
> The purpose of the ? Is to make sure you go to the top when you press the next button.

But the `next` and `back` buttons are in a different form, included in `matches.html`

rbommel:
> > The purpose of the ? Is to make sure you go to the top when you press the next button.
> 
> But the `next` and `back` buttons are in a different form, included in `matches.html`

That is not relevant. The form will copy the "#location' from the URL unless you specify in the action not to do that.

AndrewVSutherland:
Hmm, I'm not sure this is the best solution, in cases where the results all fit on the screen, clicking the download link at the top has no effect (I realize it is meant to take you to the part of the form with the download buttons).  It is also confusing in cases where the download buttons are not yet available because the search results count is unknown (a simple text comment that says "see below for download options" might be better than a button).

I'm also concerned that there are going to be a lot more merge conflicts with the PR that @jenpaulhus and her students have been working on to uniformize the display of search results across the LMFDB.  I'm going to close this PR for the moment, I think we should revisit this and agree on an appropriate solution before implementing it (apologies @rbommel this is completely not your fault, I should have vetted the issue better before tagging it for Release 1.1.1).

rbommel:
Okay, I was not aware that someone was working on the search results. Then it makes sense to wait for that.

jenpaulhus:
We should have the PR which gives uniform search result table style ready to go by early next week.

Issue #3444, opened 2019-11-03
Simplify polynomial and power series display


alexjbest:
A lot of objects that are displayed in latex as sums with many terms (e.g. q-expansions, certain polynomials) are currently implemented as a sequence of short tex strings rather than one long one (i.e. $x^3 + x+ 1$ becomes $x^3$ $+ x$ $+ 1$). If I remember correctly this was done to ensure they would overflow onto the next line gracefully when too wide for the screen or container.

This leads to a couple of annoyances, either the spacing around the plus is a bit off (in q-expansions), or we added $\mathstruts$ everywhere, and if someone copies the tex code from a page they get a string with a bunch of mathstruts in.
The most important issue (to me) is that selecting by clicking and dragging and copy-pasting is a bit weird. The user will copy a string with many short latex blocks in as we coded it which they likely do not want, this makes it hard to paste things directly into a tex document or CAS without editing a lot.

I have done some preliminary checking and it seems like KaTeX handles the overflow problem well for inline math, and the above hacks are not really necessary, so I propose to remove them. 

After this change clicking once on a q-expansion say selects the whole thing which can be copied and pasted directly into a latex file.

One potential downside is this may cause some things to hit the katex string too long issue we saw #3201 , but I think that is unlikely to happen anywhere (those fundamental units were BIG) note we have the auto-knowling of large numbers in q-expansions anyway, which does still break apart the string.

davidfarmer:
I see two problems:  how to make the display look good, and how to enable
the user to select LaTeX code that they can paste somewhere.

In the past it was impossible to satisfy both requirements with the same
text string.  We (correctly!) gave priority to the appearance, because
LMFDB is a website and we do offer download options.

I speculate that it is still impossible, it's just that the compromises
will not be as bad.  But maybe we don't have to compromise.  We can have
javascript that causes copying to actually copy some alternate text
hidden in the page.  That way, we can optimize each version separately.

I can elaborate or do a mockup if the above is not clear.

And by "speculate" I mean "am absolutely sure"

alexjbest:
@davidfarmer I have made this change in PR #3445

Can you elaborate what you mean if you do see some issue there, you sound very sure.
We already do have the copy-tex javascript plugin for katex which is what makes this all nice.

[...]

PR #3445, opened 2019-11-04
Remove many instances of web_latex_split_on_pm as katex handles this well

alexjbest:
See #3444 for discussion, maybe @edgarcosta can make a branch for people to see this, or people can test the old way of course!
I saw no issues with any page I tested with this (and I tried hard to ensure I looked at every possible impacted page). Having done this it looks like a win-win to me.

Maybe more testing to make sure I didnt break web_latex_poly would be good

### DIFF
--- a/lmfdb/characters/web_character.py
+++ b/lmfdb/characters/web_character.py
@@ -65,7 +65,7 @@ from sage.all import gcd, Rational, power_mod, Integers, gp, xsrange, cached_met
 from sage.databases.cremona import cremona_letter_code
 
 from lmfdb import db
-from lmfdb.utils import web_latex_split_on_pm
+from lmfdb.utils import web_latex
 from lmfdb.logger import make_logger
 from lmfdb.nfutils.psort import ideal_label, ideal_from_label
 from lmfdb.number_fields.web_number_field import WebNumberField
@@ -1660,7 +1660,7 @@ class WebHeckeGroup(WebCharGroup, WebHecke):
     @property
     def nfpol(self):
         #return self.nf.web_poly()
-        return web_latex_split_on_pm(self.k.polynomial())
+        return web_latex(self.k.polynomial())
 
     @property
     def codegen(self):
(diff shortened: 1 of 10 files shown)


edgarcosta:
Running on red.lmfdb.xyz

JohnCremona:
Please post some links to pages which are different now to allow people to compare.

alexjbest:
@JohnCremona apologies, here are some of the main examples:
In all cases before clicking a long math string (like a q-expansion) would select one term only to copy, now the whole expression will be selected. The important difference is that if you click and drag to select a whole expression previously and copy paste it there will be a lot of $ signs between each term and mathstruts which was not good for pasting into a tex document of CAS.
Importantly the new code should still not overflow or look weird if the string is too long for the screen, it should still break on a plus or minus and continue on the next line as before.

https://www.lmfdb.org/ModularForm/GL2/Q/holomorphic/1410/2/a/k/
https://red.lmfdb.xyz/ModularForm/GL2/Q/holomorphic/1410/2/a/k/
changes affected the q-expansion trace expansion and hecke polynomials

https://www.lmfdb.org/NumberField/30.30.20159382829191092591451779536401274948781988965475418112.1
https://red.lmfdb.xyz/NumberField/30.30.20159382829191092591451779536401274948781988965475418112.1
should affect the defining polynomial.

I can give more if you like as I made changes to many sections (hilbert mfs, hecke chars, mod l) also but those changes were the same and so these should be representative.

JohnCremona:
Thanks.  I see that (as explained in #3444) the idea is not to have the display look any different, and as far as I can see it does not so I have no objection.

Since we already provide code snippets (in some places) and download options for data, I don't think we should expend a lot of effort trying to make formulas on web pages cut-and-pastable.  But I have no objection to this being merged.

alexjbest:
@JohnCremona yes exactly, though the number field example looks ever so slightly different to my eye on chrome (with the new one looking better of course as the spacing around +- being more balanced and compact now).

I don't propose we do expend any effort doing anything more in this direction right now. But as you say the code snippets are not present for CMFs for instance, and even if they are it can take some time for a CAS to produce the data again. (You are right that someone can download data to get a q-expansion as a list for instance, I'm just lazy and like to copy paste directly from the website ;).)

Issue #2619, opened 2018-08-30
Add an extra search parsing function for (lists of) positive integers


alexjbest:
In David Roe's branch we changed parse_ints to allow negative integer values, for some purposes we wish to require values to be positive so a parse_pos_ints function should be added and used across the LMFDB. The regex for this is already in the file.

PR #3448, opened 2019-11-09
Add parse_posints search parser

roed314:


roed314:
See #2619.  I'm not sure if we should switch some parsers that are currently using `parse_ints` to use `parse_posints` instead....

PR #3453, opened 2019-11-15
Work around copy-tex selection issue

alexjbest:
by adding an invisible div with a block display style after code blocks

This should have no visible effect but if one visits www.lmfdb.xyz/Genus2Curve/Q/18272/a/36544/1 clicks Magma and triple clicks to select the first line of magma code, before the (invisible!) sage code would also be selected and appear when copy-paste, after this should not occur

See also https://github.com/KaTeX/KaTeX/issues/1981 for a report upstream.


AndrewVSutherland:
Travis failure is to an acknowledgement URL that is down at the moment, not related to this PR.

Issue #2977, opened 2019-04-24
List torsion structure instead of torsion order


edgarcosta:
When displaying search results for elliptic curves over Q we display torsion order.
I believe it would be much more helpful to display the torsion structure instead.
The same holds for the isogeny class pages.

JohnCremona:
Would we show this as a list of invariants (of length 0 1 or 2) ?   Or in latex?

davidfarmer:
I find it tedious to see it written as: trivial, or  Z/10Z or
Z/2 Z \times Z/ 3 Z.

I suggest those should be written as: "1" or "10" or "2 \cdot 3".
(Which could be in LaTeX or with the "middot" html character.

It could still be labeled as "Torsion order", but shown
in factored form.

My suggestion involves not using the notation in the
Torsion subgroup/Torsion structure knowl:
http://beta.lmfdb.org/knowledge/show/ec.torsion_subgroup

If that knowl was expanded to allow the structure to also be
written as a \cdot product, then maybe everyone would be happy?

AndrewVSutherland:
We should use invariants, as we do elsewhere in the LMFDB, e.g. for class groups of number fields and torsion subgroups of Jacobians of genus 2 curves.  We already have parsing functions for these, and we already display them in search results for number fields.

Recall that the invariants of a finite abelian group are the unique list of integers [n_1,...,n_r] with n_i dividing n_{i+1} and r the size of a minimal set of generators.  The trivial group is the empty list [].

davidfarmer:
If you want to show the group instead of the order of the group,
then I think Z/2 Z  is better than [2].  The former is immediately
understood by anyone capable of understanding it (even if they did not
already know that "Torsion" refers to a group).  The latter requires
opening a knowl.

I'm just affirming the choice on the genus 2 curves pages:

http://beta.lmfdb.org/Genus2Curve/Q/295/a/295/2

Perhaps if we are discussing entries in a column of a table,
my previous comment about Z/2 Z looking tedious, would apply.
(And I still prefer my suggestion that what is listed is the
order, and it is shown as a number or the product of two numbers.)

AndrewVSutherland:
@davidfarmer Sorry, perhaps I misunderstood, but I thought this issue was entirely about what to display in the search results column (for displaying on the home page of an object, I completely agree we should display finite abelian groups as products of Z/nZ's).   I was suggesting that we should do what is already done on number field search results, see

[...]

PR #3457, opened 2019-11-17
Use torsion structure on elliptic curve search/isogeny pages

vishalarul:
This commit is intended to address issue 2977 "List torsion structure instead of torsion order"
#2977.

The affected pages are:

https://beta.lmfdb.org/EllipticCurve/Q/?all=1 vs
http://127.0.0.1:37777/EllipticCurve/Q/?all=1

and

https://beta.lmfdb.org/EllipticCurve/Q/14/a/ vs
http://127.0.0.1:37777/EllipticCurve/Q/14/a/

@JohnCremona  I made the knowl fix (ec_torsion_structure to ec_torsion_subgroup) and I am also committing the changes from a branch of my own, as per your suggestion. Maybe the previous pull request should be deleted?


JohnCremona:
Looks good now.

Some may not like the empty list [] for trivial torsion, and one could replace it by the word "trivial" or similar, but that is not necessary at this point.

Issue #3271, opened 2019-08-23
yaml files for L-function browse pages are out of date and not used at runtime


AndrewVSutherland:
The files degree?browsetable.yaml in lmfdb/lfunctions/templates which at one time were used to generate the HTML pages for the browse pages are out of date.  They should be recreated using the content of the Degree?.html pages and we should then switch to having jinja render the yaml dynamically and get rid of the Degree?.html files.

Until this issue is addressed, the Degree?.html files should be considered the correct source of information for the L-function browse pages, not the yaml files.

edgarcosta:
Just a link to point to a time when we had these yaml files, as I will delete them in a PR to not confuse people.
https://github.com/LMFDB/lmfdb/tree/ef9532ffa78f13c84cf8c1c7c9fb9c50fbec9dcb/lmfdb/lfunctions/templates

edgarcosta:
We should use Jinja to generate those tables

AndrewVSutherland:
Why?  I'm not convinced that the effort to auto-generate them (and the time wasted every time someone goes to make an update and has to figure out what to do) is less than the effort to maintain the html files, which is minimal.  Other than the formatting changes Jen just made and I think David Farmer and I are the only ones who have touched these files in the past four yeas, and most of the changes we made (like Jen's) involved formatting changes or hiding links that would have been *more* work to implement if we had not just directly edited the html, and I doubt that collectively we have spent more than an hour or two maintaining these pages during that time.

I don't think it makes sense to spend time and effort designing and implementing a solution that is almost certainly going to involve more effort than the problem it is trying to solve.  But maybe I'm missing something?

JohnCremona:
I seem to remember that @davidfarmer used to manage these with a perl (?) script to convert the yaml to html.  Unless there are factors I don't know about I would agree with @AndrewVSutherland on this: just edit the html.

Other places where we try to autogenerate are: 
   - the sidebar.  The yaml file is easy to edit and the python which reads it is simple enough, but the html template is horrendous and almost whenever the yaml file changes, more is needed since the script does not quite work automatically enough.  Also,, the template is being filled for *every page* even though the content is the same for every page...
   - the citations list (see another issue #3450) where html is generated from a bibtex file using bibtex2html.  It may be simpler to add new entries by editing the html manually (at least, once the list has been brought up to date and only a small number are added each time).

AndrewVSutherland:
[...]

PR #3461, opened 2019-11-22
Uniform Table Styles

jenpaulhus:
This pull request uniformizes the look of search result tables, as well as updates the look of the table on the Local Number Field landing page to match the new search result table look.

We made all search result tables have no border, cellpadding 3, alternating color per row, and a line under the column title. We did not touch any of the justification, as there was less consensus on what should be justified in what way. These changes are all in the style.css file now under class ntdata.   

As for Local Number Fields, on the main landing page you can compare:
http://beta.lmfdb.org/LocalNumberField/
http://localhost:37777/LocalNumberField/

This is a table class called "number_table".  We moved all the style information for this to the style.css file (it had been locally (ha!) in the lf-index.html file). 

The search pages we changed/updated are:
Classical MoFos
Maass Forms
HMF
Bianchi MoFos
EC/Q
EC/NF
HGCWA
Ab Var (which you'll notice more white space since their rows are varying height)
Global NF
Local NF
Artin
Galois Groups
Sato-Tate Groups

One continuing issue:  Classical Modular Forms and Elliptic Curves/Q  employ two tiers of column headers which make the headers look a little strange in this new style.  Compare this to EC/NF which also have two tiers but look different.  The EC files have extra style defined locally which we did not touch either.

We did make our changes to these tables, though, as we thought the alternating color and proper spacing were important to have on all tables, but it might be worth revisiting these tables particular tables or adding a new style for those tables with multiple rows of column headings.

Thank you for your patience as my students learned the ropes, and I dealt with some unexpected college service!


roed314:
I pushed this to my blue branch so that it's visible on https://blue.lmfdb.xyz/LocalNumberField/

roed314:
I like the alternating color search result tables.  I think that the column headings are a bit too close together though: it's hard to visually distinguish where one starts and the next begins.

roed314:
I also miss the faint lines separating columns in the search results tables.  The shorter row height is nice though.

AndrewVSutherland:
First let me say thank you, this is a huge improvement!  Two comments:

- Unlike @roed314 I don't miss the vertical lines (which were not present in many of our tables, and in most journal styles one would not see vertical lines separating columns in tables), but I agree that the columns are a bit too close together (compare https://blue.lmfdb.xyz/ModularForm/GL2/TotallyReal/?all=1 to https://beta.lmfdb.org/ModularForm/GL2/TotallyReal/?all=1), for a good example of this. 
 A little more (but not too much) space between columns would be welcome.

- I would add a bold horizontal line at the bottom matching the line at the top separating the headings (like \bottomrule matching \toprule in latex).  This will help to visually distinguish the bottom of the table and is pretty common in many journal styles.

davidfarmer:
I understand that something omitted will initially be missed,
but the good advice is to use as little ink as possible, and in
particular to avoid routine use of vertical and horizontal
rules in tables.

To quote one (of many) places offering that good advice:

   If you’ve properly aligned the data points in your table,
   rules become redundant.

The local number field table is an example, with the integer entries and
right alignment of those integers.

The new table layout is a great improvement, and I applaud both the result
and the process used to get there.

Some possible tweaks:

horizontal space between columns:
https://blue.lmfdb.xyz/NumberField/?degree=4
https://blue.lmfdb.xyz/ModularForm/GL2/ImaginaryQuadratic/?field_label=2.0.8.1
https://blue.lmfdb.xyz/ModularForm/GL2/TotallyReal/?field_label=2.2.8.1
https://blue.lmfdb.xyz/ModularForm/GL2/Q/holomorphic/?weight=9-16

left/middle/right alignment of items in a column:
"Dimension" column:
https://blue.lmfdb.xyz/ModularForm/GL2/TotallyReal/?field_label=2.2.8.1
"Dim", and the heading "A"
https://blue.lmfdb.xyz/ModularForm/GL2/Q/holomorphic/?weight=9-16


I am not suggesting that those tweaks need to occur before accepting
this pull request.

AndrewVSutherland:
I agree with @davidfarmer 's comments and I think it would be worth making the tweaks he suggests before merging if they can be done quickly/easily  (which I expect they can).

Regarding the use of a bottom rule, I think we should definitely remove the horizontal blue line spanning the entire width of the screen that currently appears at the bottom of some of the tables (e.g. https://blue.lmfdb.xyz/LocalNumberField/?all=1) but not others (e.g. https://blue.lmfdb.xyz/Genus2Curve/Q/?all=1).

I do think having  a less obtrusive bottom rule that runs just the width of the table and matches the top rule used to separate the headings would be visually appealing, and not an excessive use of ink. 
 Whether that is done or not, we should remove the horizontal blue line from the tables where it currently appears.

jenpaulhus:
We'll get right on those suggestions.  I think those are all easy fixes we can get finished by Monday.   

@davidfarmer a clarification question.  Your comment on horizontal space between columns, is that reaffirming the earlier comments of Drew and David R. to add a smidgen of space between the columns, or are you suggesting additional tweaks in those situations?

davidfarmer:
I think my comments are the same as the other comments about more space
between columns.

jenpaulhus:
Ok.  We've made most of the requested updates.

1. We removed an <hr> command which gave the bottom light blue line.  There is an <hr> in templates/debug_info.html  which we kept,  as it helps make it clear where the debugging message starts, and it won't show up on beta or live.  Along that line, HMF did not have the debug commands so we added those couple lines of code so that the query which formed the search table is now posted on the search pages for HMF. Also we noticed that ECNF has its own debugging commands.

2. We made left/right padding 5px instead of 3px on the tables (to add a bit more space between columns, but not too much!).  If you want to play around with this number to see what the optimal space between columns might be, its line 661 and 662 of templates/style.css.  It's trivial to make this larger if more space is preferable. We started with a small increase.

3. We added a bold horizontal line to the ntdata table style.

4. We realized we didn't update the main tables for Degree 2, 3, and 4 L-functions.  So we made those to match the style of the search result tables. See:
http://localhost:37777/L/degree2/
http://localhost:37777/L/degree3/
http://localhost:37777/L/degree4/
@davidfarmer, there was a warning not to touch these tables because they were created with a script. I believe I only changed html code so I don't think this PR  messes anything up, but please double check that.


5. A couple things we didn't do in this PR:  we didn't touch tables on any pages in beta (Belyi, Dirichlet, Siegel MoFo, Lattices, etc). I am assuming checking "house style" will be part of the editorial process for these pages.

We did not make @davidfarmer justification suggestions.  The CMF page had a lot of code we would have to first understand to make the change, whereas the justification changes should take one of the authors of this page 2 minutes to fix .

In a similar vein, the Zeroes of Zeta table has a vertical blue bar between every line, I believe because of a <thead> </thead> command. But the scripts to create that would take us a while to dig through also.

AndrewVSutherland:
I think this is looking really good!  Regarding your comment to @davidfarmer, no worries, those yaml files are already (long) out of date (see #3271) so changing the html is fine (I think we should probably just remove the yaml files at this point).

And I agree its best to just ask the authors of the CMF page to fix the alignment issue (@roed314) after this PR is merged rather than digging into it now.

I did notice one issue with your most recent change:

- There is an extra middle centered comma in some the rows on http://localhost:37777/L/degree4/ that should be removed.

Once that is fixed I'm happy to merge (last call for anyone else to weigh in with an opinion).

roed314:
I've made a few more [tweaks](https://github.com/roed314/lmfdb/commit/01ff0badbb743675b175f61dd3a845d6007d10d5) and pushed them to my [blue branch](https://blue.lmfdb.xyz/EllipticCurve/Q/?all=1).  In particular, I made the horizontal bold line for elliptic curves over Q not end halfway across the page, and changed the horizontal padding from 5px to 7px.  Even if we like 5px more than 7, we should add a 5px padding to `<th>` tags (not just `<td>`s).

roed314:
I haven't made any changes to CMFs, since I figured that was better as a followup PR.  I'm also not sure what needs to be changed (see [here](https://blue.lmfdb.xyz/ModularForm/GL2/Q/holomorphic/?search_type=List&all=1) for the current status with this PR).

AndrewVSutherland:
@roed314 I would just get rid of the top horizontal line on https://blue.lmfdb.xyz/EllipticCurve/Q/?all=1 that separates the Curve and Isogeny Class headings from the headings on the line below, as on the current site (https://www.lmfdb.org/EllipticCurve/Q/?all=1).

I'm guessing the half-line was an attempt to imitate  \cmidule in the booktabs package, e.g. as in Table 1 on page 14 of https://arxiv.org/pdf/1410.5222.pdf, for example, but this just doesn't work that well and I think it's fine without any line at all (and more confusing with a line that goes all the way across).

AndrewVSutherland:
@roed314 I agree that addressing the CMF column alignment issue in a separate PR is better, I'd like to merge this one as soon as we can (in addition to the tweaks you made there is the problem on http://localhost:37777/L/degree4/, which is more significant, I don't want to merge that).

jenpaulhus:
I see what you mean about the commas.  They aren't new, they are just showing up on their own line now.  Since @roed314 has already made more changes to the PR, I'll just look into that now and report how to fix it (unless someone beats me to a solution!)

AndrewVSutherland:
@jenpaulhus I don't think David has pushed anything to your branch yet (and if/when he does it shouldn't conflict with any changes you make to the degree4 page unless he changed the padding there as well but I don't think he did).

BTW, I also notice that the left alignment of the "underlying object" and "N" headings on the degree 4 page don't match the alignment of their columns.

jenpaulhus:
It looks best aesthetically to me to just remove the commas in the degree 4 page
http://localhost:37777/L/degree4/
Each term is on its own row anyway.

I changed some align=center to class=center in Degree 2, 3, and 4 to force the centering of column headings (most of the other pages default to left justification of column headings, and that is in fact what the style.css file sets as default).

Also in degree 4 I made all the headers "blue header color" instead of black. I think that was the only table with the non-knowl headers still were in black instead of blue, but not underlined.

roed314:
I updated [CMFs](https://blue.lmfdb.xyz/ModularForm/GL2/Q/holomorphic/?search_type=List&all=1) and [ECQ](https://blue.lmfdb.xyz/EllipticCurve/Q/?all=1) and submitted a PR to this PR (Jen should be able to merge it into the PR).

AndrewVSutherland:
@jenpaulhus Looks good!  But I'll wait to let you merge @roed314's PR into your branch before merging.

AndrewVSutherland:
I noticed that HGM search results still has an extra blue line at the bottom, but we can fix that later.

jenpaulhus:
@AndrewVSutherland  Merge is done.   And thanks, @roed314 for your additions!

jenpaulhus:
> 

> I noticed that HGM search results still has an extra blue line at the bottom, but we can fix that later.

We didn't touch the beta parts of the webpage:  Beyli, HGM, Lattices, etc.  I assumed some of them were still being modified by their authors, so we risked doing unnecessary work by modifying a table that might get changed anyway.  On HGM, the landing page has a table similar to that on the landing page of local fields. There is a separate class we used for that sort of table.

AndrewVSutherland:
Got it, thanks!

PR #3462, opened 2019-11-22
Regulators for genus 2 curves

rbommel:
This displays the new regulator data (maybe some database changes have to be mitigated).


AndrewVSutherland:
Travis failed on genus2 stats (due to having read only access).  I refreshed the stats and ran the test locally and all seems well -- I restarted Travis so he can try again.

AndrewVSutherland:
The knowl for regulator needs to be created, but a link to it is already in the template so I will go ahead and merge the PR.

Issue #3238, opened 2019-08-19
Search elliptic curves over Q by regulator and by number of integer points


AndrewVSutherland:
From Noam Elkies: It would be nice to be able to search the EC/Q database by number of integer points and by regulator (within a specified interval).  The data to support these queries is already present (the number of integer points is the length of the list xcoord_integral_points), so this is simply a matter of adding search options for them (and possibly some indexes to support these searches more efficiently).

roed314:
@JohnCremona Any objections to adding a `num_integral_points` column to `ec_curves`?  Once that's present, it's very easy to populate, and then the front end changes shouldn't be too bad.

roed314:
I don't want to get in the way of your methods for updating the curves by doing something ad hoc.

JohnCremona:
@roed314 No objections but please do not change the ec_curves schema at this point since I am in the middle of uploading data (to be resumed after the weekend which means Tuesday).  I'll add that to a list of minor things to be done in a table rewrite.  We also need to remove the obsolete column galois_images, and add columns relating to optimality and Manin constant.  My plan is to do all that after uploading the new curves.  While we are at it we can also add * number of bad primes (useful to restrick to prime power conductor) * semistable * and more.

roed314:
Sounds good; I won't do anything to the schema for now.

JohnCremona:
I have added new columns 'num_int_pts', 'num_bad_primes', 'semistable' and the last two are already populated with a rewrite but I forgot to do the first.  I'll be doing another rewrite quite soon to add optimality and Manin constant data, so I was not planning to populate the 'num_int_pts' columns until then (but of course I could).  Rewriting takes 1h40m approx.

JohnCremona:
OK, so now all curves have 'num_int_pts' and 'num_bad_primes' and 'semistable' set.  (On dev not www at this point.)

SamSchiavone:
I've started working on adding a search feature for number of integral points. It seems that currently 'num_int_pts' is really recording the number of x-coordinates of integral points, so the number of integral points "up to sign." Is this the desired effect? If I search for num_int_pts = 2, that means I could get curves with 3 (if one of the 2-torsion points is integral) or 4 integral points. Or should I try to change it to give the actual number of integral points? I lean toward the second option, but I'm happy to do either.

JohnCremona:
[...]

PR #3464, opened 2019-11-22
Searching on number of integral points

SamSchiavone:
Addressing https://github.com/LMFDB/lmfdb/issues/3238. I've added functionality allowing one to search for elliptic curves by number of integral points.

I also recomputed the number of integral points using the script update_num_int_pts in import_ec_data.py and rewrote the data in ec_curves. Now the field num_int_pts should give the number of integral points rather than the number of integral points up to sign. I backed up a copy of the data before the rewrite under ec_backup in case there are any problems.


roed314:
It looks like pyflakes failed with
```
./scripts/elliptic_curves/import_ec_data.py:154: 'lmfdb.elliptic_curves.web_ec.make_y_coord' imported but unused
```

SamSchiavone:
Thanks, David! I've deleted that unnecessary import of make_y_coord.

AndrewVSutherland:
There is a print statement that has been added to line 249 of elliptic_curve.py that I expect was put in during development to help with debugging -- this should be removed.

JohnCremona:
I have a small concern about what has happened here.  The data (on beta) has *already* been changed, by overwriting the existing column with the new count.  I think it would have been better to create a new column for the new count.  That would have safer, though in this case the num_int_pts column is not referred to in the website code, so it is OK.  But thiat would also have allowed an option in the integral point search, of which of the two counts to search on.  Never mind that, it would be easy to recreate the old column since the old number is just the length of the x-coordinate list.

The code changes now being asked to be merged are mostly simple.  There are coercions to RR and back to ZZ in the revised code which are unnecessary (my old code did work correctly as it converted each x to ZZ first).

AndrewVSutherland:
@JohnCremona This was my call -- my thinking was that since we don't currently display or allow searching on number of integral points there was no harm in changing the value on beta, especially since the new value is arguably the correct one.   There is a backup copy of the table that we can revert to if needed -- let me know if you would like me to do that.

JohnCremona:
@AndrewVSutherland No it's fine, I had not realised that the number we stored before was never used at all.

Independently of this PR though, we really ought to have a way to propose/approve/implement changes in table schemas (including cases such as this one where the schema has only changes in a subtle way to do with the definition of the stored integer).   We have got away with it up to now mostly because each table has only been ever touched by one person (an exception was of course CMFs but then the people working on it were all talking to each other).

PR #3469, opened 2019-11-25
add missing space

alexjbest:
see https://www.lmfdb.org/EllipticCurve/2.0.11.1/243.5/a/1 also simplified code a little (https://www.lmfdb.org/EllipticCurve/2.0.11.1/240.1/a/1 for the other case)


edgarcosta:
Looks good to me

Issue #3468, opened 2019-11-25
Search parameters after a comma are sometimes ignored


jwj61:
From the feedback page:

When searching for number fields, search parameters appearing after a comma are ignored--but only (as far as I can tell) when the Galois group is also specified.  The URL I linked is the simplest case I could find; the search results omit the quadratic fields with discriminants -23, -31, and so on.  The problem does not occur if the Galois group is unspecified, or if "2,3" is changed to "2-3".  If "2,3" is replaced by "3,2", then only the fields with class number 3 appear.  Similar behavior happens if you specify the number of ramified primes instead of the class number.

I first noticed the problem here:  https://www.lmfdb.org/NumberField/?signature=%5B4%2C0%5D&galois_group=C2xC2&class_number=3%2C6.  In that search, there should be a field with class number 6 between the 6th and 7th results.  Strangely, that search works perfectly when I switch "3,6" to "6,3"--but it breaks again when I try "5-6,3".

Respond to fernando@berkeley.edu

jwj61:
I have not looked at this, but my guess is that it involves "splitting ors" in a search query, i.e., what we need to do when there is more than one "or" in a search query.  It may have been introduced in a recent fix which involved handling queries with an or in them (in the previous case it was when a branch of the or was impossible).

roed314:
I will take a look.

roed314:
See #3470.

edgarcosta:
The changes are on www.lmfdb.org and I have emailed Ravi Fernando informing him that we have fixed the bug.

PR #3470, opened 2019-11-25
Remove optimization on split_ors, change criterion for when it's applied

roed314:
The bug in #3468 was caused by an invalid optimization that's this PR removes.  I also change the criterion for using split_ors to make it more restrictive, so that it isn't actually triggered in the example from that issue.


edgarcosta:
I added two tests to verify that we address #3468.
I those pass we can merge :)

roed314:
Thanks.  I'd meant to add some tests but got distracted.  Tests pass, so I'm merging.

edgarcosta:
No problem. Was easier for me to add the tests than to run it locally :)

PR #3482, opened 2019-11-30
second wave of print fixing in scripts

fchapoton:
from scripts/hmf/ to scripts/reports/

### DIFF
--- a/scripts/hmf/check_conjugates.py
+++ b/scripts/hmf/check_conjugates.py
@@ -11,7 +11,7 @@ Note that labels contain no information about weight.
 Initial version (University of Warwick 2015) Aurel Page
 
 """
-
+from __future__ import print_function
 import sys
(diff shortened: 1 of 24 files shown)


JohnCremona:
As with the other similar PRs: changed print statements all fine; why the future imports?  and these are all scripts anyway.

fchapoton:
The future imports are useful for two things:
* first to make sure that nobody will ever re-introduce some old-style print.
* second, to have the exact same behaviour for print in py2 and py3. For simple arguments, wrapping with parentheses does not change anything. But in more complicated cases, like several arguments separated by commas, the behaviour (without the future import) would be different in py2 and py3.

roed314:
Looks good to me.  I don't think the future imports make a huge difference either way for `print_function`, and I agree with @fchapoton about the benefits.

PR #3483, opened 2019-11-30
3rd wave of python3 print fixing in scripts

fchapoton:
final wave of correction for python3-compatible prints in scripts

### DIFF
--- a/scripts/elliptic_curves/import_ec_data.py
+++ b/scripts/elliptic_curves/import_ec_data.py
@@ -14,6 +14,7 @@ Evolved during 2012-2018
 
 Postgres table ec_curves has these columns (updated 2019-08-05):
 """
+from __future__ import print_function
 
 qcurves_col_type = {
     u'2adic_gens': 'jsonb',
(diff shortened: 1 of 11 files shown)


JohnCremona:
These all look fine to me and completely harmless though I do have one query:  surely it is not necessary to have the import from future for these changes?  I have been using print as a functions for a long time now even with Sage on python 2.  I would suggest removing the imports before merging.  (I also would not mind if this PR was not merged as it only concerns scripts.)

roed314:
@JohnCremona I just checked and the print function does work without the import, though I'm not sure exactly how.

JohnCremona:
@roed314 Well I don't pretend to be a python expert but I have been using print() for years now, long before Sage's recent beta releases which are Python 3.  I also use (for example) print("x={}".format(1)) instead of print("x=%s" % (1)) or similar.

I think it is because Sage (looking at 8.9) *already* imports print_function (it is in sage.all.py).  Perhaps @fchapoton (who has masterminded Sage's upgrade to Python 3) did not know that the LMFDB code runs under Sage.

If pyflakes 3 is happy without the import functions I think we can remove them.

roed314:
Looks good to me.

PR #3486, opened 2019-11-30
some care for TAB characters

fchapoton:
by replacing them by spaces


fchapoton:
This one is even more harmless that all the python3-print requests.

PR #3489, opened 2019-12-06
convert all xrange to range, and replace some TAB by spaces

fchapoton:
as another good step towards python3

The TAB changes are easily checked by looking at the diff.

For the change of xrange, you should (I did it already) check that in the modified files, every range is either inside a for loop or has been wrapped with a list if necessary. So this is a little more work..


roed314:
Looks good to me.

PR #3491, opened 2019-12-06
a bunch of python3 fixes suggested by pyflakes

fchapoton:
another set of rather simple fixes

* get rid of reduce, which is not available in python3
* some syntax errors in raise
* etc

### DIFF
--- a/lmfdb/abvar/fq/main.py
+++ b/lmfdb/abvar/fq/main.py
@@ -107,7 +107,7 @@ def abelian_varieties_by_gqi(g, q, iso):
         return search_input_error()
     try:
         cl = AbvarFq_isoclass.by_label(label)
-    except ValueError as err:
+    except ValueError:
         flash_error("%s is not in the database.", label)
         return search_input_error()
     bread = get_bread(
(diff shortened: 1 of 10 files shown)


roed314:
These all look fine to me.  The only one that I'm a bit hesitant about is the change to `Factorization([(P,e) for P,edd in zip(badprimes,mindisc_ords)])`.  It looks like there's currently a bug that @fchapoton is fixing, but I wasn't able to find a current example where this problem appeared.

edgarcosta:
I agree with you, it looks like we are fixing a bug. 
@JohnCremona will be the best person to approve that change.

JohnCremona:
I'll take a look. The snippet quoted certainly looks bad but the exponents are to be ignored whatever identifier is used.

JohnCremona:
> These all look fine to me. The only one that I'm a bit hesitant about is the change to `Factorization([(P,e) for P,edd in zip(badprimes,mindisc_ords)])`. It looks like there's currently a bug that @fchapoton is fixing, but I wasn't able to find a current example where this problem appeared.

Yes this was definitely a bug.  It would result in the exponents being wrong in the factorization of the minimal discriminant ideal whenever the exponents were not all equal to whatever the last value of "e" was.  The new code is much simpler (and this and the similar changes have nothing much to do with the Python 3 conversion other than because P3 is stricter about the values of dummy variables not being available after a list comprehension so that under P3 "the last value of e" would not exist).

If that was all, feel free to merge.

PR #3492, opened 2019-12-06
some pyflakes changes outside of lmfdb subfolder

fchapoton:
mainly replacing "file(" by "open(" in the scripts
and a few other details

see https://docs.python.org/2/library/functions.html#file
for the relationship between file and open ; file does no longer exist in python3

### DIFF
--- a/data_mgt/lf/convert.py
+++ b/data_mgt/lf/convert.py
@@ -13,7 +13,8 @@ import lf_pb2 as lf
 fn = sys.argv[1]
 fn_new = os.path.splitext(fn)[0] + ".pb2"
 
-data = yaml.load(file(fn))
+with open(fn) as f:
+    data = yaml.load(f)
 #print data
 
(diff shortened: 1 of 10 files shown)


AndrewVSutherland:
These changes all look fine to me but I'll give the script owners a chance to chime in before merging.

JohnCremona:
Everything involving changes from file() to open() and to print() are fine.  I don't understand the two changes involving "with x as y:" -- the ecnf one makes it look more complicated: surely we don't need the additional level of nesting?

AndrewVSutherland:
@JohnCremona the "with open(blah) as blah" expression is standard python usage to ensure that opened files get closed (this will happen automatically a as soon as you leave the scope of the with).  Otherwise you have to remember to have a matching close for every open, and even if you do an exception can result in a file being left open.  The "with open" usage is generally considered best practice.

PR #3496, opened 2019-12-07
py3: some care for map in lmfdb folder

fchapoton:
various changes, most notable

* try to fix the use of map when a list is expected
* same fix for some range
* some open context manager
* one syntax error in raise
* etc

### DIFF
--- a/lmfdb/artin_representations/math_classes.py
+++ b/lmfdb/artin_representations/math_classes.py
@@ -123,7 +123,7 @@ class ArtinRepresentation(object):
         return int(self._data["Conductor"])
 
     def NFGal(self):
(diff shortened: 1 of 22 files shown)


JohnCremona:
See https://www.python-course.eu/python3_lambda.php for some explanations:  in Python3 some functions including map() and range() return iterators not lists, and if you want a list then using list comprehension is just as good, possibly clearer. (I am biassed as I always did use list comprehension rather than map()).

@fchapoton can you explain the changes of the form
```
data = file(fn).read()
```
changed to
```
with open(fn) as f:
        data = f.read()
```
?  
Why not just
```
data = open(fn).read()
```
?
Otherwise I am happy to merge this.

PR #3497, opened 2019-12-07
fix some tiny errors found by lgtm

fchapoton:
see there for the full list of problems identified:

https://lgtm.com/projects/g/LMFDB/lmfdb/alerts/?mode=list

### DIFF
--- a/lmfdb/app.py
+++ b/lmfdb/app.py
@@ -145,9 +145,6 @@ def git_infos():
         return '-', '-', '-'
 
 
-def git_summary():
-    return "commit = %s\ndate = %s\nsummary = \n%s\n" % git_infos()
-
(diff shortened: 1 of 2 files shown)


fchapoton:
maybe "git_summary" could be removed altogether, as it is not used anywhere ?

edgarcosta:
yes `git_summary` can be deleted

fchapoton:
ok, I have removed git_summary

fchapoton:
This one should be a very easy review..

JohnCremona:
Fine.

BTW I did look at LGTM's verdict on our code some time ago and was rather daunted by the large number of things found (not all of which seemed very important), but I follow all its advice for eclib (C++ not Python of course).

PR #3498, opened 2019-12-09
Fixing magma code snippets for magma relating to the disc of a number field

jwj61:
You can test by looking at any number field page.  The one suggested by the bug reporter

https://beta.lmfdb.org/NumberField/8.0.8653650625.2
http://127.0.0.1:37777//NumberField/8.0.8653650625.2

gives an example which, if you run the magma commands, makes a difference.

### DIFF
--- a/lmfdb/number_fields/code.yaml
+++ b/lmfdb/number_fields/code.yaml
@@ -39,17 +39,17 @@ signature:
 discriminant:
   sage:  K.disc()
   pari:  K.disc
-  magma: Discriminant(K);
+  magma: Discriminant(Integers(K));
 
 rd:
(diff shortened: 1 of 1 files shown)


JohnCremona:
Looks good: I used the "Download Magma code" button to get all the commands which run fine and (as far as I can tell) give the correct output (e.g. the list of ramified primes is correctly [5, 61]).

Not to delay this fix, but we should standardise the behaviour of links of the form "Download X": I prefer getting a dialogue box for where to store the file;  next best is the content opening in another tab; worst is -- as here -- the download text is displayed in the same tab as the original html page.

PR #3502, opened 2019-12-10
one duplicate import of re

fchapoton:


lmfdb/artin_representations/math_classes.py
```diff
@@ -14,11 +14,11 @@
...
 def id_dirichlet(fun, N, n):
     N = Integer(N)
-    if N==1:
-        return (1,1)
+    if N == 1:
+        return (1, 1)
     p2 = valuation(N, 2)
     N2 = 2**p2
-    Nodd = N/N2
+    Nodd = N//N2
```
edgarcosta: why the '//'?

edgarcosta: Forget, I had forgotten that in python3, 3/2 = 1.5

JohnCremona:
But the N here is not a python int it is a sage Integer.    Having //
instead of / means that the quotient returned is also a  sage integer
(rounded down if not exact) while with / it would be  a sage Rational even
if exact.

jwj61:
It is moot either way.  Looking at the lines just before it, it is N2 is the largest power of 2 dividing N, and then we divide N by that.  So, it will always be a divisor.

It will also be moot in the foreseeable future.  I am shifting this computation to magma (so precomputed before the lmfdb sees it).  At that time, this function will be removed.

PR #3504, opened 2019-12-10
comparison to None should rather use "is"

fchapoton:
this fixes another big set of lgtm warnings

also contains a python3 fix in the file install-deps.sh


roed314:
Looks good to me.  Will merge once tests pass.

PR #3505, opened 2019-12-10
Removing data_mgt folder

edgarcosta:
The data_mgt folder contains 3 deprecated folders:
- utilities -- deprecated by the transition to Postgres)
- lf -- deprecated since MongoDB? This folder tried to serialize some L-function data into proto
- pipes -- not sure what was the purpose, but clearly deprecated by the transition to Postgres


JohnCremona:
Agreed -- no idea what that stuff in lf/ was intended to be: something @haraldschilly wrote 6 years ago.  Confirmation from someone?

haraldschilly:
Hello, well, no clue and I don't remember much. However here is a link showing you the commits, i.e. there were other people involved than me. https://github.com/LMFDB/lmfdb/commits/master/data_mgt

jwj61:
Most of the commits are related to rewrite, which has now moved into the backend code.  I don't think anyone uses the rest, so it can go.

PR #3507, opened 2019-12-11
fix another lgtm warning

fchapoton:
because "zeros_starting_at_t" takes only 2 arguments


JohnCremona:
OK. @jwbober should the platt_zeros.py file be moved to the scripts/ directory, as it is not involved in the website?

JohnCremona:
I'm merging this.  The question of moving non-website code out of lmfdb/ directory is separate (and there may be more such files).

PR #3510, opened 2019-12-11
py3 : fixing the comparison of elliptic curves (no more cmp)

fchapoton:
as cmp is no longer available

in the script, I change a comparison function, so that it does no longer use cmp.
in the lmdfb file, I remove the comparison function and provide instead a comparison-key function, that can used if sorting is required. I left the sorting line commented.


JohnCremona:
Thanks!

PR #3511, opened 2019-12-11
commiting buffered cursors

### LINKED ISSUE #3494: buffered cursors leave locks behind
When looping over search results without a limit, we use buffered cursors, so one doesn't need to download the full result before starting to loop over it.
However, these cursors leave locks behind, as they start a transaction automatically, see: 
https://github.com/psycopg/psycopg2/issues/987

We need to workaround this, as it is easy for some one to leave a session open in some server, and block anyone from modifying the table.

edgarcosta:
This resolves #3494, by committing the connection instead of just closing the cursor.

To check that it works, run the following snippet with master and with the PR.
```
from lmfdb.backend.database import db, SQL
other = db.cursor()
relationid = list(db._execute(SQL("SELECT OID from pg_class where relname='g2c_curves'")))[0][0]
qactivity = "SELECT  1 from pg_locks l WHERE relation=%s" % relationid
other.execute(SQL(qactivity));
print 'locks at start: %d' % other.rowcount
for e in db.g2c_curves.search():
    other.execute(SQL(qactivity))
    print 'locks in the middle: %d' % other.rowcount
    break
other.execute(SQL(qactivity));
print 'locks at the end: %d' % other.rowcount
```

With the old code we get:
```
locks at start: 0
locks in the middle: 1
locks at the end: 1
```

with this PR you should get:
```
locks at start: 0
locks in the middle: 1
locks at the end: 0
```

This PR has one unintentional consequence, that was also present before we introduced buffered cursors.
If someone runs something like:
```
db.foo.update(..., commit = False)
c = something(db.bar.search(...))
```
outside a `with DelayCommit` the connection will be commited, on the search.

### DIFF
--- a/lmfdb/backend/database.py
+++ b/lmfdb/backend/database.py
@@ -1593,7 +1593,8 @@ class PostgresTable(PostgresBase):
         finally:
             if isinstance(cur, pg_cursor):
                 cur.close()
-                if cur.withhold:  # to assure that it is a buffered cursor
+                if (cur.withhold and # to assure that it is a buffered cursor
+                    self._db._nocommit_stack == 0): # and there is nothing to commmit
                     cur.connection.commit()
 
     ##################################################################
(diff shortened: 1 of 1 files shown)


roed314:
This is an improvement on the current state, so I'm merging.  I'll create another issue to address the problem about `commit=False`

PR #3512, opened 2019-12-12
py3: do not use "long"

fchapoton:
in the case calling long(), I used ZZ instead.

in the type testing cases, I used six.integer_types, see https://six.readthedocs.io/#six.integer_types

I also removed data['j_inv_str'] that was unused.

 and fix one string, because ur"sometext" is no longer allowed.


lmfdb/elliptic_curves/web_ec.py
```diff
@@ -197,7 +197,6 @@ def make_curve(self):
         data['j_inv_factor'] = latex(0)
         if data['j_invariant']: # don't factor 0
             data['j_inv_factor'] = latex(data['j_invariant'].factor())
-        data['j_inv_str'] = unicode(str(data['j_invariant']))
```
edgarcosta: 'j_inv_str', this field is not used anywhere else

PR #3514, opened 2019-12-12
replace all uses of basestring by six.string_types

fchapoton:
see https://six.readthedocs.io/#six.string_types


roed314:
Looks good.  Will merge once tests pass.

PR #3515, opened 2019-12-12
various minor details, some from lgtm

fchapoton:
including one division fix

some better comparison to None

removal of some useless pass

better tests for emptyness of lists


roed314:
I'm fine with all of these changes except the division one.  The behavior there is changing when you pass in an odd Sage integer: it used to give you a rational number which was half integral, and with the change it would give you the floor of that half integer.  Since it's in the scripts folder, I think it should be left unchanged.

fchapoton:
but the text above says that this is an even integer..

edgarcosta:
I agree with the division, as `hweight` stands for Hilbert modular form weight, which is a pair of even integers.

Issue #3148, opened 2019-06-07
HGCWA search of group order with genus variable


jenpaulhus:
Once upon a time, on /HigherGenus/Aut/C,  one could input an expression like 84(g-1) in the group order search box and a search would be returned that gave any example where the group order was that value for the genera which were searched over. (So in this case it would give all Hurwitz curves in the database.)

That functionality has disappeared somewhere along the line.

jenpaulhus:
You can assign me to this (I don't immediately see how to assign myself to it).

jenpaulhus:
@roed314  Can we close this?  It looks like I implemented this in #3516 however in a less than ideal way.   But to fix it permanently would first require #3580.

roed314:
Sounds good.  If we get #3580 working, we can revisit various things.

PR #3516, opened 2019-12-12
Update Group Order Search on HGCWA

jenpaulhus:
This fixes issue #3148.

It reinserts code (also slightly modified) that allows searches on group order in the higher genus automorphisms page to include any expression linear in the genus.  

So, for example, if you want to know all the Hurwitz curves (those whose automorphisms attain the maximum allowable for a given genus), you can enter  84(g-1) into the "Group Order" Search at 
http://localhost:37777/HigherGenus/C/Aut/

Other expressions  like "2g"  or "2*g" or "84g-84" will also work. 

Note that we needed to create a new search parser "parse_range2_extend" (and two types of regular expressions) which is in the main.py file of hgcwa pages. If that should get moved to the utils/search_parsing.py file, just let me know.

Finally, we took advantage of this PR to make one statistics table match the new table style.
See:
http://beta.lmfdb.org/HigherGenus/C/Aut/stats/groups_per_genus/4
vs
http://localhost:37777/HigherGenus/C/Aut/stats/groups_per_genus/4


jenpaulhus:
I see this failed check, and will look at what I think is a pyflakes error tomorrow.

roed314:
Here are some issues to look at:

* Please use 4 spaces instead of tab characters (you'll note in the code preview some places where it looks like lines are indented by two spaces; this is a consequence of using tabs)
* In the line `a = GENUS_RE.match(arg).groups()[0]`, if your text input contains a `g` but doesn't match `GENUS_RE` you'll get an error when calling `None.groups()`.  You need to check that it matches your regular expression.
* In general, I worry about the performance implications of your approach, where you create a big `OR` clause splitting things up by genus.  I understand that this is currently the only reasonable way to implement this feature using the functionality of the backend search interface.  The right way to do this is to support expressions in the generated postgres query that refer to other columns.  I can try to make this available, but it won't be immediate.

roed314:
I assume that this is just waiting until after the holidays, but let me know if you're waiting for feedback from me!

jenpaulhus:
Thanks for checking, @roed314. It is waiting until my final exam grading is done!  Maybe tomorrow.

jenpaulhus:
Grading finals took longer than I hoped (as it always seems to do!) but here's the updated PR. 

1. Got rid of the tabs.
2. Added a check for the regular expression.  Also fixed a small issue where the entry "1g" would work but "g" would not. 
3. Made issue #3580 in relation to the third point.  Since I will not be growing the data in the HGCWA in the short term, we can get by with this slightly less efficient method, and once #3580 is implemented, I will switch over to that other method.

AndrewVSutherland:
@roed314 Have you looked at this since Jen's latest commit?  I'd like to get this merged today.

roed314:
Looks good to me.  Sorry for the delay in reviewing!

PR #3518, opened 2019-12-13
py3: no iterkeys, iteritems but keys, items

fchapoton:
that is not an iterator in Python2, hence a bit slower. For python3, one will get back the previous iterator behaviour.

and also changing one "buffer" to "memoryview" (not clear to me if this will work)


fchapoton:
I have undone the change from "buffer" to "memroyview", that was breaking things.

roed314:
Looks good.

PR #3520, opened 2019-12-13
py3: no raw_input

fchapoton:
in python3, raw_input has been renamed input

see https://six.readthedocs.io/#module-six.moves


roed314:
Looks good.

Issue #2950, opened 2019-04-11
Maass form download not working correctly


jwj61:
UPDATE: the original problem has been fixed, but there still is an issue in that you can only download a few of the 5000 coefficients we have stored for most forms.

From the feedback page:

=====================
Only one of the Maass forms of level 1 has any coefficients??

Alas, I needed one of even symmetry...

http://www.lmfdb.org/ModularForm/GL2/Q/Maass/

======================

There was no name or contact e-mail.

Note, the search results page says all but 1 have no coefficients, but if you go to the individual pages, they seem to have a few.

edgarcosta:
1- I think if the collumn Numc is null, then we should say 5000 instead of 0
2- We can't download all the 5000 coefficients, as we could before: http://prod.lmfdb.xyz/ModularForm/GL2/Q/Maass/4cb8503a58bca91458000032

edgarcosta:
This might be a data conversion issue from MongoDB to PostgreSQL:
http://prod.lmfdb.xyz/api/MaassWaveForms/FS/?Level=i1
vs
http://beta.lmfdb.org/api/mwf_forms/?Level=1

edgarcosta:
We fixed the data issue.

Not the download issue.

edgarcosta:
ps: this was only an issue at level 1

AndrewVSutherland:
It looks like the data issue that caused the original problem is now fixed on beta.  @edgarcosta should we copy the modified mwf tables over to proddb (which would fix the original problem) or do you anticipate that fixing the download problem might also involve data changes?

edgarcosta:
Can I just edit the data directly on proddb?
We just did:
```
UPDATE mwf_forms SET "Numc" = 5000 WHERE "Numc" is Null;
```
note, that `Numc` was only `Null` for level 1, and in those cases was supposed to be 5000

AndrewVSutherland:
I think this is safe (and the file is small), go ahead.

edgarcosta:
Done

AndrewVSutherland:
@edgarcosta Thanks, I have removed the critical label since the originally reported problem is fixed.  I'll leave the issue open for the purpose of fixing downloads.   @jwj61 if you have an e-mail address, can you let whoever posted the issue know that it has been fixed.

edgarcosta:
@AndrewVSutherland is this still a 1.1 thing?

AndrewVSutherland:
[...]

Issue #2576, opened 2018-08-18
More coefficients in GL2 Maass form


jvoight:
In the "first" Maass cusp form with level 1 and R = 9.5...,
  http://beta.lmfdb.org/ModularForm/GL2/Q/Maass/4cb8503a58bca91458000032
I can only get 9 coefficients; but many more available in the L-function.  Where are they hiding?  Surely we should display more?

edgarcosta:
Yes, they are at least 100 of them (somewhere), as the L-functions for GL2 Maass forms are generated on the fly, and that is the threshold for us to generate the L-function

AndrewVSutherland:
The issue is that some forms have coefficients in mwf_forms, some have coefficients in mwf_coeffs, and some have both (this data was uploaded from different sources that used different conventions which likely also explains the 0/1 indexing shift noted in #2647).

So we know where the coefficients are hiding, we just need to figure out what combination of code/data changes will get them to be displayed on the screen.

rbommel:
Okay, I tried to do some research in the database to see what happened to the coefficients of http://www.lmfdb.xyz/ModularForm/GL2/Q/Maass/4cb8503a58bca91458000032. In the table where it gets its data from, _mwf_forms_, there are only 9 coefficients listed. So that is not the issue.

However, there seems to be a table _mwf_coeffs_ which contains raw data for coefficients as they were migrated from the old system. The problem is that its entries are listed by some label, which seems to be very hard to find for this particular Maass form.

I just went through the whole database to see where the coefficients are occurring and I found 6 candidate labels for this form:
c-1-0.0-1-1-9.5336952613
c-3-0-1-1-9.5336952613
c-5-0-0-1-9.5336952613
c-5-0-1-1-9.5336952613
c-9-0-0-1-9.5336952613
c-9-0-1-1-9.5336952613

Does anyone know which of these should be the right one? If we know this (in general of course, not just for this particular Maass form), then it should be possible to get more coefficients.

davidfarmer:
I have a guess at what the label notation means.  If that guess
is correct, then it is the first one that corresponds to the
Maass form on Gamma_0(1) that you linked.

And the others do not serve any useful purposes, because the
pages they correspond to no longer exist, because of these
closed issues:

https://github.com/LMFDB/lmfdb/issues/2688

https://github.com/LMFDB/lmfdb/issues/2645

Again assuming I understand the label notation, then the pattern
you will see in the labels is:

c-N1-something-decimal_number
[...]

PR #3521, opened 2019-12-13
Accessing stored Maass forms

rbommel:
This is an attempt to solve issues #2576 and #2950. It is a bit of a hack, but it is only for the period until the data is regenerated. For example, http://localhost:37777/ModularForm/GL2/Q/Maass/4cb8503a58bca91458000032 is changed to include more coefficients, but there are more. All 5000 should show up when you download.

Currently I am assuming the following (which I cannot verify because of lack of knowledge about the data):
- there are no two forms with the same value of R up to the precision with which R is used in the index;
- the first hit in the database is the one with the smallest 'Ni' in the sense David is saying in #2576; I didn't manage to get sorted output, but this seemed to be always the case.

### DIFF
--- a/lmfdb/modular_forms/maass_forms/maass_waveforms/backend/maass_forms_db.py
+++ b/lmfdb/modular_forms/maass_forms/maass_waveforms/backend/maass_forms_db.py
@@ -45,6 +45,19 @@ class MaassDB(object):
     def maassform_has_plot(self, maass_id):
         return db.mwf_plots.exists({'maass_id':maass_id})
 
+    def find_coeffs_from_dict(self, d, dictType):
+        if type(d) == dictType:
+            for i in d.keys():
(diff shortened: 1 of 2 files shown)


AndrewVSutherland:
Looks good, thank you for tackling this!

davidfarmer:
>  *  there are no two forms with the same value of R up to the precision with which R is used in
>     the index;

This can happen when N=9, and probably at no other time.

rbommel:
I just checked the ones I could find by hand. It didn't seem to cause a problem there. It looks like the 'coeff_label' has been set in that case. Then the old code is called.

AndrewVSutherland:
@rbommel Thanks for checking.  I closed the open issues because the specific problems they identified have been fixed by this PR, which is a strict improvement over what we had previously.  If any problems  do arise from this PR we can raise new issues for them.

PR #3530, opened 2019-12-14
py3: fix again the comparison to None

fchapoton:
as somebody introduced very recently the bad syntax "<> None"


AndrewVSutherland:
test_artin failure is unrelated to this PR which looks fine, merging.

PR #3536, opened 2019-12-14
Improvement of search for coefficients of Maass forms

rbommel:
This should make the search a bit better, trying both the rounded down and rounded up version of the eigenvalue.
 
Moreover, I commented away a few lines that returned 'None' in case the attribute 'Numc' is not set. I don't know what this is exactly supposed to do, so maybe someone knowing more about this 'Numc' might want to object.


edgarcosta:
`/lmfdb/modular_forms/maass_forms/maass_waveforms/backend/maass_forms_db.py:73: local variable 'nc' is assigned to but never used`

PR #3538, opened 2019-12-15
Deprecated routes

edgarcosta:
I have removed the following routes:
- https://beta.lmfdb.org/AutomorphicForm -- redirects to ModularForm
- https://beta.lmfdb.org/ModularForm/Maass  -- broken
- https://beta.lmfdb.org/ModularForm/GL2/C/Maas -- broken
- https://beta.lmfdb.org/EC -- redirects to EllipticCurve/Q/
- https://beta.lmfdb.org/NF -- redirects to NumberField/

All these are deprecated in my opinion.


AndrewVSutherland:
pyflakes isn't happy (I think you can remove an import that is no longer needed)

PR #3543, opened 2019-12-15
py3: fix the use of "unicode" in scripts, using six

fchapoton:
with the aim to have pyflakes3 passing on scripts too.


roed314:
Looks good to me.

PR #3544, opened 2019-12-15
first little wave of absolute imports (at top level)

fchapoton:
following the suggestions of

pylint --py3k *.py


JohnCremona:
check-knowl.py is obsolete.  It should be deleted.

fchapoton:
ok, removed

PR #3545, opened 2019-12-15
second wave of absolute import (top level of lmfdb/lmfdb)

fchapoton:
as suggested by pylint --py3k

see also https://six.readthedocs.io/#module-six.moves.urllib.parse


roed314:
Looks like `from .lmfdb.logger import logger_file_handler, critical` is failing.

PR #3547, opened 2019-12-16
python3 fixes for range and zip, some fixes for map

fchapoton:
only inside the lmfdb folder
found using pylint --py3k (very useful tool)


roed314:
Looks good to me.

PR #3548, opened 2019-12-16
Fixing absolute imports

edgarcosta:
I regret some changes in #3544 and #3545, that break things.
This PR reverts them.
I will merge this once the tests pass, as the current setup is broken.

@fchapoton let me know if I should fix this in a different way.


fchapoton:
I have no way to test, and this kind of change shoould rather not be done blindly. Sorry for the trouble.

Issue #3487, opened 2019-12-02
Support a base url


slel:
It would be nice for the LMFDB web server to support a base url.

In particular, that would allow LMFDB development on CoCalc.

I tried setting up LMFDB in a CoCalc project, following
the "getting started with LMFDB development" guide:

- https://github.com/LMFDB/lmfdb/blob/master/GettingStarted.md

and as suggested in the section "Running", ran:
```
sage -python start-lmfdb.py --debug
```
which starts an LMFDB server, and suggests to visit the url

- http://localhost:37777/

to access the version of the lmfdb website served locally.

This however is hard to do when working in a CoCalc project.

I asked CoCalc support:

> Can one visit such a url in CoCalc, when it is served by
> a server running in the project? And how does one do it?

and they pointed me to the "Webserver" page in the CoCalc docs:

- https://doc.cocalc.com/howto/webserver.html

and explained:

> You will have to modify the LMDB code to support
> a base url, like SageNB and Jupyter (among many other
> web servers) all do. That's what is meant by this:
>
> > When you visit that URL with your web browser,
> > your web server will get a request for
> > `/[project_id]/port/[port]/` rather than for `/`
> > like it does with `server`. Otherwise, everything
> > is the same, including the big caveat above.
>
> Many web servers run into situations where they want
> to be served from some URL under a bigger domain and
> they implement base url support. However, I really
> doubt LMDB has done so yet so you'll have to if
> you want to fully run and use it from cocalc.

Note: the quote "When you visit ..." is from the section
"Using the port proxy" of the "Webserver" page
in the Cocalc docs:

- https://doc.cocalc.com/howto/webserver.html#using-the-port-proxy

haraldschilly:
FYI: I think what william calls a base url is actually called script root.

edgarcosta:
We would love to be able to run LMFDB inside cocalc.
I tried it once and ran exactly into the same problem.

Does anyone know how to do this in flask?

edgarcosta:
Maybe something like this:
https://github.com/dtkav/connexion/blob/dec723740ed4af860f60ad4e6dcc63efeddfb427/connexion/apis/flask_api.py

edgarcosta:
The PR #3551 addresses this issue. But would still be good for us to have some documentation and perhaps a template project.

AndrewVSutherland:
Can we close this?

PR #3551, opened 2019-12-16
Running LMFDB inside Cocalc

edgarcosta:
This addresses #3487.

I set it up in a way that automatically detects if you are trying to run it inside cocalc,.

With this PR, when you run it in Cocalc you will get a message like:
```
********************************************************************************
Cocalc environment detected!
Visit
   https://cocalc.com/48cb29a8-fa63-4d7e-bd26-ce1b4c0b9b65/server/37777 
to access this LMFDB instance
********************************************************************************
```
and by following the link you will be able to interact with in same way as if you were running it locally.

ps: I only reformatted `lmfdb/utils/config.py`, as I was planning to add a couple more options, but then decided to do it automatically.

### DIFF
--- a/lmfdb/utils/cocalcwrap.py
+++ b/lmfdb/utils/cocalcwrap.py
@@ -1,0 +1,19 @@
+#-*- coding: utf-8 -*-
+import os
+
+class CocalcWrap(object):
+    """Wrap the application in this middleware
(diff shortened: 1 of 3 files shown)


roed314:
Tested on a CoCalc project and seems to work.

PR #3553, opened 2019-12-16
py3: more fixes for dicts and map

fchapoton:
some further fixe for dict.keys(), dict.values(), map()

### DIFF
--- a/lmfdb/backend/database.py
+++ b/lmfdb/backend/database.py
@@ -797,7 +797,7 @@ class PostgresBase(object):
         col_list, col_type, _ = self._column_types(table_name)
         columns_set.discard("id")
         if not (columns_set <= set(col_list)):
-            raise ValueError("{} is not a subset of {}".format(columns_set, col_type.keys()))
+            raise ValueError("{} is not a subset of {}".format(columns_set, list(col_type)))
         header_cols = self._read_header_lines(F, sep=sep)
         names = [elt[0] for elt in header_cols]
         names_set = set(names)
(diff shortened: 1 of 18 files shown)


jwj61:
The change to web_number_field does not look right (just a typo).  It has fc and cf as dummy variables, and they should be the same.

lmfdb/backend/database.py
```diff
@@ -797,7 +797,7 @@ def _check_header_lines(self, F, table_name, columns_set, sep=u"|", prohibit_mis
         col_list, col_type, _ = self._column_types(table_name)
         columns_set.discard("id")
         if not (columns_set <= set(col_list)):
-            raise ValueError("{} is not a subset of {}".format(columns_set, col_type.keys()))
```
edgarcosta: We might give the wrong message here. (Nothing to do with the PR)

lmfdb/backend/database.py
```diff
@@ -3397,7 +3397,7 @@ def upsert(self, query, data, commit=True):
                 for table, dat in cases:
                     inserter = SQL("INSERT INTO {0} ({1}) VALUES ({2})").format(
                                     Identifier(table),
-                                    SQL(", ").join(map(Identifier, dat.keys())),
+                                    SQL(", ").join(map(Identifier, list(dat.keys))),
```
edgarcosta: `list(dat.keys)`-> `list(dat)`, and remove map

edgarcosta: ~and remove map~

edgarcosta:
Once tests pass I will merge it

PR #3554, opened 2019-12-16
types python2 & 3

edgarcosta:
buffer -> memoryview 
and 
long() -> type(ZZ(2**65).__int__()) (as didn't want to rewrite the code to use `is_instance`)

EDIT: just changed travis to use python3, let's see if I got it right

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -32,7 +32,7 @@ install:
     # assert parallel
     - parallel --version
     # install the latest pyflakes
-    - pip install pyflakes --user
+    - pip3 install pyflakes --user
     # install the latest coveralls
     - pip install coveralls --user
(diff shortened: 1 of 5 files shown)


roed314:
The documentation on [memoryview](https://docs.python.org/2/library/stdtypes.html#memoryview) suggests that it has been backported to Python 2.7.  Why can't we use memoryviews for both Python 2 and 3?

edgarcosta:
That is how psycopg2 handles binary :(
> Received data is returned as buffer (in Python 2) or memoryview (in Python 3).

http://initd.org/psycopg/docs/usage.html#adapt-binary

thus we don't have the choice to only use both, unless, there is something like six.buffer_types, but I couldn't find it.

fchapoton:
looks good to me

PR #3556, opened 2019-12-17
A hack for placement of code blurbs in genus 2 curves 

### LINKED ISSUE #2518: Placement of code blurbs
When you click the Magma code to appear on the genus 2 pages, it awkwardly justifies the data below it:
screen shot 2018-06-27 at 4 03 59 pm

(It seems better behaved on other pages, e.g. elliptic curves over number fields.)

edgarcosta:
This resolves #2518 by adding a phantom column

### DIFF
--- a/lmfdb/genus2_curves/templates/g2c_curve.html
+++ b/lmfdb/genus2_curves/templates/g2c_curve.html
@@ -175,20 +175,27 @@ function show_invs(invstyle) {
 
 <table>
     <tr>
-        <td colspan="6">
+        <td colspan="5">
             {{place_code('aut')}}
         </td>
     </tr>
     <tr>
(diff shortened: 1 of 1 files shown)


AndrewVSutherland:
This seems to fix the problem in the automorphism group section, but not in the invariants section (tested on chrome).

edgarcosta:
On the first commit I was only trying to fix the Aut table issue.
Check out the new commit where I move code blurbs to end of the row, just like in EC and NF.

AndrewVSutherland:
I like this a lot, why not do the same for the aut table? (in general, whenever there is space to do so, I think it is clearer when the code blurb is to the right rather than above or below).

edgarcosta:
Because the code blurb is quite long.
In my screen would overflow.

PR #3561, opened 2019-12-18
py3: some fixes for filter and for the import of StringIO, etc

fchapoton:
- replace many filter by list comprehension
- import StringIO from six
- other fixes

Tool used: `~/lmfdb/lmfdb$ pylint --py3k -d W1618,W1619 */*.py`

### DIFF
--- a/lmfdb/abvar/fq/isog_class.py
+++ b/lmfdb/abvar/fq/isog_class.py
@@ -93,7 +93,7 @@ class AbvarFq_isoclass(object):
             raise ValueError("Label not found in database")
 
     def make_class(self):
-        self.decompositioninfo = decomposition_display(zip(self.simple_distinct, self.simple_multiplicities))
+        self.decompositioninfo = decomposition_display(list(zip(self.simple_distinct, self.simple_multiplicities)))
         self.basechangeinfo = self.basechange_display()
         self.formatted_polynomial = list_to_factored_poly_otherorder(self.polynomial, galois=False, vari="x")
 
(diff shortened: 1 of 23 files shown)


roed314:
The test failures look related to `StringIO`

fchapoton:
I am afraid this is just the tip of unicode/str/bytes hell. Given that you manipulate so many files, this will be very tricky.

fchapoton:
would it be ok to use sage helping method "bytes_to_str" ?

roed314:
I know @AndrewVSutherland has wanted to minimize dependencies when possible, but in this case I would say that using this utility function will make our lives much easier and we should go ahead with it.  @fchapoton, do you know where the source for `bytes_to_str` is viewable online?  I can find it in my local sage installation (in `$SAGE_ROOT/local/lib/python3.7/site-packages/sage/cpython/string.pxd` and `$SAGE_ROOT/local/lib/python3.7/site-packages/sage/cpython/string_impl.h`).

I'd say that this is a function that's been developed to ease the transition from Python 2 to 3 and we should take advantage of it.

fchapoton:
The code can be found here:

https://github.com/sagemath/sage/blob/develop/src/sage/cpython/string.pxd

edgarcosta:
The errors are almost all from using strIO 
```
strIO.write(bytes_to_str(s))
...
TypeError: expected bytes, unicode found
```

fchapoton:
Damn. Here is another tentative, using six

fchapoton:
now everything looks fine.

PR #3564, opened 2019-12-19
yet another bunch of py3 corrections

fchapoton:
various things:

py3 compatible imports, remove filter, etc


roed314:
Will merge once tests pass.

PR #3566, opened 2019-12-20
some further details, as suggested by lgtm

fchapoton:
various little improvements, that fix some lgtm suggestions

### DIFF
--- a/lmfdb/classical_modular_forms/download.py
+++ b/lmfdb/classical_modular_forms/download.py
@@ -383,7 +383,7 @@ class CMF_download(Downloader):
         if newform.dim == 1:
             return begin + [
                     '        Kf := Rationals();',
-                    '    end if;'
+                    '    end if;',
                     '    return [Kf!elt[1] : elt in input];',
                     'end function;',
                     ]
--- a/lmfdb/lfunctions/main.py
+++ b/lmfdb/lfunctions/main.py
@@ -867,19 +867,19 @@ def getLfunctionPlot(request, *args):
     p = line(F_interp)
 
     styleLfunctionPlot(p, 10)
-    fn = tempfile.mktemp(suffix=".png")
-    p.save(filename=fn)
-    with open(fn) as f:
+    with tempfile.NamedTemporaryFile(suffix=".png", delete=False) as fn:
+        p.save(filename=fn.name)
+    with open(fn.name) as f:
         data = f.read()
-    os.remove(fn)
+    os.remove(fn.name)
     return data
+
 
 def styleLfunctionPlot(p, fontsize):
     p.fontsize(fontsize)
-    p.axes_color((0.5,0.5,0.5))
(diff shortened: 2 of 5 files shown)


roed314:
This looks like the previous code contained a bug (I'd guess implicit multiplication was intended in the following rather than power series composition):

* Old version: `H_all = (1 + x ** 4)(1 + x ** 11) / (1 - x ** 2) ** 3 / (1 - x ** 6)`,
* Your version: `H_all = (1 + x**4) * (1 + x**11) / (1 - x**2)**3 / (1 - x**6)`.

I'd like to hear from someone more familiar with the siegel modular forms code though.  @AndrewVSutherland, do you know who the right person to ask would be?

JohnCremona:
Surely @nilsskoruppa , but I am pretty sure that you are right.  It might be a good idea to check the rest of the file for similar glitches.

AndrewVSutherland:
@roed314 This is clearly a bug in the old code (which I don't think is used anywhere).

edgarcosta:
I agree that it also looks like an error.
The code is still used to display dimensions here: https://beta.lmfdb.org/ModularForm/GSp/Q/?family=Gamma0_4&k=&j=&table=1 (that is why the tests are failing)

Do we have a way to verify which one is the correct series?
```
2 + 6*x^2 + 12*x^4 + 22*x^6 + 36*x^8 + 54*x^10 + 4*x^11 + 78*x^12 + 12*x^13 + 108*x^14 + 24*x^15 + 144*x^16 + 44*x^17 + 188*x^18 + 72*x^19 + 240*x^20 + O(x^21)
1 + 3*x^2 + 7*x^4 + 14*x^6 + 24*x^8 + 38*x^10 + x^11 + 57*x^12 + 3*x^13 + 81*x^14 + 7*x^15 + 111*x^16 + 14*x^17 + 148*x^18 + 24*x^19 + 192*x^20 + 38*x^21 + O(x^22)
```

AndrewVSutherland:
http://www.math.unt.edu/~schmidt/dimension_formulas/

AndrewVSutherland:
This formula is meant to implement the first part of Proposition 5.4 in http://www.math.unt.edu/~schmidt/dimension_formulas/papers/2003_half_integral_weight_degree_two.pdf, which makes it clear that multiplication (not composition) was intended, and that the new code is correct.

My guess is the test was simply based on existing behavior (as opposed to a mathematical check) and needs to be changed to match reality rather than the old implementation.

AndrewVSutherland:
@JohnCremona Agreed, I seriously doubt any of these formulas involve composition, but it would be easy to make this typo when copying them from a mathematical paper (the link above gives a convenient list of citations for most (possibly all) of the formulas that are used, so if anything else looks suspicious, that is the place to look).

edgarcosta:
@AndrewVSutherland, I agree with you. I have changed the test.

PR #3570, opened 2019-12-21
Whitelisting web

edgarcosta:
The goal of this PR is to limit the pages that we display on www.lmfdb.org.
If such a page is not whitelisted, then the address will be redirected to [beta.lmfdb.org](beta.lmfdb.org)

I'm at the moment running a modified version of this branch on https://red.lmfdb.xyz, where I replaced the `www.lmfdb.org` on line 226 of `lmfdb/app.py` with `red.lmfdb.xyz`.

Please look at https://red.lmfdb.xyz/forcebetasitemap and double check that I didn't miss any route.

I'm also currently doing a recursive wget on red.lmfdb.xyz, and so far these are redirects that I have hit:
```
Location: https://beta.lmfdb.org/Representation [following]
Location: https://beta.lmfdb.org/zeros/first/ [following]
Location: https://beta.lmfdb.org/L/ModularForm/GSp4/Q/Maass/1/1/12.46875_4.720951/1.34260324/ [following]
Location: https://beta.lmfdb.org/\"" [following]
Location: https://beta.lmfdb.org/L/ModularForm/GL3/Q/Maass/1/1/16.40312_0.171121/-0.4216864/ [following]
Location: https://beta.lmfdb.org/L/ModularForm/GL3/Q/Maass/4/1/8.239796_2.641226/1.066042%20+/ [following]
Location: https://beta.lmfdb.org/L/ModularForm/GL3/Q/Maass/4/1/9.632444_1.374060/0.15012282/ [following]
Location: https://beta.lmfdb.org/L/ModularForm/GL3/Q/Maass/4/1/8.954662_2.936591/0.36025530/ [following]
Location: https://beta.lmfdb.org/L/ModularForm/GL4/Q/Maass/1/1/16.89972_2.272587_-6.03583/0.55659019/ [following]
Location: https://beta.lmfdb.org/L/ModularForm/GSp4/Q/Maass/1/1/19.81936_8.531078/2.06692296/ [following]
Location: https://beta.lmfdb.org/L/browseGraph/?group=GSp4&level=1 [following]
Location: https://beta.lmfdb.org/L/browseGraph/?group=GL4&level=1 [following]
Location: https://beta.lmfdb.org/L/browseGraph/?group=GL3&level=1 [following]
Location: https://beta.lmfdb.org/L/browseGraph/?group=GSp4&level=1 [following]
Location: https://beta.lmfdb.org/L/browseGraph/?group=GL4&level=1 [following]
[...]

PR #3573, opened 2019-12-24
Comment out unimplemented history routes

AndrewVSutherland:
This PR comments out 6 unimplemented top level history routes and fixes a typo in the modular_forms_history function that was causing a server error (this function is now unreachable, but if/when the route is uncommented it won't cause a server error).

This is related to the white listing PR #3570 but independent of it.  It will have no visible effect on the website other than rendering a page not found error when someone manually types in a history URL rather than throwing a server error or rendering a blank page.

I plan to merge this as soon as Travis is happy.


roed314:
Looks good.

PR #3574, opened 2019-12-24
Tiny fix to g2c stats header text

AndrewVSutherland:


roed314:
Shouldn't `Sato-Tate group` be plural, since it would be `Distribution of Sate-Tate group` otherwise?

PR #3583, opened 2019-12-29
Fix more separator problems

roed314:
I hope change should fix the recent problems with reload, though I haven't had a chance yet to test it extensively.


AndrewVSutherland:
This seems to fix the copy_to/reload issue, I'll try it with reload_all in a bit.

AndrewVSutherland:
@roed314 This fixes the problem with reload_all when you don't specify the separator, but if you do a copy_to using the separator ":" and then a reload_all using ":" there is still a problem.

Issue #3565, opened 2019-12-19
Parsing problem for Galois groups


jwj61:
If you search for number fields with Galois group S4 in degrees 6..12

https://beta.lmfdb.org/NumberField/?degree=6..12&galois_group=S4

you get an error message saying that the degree is inconsistent with the group (which is not true).  There are really two problems
<ol>
<li> the code which tries to form queries by combining degrees and t-numbers implicitly assumes that the user specified at most one degree, which is why the search above gives an erroneous error message.  Bear in mind that the user may give multiple degrees and multiple groups.  The current approach with degree and t-number separate in the database would produce complicated queries if programmed right.  The whole thing would be simpler if (n,t) pairs were stored for the Galois group.
<li> we will never be fully accurate in saying when a group is incompatible with a degree.  For example, searching for S5 in degree 120 gets a similar error message even though they are compatible.  We just don't store enough information to always get the message right.  So, that error message should be removed and just return empty search results.
</ol>

AndrewVSutherland:
Agreed, I suggested this in my comment https://github.com/LMFDB/lmfdb/issues/2979#issuecomment-493677157 on issue #2979.  Unless we are certain (e.g. if the degree does not divide the order of the group) we should just run the search, and I see no problem with always just running the search.

jwj61:
Yes.  Unfortunately, it isn't so easy with the search code.  It tries to mesh the Galois group information (with n and t separate) with specifications on n.  This will be easy to fix if we reconnect n and t in the database, but I wanted to see if there was a good reason to not do that first.

AndrewVSutherland:
Resolved by #3587

PR #3587, opened 2019-12-29
Uniformize searching of Galois groups

jwj61:
In the "telescope/microscope" workshop, it was suggested that when searching for transitive groups (in the Galois group pages), one should be able to use the same nicknames as for Galois groups of local or global fields.

We did have searching by GAP id in the transitive group pages, which is in principle the same idea: search by abstract group.  This pull request merges searching by GAP id, group nickname, and transitive label to a single box, and uses it for Galois group searching in local fields, global fields, and transitive groups.  We accept a list of any combination of these, such as S3, [4,1], 4T4:

http://127.0.0.1:37777/GaloisGroup/?gal=S3%2C+%5B4%2C1%5D%2C+4T4

Note, when a nTj is given, we only return that transitive representation, so the results are different for 3T2, [4,1], A4 even though it is the same list of abstract groups:

http://127.0.0.1:37777/GaloisGroup/?gal=3T2%2C+%5B4%2C1%5D%2C+A4

In forming the search query, duplicates are removed (e.g., if you specify both S3 and [6,1]).

If the degree is not part of the search, we add that.  If it were already part of the search query, we would need to intersect the two specifications.  Not doing that right was the source of  Issue #3565.  So, this also fixes that problem:

http://beta.lmfdb.org/NumberField/?degree=6..12&galois_group=S4
http://127.0.0.1:37777/NumberField/?degree=6..12&galois_group=S4

### DIFF
--- a/lmfdb/artin_representations/main.py
+++ b/lmfdb/artin_representations/main.py
@@ -113,7 +113,7 @@ def artin_representation_search(info, query):
     parse_restricted(info,query,"frobenius_schur_indicator",qfield="Indicator",
                      allowed=[1,0,-1],process=int)
     parse_container(info,query, 'container',qfield='Container', name="Smallest permutation representation")
-    parse_galgrp(info,query,"group",name="Group",qfield=("GaloisLabel","Galn"))
+    parse_galgrp(info,query,"group",name="Group",qfield=("GaloisLabel",None))
     parse_ints(info,query,'dimension',qfield='Dim')
     parse_ints(info,query,'conductor',qfield='Conductor')
     parse_bool(info,query,'Is_Even')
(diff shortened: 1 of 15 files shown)


jwj61:
Will fix test failures, then reopen.

jwj61:
In number field searching, the fields degree, r2, and Galois group all affect degree which was causing some bugs.  This is not the optimal implementation, but I think it works in all cases.

AndrewVSutherland:
Reviewing this now.

AndrewVSutherland:
@jwj61 Thank you, this PR makes me very happy!

The only small suggestion I have (but I'm going to merge this PR now), is to handle group names and group ids typed in the label box in a friendlier way, either treat it as a search on group and give the use a list, or have the error message tell the user a list of transitive group labels (I think the former is probably better).  And in the rare cases where their is only one transitive group that matches (e.g. a cyclic group), just go to that transitive group page.

The reality is that many users (including me) are going to make the mistake of typing something like D5 into the Galois group label box rather than the Group box and we may as well try to help them out because we do know what they mean.

jwj61:
That's a good suggestion on the Galois group by label box.

PR #3593, opened 2019-12-31
py3: one import change, and some changes inn exceptions

fchapoton:
as another tiny step towards python3, suggested by pylint --py3k


roed314:
Looks good.

PR #3594, opened 2019-12-31
py3: last care for map, some changes (pep8 and py3) in encoding.py

fchapoton:
after that, the command

`pylint --py3k -d W1618,W1619,W1633,W1653  lmfdb/`

almost passes (remains just one annoying import problem)


roed314:
Looks good.

PR #3595, opened 2019-12-31
First large bunch of absolute imports

fchapoton:
made purely automatically using

`futurize -W --fix=libfuturize.fixes.fix_absolute_import lmfdb`

and then reverting the changes in the subfolders users/ and inventory_app/

### DIFF
--- a/lmfdb/abvar/fq/__init__.py
+++ b/lmfdb/abvar/fq/__init__.py
@@ -1,4 +1,5 @@
 # -*- coding: utf-8 -*-
+from __future__ import absolute_import
 from lmfdb.app import app
 from lmfdb.logger import make_logger
 from flask import Blueprint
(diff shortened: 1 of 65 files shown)


fchapoton:
Apparently, all doctests are ok

JohnCremona:
@fchapoton What is happening with the file in deprecated/?

Also, please add yourself to CONTRIBUTORS.yaml at some point and make a PR for that.  Thanks!

fchapoton:
will do.

About the deprecated file, I have no idea. Maybe this file was deleted elsewhere and reintroduced here ?

I can prepare a cleaner branch with no change in this file.

JohnCremona:
Yes I think this file was moved. It is from an unfinished project from a few years ago (I think).

It's not necessary to remove this change from the PR.   I did a small amount of manual testing of this branch but would be happier if someone else did more (I have no time today).

fchapoton:
ok, here is a branch without 2 problematic files. Now one needs to wait for Travis CI

fchapoton:
the failure does not seem to be related to the changes made here

fchapoton:
so this should be good to go now

fchapoton:
please merge before it bitrots

PR #3596, opened 2020-01-01
remove duplicate class twice

fchapoton:
as suggested by lgtm


fchapoton:
This one should be very easy to review.

PR #3597, opened 2020-01-02
yaml : my name as contributor

fchapoton:


AndrewVSutherland:
This travis failure is due to a change to mf_newforms (stats need to be updated but travis cannot update stats, I will take care of this shortly).  Merging.

PR #3598, opened 2020-01-04
second wave of absolute imports

fchapoton:
Here is another bunch of conversions to absolute imports


edgarcosta:
Here are some of the issues:
```
./lmfdb/zeros/__init__.py:5:19: invalid syntax

from . import zeta.zetazeros
```
```
lmfdb/inventory_app/inventory_viewer.py:4: in <module>

    from . import inventory_helpers as ih

E   ImportError: cannot import name inventory_helpers
    from . import inventory_viewer as iv

E   ImportError: cannot import name inventory_viewer
```

fchapoton:
I have undone the change in lmfdb/zeros/__init__.py for the moment

The other problems points toward a cyclic import.. This needs to be fixed somehow.

fchapoton:
It seems that "inventory_app" is a mess of cyclic imports.

JohnCremona:
> It seems that "inventory_app" is a mess of cyclic imports.

OK, so let's sort them out now then.  The people who wrote that module are longer working for the project but we need the code to work.

fchapoton:
I have no good strategy to sort that out. One would need a graph of imports inside "inventory_app" to find the cycles that needs to be broken.
Sometimes in this kind of problem, it is useful to move some imports into the functions, no longer in the global imports of a file.

edgarcosta:
I will see if I can resolve this.

davidlowryduda:
@edgarcosta I can do this with you.

fchapoton:
I would suggest, inside
"lmfdb/inventory_app/inventory_upload_data.py"
to move the import
"import inventory_live_data as ild" inside the unique method where it is used.

davidlowryduda:
I see that the inventory also imports from `scripts/...` to run things. It seems this is the only place in the LMFDB where code from scripts is used to run the site. I suspect that we should probably not do this. Does anyone disagree?

edgarcosta:
I agree with you, indeed that is a big no-no.

fchapoton:
there is a missing dependency to dbtools

fchapoton:
missing dependency to id_object now

edgarcosta:
I think it should be good now

edgarcosta:
@davidlowryduda or @roed314, could one of you have a look at this? I butchered this PR a bit too much to be the one merging it.

roed314:
Looking now....

AndrewVSutherland:
Please hold off merging this until after I push to dev and web (should happen in the next hour or two)

lmfdb/inventory_app/jsonify_pg_structure.py
```diff
@@ -1,10 +1,19 @@
-import dbtools
-import id_object
+from id_object import get_description
```
roed314: Shouldn't we be switching this to an absolute import?

edgarcosta: Agree

roed314:
Looks good, and @AndrewVSutherland has pushed to dev and web.  Merging.

PR #3599, opened 2020-01-04
some further python3 fixes, to be able to launch

fchapoton:
a short list of minor changes

including some raw-strings prefixes r (not really required right now)

with all this, it seems that "sage -python start-lmfdb.py --debug" is successful with sage 9.0


fchapoton:
I forgot to say, that was working on top of #3598

roed314:
Looks good.

fchapoton:
Thanks. One now needs to fix the hardcoded python 2.7 in the setup.py of Conrey package for Dirichlet characters.

roed314:
Do you know what it should be changed to?  If you submitted a PR to https://github.com/jwbober/conrey-dirichlet-characters we could probably get @jwbober to merge it.

fchapoton:
No, I have no idea how to make that compatible with py2 and py3. This setup.py looks very unorthodox to me, but I am no expert.

davidlowryduda:
I have now made a pull request to make setup.py compatible with both py2 and py3.

PR #3600, opened 2020-01-04
force encode

edgarcosta:
This should fix various broken download links.
For example, any download link on https://www.lmfdb.org/ModularForm/GL2/Q/holomorphic/128/2/e/b/

According to, https://github.com/pallets/werkzeug/issues/960, we should encode utf-8 strings before passing them to strIO


roed314:
I just checked that the broken download link works on this branch (I pushed this to olive.lmfdb.xyz).  Feel free to merge once tests pass.

Issue #3342, opened 2019-09-07
Weil Polynomials Knowl, L-Polynomials Display


tdupu:
The L-polynomials and Weil polynomials are getting mixed up:
    http://www.lmfdb.org/Variety/Abelian/Fq/2/4/ae_j
The objects that are displayed are for det(1-tF) and the knowls are for det(t-F). Also, it *is* the case that the Weil polynomials are the det(t-F) ones since Weil q-numbers need to be roots of Weil polynomials. 

We should either change the knowls or change the way the data is displayed. Also, I think the information in the database is agnostic about which is displayed since it only outputs an array with the first half of the coefficients which can be interpreted as weil polynomial coefficients or L-polynomial coefficients.

edgarcosta:
I prefer to keep displaying det(1-t Frob).

AndrewVSutherland:
I agree that we should continue to display det(1 - t Frob), since this is what we display elsewhere in the LMFDB, but it would perhaps make more sense to change the caption to L:polynomial and give it its own knowl with a link to the Weil polynomial knowl (and explain that the two are reciprocal).

Alternatively, we could let the user toggle between the two, as we do with Igusa-Clebsch/Igusa/G2 invariants of genus 2 curves, and make L-polynomial the default.

kedlaya:
+1 for having an option to toggle.

edgarcosta:
The (unreviewed) knowls on Beta already recursively mention each other.
What about in the search results? 

Somewhat related to this issue, we should display the polynomials expanded, at least in the abvar page.

kedlaya:
+1 for expanding the polynomials in addition to showing the factorization.

And since I didn't this earlier, if we do have a toggle for L-polynomials vs Weil polynomials, I agree with keeping the L-polynomial as the default.

AndrewVSutherland:
I would keep the search results as they are (L-polynomials) but change the header from Weil Polynomial to L-Polynomial.

tdupu:
Another, minor thing. On the page they are x's but in the knowl they are t's. In the manuscript we landed on T's. 

Best,
Taylor

edgarcosta:
I also prefer t's, but don't have strong opinions.

We have detoured a bit of the initial issue. Here is my attempt to summarize things left to do to close this item:

[...]

PR #3603, opened 2020-01-05
F_p -> F_p(T)

edgarcosta:
As discussed in #3342 we changed
F_p -> F_p(T)
in the table when displaying local factors of L-functions.

For consistency, I have also changed `Gal(F_p)` to `Gal(F_p(T))`
and 
`F_p is a polynomial of degree...`
to 
`F_p(T) is a polynomial of degree...`

However, I would prefer to stick with `Gal(F_p)` and `F_p is a polynomial...`

See:
http://localhost:37777/L/ModularForm/GL2/Q/holomorphic/2496/2/d/d/
http://localhost:37777/Motive/Hypergeometric/Q/A14.2_B9.1

### DIFF
--- a/lmfdb/hypergm/templates/hgm_family.html
+++ b/lmfdb/hypergm/templates/hgm_family.html
@@ -107,11 +107,11 @@
       <th class='weight'> ${{col_var}}$ </th>
       {% if family.degree < 12 and family.degree > 2 %}
         {% set colspan = 3 %}
-        <th class='weight'> $\Gal(F_p)$ </th>
+        <th class='weight'> $\Gal(F_p(T))$ </th>
       {% else %}
         {% set colspan = 2 %}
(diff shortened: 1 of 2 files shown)


AndrewVSutherland:
@edgarcosta I  pulled the trigger to quickly on this one (I just merged it once I was sure it worked, it didn't occur to me that you would submit a change you didn't actually want to make!)

I think on the L-function page, given that we just wrote F_p(T) it make more sense to write "F_p(T) is a polynomial..." (I agree that F_p on its own would be OK but think F_p(T) is better).

But on the hypergeometric motive page, I agree with your preference, I see no benefit to writing Gal(F_p(T)) (the splitting field does not care what you call the variable) and it makes the display busier than it needs to be.  I just reverted this change (but left in the L-function page change).

PR #3605, opened 2020-01-05
Gal(F_p(T)) -> Gal(F_p)

edgarcosta:
Undoing a part of #3603 that @AndrewVSutherland  missed


AndrewVSutherland:
@edgarcosta thanks.

PR #3611, opened 2020-01-06
Allowing more inputs for Galois group labels

jwj61:
This addresses the suggestion in pull request #3587 .

On the index page for Galois groups, one can enter a label to go directly to the transitive group's home page (like 6T3).  With this, if you enter something else like [4,1] or C6 where there is a unique answer, you still go directly to the transitive group home page.  If you enter something like D4, you get the search results page showing the different all results for that group.  Also, the input is transferred to the Group box as if the user put their input there.

There are two "rough edges" to this.  Maybe they are as good as possible, or maybe someone has a suggestion for improving them.

The first is if you enter S33 as a transitive group label.  This now takes you to the home page for S33 as a transitive subgroup of S33, but in principle, there are other transitive representations of this group.  Since they aren't in the database, we skip the search results page which would have only one result.

The second is if you enter something like banana.  As expected, you get an input error, but the message is the same as if you entered banana in the group box lower on the page, so it talks about much more general inputs which are allowed.  It also says that it is not a valid input for "Galois group" instead for a "Galois group label".  The latter would be easy to change, but it seems odd to say that the user hasn't input a Galois group label, and then go on to say that things other than labels are allowed.


AndrewVSutherland:
Thanks for this!  Regarding the rough edges, now that I see this in action I can see an argument for always displaying a list, as you do when the Group box, just so the user get the chance to see the "results are complete for degrees <= 23" message and perhaps realize that they might be missing a higher degree transitive group that also works.  But given that we don't actually promise to do anything with group names on the "search by group label" input box, I don't think this is necessary.

Regarding the error message, I think using the group error message is fine, but I do have an alternative/related suggestion (which applies to the group search as well, so it isn't really specific to this PR)..

I think we should either soften the error message or display an empty list of results if the user types something that looks reasonable into the box (e.g. a single letter followed by digits).  It's a little weird to be running through dihedral (or cyclic, alternating, symmetric, ...) groups and get matches for D46, D47, and then suddenly be told that D48 is not a valid input (it's perfectly valid, it's just not in the database right now, it may well be in the near future).

Or maybe we should just always display search results no matter what the user types (who knows, maybe one day the "banana" group will be a thing).  This has the virtue of being trivial to implement, and given how varied group names/nicknames can be, it seems hopeless to try to validate the input.

jwj61:
The last changes polish the two rough edges.

If you enter a label like S123 or C321, you are told that the group is not in the database.  This holds for An, Sn, Cn, Dn.

Also, it jumps to the homepage only when one of two conditions hold:
  1) the user entered a group label like 23T2
  2) the user entered a gap id or group nickname where the group is in the database and the group has a unique faithful transitive permutation representation.

So, C4 or Q8 will jump you to the group's hold page, but S33 will give a search result with only one entry.

AndrewVSutherland:
Looks good, and I like your approach 2 to handle cases that happen to give unique matches now but might not in the future. 

 question: I notice that C3xC3 does not work in the label box but does work in the group box, is that intentional? (if so fine, I just wanted to check).

jwj61:
That's a bug, .. working on it.

PR #3612, opened 2020-01-06
Genus 2 page update

AndrewVSutherland:
This PR is a major genus 2 update that makes the new BSD and Mordell-Weil data visible on the genus 2 curve pages and incorporates a number of UI improvements.  I am grateful to Raymond van Bommel, Edgar Costa, and Michael Stoll for all the work they put in to these computations.

This branch is up and running on

https://red.lmfdb.xyz/Genus2Curve/Q/

and incorporates almost all of the feedback I received on Saturday at the collaboration meeting (thanks to everyone who was there for the helpful feedback!).  The only thing I have not done is add links on the isogeny class pages to AV/Fq pages -- this should definitely be done but is independent of this PR which really doesn't change the isogeny class pages at all other than adding the Mordell-Weil rank (which is now rigorously known in all but 89 cases).  I have also added statistics and searching on analytic sha (now computed for all but 2 curves), which was not in Saturday's demo, see

https://red.lmfdb.xyz/Genus2Curve/Q/stats

There is still work to be done on the knowls (so don't panic if you see a TBA pop up), but the links are all in place so this work will not require any code changes.

This PR depends on data changes that have already been made on beta and copied to the cloud (the old code still works fine on the new data).


AndrewVSutherland:
I need to make a small change to correctly display the Mordell-Weil group for the 2 curves of analytic rank 1 where we don't know an infinite order generator.  I'm going to close and resubmit after lunch.

AndrewVSutherland:
OK, I think this is now ready to go.  In situations where we are missing a generator, such as

https://red.lmfdb.xyz/Genus2Curve/Q/253847/a/253847/1

It will display only a conditional group structure for the Mordell-Weil group and omit the table of generators completely (and only rigorous MW groups are listed in the property box, so it is omitted in the example above).  The code now also handles missing BSD data in a cleaner way.

I'd be grateful if someone could review this soonish, because I would like to have this (and an upcoming PR related to the other demo I did on Saturday) in production before the upcoming meeting in New York.  Some pages to look at:

https://red.lmfdb.xyz/Genus2Curve/Q/
https://red.lmfdb.xyz/Genus2Curve/Q/stats
https://red.lmfdb.xyz/Genus2Curve/Q/random
https://red.lmfdb.xyz/Genus2Curve/Q/900617/a/900617/1
https://red.lmfdb.xyz/Genus2Curve/Q/519829/a/519829/2
https://red.lmfdb.xyz/Genus2Curve/Q/8190/a/982800/1
https://red.lmfdb.xyz/Genus2Curve/Q/864569/a/864569/1
https://red.lmfdb.xyz/Genus2Curve/Q/989041/a/989041/1

All the changes in this PR are all local to genus 2 curves (the one change to template.html simply duplicates the change in #3607 that has already been merged and pushed to prod).

JohnCremona:
I'll try to look at this on Tuesday.

AndrewVSutherland:
@JohnCremona Thanks!  I have now written at least a rough draft of all the knowls linked to by the new pages, which should help your review (feel free to edit them as you see fit).

BTW (as you probably know) the fact that the top line of the ntclass tables disappears after you open and close a knowl in a header is a known issue that is independent of this PR (if and when we figure out the write CSS incantation to fix it, it will be fixed everywhere).

JohnCremona:
1. On https://red.lmfdb.xyz/Genus2Curve/Q/ three of the search box labels have asterisks which I suppose is related to these quantities possibly being known only conditionally, but I think these stars need an explanation on that page.
   2. random took me to https://red.lmfdb.xyz/Genus2Curve/Q/171452/a/342904/1  where I needed the "generator" knowl to explain "zx=0, y=-x^3".   To be consistent with the description in the knowl I think that u=zx and v=-x^3, but is the knowl confusing (u,v) and (a,b)?  I'm not sure what to suggest but what is displayed would seem somewhat mysterious to a non-expert.

Apart from that all looks very good.

Are there cases where the MW group structure is known and you have the right number of independent generators modulo torsion but do not know whether they are saturated?

AndrewVSutherland:
@JohnCremona Thanks for the quick review.  I have addressed both your comments as follows:

1. I added a footnote on the front search page explaining that asterisked items are only known conditionally, and that searches will match these conditional values (I also added an asterisk to the caption for rational points).  Each of the knowls explains the nature of the condition (rational points may only be a lower bound, analyic rank is typically only an upper bound, squareness of sha assumes finiteness of Sha).  One could regard analytic sha as non-conditional in principle, but given that our computations of it depend on a conditional analytic rank, it deserves a star.

2. I corrected the typo in the generator knowl (yes I wrote u,v in one place where I meant to write a,b, good catch!)

Regarding your question about saturation, while I expected there to be cases where this would happen, but in fact we (by whichI  mean Michael Stoll) were always able to saturate, so the only uncertainty about the MW group arises in cases where either we cannot find an independent point of infinite order that we know should be there (2 cases), or we cannot prove a tight upper bound on the Mordell-Weil rank but the lower bound we have matches our upper bound on the analytic rank (so correct under BSD), which occurs in 87 cases.

PR #3613, opened 2020-01-07
Ntdata border

alexjbest:
In #3612 @AndrewVSutherland complained about this weird table header thing; there are maybe two ways of fixing it: an easier one I just thought of and a harder one which is this branch:

- Instead of setting `table.ntdata > thead > tr:last-child {   border-bottom: 2px solid #0D47A1;}` in style.css (line 678) just set `table.ntdata > thead {   border-bottom: 2px solid #0D47A1; }`?

- The idea for the one in the PR is that slideToggle gives the tr containing the knowl the "display: hidden;" css which is great, except that the bottom border on the last row we still want to see, so we toggle again to make it displayed and then set visibility to collapse so that we can still see the border this is a bit jerky as everything needs to be done in the right order so its only worth doing if the first one doesn't work for some reason?


AndrewVSutherland:
Your first suggestion works!  I added this to #3612 and clicked through every page I could find with an ntdata table and the all work as desired.

Thank you!!

PR #3615, opened 2020-01-07
Minor genus 2 tweaks

AndrewVSutherland:
This PR is a follow up to #3612 with 3 small fixes/improvements:

1. The group names in the drop down for the geometric automorphism group have been changed to match the names used in the group knowls that are now used (except for V_4 which is as it was).
2. You can specify "None" in the search box for analytic sha to find the two curves where this has not been computed -- this change was made so that the corresponding link on the stats page will work.
3. In the 8 cases where the regulator is known but generators for the Mordell-Weil group are not yet stored in the database (this should happen soon) the regulator is now correctly displayed.

These changes can be seen at

https://red.lmfdb.xyz/Genus2Curve/Q/ (click the Q-bar Automorphism drop down)
https://beta.lmfdb.org/Genus2Curve/Q/ (old group names)

https://red.lmfdb.xyz/Genus2Curve/Q/stats (click on the 2 in the "None" box for analytic sha)
https://red.lmfdb.xyz/Genus2Curve/Q/?analytic_sha=None (result of clicking on the 2 now)
https://beta.lmfdb.org/Genus2Curve/Q/?analytic_sha=None (what used to happen)

https://red.lmfdb.xyz/Genus2Curve/Q/519829/a/519829/2 (page where regulator now appears)
https://beta.lmfdb.org/Genus2Curve/Q/519829/a/519829/2 (old page with regulator unknown)


JohnCremona:
I'm onto it...  these issues will require very similar solutions for elliptic curves when we have incomplete data, such as for curves over number fields (most existing ones in the database) and over Q once we start to extend beyond the range of my collection.

JohnCremona:
Looks good.  One small thing: on the stats page Sha section,  would it be better to head the column"Unknown" than "None"?  And why the column headed "1-"?  Perhaps that should not be there.

It makes life very easy for the reviewer to have these dedicated URLs such as red.lmfdb.xyz.  Is that only possible for people at MIT?

AndrewVSutherland:
I agree that Unknown would be better than None, but this is not controlled by the genus 2 code (which is just using the generic stats implementation in db backend) so this change should really be in a separate PR.  The 1- column is a count of the number of records with analytic_sha >= 1 (all but the 2 Nones) which really should not appear -- @roed314 is aware of the issue and looking into it (but again the fix will not be in the genus 2 code).

Regarding the dedicated URLs, we can easily set this up for anyone who wants one -- the simplest approach in the short term is to push to one of roed314's "colored branches" (but check with David first to make sure you aren't overwriting someone else).  But if you want a dedicated server for a branch you are going to be developing/reviewing for a significant period of time, let me know and we can set up a server that automatically fetches from your branch (it will update itself every few minutes once this assignment is made).

AndrewVSutherland:
I take back what I wrote above, I can switch None to Unknown using a formatter, give me a minute to get this working.

edgarcosta:
> Regarding the dedicated URLs, we can easily set this up for anyone who wants one -- the simplest approach in the short term is to push to one of roed314's "colored branches" (but check with David first to make sure you aren't overwriting someone else). But if you want a dedicated server for a branch you are going to be developing/reviewing for a significant period of time, let me know and we can set up a server that automatically fetches from your branch (it will update itself every few minutes once this assignment is made).

What about we have an url for each main editor gets a branch, let's say `test` or `live`, and url username.test.lmfdb.xyz? 
This way they can always force push to their branch to review stuff.
Adding more URLs that follow other branches is not complicated but is a bit tedious and bothersome, as requires editing various scripts and config files, thus I would prefer to do all of them in one go.

AndrewVSutherland:
Actually this doesn't quite work (and it would be better to change the handling of None generically rather than hacking it in the genus 2 code).  Let's leave this PR as is and merge as soon as Travis finishes.

AndrewVSutherland:
@edgarcosta Hmm, I think for developers who are generating lots of PRs it might make sense to have them do this rather than waiting for a reviewer to do it -- the whole point is to make PR's as easy to review as possible.  This way the PR is immediately available for everyone to view (without needing to do anything), and the submitter of the PR can provide links that people can just click on (which I think is one of the main features of this workflow and better than having links that begin 127.0.0.1 that are only going to work if you have pulled the branch and are currently running it).

AndrewVSutherland:
@edgarcosta So my suggestion would be to do exactly what you said, but have the "test" branches be on the submitters repo rather than the reviewers.

JohnCremona:
Sorry I cluttered up this PR with an independent topic.  And I saw the original change from C_{10} to C_10 (which is now undone) and thought it was fine for the drop-down list.  We do need some consistency though:  at https://beta.lmfdb.org/EllipticCurve/Q/ (torsion group drop-down) I did not use underscores; similarly https://beta.lmfdb.org/EllipticCurve/.

AndrewVSutherland:
@JohnCremona I agree that C10 would be better, I changed it back because it messed up the title on the stats page. I will figure out how to decouple these (and deal with the None vs Unkonwn) in a later PR.

JohnCremona:
@AndrewVSutherland That makes sense, if he strings are being used as latex also.  So you need some underscores in C3:D4 also (see https://red.lmfdb.xyz/Genus2Curve/Q/stats).

JohnCremona:
> @edgarcosta So my suggestion would be to do exactly what you said, but have the "test" branches be on the submitters repo rather than the reviewers.

Do we  have a URL always showing the website with the current master branch?  That would make sense, and was something I always wanted to set up when beta and prod were served from Warwick, as it would show the current state better than beta, and would show exactly what the effect of merging to beta would be before it happened.

PR #3617, opened 2020-01-07
Fix genus 2 curve stats

roed314:
This PR should fix the following issues with statistics pages:
* Ranges won't show up as column headers
* None will print as "Unknown."


AndrewVSutherland:
Thanks!

PR #3633, opened 2020-01-10
fix reload_all with nondefault sep

### LINKED ISSUE #3586: reload_all does not work with non-default separator
PR #3583 fixed the problem when using a default separator, but with a specified separator reload_all no longer works correctly (it did work prior to #3572).

edgarcosta:
This resolves #3586


roed314:
This looks good; thanks!

PR #3634, opened 2020-01-10
Better handling of direct products

jwj61:
Currently, in the list of nicknames for direct products, we have to list things like C3 x S3 separately from S3 x C3.  This modifies the code to look up a group nickname so that we only need one entry in the alias list.  Basically, it sorts the factors so it just has to compare to one nickname.  This is safe unless we get a group nickname containing an X which does not mean direct product.

To test, enter group nicknames like C3xS3 and S3xC3 in the group label box (or the Group box) on the transitive group page.  Both should take you to the same search page.  (I tested every direct product currently in the nickname list.)


roed314:
Looks good.

PR #3636, opened 2020-01-10
simplify comparisons to empty things

fchapoton:
This is a trivial cleanup merge request.

Direct test for emptyness is faster than computing the length and comparing to 0

Also another kind of fix in belyi/main


roed314:
Looks good.

PR #3637, opened 2020-01-10
tentative usage of "autopep8" on one single file

fchapoton:
This was done purely by the command

`autopep8 lfunctions/HodgeTransformations.py --in-place`

that fixes the indentation and some other pep8 conventions

### DIFF
--- a/lmfdb/lfunctions/HodgeTransformations.py
+++ b/lmfdb/lfunctions/HodgeTransformations.py
@@ -8,158 +8,172 @@
 
 from sage.rings.integer_ring import ZZ
 
-def hodge_structure(wt,gamma): # gamma is motivic, max is 1
- """
- Input wt an integer
-       gamma a list of integers.
 
- Output : list of lists of integers
- """
- t=1
- H=[]
- while 2*t > 2-wt:
-  m=gamma.count(t)
-  n=gamma.count(t-1)
-  if n < m:
-   raise ValueError('Gamma factors are not Hodge')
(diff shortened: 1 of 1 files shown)


roed314:
Looks good.

JohnCremona:
We do have the script codestyle.sh which Harald supplied in 2012 but is
hardly every used, probably because it introduces a lot of whitespace
changes which is liable to break work in progress.  Per haps we need to
make an executive decision as to weather we will at some point insist on
this standard, but I would not give that a high priority.

edgarcosta:
I think fixing the code style and then automatically enforcing (something a bit less strict than pep8) it would help us in the long run, as must code in the LMFDB grows through copy and paste.

edgarcosta:
On the same topic, whenever I  rewrite some file I use Black https://pypi.org/project/black/ to reformat it.
It isn't perfect, e.g., `2**3` becomes `2 ** 3`, but I find it very useful.

edgarcosta:
We can also partially integrate pep8 in the PRs with https://github.com/marketplace/pep-8-speaks

jwj61:
I would be concerned that enforcement of style would turn off potential contributors.  People join because of their expertise in math, so if you tell them that they need to fix their pull request because they don't have spaces after commas, they might see the project as being overly fussy about minutia.

edgarcosta:
I fully agree with you! Maybe we could just try to run some tool once in a
while.

PR #3641, opened 2020-01-11
.get_data(as_text=True)

edgarcosta:
This gets us ready for Sage 9.0, as `tc.get(foo).data` returns a byte string, thus all of our tests fail
This PR was generated with:
```
find lmfdb -name "test_*.py" | xargs sed -ie '/assert/s/\.data/.get_data(as_text=True)/'
```


roed314:
Looks good.

PR #3642, opened 2020-01-12
Correct sage code for torsion subgroup

alexjbest:
for ecnf, see for example https://www.lmfdb.org/EllipticCurve/6.6.300125.1/1.1/a/2 and click sage code then  `sage: E.torsion_subgroup().gens()`  appears under structure and generators

### DIFF
--- a/lmfdb/ecnf/code.yaml
+++ b/lmfdb/ecnf/code.yaml
@@ -73,7 +73,7 @@ gens:
   magma: Generators(E); // includes torsion
 
 tors:
-  sage:  E.torsion_subgroup().gens()
+  sage:  E.torsion_subgroup()
   pari:  elltors(E)[2]
   magma: TorsionSubgroup(E);
 


JohnCremona:
I do not like this change.  The code was not incorrect and now is worse. We no longer show how to see the group structure as is displayed on the page. We show torsion generators twice instead of showing generators and the list of torsion points.  I will follow this up with something better.

AndrewVSutherland:
@JohnCremona I apologize, and I agree with you, I did not look carefully at the snippet in the context of the page and should not have merged this PR.  It is the Magma code snippet that should change, not the Sage code snippet.

I think it would be helpful for both snippets to create both the torsion subgroup and the generators, e.g.

sage:  T=E.torsion_subgroup(); T.gens()
magma: T=TorsionSubgroup(E); Generators(T);

Does that sound reasonable?

JohnCremona:
Yes that's reasonable, but since the page shows first the structure then the gens we can have two code lines each producing output, for both sage and magma.

alexjbest:
> I do not like this change. The code was not incorrect and now is worse. We no longer show how to see the group structure as is displayed on the page. We show torsion generators twice instead of showing generators and the list of torsion points. I will follow this up with something better.

I guess its moot now but I'm confused by your comment: this change removed the doubling of `.gens()` being displayed to make it display the structure when run in a Sage session 
```python
sage: EllipticCurve("11a1").torsion_subgroup()
Torsion Subgroup isomorphic to Z/5 associated to the Elliptic Curve defined by y^2 + y = x^3 - x^2 - 10*x - 20 over Rational Field
```

JohnCremona:
I'm sure you are right, sorry to have overdone the criticism.  Anyway, the new PR leaves everyone happy I think.

PR #3643, opened 2020-01-12
Tweak torsion subgroup snippets

AndrewVSutherland:
This PR is meant to address the concerns raised after I prematurely committed #3642.  The changes all relate to the code snippets on torsion subgroups.  You can see the changes on https://red.lmfdb.xyz. 

Some pages to compare:

https://beta.lmfdb.org/EllipticCurve/2.2.5.1/76.1/b/2 (old)
https://red.lmfdb.xyz/EllipticCurve/2.2.5.1/76.1/b/2 (new)

https://beta.lmfdb.org/EllipticCurve/3.1.23.1/19379.1/A/1 (old)
https://red.lmfdb.xyz/EllipticCurve/3.1.23.1/19379.1/A/1 (new)

https://beta.lmfdb.org/EllipticCurve/2.2.40.1/142.1/b/1 (old)
https://red.lmfdb.xyz/EllipticCurve/2.2.40.1/142.1/b/1 (new)

https://beta.lmfdb.org/EllipticCurve/4.4.1125.1/45.1/b/8 (old)
https://red.lmfdb.xyz/EllipticCurve/4.4.1125.1/45.1/b/8 (new)

Try clicking on code snippets for Magma / PariGP / SageMath, and also the download all code options.  You will see that the torsion structure code snippets now actually display the structure (i.e. what you see on the page), rather than just computing the torsion subgroup and/or a list of generators for it and expecting the user to figure out what to do from there.

In addition to changing the torsion subgroup code snippets, I modified the (infinite order) generators snippet for magma to only take the generators of infinite order (to be consistent with Sage and what is displayed on the page) and modified the regulator snippet accordingly.

I also added a new code snippet for heights, which Sage and Magma are both happy to compute.

I changed the ecnf-curve.html template to only show the code snippet for the structure of the torsion subgroup, rather than structure and order, since the page doesn't display the order, but you still get the snippet for the torsion order in the download code options.  I also changed the display to only put code snippets for quantities that are actually displayed on the page (again the download code options will return everything), because otherwise it gets confusing.

Aside: on the second example page above the rank is marked as not available, but if you run the code snippets you will see that Magma and Sage are both able to compute it (and say it is zero).

### DIFF
--- a/lmfdb/ecnf/code.yaml
+++ b/lmfdb/ecnf/code.yaml
@@ -1,37 +1,33 @@
 prompt:
-  sage:   'sage'
-  pari:   'gp (2.8)'
-  magma:  'magma'
+  sage:  'sage'
+  pari:  'gp (2.8)'
+  magma: 'magma'
 
 logo:
-  sage: <img src ="http://www.sagemath.org/pix/sage_logo_new.png" width = "50px">
-  pari: <img src = "http://pari.math.u-bordeaux.fr/logo/Logo%20Couleurs/Logo_PARI-GP_Couleurs_L150px.png" width="50px">
+  sage:  <img src ="http://www.sagemath.org/pix/sage_logo_new.png" width = "50px">
+  pari:  <img src = "http://pari.math.u-bordeaux.fr/logo/Logo%20Couleurs/Logo_PARI-GP_Couleurs_L150px.png" width="50px">
   magma: <img src = "http://i.stack.imgur.com/0468s.png" width="50px">
 
 not-implemented:
-  sage: |
+  sage:  |
     # (not yet implemented)
-  pari: |
+  pari:  |
     \\\\ (not yet implemented)
   magma: |
     // (not yet implemented)
 
 field:
-  sage: x = polygen(QQ);  K.<a> = NumberField(%s)
-  pari: K = nfinit(%s);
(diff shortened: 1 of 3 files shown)


JohnCremona:
Thanks -- reviewing now.

JohnCremona:
Looks great.  I tried several of the complete code downloads and all was well.

As we are doing this (and even already changing the relevant line) I suggest that we can replace the prompt "gp (2.8):" with plain "gp: ".  When Aurel originally wrote the gp snippets, support for elliptic curves over number fields was not yet in the stable release, only in 2.8 which was then a pre-release.  Since the current stable is now 2.11.1, we hardly need include the version number.

I note your final aside!   In the vast majority of cases where the rank and generators have not yet been computed and added to the database they are easy to compute, it just needs someone to do this systematically (and deal with the hard cases and possibly bugs which arise).  Perhaps the thing to do is design a first run through which gives up quickly so we get all the easy cases in sooner.

AndrewVSutherland:
Thanks for the quick review.  I removed the gp version number, feel free to merge as/when you like.

Regarding the rank computations, yes, running through all the curves with a gradually increasing timeout is the way to go (parallel makes this easy to do).  This is something we can easily run on one of the MIT machines.

Issue #3646, opened 2020-01-14
HMF Magma code errors


JohnCremona:
The web page for a Hilbert Modular Form includes in its Downloads section "modular form to Magma", giving a file of Magma code which should be runnable. At the end of the file is some commented-out code for actually computing the newform in Magma, and this code has errors.

Example:

https://beta.lmfdb.org/ModularForm/GL2/TotallyReal/6.6.1397493.1/holomorphic/6.6.1397493.1-71.2-c

   1. The line ```newforms := [Eigenform(U) : U in newspaces];``` gives the error "Runtime error in sequence construction: No valid universe containing all  elements".
   2. The loop should increment ppind.

I will post a correction.

JohnCremona:
Fixed by #3647

PR #3647, opened 2020-01-14
fix for HMF magma download code (issue #3646)

JohnCremona:
To test this go to a HMF home page such as 
http://localhost:37777/ModularForm/GL2/TotallyReal/6.6.1397493.1/holomorphic/6.6.1397493.1-71.2-c
or a random one, e.g.
http://localhost:37777/ModularForm/GL2/TotallyReal/2.2.28.1/holomorphic/2.2.28.1-531.1-m
then click on the "Modular form to Magma" link on the right, save the file to the default name magma.txt, then at your command line type magma magma.txt and then make_newform();

But don't test this yet since the above random one just failed for me.


JohnCremona:
The second commit fixes one problem but I still fail to find the match for http://localhost:37777/ModularForm/GL2/TotallyReal/3.3.1384.1/holomorphic/3.3.1384.1-64.3-d
because the Hecke fields are equal but we also need to find an isomorphism between them which maps one eigenvalue sequence to the other.  To be continued...

JohnCremona:
OK, ready for testing.  As well as the minor bugs I set out to fix, the code now correctly (I hope) first selects newforms with the correct Hecke field (up to isomorphism), then computes all ismorphisms between the downloaded Hecke field and the newforms' fields, and tries them all.  Allowing for Q being a special case.

edgarcosta:
@JohnCremona, are there a couple of examples that one could add to the tests suite using the magma free calculator?
See:
https://github.com/LMFDB/lmfdb/blob/429aafa5fb7d918c8490bc15275aa0a9eb3de35a/lmfdb/classical_modular_forms/test_cmf.py#L610

JohnCremona:
OK I'll try that out. I didn't know that it was possible.

JohnCremona:
I have spent 2 hours trying to get this (testing using the online Magme calcualtor)  to work and am ready to give up.  I have everything working perfectly *except* that the output from magma_free is always empty.  I think that is possibly because there is a rather short timeout, and these computations are not so trivial.  

My evidence is that when I save the string magma_input to a file and run it with a local magma ('magma test.m' at the bash command line) it works fine, but if I run it within sage using magma_free(magma_code) the putput after 1s is empty.

I can update the PR with the revised test_download_magma() function if you want but there seems little point.

JohnCremona:
On second thoughts, I could rewrite the test to use a local magma in a try/except block so that we can run the test on a local machine with magma installed but Travis will still work.  Watch this space...

JohnCremona:
I'll resolve the conflicts.  The new test runs fine locally on legendre, and I hope will not fail on Travis.

JohnCremona:
Success!  @edgarcosta , I hope you are now happy.

edgarcosta:
Happy indeed. I didn't mean for you to spend hours on this, but without a question that you have certainly improved the situation!!
We rarely test our download scripts, and without having some tests in place, it usually takes too long to figure out that we have broken something.

Issue #3618, opened 2020-01-07
EC/Q: show all integral points not just half


JohnCremona:
We currently only show one of a pair \pm P, but we should show both (when they are not equal).  In case of a curve with a_1=a_3=0 these should be displayed as (x,\pm y), otherwise as (x,y_1), (x,y_2).

davidfarmer:
Glad to see this suggestion.  It had been argued for this
before (possibly by me), but unsuccessfully.  I have been bitten
by this more than once.

An edge case is this one:

https://www.lmfdb.org/EllipticCurve/Q/106/c/2

There are two integral points:  (0,1) and 0,-1) ,
but would those be shown as (0, \pm 1) even though a_1 \not= 0 ?

JohnCremona:
My intention (which was actually @jvoight 's suggestion) was only to use the \pm notation in the case I stated.  In general the two y's add up to -(a1*x+a3) which can of course be 0 even when a1, a3 are not both zero as in your example (a3=x=0).

What convinced me (at last) to do this was to be consistent with hyperelliptic curve points, though there the \pm abbreviation is not used in the analogous cases (see https://beta.lmfdb.org/Genus2Curve/Q/400/a/409600/1 for example).

The current record is 84 integral ponts on https://beta.lmfdb.org/EllipticCurve/Q/232050/gy/1 where we only show 42 currently.  In this case the new display would show all 84.  As for hyperelliptic points, the display would be a list if under a certain length, otherwise a table to look neater.

PR #3649, opened 2020-01-14
EC/Q: display all integral points, not just up to sign (Issue #3618)

JohnCremona:
When a_1=a_3 (only) we display pairs as (x,\pm y) for nonzero y.

Examples to try:

http://localhost:37777/EllipticCurve/Q/410550/ep/1 (random, no symmetry or 2-torsion)
http://localhost:37777/EllipticCurve/Q/11/a/3 (similar)
http://localhost:37777/EllipticCurve/Q/15/a/3 (None)
http://localhost:37777/EllipticCurve/Q/200/b/2 (symmetric equation and a 2-torsion point)
http://localhost:37777/EllipticCurve/Q/172/a/1 (symmetric equation, no 2-torsion)
http://localhost:37777/EllipticCurve/Q/106/c/2 (@davidfarmer 's example where \pm could be used but is not, by design)

The last one is arguable since the two points are negatives of each other, but checking for this in general would be a little tedious (though not hard).

### DIFF
--- a/lmfdb/elliptic_curves/web_ec.py
+++ b/lmfdb/elliptic_curves/web_ec.py
@@ -98,36 +98,30 @@ def EC_ainvs(E):
     """
     return [int(a) for a in E.ainvs()]
 
-def make_y_coord(ainvs,x):
+def make_y_coords(ainvs,x):
     a1, a2, a3, a4, a6 = ainvs
     f = ((x + a2) * x + a4) * x + a6
     b = (a1*x + a3)
     d = (RR(b*b + 4*f)).sqrt()
     y = ZZ((-b+d)/2)
-    return y, ZZ(d)
+    return [y, -b-y] if d else [y]
+
+def pm_pt(P):
+    return r"\(({},\pm {})\)".format(P[0],P[1]) if P[1] else web_latex(P)
 
 def make_integral_points(self):
-    ainvs = self.ainvs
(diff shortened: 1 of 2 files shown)


lmfdb/elliptic_curves/web_ec.py
```diff
@@ -98,36 +98,30 @@ def EC_ainvs(E):
...
-    num_int_pts = 0
-    for x in c["xcoord_integral_points"]:
-        y, d = make_y_coord(ainvs,x)
-        if d == 0:
-            num_int_pts += 1
-        else:
-            num_int_pts += 2
-    return num_int_pts
+    xcoords = c['xcoord_integral_points']
+    return sum([len(make_y_coords(ainvs,x)) for x in xcoords])
```
roed314: Why not just use `c['num_int_pts']` here?

JohnCremona: That function was written (not by me originally) in order to populate the new 'num_int_pts' column a few weeks ago and is not used by the website, only by the one-off rewrite script.  It will need to be used next time data is uploaded (when c will be a dictionary without the key 'num_int_pys' yet) so it might as well be simplified here.

roed314: Ok, thanks for clarifying.

PR #3650, opened 2020-01-14
merged two lists of sponsors into one

JohnCremona:
See http://localhost:37777/acknowledgment

The wording looked odd to me, this merges (concatenates) two lists into one.


edgarcosta:
I like the changes, but I will let others chime in.
Do you think we could also add the years for OpenDreamKit grant? (2015-2019)?

JohnCremona:
Yes they should all have years. I'll fix that. I ran this past the other man.eds. first.

jwj61:
Looks good, and agree on adding years for all.

JohnCremona:
I added the dates requested.  Can be merged once tests pass.

PR #3651, opened 2020-01-14
Fix bug in random_sample's CHOICE method

roed314:
There was a bug in `random_sample`.  Since `random_sample` isn't currently used in the production site, this change should have no effect on the website.

To see the difference, compare the results of 
`db.ec_curves.random_sample(1/3000, mode='choice', projection=['num_int_pts', 'xcoord_integral_points', 'label', 'ainvs'])` before and after this change.

### DIFF
--- a/lmfdb/backend/database.py
+++ b/lmfdb/backend/database.py
@@ -2294,7 +2294,7 @@ class PostgresTable(PostgresBase):
             count = int(len(results) * ratio)
             if repeatable is not None:
                 random.seed(repeatable)
-            return self._search_iterator(random.sample(results, count), search_cols, extra_cols, projection)
+            return random.sample(results, count)
         elif mode in ['SYSTEM', 'BERNOULLI']:
             if extra_cols:
                 raise ValueError("You cannot use the system or bernoulli modes with extra columns")


JohnCremona:
Your sample query takes about 2 minutes since the null query has >3M rows.  Was that intended?

Anyway, having waited (twice!) for the function to complete, what I see is: (before) it returns a generator object, all of whose yield values are
```
{'ainvs': 'ainvs',
 'label': 'label',
 'num_int_pts': 'xcoord_integral_points',
 'xcoord_integral_points': 'num_int_pts'}
```
which is certainly not what is intended, while (after) it returns a long list,  of ~3M/3000 =~1000 entries, each looking like
```
{'ainvs': [1, -1, 1, -7233755, -7486571253],
  'label': u'486450ea2',
  'num_int_pts': 0,
  'xcoord_integral_points': []}
```
which looks a lot better.

If that is what you intended the change to do then I'm happy with it.  The function is used in verify/verification.py (as you presumably knew).

roed314:
> Your sample query takes about 2 minutes since the null query has >3M rows. Was that intended?

Sorry about that.  I should have given a faster example rather than just copy the one I was using in looking at #3649.

> Anyway, having waited (twice!) for the function to complete, what I see is: (before) it returns a generator object, all of whose yield values are
> 
> ```
> {'ainvs': 'ainvs',
>  'label': 'label',
>  'num_int_pts': 'xcoord_integral_points',
>  'xcoord_integral_points': 'num_int_pts'}
> ```
> 
> which is certainly not what is intended, while (after) it returns a long list, of ~3M/3000 =~1000 entries, each looking like
> 
> ```
> {'ainvs': [1, -1, 1, -7233755, -7486571253],
>   'label': u'486450ea2',
>   'num_int_pts': 0,
>   'xcoord_integral_points': []}
> ```
> 
> which looks a lot better.
> 
> If that is what you intended the change to do then I'm happy with it. The function is used in verify/verification.py (as you presumably knew).

Yep, that's the intention.  I'll go ahead and merge this then.  Thanks for the review.

PR #3655, opened 2020-01-17
delimiter=sep --> delimiter=str(sep) 

edgarcosta:
CSV reader in python2 doesn't support utf8, we may help it out by casting it to str

@AndrewVSutherland, this should hopefully fix your bug that you noticed this morning


AndrewVSutherland:
@edgarcosta I can confirm that this fixes the problem.

PR #3657, opened 2020-01-18
More genus 2 tweaks

AndrewVSutherland:
This PR changes the display of Mordell-Weil generators to display divisors as sums of rational points when possible,   To see the difference go to

https://red.lmfdb.xyz/Genus2Curve/Q/

and click on each of the "interesting curves" (they all hit different cases, depending on whether divisors have rational affine/infinite parts and the degree of the affine part).

The is also a new row in the BSD invariants table that indicates whether the Hasse-Weil conjecture has been verified the the L-function of this curve (this currently applies only to cases where the Jacobian is of GL2-type or Q-isogenous to a product of elliptic curves over Q, or the paramodular conjecture is known).  More cases will be added in the future, but this will only involve data changes.  See

https://red.lmfdb.xyz/Genus2Curve/Q/169/a/169/1 (verified)
https://red.lmfdb.xyz/Genus2Curve/Q/277/a/277/1 (verified)
https://red.lmfdb.xyz/Genus2Curve/Q/563011/a/563011/1 (unverified)

The analytic rank display now checks the "analytic_rank_proved" flag and will not display upper bound when this is set (this applies, for example, to rank 2 Jacobians isogenous to a product of elliptic curves over Q), see

https://red.lmfdb.xyz/Genus2Curve/Q/3721/a/3721/1 (new)
https://beta.lmfdb.org/Genus2Curve/Q/3721/a/3721/1 (old)

I also changed the code snippets to get rid of some duplicates and moved them so they appear after the relevant entry on the page rather than before, which is consistent with elliptic curves.

### DIFF
--- a/lmfdb/genus2_curves/main.py
+++ b/lmfdb/genus2_curves/main.py
@@ -322,6 +322,8 @@ def genus2_curve_search(info, query):
         mode = 'exact'
     elif info.get('bad_quantifier') == 'exclude':
         mode = 'complement'
+    elif info.get('bad_quantifier') == 'subset':
+        mode = 'subsets'
     else:
         mode = 'append'
     parse_primes(info, query, 'bad_primes', name='bad primes',qfield='bad_primes',mode=mode)
--- a/lmfdb/genus2_curves/templates/g2c_browse.html
+++ b/lmfdb/genus2_curves/templates/g2c_browse.html
@@ -150,9 +150,9 @@ A <a href={{url_for('.random_curve')}}>random genus 2 curve</a> from the databas
 {{KNOWL('g2c.good_reduction', title='Bad primes')}}&nbsp;&thinsp;
   <select name='bad_quantifier'  style="width: 122px">
   <option value=''>include</option>
-  <option value="exclude">exclude</option>
+  <option value='exclude'>exclude</option>
   <option value='exactly'>exactly</option>
-  <option value='subset of'>subset</option>
+  <option value='subset'>subset of</option>
   </select>
 </td>
 <td><input type='text' name='bad_primes' size=10 example='5,13'></td>
--- a/lmfdb/genus2_curves/templates/g2c_browse.html
+++ b/lmfdb/genus2_curves/templates/g2c_browse.html
@@ -152,6 +152,7 @@ A <a href={{url_for('.random_curve')}}>random genus 2 curve</a> from the databas
   <option value=''>include</option>
   <option value="exclude">exclude</option>
   <option value='exactly'>exactly</option>
+  <option value='subset of'>subset</option>
   </select>
 </td>
 <td><input type='text' name='bad_primes' size=10 example='5,13'></td>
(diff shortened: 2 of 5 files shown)


AndrewVSutherland:
Travis failed due to not being able to update stats (which didn't really need to be updated but were likely flagged as out of date after I made updates to g2c_curves and g2c_ratpts).  I just restarted Travis.

roed314:
There's a typo in the code snippets for Sage: it should be `HyperellipticCurve` rather than `Hyperellipticcurve` (see https://red.lmfdb.xyz/Genus2Curve/Q/169/a/169/1 for example).  It would also probably be good not to have the definitions of `R`, `C` and `X` all on one line: with my default screen width showing the Sage code snippets causes it to drop down below the properties box.

AndrewVSutherland:
@roed314 Thanks, I fixed the typos and split both the sage and magma curve snippets into two pieces (apparently you can't embed a linebreak in a single snippet?  In any case I think two snippets is fine).

JohnCremona:
> @roed314 Thanks, I fixed the typos and split both the sage and magma curve snippets into two pieces (apparently you can't embed a linebreak in a single snippet? In any case I think two snippets is fine).

I'm sure you can and I have done this.  In elliptic_curves I put the snippets in a yaml file and some of those have more than one line, so it must be possible directly too.

AndrewVSutherland:
@JohnCremona Thanks, I'm sure you are right.  Given that these are in some sense two separate things (one snippet for the minimal model, another for the simplified model), I think it makes sense to leave them as separate snippets, but there are definitely other cases where mult-line snippets will be necessary.  if/when Magma fixes the bug that is preventing me from including a snippet to create the generators for the Mordell-Weil group of the Jacobian (without recomputing them), I will add a snippet that will likely require this.

JohnCremona:
A big +1 for the expressions of generators as sums of rational points where possible, i.e. where the divisor's support is defied over Q.

Do you mean the BSD conjecture rather than the Hasse-Weil?  I am confused otherwise, but this would require a change to the template if I am right.

Bad primes subset option:  in the search-again template there are some superfluous quotes around 'subset of' which I think should just say subset as on the main search page; also the search-again template goes back to the default for this switch instead of pre-selecting the last value.

A mathematical question.  In https://beta.lmfdb.org/Genus2Curve/Q/3721/a/3721/1 the two generators have the same height (to the precision shown).  I think that there is a conjecture that this can happen only if there's an automorphism taking one to the other (there certainly is such a conjecture for elliptic curves, even a stronger one about independence of heights).  Is that the case here?  It may be sice the curve has 6 autos, so not just the hyperelliptic involution.

AndrewVSutherland:
@JohnCremona Thanks, I fixed the subset quotes issue (and it now says subset of)

The caption Hasse-Weil conjecture is correct.  We added this (after some discussion with John Voight and Andy Booker) because all computations of analytic ranks are made under the assumption that the Hasse-Weil conjecture holds for the L-function of the curve.  Tthis is stated explicitly in the knowl for the analytic rank that is displayed immediately below, and in the reliability knowl.

Whether BSD holds or not, it doesn't really make sense to talk about or display a number for the analytic rank without assuming that the L-function has an analytic continuation to the central point (without the Hasse-Weil conjecture we can't even say the L-function is meromorphic at the central point, never mind holomorphic, and it makes no sense to talk about its order of vanishing there).

For genus 2 curves of GL2-type, products of elliptic curves, and cases where the paramodular conjecture has been verified (e.g. conductor 277) we know that the Hasse-Weil conjecture holds, but in most cases we do not.  But we still display the analytic rank, and having the line "Hasse-Weil conjecture: unverified" immediately above it is meant to alert the user to the fact that the analytic rank was computed under an assumption that has not been verified.

This issue is independent of the fact that in general analytic ranks greater than 1 are usually only known to be upper bounds, even if the Hasse-Weil conjecture has been verified, this is indicated by the parenthetical (upper bound).

JohnCremona:
Sorry for misunderstanding, I should have read the Hasse-Weil knowl.

PR #3661, opened 2020-01-21
EC/Q: added subsets option for bad primes search selector

JohnCremona:
@jvoight asked 

>  How do I get the LMFDB to produce the list of elliptic curves with
good reduction away from 2,3?  In "bad primes", I have to choose from
'include', 'exclude', 'exactly', but then I guess I'd have to choose
'exactly' and merge 3 outputs?  And if I wanted to do this for
2,3,5,7, I need 15 outputs?

which raises a serious shortcoming, fixed by this PR.  Try selecting 'subset of' now with primes 2,3 and you should see all 752 curves listed here http://homepages.warwick.ac.uk/staff/J.E.Cremona/ftp/data/extra/E-2-3.txt

### DIFF
--- a/lmfdb/elliptic_curves/elliptic_curve.py
+++ b/lmfdb/elliptic_curves/elliptic_curve.py
@@ -276,6 +276,8 @@ def elliptic_curve_search(info, query):
         mode = 'exact'
     elif info.get('bad_quantifier') == 'exclude':
         mode = 'complement'
+    elif info.get('bad_quantifier') == 'subset':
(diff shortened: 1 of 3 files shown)


JohnCremona:
PS Genus 2 curves need something similar

AndrewVSutherland:
@JohnCremona Thanks, I just added this to #3657  (if someone would review this and merge it once travis finishes, I would be grateful -- it is installed on https://red.lmfdb.xyz)

Issue #3338, opened 2019-09-06
Table headers of search results should be knowls


grundman7:
After a search in global number fields, a table is returned.  The header "Label" on the first column of the table is not a knowl.  So I don't see a way to easily get a reminder as to what the parts of the field label mean, without leaving the page.  (I did fine a reference for it on the right-hand side bar, but it takes me to another page, rather than being a knowl.  And, in any case, it would be easier to find if the column header was a knowl.)  

Example of search return: http://www.lmfdb.org/NumberField/?ram_quantifier=exactly

Example of different part of lmfdb where a search yields a table with knowl headings: http://www.lmfdb.org/EllipticCurve/Q/?surj_quantifier=include

EDIT: the issue for global number fields has been addressed, but the exact same issue is sill present on the following search result pages:

https://www.lmfdb.org/ModularForm/GL2/TotallyReal/?all=1
(Label column header should link to the knowl mf.hilbert.label)

https://www.lmfdb.org/ModularForm/GL2/ImaginaryQuadratic/?all=1
(Label column header should link to the knowl mf.bianchi.labels and the Field column header should link to the knowl nf, as it does on the HMF search results page)

(the individual field labels in the Field columns on both pages should link to number field home pages, but this is a separate issue, see #3534)

There is a similar issue for the Polynomial column on the local fields search results page:
https://www.lmfdb.org/LocalNumberField/?all=1
(Polynomial column header should link to the knowl lf.defining_polynomial)

edgarcosta:
@grundman7 Thanks!

Here is a good example to follow: http://www.lmfdb.org/ModularForm/GL2/Q/holomorphic/?search_type=List&all=1

AndrewVSutherland:
Resolved by #3340.

edgarcosta:
@AndrewVSutherland, #3340 resolved this for number fields, but not the BMF, HMF and local field.

AndrewVSutherland:
@edgarcosta Other than the labels you added, there is nothing in the description of the issue or the comments that makes this clear.  If you are going to mark an issue as a good first issue it should be blindingly obvious exactly what needs to be done, and in general, if you don't think an issue is ready to be closed you should edit the description to make this clear (as I have just done).  Any issue for which the problem described in its description has been resolved is fair game for closing.

edgarcosta:
@AndrewVSutherland, I agree with you. I should have also commented instead of just adding labels.

I still think we should that local fields should have the knowl 
[...]

Issue #3534, opened 2019-12-14
Field labels in HMF and BMF search results should link to number field home page


AndrewVSutherland:
On the Bianchi and Hilbert modular forms search results pages

https://www.lmfdb.org/ModularForm/GL2/ImaginaryQuadratic/?all=1
https://www.lmfdb.org/ModularForm/GL2/TotallyReal/?all=1

The field labels listed in the "Field" column should all be links to the homepage of the corresponding number field, as on 

https://www.lmfdb.org/EllipticCurve/?all=1
https://www.lmfdb.org/EllipticCurve/?field=2.2.5.1

for example.  Also, I think it would be better to have the column header be "Base Field" to avoid possible confusion with the coefficient field (Hecke eigenvalue field), which coincides with the base field only when the dimension is 1.

JohnCremona:
I think the links from field labels should be to field knowls not home pages.
I agree about "base field"  for the column header but it has no relation to the Hecke field which is Q for dimension 1 forms.

AndrewVSutherland:
@JohnCremona Thanks, they should of course be field knowls, as they are on the ECNF pages, I have edited the description accordingly.

edgarcosta:
I have edited the description to reflect the "field knowls" and progress from #3670 and #3672

edgarcosta:
@samuel-rice is working this

PR #3670, opened 2020-01-24
Added knowl to field and label

apeng16:
Address part of #3338 for Bianchi modular form search results, added knowl to field and label


AndrewVSutherland:
Looks good, will merge once Travis finishes.

apeng16:
did part of #3534 by changing "Field" to "Base field" (and also "Base Field" to "Base field" to be consistent with https://www.lmfdb.org/EllipticCurve/?all=1 )

PR #3673, opened 2020-01-25
Using search boxes in G2C

edgarcosta:
the goal of this PR is to build on the work of @roed314 and make use of search boxes in genus 2 curves. 

This PR is running on scratch.lmfdb.xyz

1- I tweaked the search boxes, so the default widths match CMFs, which affects the looks of search/refine for abvar/fq  but the layout didn't change
Compare:
-https://beta.lmfdb.org/Variety/Abelian/Fq/
-https://beta.lmfdb.org/Variety/Abelian/Fq/?search_type=List
vs
-https://scratch.lmfdb.xyz/Variety/Abelian/Fq/
-https://scratch.lmfdb.xyz/Variety/Abelian/Fq/?search_type=List

2- used search boxes in genus 2 curves
in the browse page, the layout is the same, but now the input boxes are wider
and we added the random button at the end
https://beta.lmfdb.org/Genus2Curve/Q/
vs
https://scratch.lmfdb.xyz/Genus2Curve/Q/

in the refine search the layout changed a bit, as the default is to have 5 boxes and not 6
compare
https://beta.lmfdb.org/Genus2Curve/Q/?all=1
vs
https://scratch.lmfdb.xyz/Genus2Curve/Q/?search_type=List&all=1
if unhappy with the layout I picked, this is now trivial to tune

3- Fixed a minor bug
 https://beta.lmfdb.org/Variety/Abelian/Fq/2/9/ak_bl takes you to a fake search page
while
https://scratch.lmfdb.xyz/Variety/Abelian/Fq/2/9/ak_bl
gives a 404 and a clear error message that we don't have such an isogeny class.

4- inadvertently reformatted `lmfdb/utils/search_wrapper.py`, if desired I can remove these white space changes


AndrewVSutherland:
Clicking on the random result button seems to do the same thing as clicking on the list results button (both in genus 2 and avbar)

edgarcosta:
@AndrewVSutherland, it's fixed now. last night I tried to undo some changes on `lmfdb/utils/search_wrapper.py` and had forgotten that I had indeed touched for a good reason.

AndrewVSutherland:
Searching on torsion structure appears to be broken (every curve is returned), compare:

https://beta.lmfdb.org/Genus2Curve/Q/?torsion=%5B2%2C8%5D

vs,

https://scratch.lmfdb.xyz/Genus2Curve/Q/?hst=List&torsion_structure=%5B2%2C8%5D&search_type=List

I think the problem is simply that you are using "torsion_structure" rather than "torsion", since

https://scratch.lmfdb.xyz/Genus2Curve/Q/?hst=List&torsion=%5B2%2C8%5D&search_type=List

works.  But this makes me realize that before we merge this we need to individually test every input box individually to make sure it is generating the correct query..

edgarcosta:
I have fixed that typo.
I also went through every box and checked that I wasn't getting the whole database, and that the name on the url, somewhat matched what I was searching on. 
So hopefully no more of those bugs.

AndrewVSutherland:
Looks good for genus 2 curves.  Someone else (perhaps @roed314 or @kedlaya) should take a look at abelian varieties.

edgarcosta:
I would like to emphasize, that for abvars I only changed the display, by making the box size to match other places.

AndrewVSutherland:
I notice that the capitalization is inconsistent on the refine search results page of abvar (first word of captions should be capitalized, as on the genus 2 refine search page).

Also, the width of the Qbar-invariants search box on the genus 2 browse change is different from
www.lmfdb.xyz -- the right edge should line up the the boxes below (at least on chrome).

edgarcosta:
Both issues addressed.
We should figure out a way to do make the input box take the width of the column.
Perhaps with something like https://stackoverflow.com/questions/7525675/text-input-with-100-width-inside-a-td-expands-continues-outside-the-td
But for the moment this works

AndrewVSutherland:
Thanks @edgarcosta (I note that it still doesn't align in firefox, but neither does the existing version).

For my part, I'm happy to merge this, but I still think someone should make sure the abelian variety searches all still work correctly (I don't have time to do this right now).  I just want to be sure there isn't another case where a typo causes the search to ignore the constraint and return everything (as happened with torsion structure).

edgarcosta:
>  I just want to be sure there isn't another case where a typo causes the search to ignore the constraint and return everything (as happened with torsion structure).

I only changed cosmetic looks for the genus 2 curves, so that won't happen for abvars.

However, please don't merge this yet, I realized there is something broken in dynamic stats.
I will fix it tomorrow.

AndrewVSutherland:
I'm not sure why Travis keep restarting, but the tests pass fine, I]m going to merge this.

PR #3674, opened 2020-01-25
activate pylint check for python3 compatibility

fchapoton:
this adds pylint --py3k to the travis checks

for the moment, with some exceptions in the checks, that should be taken care of later

this should help not to move backwards about python3 compatibility

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -56,9 +56,9 @@ install:
     # what is the load on the server
     - uptime
     - wget  --no-verbose http://grace.mit.edu/sage-lmfdb/sage-${SAGE_VERSION}-Ubuntu_14.04-x86_64.tar.bz2
-    # travis_wait extends the default 10 minute timeout to 40 minutes
     # what is the load on the server
     - uptime
+    # travis_wait extends the default 10 minute timeout to 40 minutes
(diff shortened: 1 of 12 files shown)


fchapoton:
When succesfull, pylint, unlike pyflakes, does return something. How to handle that ?

fchapoton:
worse, it seems that the option --py3k has been removed recently from pylint

edgarcosta:
One could check for the exit code.

fchapoton:
The output should be ok, using the option "score=no". Now it's the install that is failing, do you see a way to fix it ? Should we add "pylint" to requirements.txt ?

edgarcosta:
We must install pylint in python2 to have py3k option.
The current commit will fail, but now for a good reason: https://travis-ci.org/edgarcosta/lmfdb/builds/641826230#L1152

fchapoton:
Damn, where are all these bad print coming from ???

edgarcosta:
I figure it out, pylint is just complaining that we are not importing the print_function

edgarcosta:
We should be good now.

fchapoton:
Looks good to me.

Issue #3580, opened 2019-12-26
 support expressions in queries that refer to other columns


jenpaulhus:
This is a request related to the group order search on HGCWA.  Currently, to search for linear expressions for the group order in terms of the genus, a large "or" statement is created for each genus in the database.  As @roed314 says in PR #3516, the better way to do this would be to,  "support expressions in the generated postgres query that refer to other columns".

AndrewVSutherland:
Agreed.  For the benefit of those who might not be aware of this.  It is possible to perform queries that compare two columns by going directly to postgres, e.g. to get a list of the 11 rows in hgcwa_passports for which group_order = 84*(genus-1) you can do
```
from psycopg2.sql import SQL
from lmfdb import db

cur = db._execute(SQL("SELECT * from hgcwa_passports WHERE group_order = 84*(genus-1)"))

```
but there is currently no way to do this using the standard db backend interface.  A quick and dirty way to do this would be to simply accept a WHERE clause in place of the query parameter of lucky, count, and search, e.g. the above search could be specified as

`results = db.g2c_curves.search ("group_order = 84*(genus-1)")`

This would be trivial to implement and would provide a lot of functionality (e.g. if I wanted to find all the newforms that are the designated representative of their twist equivalence class I could just do

`results = db.mf_newforms.search("label=minimal_twist")`

What do you think @roed314 ?  this would only take a few lines of code in db backend and seems a lot easier to me than trying to adapt the current query dictionary interface to support more general postgres queries.

roed314:
I'm hesitant about the specifics of your suggestion for a few reasons.  First, we need to do some kind of validation to prevent SQL injection attacks (we can't accept arbitrary strings).  Second, while this works fine for interactive use, it doesn't behave very nicely with the infrastructure we have set up for parsing inputs from input boxes, where we have search parsers that are called consecutively to create a dictionary.

Here's an alternative suggestion which I think will work better on both fronts.  We add a new special parsing command `"$raw"`, which takes a string.  This string is split up and every string of alphabetic characters is wrapped in double quotes (this will make Postgres treat them as a name, for example a column name).  Aside from these column names, the only other allowable characters in the string are arithmetic operations: `+-*/^`.  We'd need to think a bit further if we want to support congruence conditions in this way, but I think it already does a lot of what we need and isn't that bad to implement.  So, to do the search requested above, we would use
`db.g2c_curves.search({'group_order':{'$raw':'84*(genus-1)'}})`.

[...]

PR #3675, opened 2020-01-27
Raw search

roed314:
Working toward #3580, this PR adds two features
* The ability to pass a `raw` keyword argument to the search and lucky functions.  This is intended ONLY for interactive use and should never be used with text entered by a user on the front end
* The addition of a `$raw` keyword to the query parser, which allows for the user to enter expressions involving other columns.  

I haven't enabled the second feature for any actual search input so that this PR is safe to merge.  I will have a followup PR that enables raw search.

I'd like to get feedback, especially around the security implications.  I tried to ensure that these changes don't open us up to any SQL injection vulnerabilities by aggressively whitelisting, but I'm not an expert yet so I don't know if something might have slipped through.


edgarcosta:
I pushed this to https://teal.lmfdb.xyz/
(in a read only mode)
but 
https://teal.lmfdb.xyz/Variety/Abelian/Fq/?hst=List&g=2-&abvar_point_count=q%5Eg
doesn't work for me.

edgarcosta:
@alexjbest, @roed314  and I checked this carefully and we believe that everything looks good.
Nonetheless, we are keeping this feature disabled for the time being.

PR #3677, opened 2020-01-27
Simplifying Cocalc instructions

edgarcosta:


AndrewVSutherland:
It would be great if some non-expert user tested this, I'm going to merge it to make it easier to do so (this is just a documentation change).  To test just follow the co-calc instructions at

https://github.com/LMFDB/lmfdb/blob/master/GettingStarted.md#cocalc

edgarcosta:
FYI, knowls are still broken in cocalc and conrey-dirichlet-characters still fails to compile

PR #3687, opened 2020-01-30
A better fix for abvar stats

edgarcosta:
Stats code was not ready to handle a bucket of the shape '{'$gt': -1}' for something that just takes -1, 0, or 1.
I fixed it by avoiding the row that was triggering this, but I'm don't really understand the underlying issue.
Hopefully, @roed314 will have a much better idea.

Meanwhile, this restores abvar stats to all of its content.


roed314:
I'm taking a look now....

edgarcosta:
FYI, the tests might not succeed. The server that serves our sage image to Travis is under a minor denial of service attack and once in a while, Travis gets caught in this mess, and its ip gets blocked.

Issue #3424, opened 2019-10-27
Sluggish genus 2 page load times


AndrewVSutherland:
The load time for genus 2 pages is longer than it should be, and longer than it used to be.  According to Google analytics, the average load time for 

http://www.lmfdb.org/Genus2Curve/Q/277/a/277/1

is over 3 seconds (it is a lot faster than that for me, but still noticeably slower than EC/Q, number fields, and CMFs, and there is no reason why this should be the case).  The page above has a top knowl which might be slowing things down a bit, but so does

http://www.lmfdb.org/EllipticCurve/Q/11a1/

and the genus 2 page is consistently about 50 percent slower to load (around 800 ms versus 500 ms).   I don't know if this has anything to do with the recent changes to add new search features (I suspect not).

For reference, our target load time is 400 ms (which is easily achieved on MIT's internet connection by pages like

http://www.lmfdb.org/LocalNumberField/5.4.2.1
http://www.lmfdb.org/SatoTateGroup/1.4.6.1.1a
http://www.lmfdb.org/GaloisGroup/36T109887

Number fields, elliptic curves, and CMFs all come pretty close to this on pages that do not have an overwhelming amount of data (typically around 500 ms).

edgarcosta:
In both cases, EC and G2C generating the plot take more than 30% of the time.
We should perhaps consider precomputing these.
The time spent doing queries to the database seem meaningless in comparison.



These are the queries done to the db sorted by time:
```  3 INFO:g2c_endomorphisms@2019-10-29 16:34:49,683: SELECT "factorsRR_base", "spl_facs_condnorms", "factorsQQ_geom", "spl_facs_labels", "is_simple_base", "spl_fod_label", "ring_geom", "st_group_base", "spl_facs_    coeffs", "factorsQQ_base", "fod_label", "fod_coeffs", "spl_fod_coeffs", "is_simple_geom", "lattice", "st_group_geom", "label", "ring_base", "spl_fod_gen", "factorsRR_geom" FROM "g2c_endomorphisms" WHERE "lab    el" = '277.a.277.1' LIMIT 1 ran in  0.0657460689545s  [database.py:404]
  2 INFO:db_knowl@2019-10-29 16:34:53,077: SELECT DISTINCT ON (id) "id", "title" FROM kwl_knowls WHERE status >= 0 ORDER BY id, timestamp DESC ran in  0.0123329162598s  [database.py:404]                         
[...]

PR #3691, opened 2020-01-31
Faster genus 2 plots

rbommel:
I have added all plots for the genus 2 curves in a new table in the database. The code now attempts to load the plot from the database and if this fails, it will use the old code. Moreover, I added the function that I used to extract the plots from the data. This handles issue #3424.


edgarcosta:
`g2c_plots` has been copied to proddb

edgarcosta:
I'm happy to merge it when the tests pass

PR #3692, opened 2020-01-31
Add Sage parser

roed314:
Some updates to the new raw parsing functionality that uses a sage symbolic parser to catch errors earlier and thus provide better error messages to the user (and also help further alleviate SQL injection attempts, since the text must be a valid symbolic expression).


roed314:
This should also be safe to merge since it doesn't enable any of the raw parsing from the website.

PR #3695, opened 2020-02-02
Splitting lmfdb/backend/database.py in various files

edgarcosta:
The file `lmfdb/backend/database.py` was getting out of control.
We decided to split it per class. Along the way, we also found some bugs :)

@roed314 might still add some commits to fix some formatting.


AndrewVSutherland:
I don't think this problem was caused by this PR but refresh_stats is currently broken, and it probably makes sense to fix it in this PR.

To reproduce just run db.mytable.stats.refresh_stats() and you will hit a TypeError (Composed elements must be Composable).

roed314:
I fixed the issue @AndrewVSutherland observed, and I'm happy with the formatting.  Once tests pass I'll merge.

edgarcosta:
:)

PR #3697, opened 2020-02-07
CMF Twists

AndrewVSutherland:
This PR makes available the twisting data for CMFs that is now available in the LMFDB in the tables mf_twists_nf (1456304 twists of newform orbits) and mf_twists_cc (49165089 twists of embedded newforms).  This includes a complete decomposition of all the CMFs in the LMFDB into twist equivalence classes and the complete graph of twists within each twist equivalence class.  This data was also used to compute inner twist data for all CMfs (not just those of dimension 20), and this data is now rigorous in all cases.

In order to make this possible 131 additional newspaces containing 746 newforms were added to the database to ensure that for every twist equivalence class that inersects the database the designated representative of that twist class (always a twist minimal form of minimal character) is in the database.

This PR is up and running on blue.lmfdb.xyz.  If you go to the home page of any newform orbit or embedded newform and scroll down to the bottom you will see tables of twists, and for newforms that are not the designated representative of their twist class, the designated representative (called "the minimal twist" will be listed as a related object).

The merge conflicts in this PR only relate to changes in the test scripts related to the python3 transition and should all be very easy to resolve (keep the lines with the as_text parameter specified), which I will do now.


davidfarmer:
I poked around a bit and liked what I saw.

On this page:

https://blue.lmfdb.xyz/ModularForm/GL2/Q/holomorphic/49/12/c/b/

it is slightly confusing that row 7.c in the first Twists
table has "inner" in the Type column, but in the Inner twists
table above it, the corresponding Type column is empty.

Looking at another example:

https://blue.lmfdb.xyz/ModularForm/GL2/Q/holomorphic/39/1/d/a/

it strikes me that maybe the problem is that we need another
option beyond trivial, RM, and CM to put in the Inner twists
table.  (I can't think of a good term.  Maybe "generic"?)

To an extremely naive user it may be slightly confusing that
in the Twists tables the form for the current page is not
highlighted (it is indicated by the word "trivial").

AndrewVSutherland:
@davidfarmer I added the text "inner" to the non-self twists listed in the inner twist table to make it consistent with the twist tables.  I did not do anything to further highlight the trivial twist (I don't think this is necessary and I worry that adding something more would make the page busier and possibly more confusing)

On a separate note, I could see an argument to remove the inner twists table entirely, since the twists table includes the same information, but this seems like a big change and it probably makes sense to get this PR up and running  for a while before contemplating such a change (there is also an argument for keeping the inner twists table, since there is an underlying group structure that we might eventually want to say more about).

AndrewVSutherland:
The travis failure is due to a problem with stats that is not related to this PR (CMF stats are currently broken on master as well).

All the data related to twists has now been copied to the cloud.  This PR is ready to be merged.

roed314:
I think I've fixed the stats problem in #3702.  If you want to merge that code into this PR I think travis will pass.

JohnCremona:
I wanted to restart Travis but cannot see the right button to press.

edgarcosta:
I have restarted it

roed314:
I think this travis run will fail again, because we need to merge the changes from #3702 into this branch if we want the stats to pass.

JohnCremona:
I thought that after changes in master such as the one I made by merging your other pr, this pr would be updated to a merge of its commits with the new master. No?

roed314:
This is done for Sage (the commit previews on trac tickets are based on merging with the most recent master, and I think the patchbot does this as well).  My guess is that travis doesn't do so though: I think it just checks out the branch on this ticket and runs tests.  We'll find out soon. :-)

roed314:
In the travis log (line 2080) we get `lmfdb/classical_modular_forms/test_cmf.py::CmfTest::test_stats PASSED` so maybe I'm wrong.

edgarcosta:
Travis has passed

roed314:
Great!  Glad I learned something about how travis works.

PR #3698, opened 2020-02-07
Belyi download buttons

SamSchiavone:
I added download buttons for Galois orbits of Belyi maps to download code to SageMath and all data to text, as well as tests.

(The TODO list for Belyi is also updated.)


edgarcosta:
Regarding labels, I think it should be: `7T7-6.1_5.2_4.2.1`

Issue #2759, opened 2019-03-15
Hecke polynomials for small primes on newform and newspace pages


roed314:
It would be nice to see the characteristic polynomial of Hecke operators for several small primes (e.g. 2, 3, 5 and 7).  This could be done both for spaces and forms, for dimension up to some bound (e.g. 100).  For forms of dimension up to 20, this information is present in the algebraic q-expansion, but not easily extracted.

jvoight:
Agree with this!  Especially for spaces, this was one of the first things that people tabulated, and it is a nice snapshot of all the forms in a space.

edgarcosta:
FYI, the data is available in the table `mf_hecke_lpolys`, see  http://www.lmfdb.org/api/mf_hecke_lpolys/

jvoight:
@edgarcosta, does that make this a good first issue for tomorrow or another LMFDB Friday?

edgarcosta:
I think so.
Do we agree that we would like the display to look like HGM families at the
end of the page?
eg: https://beta.lmfdb.org/Motive/Hypergeometric/Q/A14.2_B8.3.1

assaferan:
Hi, I'll be taking this one

assaferan:
This is resolved in pull request #3700.

assaferan:
@roed314 - do you agree we can close this issue?

roed314:
I see them on newform pages, but somehow they're no longer present on [newspace](https://beta.lmfdb.org/ModularForm/GL2/Q/holomorphic/117/6/a/) or [full $$\Gamma_1(N)$$ spaces](https://beta.lmfdb.org/ModularForm/GL2/Q/holomorphic/117/6/).  I see the changes in #3700, but they seem not to be there anymore, so I'm not sure what happened.

assaferan:
If you look at [Drew's remark](https://github.com/LMFDB/lmfdb/issues/4196#issuecomment-72345993) on issue #4196 you will see that Barinder followed through and removed them from the spaces pages.

roed314:
Great.  Closing.

PR #3700, opened 2020-02-09
2759

assaferan:
Resolving issue #2759.


lmfdb/classical_modular_forms/web_newform.py
```diff
@@ -954,6 +954,55 @@ def td_wrap(val):
...
+      # replaced by a call to a single class/function with some parameters
+        def th_wrap(kwl, title):
+            return '    <th>%s</th>' % display_knowl(kwl, title=title)
+        def td_wrap(val):
+            return '    <td>$%s$</th>' % val
+        data = db.mf_newforms.lookup(self.label)
+        hecke_orbit_code = data['hecke_orbit_code']
+        self.heckepolys = []
+        R = PolynomialRing(ZZ, 'T')
+        T = R.gen(0)
```
edgarcosta: This line needs to be deleted as the variable `T` is never used.

assaferan: Oops, sorry. Also noticed that I should add the change to new spaces as well. On it.

assaferan:
Ok, I think that now it's good to go - it displays the polynomials for newform spaces (also for Gamma1 spaces) and for the newforms themselves. Also, I got rid of the unused variable.

AndrewVSutherland:
The travis failure is pylint complaining about line 197 of web_space.py:

web_space.py:    heckepolys = hecke_polys_orbits.items()

The variable heckepolys is used in two places below

web_space.py:    if len(heckepolys) == 0:
web_space.py:    for p, lpoly in heckepolys:

While this warning is pretty mild, and I think the code you wrote is valid, rather than disable the warning I think it would be better to change your code to avoid it.

You can replace "if len(heckepolys) == 0" with "if not hecke_polys_orbits" (which is arguable better in any case), and then replace heckepolys with hecke_polys_orbits.items() in the for statement in line 295.

assaferan:
Thanks!

AndrewVSutherland:
So this works and I like how it looks, but there is a performance issue that we need to deal with before this PR can be merged.  In high weight cases where the Hecke polys have huge coefficients the pages take a very long time to load, e.g.

http://127.0.0.1:37777/ModularForm/GL2/Q/holomorphic/1/254/a/a/

takes about 25 seconds, vs less than 1 second for

https://www.lmfdb.org/ModularForm/GL2/Q/holomorphic/1/254/a/a/

This is an extreme example (we don't even try to compute the rational L-function for newforms with weight over 200),  but a slightly less extreme example is

http://127.0.0.1:37777/ModularForm/GL2/Q/holomorphic/1/198/a/a/

but you can see the same L-polynomials much more quickly in a condensed display that uses bigint knowls on the rational L-function page:

https://www.lmfdb.org/L/ModularForm/GL2/Q/holomorphic/1/198/a/a/

Now this page is still extracting just as much data from the database and sending just as much data to the browser (more in fact), so the code that is generating the Hecke poly must be less efficient in some way.

Before we can merge this we need to (1) suppress the display in cases where the charpolys are simply too big to load in a reasonable amount of time (or maybe just give the first 5 rather than 25 of them), (2) figure out why it is taking so much longer to construct and display this table than it is on the rational L-function pages, and (3) if we don't decide to suppress display of L-polynomials with large coefficients, we should use bigint knowls to keep the amount of horizontal scrowling reasonable (and possibly this also speeds things up and addresses (2), it would be wroth trying in any case).

@edgarcosta do you have any thoughts on this?

assaferan:
Thanks Drew - meanwhile I added the displaying of the coefficients as bigints.
After profiling it seems that the factoring is taking most of the time (in the above example http://127.0.0.1:37777/ModularForm/GL2/Q/holomorphic/1/198/a/a/ - it takes 6.5 out of the 7 seconds)
I noticed that in the L functions data the Euler factors are already factored, and so the server doesn't need to do any of this work. Is there a reason we only have it for the L-function access and not for the mf_hecke_lpoly ?

Also, in the L-functions page, I saw that the scrolling is inside the table (and not all over the page) - I'll try to do it here as well.

edgarcosta:
The reason is exactly what @assaferan already found out, factoring takes a long time.
The clean solution is to add the factorization to mf_hecke_lpoly, or to only store the factorization.
The hacky solutions is to grab the factorization from lfunc_lfunctions.

assaferan:
How do I add things to mf_hecke_lpoly, then?

AndrewVSutherland:
It probably makes sense for me to add the factorizations -- the mf_hecke_lpoly table is populated from an external file created by an external program (FormatNewformData in https://github.com/JohnCremona/CMFs/blob/master/magma/mfpgdata.m) that generates all of the newform tables from (many gigabytes worth) of raw data files.

It would not be hard for me to update this program to also compute factorizations of the hecke lpolys and include them as a column in mf_hecke_lpoly table, and rerun the table generation script (it only takes about 3 hours running on 64 cores), but this will need to wait until next week.

AndrewVSutherland:
The table mf_hecke_lpolys now contains the column lpoly_factorization, which contains the factorization of the L-polynomial into irreducible factors as [[[list-of-coeffs],exponent],...]. See

https://beta.lmfdb.org/api/mf_hecke_lpolys/

Every factor should have constant coefficient 1, and they should be sorted by degree, then exponent, then by coefficients (lex order).

assaferan:
OK - the code now retrieves the factored polynomials and works much faster. Do you think it's good to go?

AndrewVSutherland:
Looks good!

PR #3702, opened 2020-02-10
Fix some stats problems

roed314:
Fix a few issues with updating stats pages when new data exists.


AndrewVSutherland:
This looks good, I will merge this as soon as Travis finishes.

PR #3705, opened 2020-02-10
using bigints in mw gens

edgarcosta:
Compare
http://localhost:37777/Genus2Curve/Q/900617/a/900617/1
vs
https://beta.lmfdb.org/Genus2Curve/Q/900617/a/900617/1


AndrewVSutherland:
Looks good, I think this is fine to merge once Travis finishes.

Issue #3666, opened 2020-01-24
knowl popups on cocalc don't show up


apeng16:
When clicking the Knowl it looks like we are trying to render 
https://cocalc.com/knowledge/render/lf.defining_polynomial?

instead of perhaps the right url, something like

https://cocalc.com/32aa118f-085c-49f7-ac86-b575c334f469/server/37777/knowledge/render/lf.defining_polynomial?

roed314:
This was fixed by #3711.

PR #3711, opened 2020-02-14
Cocalc knowl

chrisxu3:
This was an attempt to fix the issues running LMFDB on CoCalc; namely, clicking on knowls didn't work (see Issue #3666 ).

Additionally, I fixed some issues in universe.html where <script src="<url>"> had <url> with respect to the base domain. I've switched <url> to url_for("static",<file>) to fix this.


lmfdb/classical_modular_forms/templates/cmf_browse.html
```diff
@@ -41,7 +41,7 @@ <h2> Browse {{ KNOWL('cmf.newform','newforms')}}</h2>
 {% endfor %}
 
 <p>
-  A <a href="?search_type=Dimensions&dim=1">table</a> of newforms with rational {{ KNOWL('cmf.q-expansion', title='\(q\)-expansion') }}
+  A <a href="?search_type=Dimensions&dim=1">table</a> of newforms with rational {{title}} {{ KNOWL('cmf.q-expansion', title='\(q\)-expansion') }}
```
alexjbest: Was this intentional? It looks like maybe it was something I did when experimenting rather than an actual change?

lmfdb/templates/knowl-defs.html
```diff
@@ -39,18 +40,19 @@
  {% with hover = "%s [%s]"|format(ktitle, kid) -%}
   {% if title is none -%}
     {% if ktitle|length > 0 -%}
-     <a title="{{ hover }}" knowl="{{ kid }}" kwargs="{{ kwargs|urlencode }}">{{ ktitle|safe }}</a>
+     <a title="{{ hover }}" prefix="{{ kurlprefix }}" knowl="{{ kid }}" kwargs="{{ kwargs|urlencode }}">{{ ktitle|safe }}</a>
```
alexjbest: I wonder if kurlprefix could just be a javascript variable set in the header of each page, that is accessed by lmfdb.js, that way the prefix won't be repeated for every knowl?

alexjbest:
I attempted to make the changes I suggested at alexjbest/lmfdb@3921208c3b3cddda8261774538cf239738699d7a

edgarcosta:
We need to fix this test:
```
self.check_args("/Variety/Abelian/Fq/3/4/e_q_bg", 'not <a title="Simple abelian variety [av.simple]" knowl="av.simple" kwargs="">simple</a>')
```

alexjbest:
@edgarcosta  with my latest version I don't believe we will need to? Each knowl should have the same generated code as before, and only the js has changed.

edgarcosta:
@alexjbest,  I have pushed your changes. sorry, I didn't realize you could not do that.

alexjbest:
Travis failure looks unrelated? I just tried again on cocalc and it has stopped caching the js so I could finally test properly that it looks to be working fine.

PR #3712, opened 2020-02-14
Allow search on rational j-invariant for ecnf

### LINKED ISSUE #3558: Server error when searching ecnf by j-invariant
Searching elliptic curves over number fields by j-invariant without specifying the field produces a server error. (It also gives the message: "Error: 0 is not a valid input for j-invariant. You must specify a field when searching by j-invariant".)

For example, https://www.lmfdb.org/EllipticCurve/?jinv=0.

alexjbest:
And fix a server error for the same pages. Fixes #3558

Compare the pages
https://beta.lmfdb.org/EllipticCurve/?start=50&hst=List&jinv=1728
after this change 
  /EllipticCurve/?start=50&hst=List&jinv=1728
will show a list of 911 curves with j-invariant 1728 defined over various fields, one can then exclude the base changed ones to get a list of all non-rational twists of these curves in the lmfdb.

### DIFF
--- a/lmfdb/ecnf/main.py
+++ b/lmfdb/ecnf/main.py
@@ -17,7 +17,7 @@ from lmfdb.utils import (
     to_dict, flash_error,
     parse_ints, parse_noop, nf_string_to_label, parse_element_of,
     parse_nf_string, parse_nf_elt, parse_bracketed_posints,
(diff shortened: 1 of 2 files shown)


roed314:
Looks good to me.

PR #3713, opened 2020-02-14
add friend for the associated quadratic field of a dirichlet character

### LINKED ISSUE #2769: Dirichlet characters that are Kronecker symbols should link back to the field
For example, http://www.lmfdb.org/Character/Dirichlet/5/4 should link to http://www.lmfdb.org/NumberField/2.2.5.1

alexjbest:
Fixes #2769 see friends of http://127.0.0.1:41017/Character/Dirichlet/8/3 where you can go to the number field to the artin rep and back to the character again!

### DIFF
--- a/lmfdb/characters/web_character.py
+++ b/lmfdb/characters/web_character.py
@@ -1166,6 +1166,13 @@ class WebDBDirichletCharacter(WebChar, WebDBDirichlet):
             )
         if len(self.vflabel) > 0:
             friendlist.append( ("Value Field", '/NumberField/' + self.vflabel) )
+        if self.symbol_numerator():
+            if self.symbol_numerator() > 0:
+                assoclabel = '2.2.%d.1' % self.symbol_numerator()
+            else:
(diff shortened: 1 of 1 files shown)


jwj61:
Why just quadratic?

If we find an Artin rep to match a Dirichlet character, then get it's Artin field and that should be the field cut out by that character.

jwj61:
One other comment -- it looks like it does not check to see if the field is in the database.

alexjbest:
@jwj61 Thanks for the comments, I like the idea of expanding this to more cases, it should be pretty doable. I'll take a look on friday.

As for the field being in the  database, I _think_ that because I only added this to the dirichlet characters which are in the database (modulus <10^5) the corresponding fields all should be too? Maybe this should be expanded though.

AndrewVSutherland:
These fields are all in the database.  While there is certainly more that could be done here (I like the idea of adding the Artin field in general), I don't see any reason not to go ahead an d merge this.

PR #3718, opened 2020-02-17
Use BytesIO instead of StringIO for Python3 downloads

roed314:
Many downloads currently fail in Python3.

### DIFF
--- a/lmfdb/abvar/fq/main.py
+++ b/lmfdb/abvar/fq/main.py
@@ -2,7 +2,7 @@
 
 import ast
 import re
-from six import StringIO, BytesIO, PY3
+from six import BytesIO
 import time
 
 from flask import abort, render_template, url_for, request, redirect, send_file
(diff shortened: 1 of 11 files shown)


roed314:
Actually, looking at classical modular forms I think we might just be able to use the Python 3 solution in Python 2 as well.

edgarcosta:
We should just import `BytesIO` from `six`:
```
This is a fake file object for binary data. 
In Python 2, it’s an alias for StringIO.StringIO, 
but in Python 3, it’s an alias for io.BytesIO
```
https://six.readthedocs.io/#six.BytesIO

and we can always encode it:
https://github.com/LMFDB/lmfdb/blob/master/lmfdb/utils/downloader.py#L143

roed314:
Okay, switched to using `BytesIO` all the time.

AndrewVSutherland:
@roed314 This PR seems to have broken all Maass form downloads due to the change you made at line 227 of lmfdb/modular_forms/maass_forms/maass_waveforms/views/mwf_main.py, where you are trying to execute the encode method of a ditctionary (which does not have such a methid).  Can you please submit a PR to fix this as soon as you can (getting the download working again would be a big help in trying to sort out some of the recent Maass form problems).

edgarcosta:
Checkout #3786

roed314:
I'll take a look in an hour or so.
David

PR #3720, opened 2020-02-18
Remove unnecessary CMF knowls

AndrewVSutherland:
Removes links to two CMF knowls that are potentially confusing and not really needed.  One is the qualifier on the stats page and the other is the "sort order" knowl on the refine search results page.

http://127.0.0.1:37777/ModularForm/GL2/Q/holomorphic/stats
http://127.0.0.1:37777/ModularForm/GL2/Q/holomorphic/?search_type=List&all=1

### DIFF
--- a/lmfdb/classical_modular_forms/main.py
+++ b/lmfdb/classical_modular_forms/main.py
@@ -1141,8 +1141,6 @@ class CMF_stats(StatsDisplay):
     def summary(self):
         return r"The database currently contains %s (Galois orbits of) %s and %s nonzero %s, corresponding to %s modular forms over the complex numbers.  In addition to the statistics below, you can also <a href='%s'>create your own</a>." % (self.nforms, self.newform_knowl, self.nspaces, self.newspace_knowl, self.ndim, url_for(".dynamic_statistics"))
 
-    extent_knowl = 'cmf.statistics_extent'
-
     table = db.mf_newforms
     baseurl_func = ".index"
     buckets = {'level':['1','2-10','11-100','101-1000','1001-2000', '2001-4000','4001-6000','6001-8000','8001-%d'%level_bound()],
--- a/lmfdb/classical_modular_forms/templates/cmf_refine_search.html
+++ b/lmfdb/classical_modular_forms/templates/cmf_refine_search.html
@@ -172,7 +172,7 @@ function set_buckets(col_selecter, bucket_id) {
     <td></td>
     {% endif %}
     {% endfor %}
-    <td>{{ KNOWL('cmf.sort_order', title='Sort order') }}</td>
+    <td>Sort order</td>
   </tr>
   {% else %}
   <tr style="height: 22px;"></tr>


roed314:
I agree that these knowls are a bit awkward, but I do think they serve a purpose.

For the first, we discussed last year not wanting to mislead a user by claiming that the statistics presented were over a representative sample of newforms (thus the language of the `cmf.statistics_extent` knowl).  I'm not particularly happy with our current mechanism for doing so, but I'm also not convinced that we should just give on the effort by deleting the knowl.  I see that John commented that this isn't really a knowl but could be included in the template.  That's certainly true, but we have other places in the LMFDB where we use knowls to make a certain part of a template more easily editable, even if this usage doesn't satisfy the knowl definition of "context free knowledge" (grep for `KNOWL_INC` to see examples, which mostly seem to have to do with extents).  

For the second, I'm not opposed to removing the knowl, but it does let the user know what the full sort order is (since it was unwieldy to include the complete definition of the ordering in the dropdown).

AndrewVSutherland:
@roed314 I am sympathetic to all the points you make.  Indeed, this PR arose from a discussion with John Voight in which I was basically arguing your case, but he persuaded me that we should remove them.  In summary his arguments were (1) in both cases the knowls are potentially confusing (the sort order knowl referred to "displayed columns" but there are displayed columns you cannot sort on, and the statistics knowl referred to a "representative sample" when it is not even clear what a representative sample of an infinite set means), and (2) neither is a knowl in the sense of being "context free knowledge".

Regarding (1), I have just edited both knowls to make them less confusing.  Regarding (2) I agree with you that there are plenty of situations where it makes to use the knowl user interface to display information that is not a "knowl" in the sense of being a context free bit of knowledge.  I think we could alleviate the concerns expressed by John and others if we tagged such knowls in the database so that by default they do not appear when browsing the knowledge database (similar to what we do for top and bottom knowls).

I suggest we give @jvoight a chance to weigh in before deciding how to proceed.  I would be happy to close this PR if he and others are satisfied with the changes I made to the knowls (I can also raise an issue for hiding "knowls" that are not knowls when browsing knowledge).

roed314:
Here are two ideas for tagging knowls to not show up in the knowledge list:
a) Have a convention on the knowl id so that knowls satisfying a certain pattern aren't shown unless you're logged in (e.g. `interface.cmf.stats_extent`).
b) Manually tag knowls using a function from the db interface.

AndrewVSutherland:
@roed314 I think I like (a) better.  It would mean changing the code references to these knowls, but there are not very many of them (and I could amend/replace this PR to do this).

AndrewVSutherland:
I'm going to close this for now.

jvoight:
I'm sorry I didn't chime in earlier, this has been a busy week and this fell down in the queue.  

Let's do (a)!  

I understand we may want to use knowl code when it is convenient (drop down, etc.) even when they are not knowls (in the sense Drew described), but I don't agree that a good criterion is that it is "easily editable"--we're talking about changing a template instead!  

I like Drew's replacement of the phrase "representative sample" with a link to the extent knowl--and extents seem like solid knowls to me.  

But really, I don't want to be a "knowl cop"!  :)  Let's call these 'pseudo-knowls' or something, label them with a different starting prefix, so they'll only show up in that category, and think carefully about how we use them.

PR #3726, opened 2020-02-28
Ec q regulator search

chrisxu3:
Adds regulator search to elliptic curves over Q, resolving #3238
Right now the code fails because 'regulator' is a column in ec_curves_extras, but I'm told David is working on getting it moved to ec_curves.


roed314:
I've moved `regulator` from `ec_curves_extras` to `ec_curves` so this should work now.

lmfdb/elliptic_curves/templates/ec-index.html
```diff
@@ -210,6 +210,15 @@ <h2> Search </h2>
     <td>
     <span class="formexample"> e.g. 2 or 4-15</span>
     </td>
+    <td>
+    {{ KNOWL('ec.q.regulator',title="Regulator") }}
+    </td>
+    <td>
+    <input type='text' name='regulator' example="8.4-9.1" size=10>
```
alexjbest: I think the HTML is a bit funny here, there should be a closing `</td>` before another opening one

alexjbest:
The file `lmfdb/templates/.knowl-defs.html.sage-chat` should  not be commited.

lmfdb/elliptic_curves/templates/ec-index.html
chrisxu3: Fixed, thanks

Issue #996, opened 2016-03-31
Local fields: download


AurelPage:
From /roadmap: Allow download of fields coming from a search.

jwj61:
It looks like buttons were added for doing the download, but they were never connected to code, so now it is a bug.

rbommel:
I am trying to do something about this, but a bit stuck. Will continue next week.

rbommel:
Pull request #3728 hopefully solves this.

PR #3728, opened 2020-03-06
Download pages for local fields

rbommel:
This should enable download pages for local fields. For Sage, some of the functionality is still to be implemented, but according to @roed314 this is in the pipeline. This hopefully solves issue #996.

### DIFF
--- a/lmfdb/local_fields/main.py
+++ b/lmfdb/local_fields/main.py
@@ -10,7 +10,7 @@ from lmfdb.app import app
 from lmfdb.utils import (
     web_latex, coeff_to_poly, pol_to_html, display_multiset,
     parse_galgrp, parse_ints, clean_input, parse_rats, flash_error,
-    search_wrap)
+    search_wrap, Downloader)
 from lmfdb.local_fields import local_fields_page, logger
 from lmfdb.galois_groups.transitive_group import (
     group_display_knowl, group_display_inertia,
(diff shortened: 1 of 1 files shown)


jwj61:
Not merging because of the travis error, but the implementation is different from what I would expect.  I would expect to get at least as much data as is in the search table, and at least with pari/gp, would not expect to have to do anything after loading the file into gp.  It could be a list of lists, each inner list corresponding to a search result.

rbommel:
Travis just timed out. Should probably just be restarted.

I don't know what you mean with your first comment. I seem to get as many fields as results in the queries I did. Do you have some example where you did not get everything?

Concerning pari/gp, I am not an expert at this. I searched a bit in the documentation and was under the impression that there is no implementation for extensions of Qp. Maybe I am wrong. In that case, could you give me a pointer to some command that I should use? That would be very helpful!

jwj61:
My first comment was in regard to the other columns of the search results -- it does not give e, f, c, G, or the slope content.

rbommel:
This has not been done consistently for other download pages of the LMFDB (e.g. for genus 2 curves). We had a small discussion here with @roed314 and @AndrewVSutherland. It does not seem unreasonable to us to put the columns in the download data. In the current implementation, there is no processing done: if the data in the database is stored as a string, for example, then a string will be returned. If this is the desirable outcome, then we can work on implementing this across all download pages of the LMFDB.

AndrewVSutherland:
We discussed this at today's editorial meeting and the conclusion was that download pages should include all the columns in the search results (as a list of lists).  The download file should include a list of column headers so that the meaning of the data is clear.

AndrewVSutherland:
Would it make sense to merge this as is and plan to expand the search results later, or are you planning to update this PR @rbommel ?

jwj61:
Would be ok with me

edgarcosta:
@rbommel is travelling right now

rbommel:
> Would it make sense to merge this as is and plan to expand the search results later, or are you planning to update this PR @rbommel ?

This is fine, indeed. Due to upcoming changes in the search features, I am postponing the other changes.

PR #3730, opened 2020-03-06
Portraits

SamSchiavone:
I added portraits to the properties box of the Belyi maps pages. (Only for hyperbolic maps, though; not spherical or Euclidean.)


SamSchiavone:
Branches got a little mixed up. Making a new request.

PR #3734, opened 2020-03-09
Cmf search boxes

roed314:
Update search boxes code to be able to handle classical modular forms, switch to using it there.


roed314:
Fixing some abvar issues....

roed314:
The Sage 8.8 job finished successfully, the 9.0 job exceeded Travis' time limit.

edgarcosta:
The worrisome part that both tests taking so long!
The Sage 8.8 test took 48min 38s, very close to timing out.
I have restarted the 9.0

roed314:
Strange.  The restarted Sage 9.0 test only took 13 minutes.

edgarcosta:
I will review this

edgarcosta:
The only thing that seems amiss is the refine search for traces, e.g.:
http://localhost:37777/ModularForm/GL2/Q/holomorphic/?hst=Spaces&dim=1-&num_forms=55&search_type=SpaceTraces

roed314:
I fixed the error you noted and finished PRs for all sections of the LMFDB except Maass forms (which I don't plan on doing).  You can see them all merged together at https://olive.lmfdb.xyz/.

AndrewVSutherland:
https://olive.lmfdb.xyz/Genus2Curve/Q/ leads straight to search results, it should go to the front browse page.

edgarcosta:
@AndrewVSutherland there was a typo, it has been fixed

edgarcosta:
I think we are good to merge once the tests pass

AndrewVSutherland:
Why is there a grey background on the column headers?  This isn't in the current UI and makes them harder to read, compare:

https://olive.lmfdb.xyz/EllipticCurve/Q/?search_type=List&all=1
https://www.lmfdb.org/EllipticCurve/Q/?all=1

This PR should not be making changes to the UI other than uniformization.

If we want to make changes to the look and feel in the future, this PR will make it easier to do so uniformly but we should not be making such changes now as part of an infrastructural change.

edgarcosta:
Should `Bad p` for CMFs also match the new standard "include/exclude/exactly/subset of" ?

edgarcosta:
Another minor thing, we need to delete the `e.g.` from `e.g. wt. 1 only` in the browse page

AndrewVSutherland:
I think so, since this gives strictly more functionality (include = superset of, but we also get exclude which is new functionality).

The number field case is a more complicated, since we do want to have the ability to specify both ramified and unramified primes (so we should probably keep subset of / superset of / exactly for ramified primes on number field searches)

edgarcosta:
> Why is there a grey background on the column headers? This isn't in the current UI and makes them harder to read, compare:
> 
> https://olive.lmfdb.xyz/EllipticCurve/Q/?search_type=List&all=1
> https://www.lmfdb.org/EllipticCurve/Q/?all=1
> 
> This PR should not be making changes to the UI other than uniformization.
> 
> If we want to make changes to the look and feel in the future, this PR will make it easier to do so uniformly but we should not be making such changes now as part of an infrastructural change.

I have just changed the color on olive, so we compare it in the same color scheme.
I don't see an different in the background. 
ps: olive is running a couple of prs ahead #3758

AndrewVSutherland:
@edgarcosta If this is just a side effect of the color scheme, that's fine then (but maybe we should consider tweaking the olive color scheme)

edgarcosta:
We should create an issue for trying to fix or a create a couple of color
schemes so they are a bit more usable in these kind of comparisons, but at
the same time, I don't think that is high priority

JohnCremona:
Is olive configured to always run current master branch?

AndrewVSutherland:
@JohnCremona No. It's being used to demo this PR

edgarcosta:
This can be merged after the tests pass.

If so, we should create a ticket regarding the tweaking of Bad p.
But perhaps that makes more sense after PRs have been merged.

roed314:
Yeah, there are various changes to select boxes that I'd like to address, possibly in a follow up ticket.

PR #3736, opened 2020-03-10
Changing citation link

jwj61:
Click on the citation link at the bottom of any page.  The resulting page now has a link to the Google Scholar profile for the LMFDB instead of our hand-curated list of citations.


AndrewVSutherland:
The Travis errors are timeouts, merging.

JohnCremona:
I would also have deleted these static files in ./templates: citations.html (*not* citation.html), citations_content.html, citations_content_bib.html, citations_update.html.

PR #3737, opened 2020-03-11
New artin representation labels

jwj61:
This implements code to deal with the new labels for Artin representations and their orbits.

First, one should upload the postgres tables artin_field_data_new, artin_reps_new, and artin_old2new_labels to the cloud so that whenever this gets pushed to production down the road, the data is there.

To test, search for Artin reps, go to the home page for one, use friends to visit its determinant and Galois orbit (Galois group A5 or cyclic will give examples with conjugate representations).  Then go to the Artin field (from friends), scroll to the bottom to see that the representations are there, and click to go back.

You can go directly to a repn by entering its label on the index page.  This accepts both old and new labels, and converts to new.

The link to L-functions still works, as do friend links from weight 1 modular forms.

That said, I probably missed some things; just let me know what they are.

### DIFF
--- a/lmfdb/artin_representations/__init__.py
+++ b/lmfdb/artin_representations/__init__.py
@@ -25,5 +25,5 @@ register_search_function(
     "artin_representations",
     "Artin representations",
     "Search over Artin representations",
-    auto_search = 'artin_reps'
+    auto_search = 'artin_reps_new'
 )
--- a/lmfdb/artin_representations/main.py
+++ b/lmfdb/artin_representations/main.py
@@ -82,6 +82,9 @@ def parse_artin_label(label, safe=False):
             return ''
     raise ValueError
 
+def both_labels(label):
+    return list(db.artin_old2new_labels.lucky({'$or': [{'old':label}, {'new': label}]}).values())
+
 # Is it a rep'n or an orbit, supporting old and new styles
 def parse_any(label):
     try:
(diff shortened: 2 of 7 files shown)


jwj61:
Looks like a problem with L-functions when using the old labels

edgarcosta:
Pyflakes failed 
```
./lmfdb/artin_representations/main.py:15: 'lmfdb.artin_representations.artin_logger' imported but unused
```
and two tests for L-functions failed.

I think has to do with the L-functions on the DB only having the old label, but I honestly didn't look too deep into it.

jwj61:
Recent commits fixed a problem with bread if the url is entered directly and is for an old label, and should fix L-functions.  In one case, I changed a test which is supposed to check the error message if it is given an invalid label, and changed it to a new invalid label (an old invalid label would give a different error message, but it is not worth preserving that).

edgarcosta:
I think we are trying to support old labels.
However, when I access:
http://localhost:37777/ArtinRepresentation/2.23.3t2.1
I get
`Error: Galois orbit of Artin representations 2.23.3t2.1 is not in database`

jwj61:
Right, I will fix that.

The url you give is for a Galois orbit of Artin representations, which I overlooked earlier.

jwj61:
Was there something not working?  I thought I had fixed the issue which was raised

roed314:
You fixed the issue where the old urls weren't loading, but we found another problem with the friends link.  We're currently talking in https://mit.zoom.us/j/208442043 if you want to join us.

jwj61:
I am in a committee meeting here via zoom

AndrewVSutherland:
@jwj61 The change had to do with the code that is populating the list of friends.  L-function friends (including weight 1 modular forms) are found through the L-functions instances table, and until we covert the Artin labels there (which we don't want to do until the new code is up and running on both beta and production) we need to check for both old and new labels.

PR #3740, opened 2020-03-13
Fix SMF space names on drop down

AndrewVSutherland:
Since the switch to Katex we no longer support latex on drop down lists, this PR changes the SMF space drop down on the main page to use a new column "plain_name" which has been added to the smf_families_table, rather that the "latex_name" column.

To see the difference, compare the results on

https://beta.lmfdb.org/ModularForm/GSp/Q/
https://red.lmfdb.xyz/ModularForm/GSp/Q/

when you click on the space dropdown.

Note: the smf_families table needs to be copied to production before this PR is merged to web, I will take care of this.


AndrewVSutherland:
The update smf_families table has been copied to production.

PR #3742, opened 2020-03-13
Ecnf search boxes

roed314:
Includes changes from #3734 so that should be reviewed first.  Changes some layout for ecnf search inputs and enables searching for a random curve with given attributes.


edgarcosta:
I will review this

roed314:
I'm confused by the travis failures.  They're failures in L-functions of Artin representations, which don't seem to have anything to do with the changes in this ticket.  They only occur in Sage-9.0, but it's repeatable: three different test runs on different branches are showing the same kind of error.  But the the branch at #3734 (which is the intersection of these three branches) is passing!

roed314:
I'm going to restart the build here to see if it changes anything.

roed314:
The error is gone.  I'm copying the one from #3744 for future reference, and then I'll restart travis on the other two tickets.

```
def test_LartinPlot(self):
    L = self.tc.get('/L/Zeros/ArtinRepresentation/2.2e2_17.4t3.2c1/')

...
lmfdb/artin_representations/math_classes.py:550: in good_factor
    return self.from_conjugacy_class_index_to_polynomial(self.nf().from_prime_to_conjugacy_class_index(p))
lmfdb/artin_representations/math_classes.py:868: in from_prime_to_conjugacy_class_index
    return self.from_cycle_type_to_conjugacy_class_index(self.frobenius_cycle_type(p), p)
lmfdb/artin_representations/math_classes.py:865: in from_cycle_type_to_conjugacy_class_index
    return fn_to_use(p)
lmfdb/artin_representations/cyc_alt_res_engine.py:191: in RES_from_cycle_type_to_conjugacy_class_index
    return polynomial_conjugacy_class_matcher(alpha)

    def polynomial_conjugacy_class_matcher(alpha):
        """
        A function that has an internal list of pairs (pol, return_val). Given alpha, finds the pol that it is a root of and returns
        the associated return_val
        """
        for pol_fn, conjugacy_class_index in fn_cc_pairs:
            if pol_fn(alpha) == 0:
                return conjugacy_class_index
>       raise AssertionError("alpha = %s is supposed to be root of one of %s" % (alpha, input))
E       AssertionError: alpha = 265*a + 71 is supposed to be root of one of [{'RootOf': ['-1', '1'], 'ConjugacyClass': 2}, {'RootOf': ['10', '6', '1'], 'ConjugacyClass': 3}]
```

edgarcosta:
@JohnCremona and @AndrewVSutherland, can you look into to refine search layout?
https://purple.lmfdb.xyz/EllipticCurve/?search_type=List&all=1
vs
https://beta.lmfdb.org/EllipticCurve/?all=1

edgarcosta:
@AndrewVSutherland or @JohnCremona, Can I get a yay from you regarding the layout implications of this PR?
https://purple.lmfdb.xyz/EllipticCurve/?search_type=List&all=1
vs
https://beta.lmfdb.org/EllipticCurve/?all=1

AndrewVSutherland:
@edgarcosta LGTM, my only slight concern would be the width of the j-invariant text box, but I don't think this is critical (even in the previous version it was too narrow to fit all inputs).  But I'll defer to @JohnCremona on this.  The alternative would be to move the j-invariant input to a third line and make it twice as wide (which would be wider than the previous version).

I would then feel motivated to find another search input parameter to fill the gap (call me obsessive), and I note there is an obvious candidate -- we support regulator searches for EC/Q and should support them on EC/K!=Q (the column is reg)

JohnCremona:
Looking at it now...

JohnCremona:
The format is fine, I think that searching by j will not be common anyway especially over number fields.

An even more obvious choice for an extra search parameter would be rank, so here's why we do not yet have this:   for the vast majority of curves over K!=Q we have not computed the rank (or the regulator obviously).  I would be happy to include a rank search once we have most ranks, not necessarily all, and that will need to be done carefully since there will be curves (there already are) for which we only have (different!) lower and upper bounds.  So I put off working out how to do that unil there are more.

After looking at some random curves, an easy new search parameter did not occur to me, and I don't mind the gap!

AndrewVSutherland:
@edgarcosta @JohnCremona I agree that j-invariant searches are rare and typing a j-invariant that is longer than the text box works fine, you just can't see it all at once.  I vote we go with the layout as it is on https://purple.lmfdb.xyz/EllipticCurve/?search_type=List&all=1 and if/when we add a new search parameter we can consider making the j-invariant double-wide.

edgarcosta:
Thanks!

PR #3744, opened 2020-03-14
Nf search boxes

roed314:
Includes changes from #3734 so that should be reviewed first. Changes some layout for number field search inputs and enables searching for a random field with given attributes (though some such searches are very slow since the number field table is large).


roed314:
Restarting travis to see if it gets rid of the intermittent bug in euler factors....

edgarcosta:
@jwj61, could you look at the layout and see if you agree?
We have it running here: https://olive.lmfdb.xyz/NumberField/

edgarcosta:
we can merge this, once we get an yay from @jwj61

roed314:
The test failure is the same intermittent one from artin representations that we've been seeing a lot.  It isn't a consequence of the changes in this ticket.

jwj61:
Looks fine.

jwj61:
Any thoughts on the intermittent artin rep failure?  This is still the old code, so the presence of new tables should not affect it.

edgarcosta:
@jwj61, @roed314  and I are honestly confused.
It only happens with Sage 9.0, not all the times, and we can't recreate it locally.
The two L-functions that always fail are:
http://localhost:37777/L/ArtinRepresentation/2.2e3_17.4t3.4c1/
http://localhost:37777/L/ArtinRepresentation/2.2e2_17.4t3.2c1/
and they take quite a while to compute.
While on beta:
https://beta.lmfdb.org/L/ArtinRepresentation/2.2e3_17.4t3.4c1/
https://beta.lmfdb.org/L/ArtinRepresentation/2.2e2_17.4t3.2c1/
they are quick.

Could it be something regarding the dictionary ordering in python3?

roed314:
I've created #3760

PR #3746, opened 2020-03-15
Hgcwa search boxes

roed314:
Includes changes from #3734 so that should be reviewed first. Changes some layout for higher genus curve search inputs and enables searching for a random hgcwa with given attributes


edgarcosta:
@jenpaulhus, can you look at 
- https://olive.lmfdb.xyz/HigherGenus/C/Aut/ 
- https://olive.lmfdb.xyz/HigherGenus/C/Aut/?search_type=List&all=1
and to see if you agree with the new layout?

jenpaulhus:
Looks good to me. I did limited testing and everything seems to work fine. 

This may not be the place to ask this question since it references an older layout change, but why was the search box to change the number of results shown removed from the search results page?  I can well imagine a scenario where someone is curious about, say, genus 3 curves on the hgcwa pages and finds, upon searching, that there are 186 of them. They may then think "Oh, I'd like to see them all on one page so I can sort them in a certain way".   Right now to do that, they would have to go back to the main hgcwa page and enter that 186 on the main page and search again.   

Stay safe out there!

AndrewVSutherland:
@jenpaulhus It was done to save screen real estate and because it creates the potential for confusion when you are paging through results (if I'm looking at 51-100 and I want to see 51-151 I might think I could just change the number of results per page to 100).

jenpaulhus:
Thanks, Drew.

PR #3747, opened 2020-03-15
Dirichlet character search boxes

roed314:
Includes changes from #3734 so that should be reviewed first.  Isn't as thorough as other sections since Dirichlet characters don't use search wrappers.


edgarcosta:
There is an issue here.
The modulus box doesn't display what has been searched:
https://olive.lmfdb.xyz/Character/Dirichlet/?modulus=13&search_type=List

roed314:
Dirichlet characters don't use the same model as most of the rest of the LMFDB, and it's causing a lot of trouble with trying to incorporate the general framework.  I think my change fixes the bug you directly observed, but there are more.  For example, if you search for modulus 13, then add the condition conductor=5, then remove it back to just modulus=13 you still get no results.  I think this is a confusion due to "count" being used both for the number of results and for the max number of results to display.

I'm somewhat tempted to just abandon this PR and leave Dirichlet characters as is.

edgarcosta:
@davidlowryduda, please let us know if you an idea how to address the issues mentioned above.

PR #3749, opened 2020-03-16
Clarify range of Galois image data for CM E/Q and E/K!=Q

AndrewVSutherland:
This PR clarifies the rigor of the mod-p Galois image data in the LMFDB.  For non-CM E/Q we provably know all p for which the mod-p rep is nonsurjective, but in all other cases we have only checked primes p < 1000.  It is in principle possible to rigorously extend these to all p, but this would require a non-trivial amount of (mathematical and computational) work and/or assuming GRH.

With this change on the home pages for E/Q with CM and E/K!=Q the statement  "for all primes p " now reads "for all prime p < 1000", see

http://127.0.0.1:37777/EllipticCurve/Q/27/a/1
http://127.0.0.1:37777/EllipticCurve/2.0.10691.1/1.0.1/a/1

for example.  There is no change on pages for E/Q without CM, see 

http://127.0.0.1:37777/EllipticCurve/Q/11/a/1

for example.

### DIFF
--- a/lmfdb/ecnf/templates/ecnf-curve.html
+++ b/lmfdb/ecnf/templates/ecnf-curve.html
@@ -329,7 +329,7 @@ The mod \( p \) Galois Representation data has not yet been computed for this cu
 <p>
 The mod \( p \) {{KNOWL('ec.galois_rep', title='Galois Representation')}}
 has {{KNOWL('ec.maximal_galois_rep', title='maximal image')}}
-for all primes \( p \)
+for all primes \( p < 1000 \)
 {% if ec.galois_data %} {# there are non-maximal primes #}
 except those listed.
 </p>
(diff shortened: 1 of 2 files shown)


AndrewVSutherland:
Clarification (and edited in my previous comment):

With this change on the home pages for E/Q with CM and E/K!=Q the statement "for all primes p " now reads "for all prime p < 1000"

This applies both to cases where there are exceptional primes listed, and when there are none.  For examples of this see:

http://127.0.0.1:37777/EllipticCurve/Q/288/d/4
http://127.0.0.1:37777/EllipticCurve/2.0.11.1/4068.2/a/1

JohnCremona:
@AndrewVSutherland Am I right that over number fields other than Q, regardless of CM or base-change, we always have the qualifier p<1000?  Also, I thought there would be a revision to the EC/Q reliability knowl, but that knowl is not in beta status so perhaps it was good already?  The corresponding ECNF knowl should probably be expanded on this too: GalReps are not mentioned on the Reliability knowl, they are on the Source knowl but perhaps that sentence should be expanded?

AndrewVSutherland:
@JohnCremona Yes, the qualifier should be there for all elliptic curves (CM or not) overnumber fields K != Q.  The EC/Q reliability knowl has been modified, see the language that I added to  the Galois representation section of

https://beta.lmfdb.org/EllipticCurve/Q/Reliability

Note that you won't see this modification on www.lmfdb.org because the knowl has not been marked as reviewed, but once you mark it as reviewed it will appear there as well.

PR #3751, opened 2020-03-16
Hmf search boxes

roed314:
Includes changes from #3734 so that should be reviewed first. Changes some layout for hilbert modular form search inputs and enables searching for a random form with given attributes.


edgarcosta:
@jvoight, this PR reorganizes the layout of the search boxes on the browse page for HMFs.
Do you agree with this new layout?
Compare https://olive.lmfdb.xyz/ModularForm/GL2/TotallyReal/ vs https://beta.lmfdb.org/ModularForm/GL2/TotallyReal/

edgarcosta:
This is good to merge if we agree on the new layout.

PR #3753, opened 2020-03-16
Artin search boxes

roed314:
Includes changes from #3734 so that should be reviewed first. Enables searching for a random artin rep with given attributes.


AndrewVSutherland:
Happily no merge conflicts, merging.

PR #3754, opened 2020-03-17
Hgm search boxes

roed314:
Includes changes from #3734 so that should be reviewed first. Enlarges search inputs for hypergeometric motives and enables searching for a random family/motive with given attributes.


edgarcosta:
I'm a bit confused about why Hodge vectors show up twice, and why the first one doesn't use the same knowl as the second.
I'm merging, as these are issues that were present before, and hopefully will be fixed in June

PR #3762, opened 2020-03-20
Genus 2 Equation Lookup

AndrewVSutherland:
This is a test PR to see if we can merge this 18 month old branch or if it would be easier to create a new one.


edgarcosta:
This is now running on https://red.lmfdb.xyz/Genus2Curve/Q/

I added some code to handle the inavialability of magma.

PR #3777, opened 2020-03-27
Curves related to Belyi maps

SamSchiavone:
I wrote scripts to check which Belyi maps are defined on curves in the LMFDB and, for those maps that are, added a link to the curve in the Related objects box on the Belyi map page.

I also had to modify names_and_urls in order to get pages for individual curves, not just isogeny classes.


AndrewVSutherland:
pyflakes not happy with a python 2 print (and this print looks like a debug line that should be taken out in any case):

./scripts/belyi/friends/find_friends.py:49:39: invalid syntax
    print "Curve not found in database"

SamSchiavone:
Thanks, Drew. I cleaned up the print statements and (hopefully) fixed all the other complaints pyflakes had.

roed314:
There are still pyflakes failures.

You can run it locally by doing
```
$ sage -pip install pyflakes
$ sage -python -m pyflakes PATH_TO_LMFDB_ROOT
```

SamSchiavone:
Ah sorry, I forgot about the missing imports. It should be fixed now.

scripts/belyi/find_friends.py
```diff
@@ -0,0 +1,104 @@
...
+        E2 = EllipticCurve(ainvs2)
+        if E.is_isomorphic(E2):
+            return r['lmfdb_label']
+    #print("Curve not found in database")
+    return None
+
+def genus1_lookup_equation_nf(rec):
+    assert rec['g'] == 1
+    # make base field
+    nf_matches = list(db.nf_fields.search({'coeffs':rec['base_field']}))
```
roed314: Is `coeffs` always in polredabs form?

SamSchiavone: Yes, we always store the polredabs coefficients.

PR #3779, opened 2020-03-28
Search box fixes

roed314:
This PR includes various components, most related to uniformizing the search inputs across different sections of the LMFDB.  They are viewable on https://teal.lmfdb.xyz

* Many select boxes are now blank by default
* The select boxes for the quantifier on an input box for primes have been standardized
* The input for the number of results to count has been standardized and put in the lower left (see genus 2 curves)
* The jump boxes have been standardized in format, with knowls added to the header for label
* The jump box for hypergeometric motives now appropriately redirects, rather than leaving `?jump=` in the url.
* The select boxes for Galois groups have been rearranged, though the formatting is a bit awkward because of the very long example span for the Group input.
* The abelian varieties jump box now accepts a polynomial


roed314:
There are real failures; I'll try to fix them soon.

roed314:
Fixing some bugs, but one of the problems seems to be revealed by this PR. Compare
* https://beta.lmfdb.org/ModularForm/GL2/Q/holomorphic/?level_primes=7&count=100&search_type=List
* https://teal.lmfdb.xyz/ModularForm/GL2/Q/holomorphic/?level_primes=7&count=100&search_type=List

I think teal is correct; I'm not sure what change caused us to pick up more results but it was something in the handling of the quantifiers for the primes dividing the level.

roed314:
Ok, I understand: we changed the default for that box from exactly to include.  Just pushed fixes; letting travis run.

AndrewVSutherland:
@roed314 Travis seems happy, is this now ready to review?

roed314:
Yep, this is ready to review.

AndrewVSutherland:
I see {{ info.jump }} inside all the jump boxes (tested on brave, chrome, and firefox).

roed314:
Oops.  I changed that at the last minute to accomodate some error handling code in elliptic curves, so didn't notice it in my testing.

I've fixed the problem and pushed the result to https://teal.lmfdb.xyz

AndrewVSutherland:
I think this looks good.  One small suggestion would be to widen the option box for include/exclude/exactly and include/exclude/exactly/subset but maybe 4 or 6 pixels.  Currently (on chrome/ubuntu) the last letter of the selected option is partially cut off.

AndrewVSutherland:
Artin and number field jump box caption begins "Go to" rather than "Find", which is used every where else (I would make them all "Find").

AndrewVSutherland:
One other thing: the number of results to display box should have the default value 50 filled in everywhere (not just as a grayed out example), as on https://www.lmfdb.org/EllipticCurve/Q/, for example.  Can you add this?

edgarcosta:
@AndrewVSutherland I have addressed the count issue

roed314:
Ok, I think we've addressed all of your suggestions.

AndrewVSutherland:
@roed314 Looks good, I'll merge as soon as Travis finishes.

roed314:
The travis failure is the same sporadic artin representation failure that we've been seeing elsewhere, and not due to this PR.

AndrewVSutherland:
Yeah, it passed on Sage 8.8.

PR #3786, opened 2020-03-31
fixing maass form downloads

edgarcosta:
This PR fixes download for Maass forms


AndrewVSutherland:
This works, and given that it only touches one line in Maass forms (which has no useful tests), I am going to merge it now rather than waiting for Travis) -- thanks @edgarcosta

Issue #3780, opened 2020-03-28
Knowl identifier link


roed314:
When editing a knowl, the Identifier link leads to the rendered version of the knowl; it would be better to have it lead back to the show page.

roed314:
@AndrewVSutherland This is an easy fix, but I think it might actually be better not to be a link at all.  If you follow the link and go back to the show page, it won't cancel the edit and will leave you with a lock on that knowl.  And there's already a link to the show page on the edit bar up above.

AndrewVSutherland:
I think I agree, no link is better (the link isn't necessary and just creates the potential for confusion).

PR #3789, opened 2020-03-31
Remove knowl identifier link when editing

roed314:
Addresses #3780.


AndrewVSutherland:
I can't really test this, but the change seems safe and straight-forward, so I will merge it.

Issue #3724, opened 2020-02-28
Search for CM elliptic curves by CM field


jwj61:
From the feedback page:

Given a quadratic imaginary extension k/Q, find all elliptic curves in the database whose endomorphism ring is an order in k. Thank you.

See feedback spreadsheet for e-mail to respond to.

AndrewVSutherland:
This is essentially addressed by #3792 except when the CM field has more than one order of class number one (the fields of discriminant -3, -4, -7).  An easy way to handle this would be add 3 more options to the CM discriminant drop down that will let you search on [-3,-12,-27], [-4,-16], and [-7,-28].

AndrewVSutherland:
Resolkved via #3961

PR #3792, opened 2020-04-02
Search for elliptic curves by CM discriminant

SamSchiavone:
I added the ability to search by CM ~~field~~ discriminant for elliptic curves over QQ and number fields, as requested in https://github.com/LMFDB/lmfdb/issues/3724. Included tests for these search features, too.

I also took out a few extra e.g.s on the HMF pages that arose during the shift to the new search boxes system.


lmfdb/ecnf/main.py
```diff
@@ -751,6 +751,13 @@ def __init__(self):
             name="include_cm",
             label="CM",
             knowl="ec.complex_multiplication")
+        cm_field = TextBox(
+            name="cm_field",
+            label= "%s field %s" % (display_knowl("ec.complex_multiplication", "CM"), display_knowl("nf.discriminant", "discriminant")),
+            example="-4",
+            example_span="-4 or -3,-8",
+            knowl="None"
```
roed314, line 759: You  should be able to just omit this.

lmfdb/ecnf/test_ecnf.py
```diff
@@ -105,6 +105,14 @@ def test_isodeg(self):
         L = self.tc.get('EllipticCurve/?start=0&torsion=1&isodeg=2')
         assert 'No matches' in L.get_data(as_text=True)
 
+    def test_cm_field_search(self):
+        r"""
+        Test that searching for CM field discriminant works
+        """
+        L = self.tc.get('/EllipticCurve/?cm_field=-4')
+        assert '1024.1-c1' in L.get_data(as_text=True)
+        assert '1.0.1-a1' not in L.get_data(as_text=True)
```
roed314, line 114: Better to use `self.check_args('/EllipticCurve/?cm_field=-4', ['1024.1-c1', '1.0.1-a1'])`

lmfdb/elliptic_curves/elliptic_curve.py
```diff
@@ -717,20 +718,30 @@ def __init__(self):
...
+            [num_int_pts, semistable],
+            [count] 
+            ]
 
         self.refine_array = [
             [cond, jinv, rank, torsion, torsion_struct],
             [sha, isodeg, surj_primes, nonsurj_primes, bad_primes],
-            [num_int_pts, regulator, cm, optimal, semistable]]
+            [num_int_pts, regulator, cm, cm_field, optimal],
+            [semistable]
```
roed314, line 746: Maybe switch semistable and optimal for consistency with ecnf?

lmfdb/elliptic_curves/test_ell_curves.py
```diff
@@ -83,6 +83,11 @@ def test_num_int_pts_search(self):
         assert '14.a2' in L.get_data(as_text=True)
         assert not('11.a1' in L.get_data(as_text=True))
 
+    def test_cm_field_search(self):
+        L = self.tc.get('EllipticCurve/Q/?cm_field=-4')
+        assert '32.a3' in L.get_data(as_text=True)
+        assert not('11.a1' in L.get_data(as_text=True))
```
roed314, line 89: As above: use `self.check_args('EllipticCurve/Q/?cm_field=-4', '32.a3')` etc.

lmfdb/ecnf/test_ecnf.py
roed314: Oops, the ability to put a list there depends on a feature in a branch of mine that hasn't been merged yet.  You can still do it in two `check_args` statements.

JohnCremona:
When you have resolved the issues, can you allow me to take a look?  Thanks

lmfdb/ecnf/main.py
SamSchiavone: Sorry, which part can I omit? The knowl="None"?

lmfdb/ecnf/test_ecnf.py
SamSchiavone: Thanks, I didn't know about `check_args` and `not_check_args`!

lmfdb/ecnf/main.py
roed314: Yep.  I'm not sure why you would need to have knowl set to `"None"`....

roed314:
The travis failure is due to a timeout, though tests actually just finished successfully.

This looks good to me, but I'll let @JohnCremona take a look.

JohnCremona:
Sorry for delay, looking now.

JohnCremona:
A couple of minor points:
   1. The search box can handle single values and comma'd lists but not ranges e.g. -4..-11 (or -1..-4).
   2. The knowl on "discriminant" points to a general definition of the discriminant of a number field as a determinant.  This is most unhelpful in this context.  I suggest that the whole of "CM field discriminant" should just be "CM discriminant" and point to the same knwl as just "CM" does now.  Separately, that knowl should also define the CM discriminant (as the discriminant of the Endomorphism ring) and give a list of the 13 CM discriminants over Q, just as it lists the CM j-invariants.  If it is possible to have tables in knowls then that would look best.

I volunteer to edit the knowl at some point unless someone else gets there first, so the only thing I definitely want to see changed before this is merged is my point 2.

edgarcosta:
The column `cm` at the moment stores the discriminant of the endomorphism ring.
Doesn't it make more sense for a beginner to search on the discriminant of the field?
Otherwise, I could just search by j-invariant.

SamSchiavone:
Ah, I didn't realize that the data stores the discriminant of the order (End(E)), not just the discriminant of the field (End(E) \otimes QQ). I've changed the label and knowl as you suggested.

The fact that -11..-4 and -11--4 don't work is a bug with the search parsing. I'll see if I can write a new regular expression that will deal with these.

SamSchiavone:
All right I updated `parse_ints` to use a new regex. Please make sure I didn't mess it up! It seems to handle things like -12..-4 and -12--4 now.

JohnCremona:
It makes more sense to have searching on the cm discriminant not field discriminant since in effect that is searching by the Endomorphism ring itself.

Sorry to not let this go but would it make sense to have a drop down list as there are precisely 13 values?  As we do for torsion structure?  That conveys more information. A beginner might otherwise enter a value not one of the 13 and find no curves but not realise that's because of a theorem.

Edgar said this was equivalent to searching by 13 specific j invariants.  Edgar, I hope you know the j invariant of -163 off by heart...

edgarcosta:
I think the drop-down list is a great idea!

@JohnCremona what kind of expert are you? It is -262537412640768000. I know it by heart, just like [0118 999 881 999 119 7253](https://www.youtube.com/watch?v=ab8GtuPdrUQ)...

AndrewVSutherland:
@edgarcosta And I'm sure you can rattle off the Igusa-Clebsch invariants of the 19 genus 2 curves over Q with CM by a quartic field.  Can you remind me what the one with CM by Q(sqrt(-61+6*sqrt(61)) is again?

I always get this one and the one with CM by Q(sqrt(-53+2*sqrt(53)).confused :)

JohnCremona:
> I think the drop-down list is a great idea!
> 
> @JohnCremona what kind of expert are you? It is -262537412640768000. I know it by heart, just like [0118 999 881 999 119 7253](https://www.youtube.com/watch?v=ab8GtuPdrUQ)...

I only saw that was a youtube link on my desktop this morning :)

SamSchiavone:
All right I changed the CM discriminant search box to be a drop-down for elliptic curves over QQ.

I guess this could be done for curves over a number field (thanks @edgarcosta for pointing out the Sage function cm_j_invariants_and_orders), but we would have to know the number field in advance. I think it makes sense to leave it as a text box for now.

JohnCremona:
Guess who wrote that Sage function :)

JohnCremona:
Travis is reporting a pyflakes error but I see none when I run test locally.

The drop-down list only contains the 9 fundamental discriminants.  There are 4 more: -16, -12, -27, -28.  

Also, in the Search Results page, the dropdown list does not "remember" if a particular discriminant had been selected on the main page.

SamSchiavone:
Thanks for pointing out those two problems. I think I've fixed both of them now. Not sure what is going on with Travis, but I don't see any pyflakes errors either.

edgarcosta:
Travis is happy

lmfdb/elliptic_curves/elliptic_curve.py
```diff
@@ -717,20 +720,40 @@ def __init__(self):
             label="Semistable",
             example="Yes",
             knowl="ec.semistable")
+        cm_opts = []
+        for t in cm_j_invariants_and_orders(QQ):
+            el = (str(t[0]*t[1]**2), t[0]*t[1]**2)
+            if not el in cm_opts:
```
roed314, line 726: As a Python note, you can also write this as `el not in cm_opts` using Python's `not in` operator, which reads a bit more naturally.

lmfdb/elliptic_curves/elliptic_curve.py
```diff
@@ -717,20 +720,40 @@ def __init__(self):
...
+                cm_opts.append(el)
+        cm_opts = sorted(cm_opts, key =lambda el: el[1], reverse = True)
+        cm_opts.insert(0, ("",""))
```
roed314, line 729: You should put this in the array to begin with (where you set `cm_opts=[]`), that way you don't need an extra line to insert it.

lmfdb/elliptic_curves/elliptic_curve.py
```diff
@@ -717,20 +720,40 @@ def __init__(self):
...
+        #cm_opts = [('', ''), (-3, -3), (-4, -4), (-7, -7), (-8, -8), (-11, -11), (-12, -12), (-16, -16), (-19, -19),
+                #(-27, -27), (-28, -28), (-43, -43), (-67, -67), (-163, -163)]
```
roed314, line 731: If you want to leave this in as a comment I would just change it to a the list of CM options (rather than pairs), and include the ones you were missing.

lmfdb/utils/search_parsing.py
```diff
@@ -339,7 +340,7 @@ def parse_rational(inp, query, qfield):
 
 @search_parser(clean_info=True, prep_ranges=True) # see SearchParser.__call__ for actual arguments when calling
 def parse_ints(inp, query, qfield, parse_singleton=int):
-    if LIST_RE.match(inp):
+    if SIGNED_LIST_RE.match(inp):
```
roed314, line 343: You're changing the behavior of `parse_ints` everywhere in the LMFDB.  Now, I agree that the name of `parse_ints` is not ideal (we should perhaps change it to `parse_positive_ints`, and also change the name of the regular expression from `LIST_RE` to `POSITIVE_INTS_RE`.  But you've removed some error checking that was useful.

Besides, you're no longer using the parser in your application.  I would probably just undo your changes to this file and leave it for another day, unless you'd like to change various names to make it clearer what is being done.  That's probably better for another PR anyway.

roed314:
See my comments.

JohnCremona:
My only new comment is that you should just hard wire the 13 values into the code as they will never change. Don't call the Sage function (which for QQ just returns a hard wired list anyway).

More thought will be needed to add this facility to ecnf, where I recommend picking up the list of cm discriminants in the database from the database itself.

JohnCremona:
This seems to have become vastly more complicated than necessary.  Now that we have gone for a drop-down list there is no need for any parsing, so only one file needs to be changed (OK: 2, counting the search again page).  In that file just have a drop down list of 13 presribed values which get put directly into the search query.  That's it.

lmfdb/elliptic_curves/elliptic_curve.py
SamSchiavone: If I start with `cm_opts = [('','')]` then the sorted doesn't work.

SamSchiavone: In any case, I'll just hard code the values and avoid this entirely.

lmfdb/utils/search_parsing.py
SamSchiavone: I still use `parse_ints` in the ecnf version unless I'm mistaken...

SamSchiavone: In any case, I'm happy to revert my changes and make a separate issue.

SamSchiavone:
> My only new comment is that you should just hard wire the 13 values into the code as they will never change. Don't call the Sage function (which for QQ just returns a hard wired list anyway).
> 
> More thought will be needed to add this facility to ecnf, where I recommend picking up the list of cm discriminants in the database from the database itself.

Okay, sounds good. I was going to leave the script there in the hopes that one day it could be used for a drop-down for ecnf. But pulling the values from the database makes sense.

I think there is still parsing in the ecnf version of the CM discriminant search, but I'll just leave it as is and make a separate issue for the search parsing.

JohnCremona:
Thanks Sam.  For future reference, the lines
`from sage.schemes.elliptic_curves.cm import discriminants_with_bounded_class_number`
`disc = dbcn(6)`

define a dictionary with keys h=1,2,3,4,5,6 and values the list of pairs (D,f) such that D*f**2 has class number h.  The numbers are [13, 29, 25, 84, 29, 101] for h up to 6 (total 281):
`[len(discs[h]) for h in range(1,7)]`
So far the nfcurves collection (which only has curves over fields of degrees up to 6) only contains 39 different discriminants out of a possible 281.

lmfdb/utils/search_parsing.py
SamSchiavone: Okay, I've reverted my changes to `parse_ints`. The new issue is here: https://github.com/LMFDB/lmfdb/issues/3825

PR #3794, opened 2020-04-02
Artin tests and Degree n L-function pages

jwj61:
The pages for degree n L-functions, for n=2,3,4 have links to Artin representations and their L-functions.  This updates those links to new labels.

This also updates tests of Artin representations.  I left some with both old and new labels, but mostly adjusted them to new labels.


jwj61:
Travis failures are timeouts.

Do we need to extend how long travis has to run the tests (if possible), or break them into multiple suites which can hopefully each get an hour?  Otherwise, we might have to start removing tests rather than adding them.

edgarcosta:
Before removing tests, we should add indexes, as we are seeing a bunch of slow query warnings, in the travis logs.
I will do that before restaring the tests.

We can also run tests in parallel, in different travis jobs, like we could have a job per directory.
I will try that later on today.

PR #3797, opened 2020-04-03
try both labels

edgarcosta:


AndrewVSutherland:
Looks good

edgarcosta:
I will merge once tests pass.
And then to web and dev.

jwj61:
I haven't studied the logic of this -- I know the routing is complicated.

But, when new representations are added, they won't be in the old/new table.  That was supposed to be just for translating old labels to new ones.  Will that cause a problem when new representations are added?

edgarcosta:
I hope this fixes it. The bug is indeed due to a rep not being in the table.

PR #3799, opened 2020-04-05
Parallelizing travis

edgarcosta:
Travis allows us to run 5 concurrent jobs at the same time.
This PR breaks down the tests into two parts, so we are running 4 jobs, 2 per sage version.

Once we move to sage 9, I will split it further to use 5 concurrent jobs.

To review this, you just need to check that Travis still runs 431 tests and skip 6 per sage version.


edgarcosta:
I'm going to merge this myself.
```
⚡️ wget https://api.travis-ci.org/v3/job/671273964/log.txt
⚡️ wget https://api.travis-ci.org/v3/job/671273965/log.txt
⚡️ grep passed log.txt* | awk -F' ' '{s+=$2} END {printf "%.0f", s}'
431
grep skipped log.txt*
⚡️ grep skipped log.txt* | cut -f2-6 -d ' '
1 skipped in 2.44 seconds
1 skipped in 2.25 seconds
2 skipped in 2.64 seconds
1 skipped in 2.63 seconds
5 passed, 2 skipped in
42 passed, 1 skipped in
```

The 2 extra skipped tests come from:
https://github.com/LMFDB/lmfdb/blob/d2214ba3e52772f9f5cd6a6a8d9cea9b70964e41/.travis.yml#L90

PR #3813, opened 2020-04-13
Changing table names

jwj61:
This removes "new" from the table names for Artin representations.  It restores the names to how they were before we changed the labels.

There are no special pages to check, but you could look at representations for the Galois group F_7, since the tables without _new have more examples in them.

First, the tables artin_data and artin_reps to production.  It currently doesn't reference those tables, but when this code change ultimately gets to prod, it will be ready.


JohnCremona:
Looks good to me.

Issue #3306, opened 2019-09-05
Search number fields by subfield


jwj61:
Add a search box for global number fields where the user can specify a subfield.

The plan is to rewrite the database so that subfields will be stored in two parallel arrays, the fields and their multiplicities.  Users can specify the subfield either by label or by polynomial.

jwj61:
This is held up by what seems to be a problem with rewrite.

JohnCremona:
Good to close -- do we wait until master has been nerged into prod?

PR #3817, opened 2020-04-15
Number fields search by subfield

jwj61:
This addresses issue #3306.  It adds a search box for a subfield when doing a number field search.  When testing, try fields given by labels, polynomials, and nicknames.  It will work with polynomials which are not polredabs'ed.

Bad inputs it will catch are reducible polynomials, labels not in the database, malformed nicknames, and variations which would effectively ask for Q (like Q(zeta_4+)), and general garbage (like banana).


JohnCremona:
Looks good.  It wasn't fooled by x^2-4 and di the right thing for x^2+4.  I did try a variety of nicknames and good/bad labels.  And fruit.

PR #3836, opened 2020-05-14
ECNF stats fix

edgarcosta:
The `_fields_by` function should only return fields with nonzero counts.

For some reason, the label `3.1.972.2`, which we have 0 curves over, made into the counts table, and was breaking :
https://beta.lmfdb.org/EllipticCurve/browse/3/1/
https://beta.lmfdb.org/EllipticCurve/browse/3/


JohnCremona:
That's weird. I'm sure I have never seen an elliptic curve defined over
that field and the database has never had any.

PR #3848, opened 2020-05-28
Artin import script

jwj61:
Update to Artin representation import script to include data on the projective image of a representation.


jwj61:
The hobbes workshop link keeps failing in one test, but every time I then put the url into my browser it pops right up.

PR #3853, opened 2020-05-29
Force refresh of degree 1 lfunction graph

### LINKED ISSUE #3852: Slider for degree-1 L-function browse page broken on Chrome
When using Chrome (or Brave or Chromium) the slider on

https://www.lmfdb.org/L/degree1/

is not working correctly.  Setting it to values above 20 does not change the display, and setting it to values N below 20 truncates the graph in the wrong way (it seems to cut off the L-functions whose conductor is less than 20-N, which is bizarre).

This problem is browser specific, it does not occur on Firefox.

alexjbest:
Fixes #3852 but needs more testing (on different browsers), and possibly this change should be propagated to other graphs with sliders etc. in LMFDB.


alexjbest:
Ok this now looks to work well on firefox safari and chrome, so ready for review.

It doesn't seem there is anywhere else in the lmfdb that uses the same method to update a graph, (maybe something in crystals but this code seems completely unused anyway).

edgarcosta:
Looks good to em

PR #3860, opened 2020-05-30
Artin projective images

jwj61:
This adds display of the projective image of Artin representations, both the group and a representative field, and the ability to search based on the projective image group.

Although this was prompted by the parallel feature for weight 1 modular forms, it makes sense for any Artin representation, so the data is there for all of them.

Reviewing this mainly involves searching for different things under Artin representations.  There is a lot of flexibility in searching for the group, so one can enter things like D5 and A4, but also S9, a gap code, or something in nTj form.

One awkward spot is the drop-down for selecting the projective image type.  The values cover only 2-dimensional representations even though there are classifications for other values of n.


jwj61:
Getting the sage version failed in the tests.

edgarcosta:
It looks like a fluke on MIT side. I have restarted the jobs, if this
persists I will talk to MIT to figure out what is going on.

jwj61:
SSL certificate expired.

edgarcosta:
The root of the problem is this:
https://calnetweb.berkeley.edu/calnet-technologists/incommon-sectigo-certificate-service/addtrust-external-root-expiration-may-2020

I don't understand why Travis would be in such a situation.
I'm trying to see what I can do. Worst case scenario we add
"--no-check-certificate" to the wget.

edgarcosta:
Now the test failures are honest pyflakes failures :)

jwj61:
Back to wget problem

edgarcosta:
I redid my changes again.

AndrewVSutherland:
@jwj61 This looks fantastic, thank you!  And apologies for the delay in reviewing this.

PR #3861, opened 2020-06-03
minor bug in reload_all

edgarcosta:


edgarcosta:
I fixed travis in #3860

Issue #3867, opened 2020-06-05
Server error when saving the test knowl


AndrewVSutherland:
If you try to edit https://beta.lmfdb.org/knowledge/edit/test.text and save your changes it throws a server error due to a Byte/str issue as with many other recently reported issues related to the switch to python3:

Here is the traceback

`Exception on /knowledge/save [POST]
Traceback (most recent call last):
  File "/home/sage/sage-9.1/local/lib/python3.7/site-packages/flask/app.py", line 2447, in wsgi_app
    response = self.full_dispatch_request()
  File "/home/sage/sage-9.1/local/lib/python3.7/site-packages/flask/app.py", line 1952, in full_dispatch_request
    rv = self.handle_user_exception(e)
  File "/home/sage/sage-9.1/local/lib/python3.7/site-packages/flask/app.py", line 1821, in handle_user_exception
    reraise(exc_type, exc_value, tb)
  File "/home/sage/sage-9.1/local/lib/python3.7/site-packages/flask/_compat.py", line 39, in reraise
    raise value
  File "/home/sage/sage-9.1/local/lib/python3.7/site-packages/flask/app.py", line 1950, in full_dispatch_request
    rv = self.dispatch_request()
  File "/home/sage/sage-9.1/local/lib/python3.7/site-packages/flask/app.py", line 1936, in dispatch_request
    return self.view_functions[rule.endpoint](**req.view_args)
  File "/home/sage/sage-9.1/local/lib/python3.7/site-packages/flask_login/utils.py", line 272, in decorated_view
    return func(*args, **kwargs)
  File "/home/lmfdb/lmfdb-git-dev/lmfdb/knowledge/main.py", line 598, in save_form
    k = Knowl(ID, saving=True, renaming=bool(NEWID))
  File "/home/lmfdb/lmfdb-git-dev/lmfdb/knowledge/knowl.py", line 730, in __init__
    self.sed_safety = knowldb.check_sed_safety(ID)
  File "/home/lmfdb/lmfdb-git-dev/lmfdb/knowledge/knowl.py", line 463, in check_sed_safety
    matches = subprocess.check_output(['git', 'grep', """['"]%s['"]"""%(knowlid.replace('.',r'\.'))]).split('\n')
TypeError: a bytes-like object is required, not 'str'
[2020-06-05 22:06:43 UTC] 500 error on URL http://beta.lmfdb.org/knowledge/save ()`

davidfarmer:
[...]

Issue #3866, opened 2020-06-05
More python 3 errors


jwj61:
Trying to restart the server with https://beta.lmfdb.org/restartserver causes an error.  The flask log indicates a byte vs str conflict.

AndrewVSutherland:
I believe the fix is to add encoding='utf8' to the Popen command on line 336 of lmfdb/users/main.py

Issue #3864, opened 2020-06-05
knowls for dirichlet characters are broken on beta


edgarcosta:
https://beta.lmfdb.org/ModularForm/GL2/Q/holomorphic/768/2/f/a/
![Screen Shot 2020-06-04 at 21 25 08](https://user-images.githubusercontent.com/322822/83826429-e03d0280-a6a9-11ea-8101-3cebdf57ea6e.png)

jwj61:
At some point there was a change so that html in dynamic knowls needed to have "| safe" put in.  I just tried to do that with this knowl, but kept getting an error when trying to save it.

jwj61:
Looking at flasklog, this looks like it is caused by the shift to python3 (byte vs str).

AndrewVSutherland:
Once #3867 is fixed we can edit the knowl to fix this.

PR #3869, opened 2020-06-07
Fix python3 bytes not str errors

AndrewVSutherland:
This should address #3866 and #3867 and make it possible to address #3864 by adding encode='utf-8' in several places where it is needed.

To ensure backward compatibility I added an explicit version check.  If someone knows a cleaner way to do this using six, please let me know and I will change it.

### DIFF
--- a/lmfdb/knowledge/knowl.py
+++ b/lmfdb/knowledge/knowl.py
@@ -446,9 +446,9 @@ class KnowlBackend(PostgresBase):
         for kid in kids:
             try:
                 if sys.version_info[0] == 3:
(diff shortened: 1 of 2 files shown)


AndrewVSutherland:
Travis (using the new wget tests @edgarcosta added, so kudos!)  caught a bug that I did not introduce in this PR but appears to be due to a previous attempt to address python3 compatibility.  I just pushed a commit that should fix this.

AndrewVSutherland:
Given that all of the checks have passed I'm going to go ahead and merge this.

edgarcosta:
wget tests?

Issue #3863, opened 2020-06-04
CMF search/browse drop-down qualifiers slightly too narrow


AndrewVSutherland:
On Chrome the drop down boxes for the parity of the weight and character and for the include/exclude bad primes are not quite wide enough, they need another 10-20 pixels to display the entire default option without cutting off the last character ("any parity" would look nicer than "any parit").

tomgrubbmath:
I'm not sure how the workflow with LMFDB usually works but to try and avoid redundant work I thought I'd comment to say I have created a new local branch to address this issue

PR #3872, opened 2020-06-07
Widened CMF Search Quantifier Boxes

tomgrubbmath:
Addresses issue #3863 ; widened drop down quantifier boxes for Classical Modular Forms so that every option would display fully in the box with no cutoff (in addition to the issues raised in 3863, the "Level" box was too narrow for "prime power"). Decided to make all such boxes a uniform width (115 px), as opposed to each having it's own, as it looks nicer to me; happy to revert to each box having it's own optimized width if desired.


roed314:
Looks good to me.  Thanks @tomgrubbmath!

PR #3877, opened 2020-06-17
fixing login issues without the remember

edgarcosta:
If you try to log in without ticking the "remember" box you get a page error.
This patch fixes that.


edgarcosta:
I will merge this once tests pass and then push it to dev.

Issue #2773, opened 2019-03-15
Formatting of central value


edgarcosta:
L(\chi, 1/2) is wrongly formatted
http://www.lmfdb.org/L/Character/Dirichlet/971/870/

tomgrubbmath:
This seems to track back to lmfdb/utils/utilities.py , where the display_complex() function occasionally spits out a number in scientific notation. There is a sloppy way around this by just adding a wrapper function into lmfdb/lfunctions/Lfunctionutilities.py which I have implemented (using string manipulation to replace the letter "e" with the appropriate tex code for "x10^()") but I've held off on a pull request in case there is a cleaner/more systematic way that you'd prefer to fix this. 

(Also, L(\chi, 1 ) is not in dollar signs for math formatting, so that should go in whatever fix is ultimately adapted)

davidfarmer:
I think it is reasonable to do those string manipulations, looking for the "e".

Is the central value unusually small?  The answer involves looking at how many
Dirichlet L-functions have smaller conductor.  If it is unusually small, it might be
worth putting on some list of interesting examples.

edgarcosta:
This L-function is extremely special as its first zero is really close to
zero :)

davidfarmer:
Yes, but my question is whether it is surprisingly close
to zero.  It is a unitary family, so randomly you will
occasionally get zeros close to the critical point.

edgarcosta:
@tomgrubbmath, I think your workaround makes sense

PR #3898, opened 2020-07-07
Improved formatting of special values

tomgrubbmath:
This deals with Issue #2773. The 'display_complex()' function occasionally returned a string in scientific notation, which was not displaying correctly. I dealt with this by 

1.) Writing a function 'scientific_notation_helper' in Lfunctionutilities.py to look for the letter 'e' and replace it with '\\times 10^{}.'
2.) Importing this function into Lfunction_base.py so that it may be called on svt_crit and svt_edge
3.) Modifying templates/Lfunction.html to make sure dollar signs were used when needed. 

The 'before and after' can be seen best on this example: https://www.lmfdb.org/L/Character/Dirichlet/971/870/


lmfdb/lfunctions/Lfunctionutilities.py
```diff
@@ -740,6 +740,34 @@ def specialValueTriple(L, s, sLatex_analytic, sLatex_arithmetic):
...
+            l_beg = lval_string[:efind1]
+            l_mid = lval_string[efind1+1:efind2]
+            l_end = lval_string[efind2+1:]
+            new_real = l_beg+'\\times 10^{'+l_mid[:l_mid.find(' ')]+'}'
+            new_im = l_mid[l_mid.find(' '):] + '\\times 10^{' + l_end[:l_end.find('i')]+'}i'
+            new_l_string = new_real+new_im
+        else:
+            l_beg = lval_string[:efind1]
+            l_end = lval_string[efind1+1:]
+            new_real = l_beg + '\\times 10^{' + l_end[:l_end.find(' ')]+'}'
```
roed314, line 763: Is there always a space at the end?  What if there's something useful after the space?

I worry that this code is too sensitive to the exact format of the input.  What about something like the following?
```python
lval_string = re.sub(r"[Ee](-?\d+)", r" \times 10^{\1}", lval_string)
```

davidfarmer: The regular expression substitution looks good to me, although it is always
necessary to check these things on actual examples.

For the math mode delimiters, I suggest changing them to \( and \),
which is more commonly used in the LMFDB (and perhaps we should
consider only using those and adding that to the coding standards).

tomgrubbmath: Thanks for the comments. @roed314 I had not worked with Python's re library before so I built it specific to the output of display_complex; I think yours looks much cleaner and more robust; I will do some tests as David suggests and then incorporate. @davidfarmer I will switch to the \( and \) standard. 

Sorry for the bad PR.

roed314: It's not a bad PR!  We're just trying to make it better. 😄

tomgrubbmath:
PR updated to incorporate improvements from @davidfarmer and @roed314 ; thanks for your help! I learned something new today.

Changes:

Used David Roe's re.sub() idea to simplify the parsing/replacement function. It needed an extra backslash in the replaced string to properly display the \times character. 
For my own education, it works like this:
re.sub(r"[Ee](-?\d+)",r"\\times10^{\1}", lval_string): r prefix is for string literals. Brackets [Ee] matches an instance of "E" or "e;" parens () form a group later backreferenced by \1 in the replacement ; -? matches a negative sign if it is there, and \d+ matches the full string of digits afterwards. 

The other change replaces dollar signs with parenthetical delimiters following David Farmer's advice.

PR #3899, opened 2020-07-08
Added equivalence class options for HGCWA

jenpaulhus:
This pull request is to add a new feature to HGCWA pages.  On a given refined passport page there may be many distinct generating vectors (up to simultaneous conjugation) but those vectors may be equivalent under a braid action.  Similarly the family pages may display distinct refined passports which actually represent topologically equivalent actions.  Researchers may not need all the distinct generating vectors or refined passports and may prefer to see only representatives of the inequivalent actions.  We have created buttons which allow people to toggle  what they want to see on the page (and also created options for what to download in gap or magma).

For example 
http://localhost:37777/HigherGenus/C/Aut/3.8-2.0.2-2-4-4
There are 12 different refined passports, but only 3 up to topological equivalence

Or
http://localhost:37777/HigherGenus/C/Aut/3.8-3.0.2-2-4-4.1
There are 2 different generating vectors but only 1 up to braiding.

The default is the button which shows the representatives not the full data.

We also created new pages which show, for each topological equivalence, class how all the generating vectors in that topological equivalence class fit into braids. Clicking on the blue "3.8-3.0.2-2-4-4.1.1"  at
http://localhost:37777/HigherGenus/C/Aut/3.8-3.0.2-2-4-4
will bring you to a page that summarizes the unique action up to topological equivalence.
http://localhost:37777/HigherGenus/C/Aut/3.8-3.0.2-2-4-4/T.1.1
showing that there are two braid inequivalent actions which each have two generating vectors in them.

Not all genera have equivalence classes computed in the database.  I have updated the data through genus 7 (and I updated a last one today so the data may need to be updated on the production servers).  I need to be a bit more mathematically clever in finding the equivalence classes for a few examples in genus >7 which will take a bit of work/time  so for genus >7, or quotient genus >0 currently you will just see the old pages.

We added a few tests to make sure the inequivalent actions were showing up.

I also still need to write two knowls explaining what the  braid equivalence and topological equivalence are.  But I wanted to do this pull request now in case there are issues, or concerns about how the data is being presented, phrasing I'm using, etc.


jenpaulhus:
I see this pyflakes error and will fix.  

My local pyflakes did not complain so I'll also look into why locally this error wasn't caught.

lmfdb/higher_genus_w_automorphisms/main.py
```diff
@@ -879,6 +949,65 @@ def render_passport(args):
...
+    downloads = [('Download Magma code', url_for(".hgcwa_code_download",  label=representative, download_type='rep_magma')),
+                      ('Download Gap code', url_for(".hgcwa_code_download", label=representative, download_type='rep_gap'))]
+
+    Lbraid={}
+
+    for element in topo_class:
+       if str(element['braid']) in Lbraid:
+           Lbraid[str(element['braid'])].append([element['passport_label'],
+                                              element['total_label'],
+                                              ' '])
```
roed314, line 987: Why is the third entry a space here?  Is it just not used?

jenpaulhus:
Ah yes.  That was a workaround I came up with for a small display issue I was having.

Consider this page:
http://localhost:37777/HigherGenus/C/Aut/3.8-3.0.2-2-4-4/T.1.1

There are times when one topological equivalence class has multiple inequivalent braid equivalence classes in it, and those braid classes may have several different generating vectors in them. But braids classes preserve conjugacy classes, so the different generating vectors will have the same conjugacy classes. On the page above, without the space, it was printing the "3,3,5,5" and "4,4,5,5" twice, which looked silly.  

I tried messing around on the html side to have it only printed once, but I couldn't find a good solution.  So instead I'm passing in a blank space to print.  

If there is a better way to do this, I will happily change it!

lmfdb/higher_genus_w_automorphisms/main.py
jenpaulhus: Ah yes.  That was a workaround I came up with for a small display issue I was having.

Consider this page:
http://localhost:37777/HigherGenus/C/Aut/3.8-3.0.2-2-4-4/T.1.1

There are times when one topological equivalence class has multiple inequivalent braid equivalence classes in it, and those braid classes may have several different generating vectors in them. But braids classes preserve conjugacy classes, so the different generating vectors will have the same conjugacy classes. On the page above, without the space, it was printing the "3,3,5,5" and "4,4,5,5" twice, which looked silly.  

I tried messing around on the html side to have it only printed once, but I couldn't find a good solution.  So instead I'm passing in a blank space to print.  

If there is a better way to do this, I will happily change it!

jenpaulhus: Here's maybe a better example:
http://localhost:37777/HigherGenus/C/Aut/7.24-14.0.2-2-2-2-2/T.1.1

PR #3910, opened 2020-07-14
Fix links in abvar counts table

roed314:
The links on https://www.lmfdb.org/Variety/Abelian/Fq/?search_type=Counts are currently broken.


edgarcosta:
I know where to fix it

edgarcosta:
Ups, you already fixed it :)

Issue #3911, opened 2020-07-14
api2 search examples return API errors


AndrewVSutherland:
None of the example searches on https://www.lmfdb.org/api2/ are currently working (this is also true on beta).  In particular, the links

https://www.lmfdb.org/api2/description/elliptic_curves_q
https://www.lmfdb.org/api2/inventory/elliptic_curves_q
https://www.lmfdb.org/api2/data/elliptic_curve_q?label=%22100a3%22&_fields=%222adic_index%22

all return API errors.

davidlowryduda:
I have a partial fix in the pull request #3918. On my branch, I fix the first two pages not resolving, and the page http://localhost:37777/api2/data/elliptic_curves_q?label=%22100a3%22.

But the last example on the api2 page, restricting fields, still doesn't work. This is the example

https://www.lmfdb.org/api2/data/elliptic_curve_q?label=%22100a3%22&_fields=%222adic_index%22

The reasons this one doesn't work seem to be deeper problems.

AndrewVSutherland:
closing, since api2 is no longer available

PR #3917, opened 2020-07-17
Api2 fix part 1

davidlowryduda:
This is a partial resolution to the issue in #3911. After this, *almost* all examples on the api2 page now work.

This was caused by a string-to-bytes issue and presumably hasn't worked since we updated to python3.

This also updates the html in the page to make the last example (for returning certain fields) of the correct form, even though this functionality does not yet actually work. I will post more on that in issue #3911.


davidlowryduda:
(I'm closing this since I didn't use a clean branch --- and making a new one)

Issue #3406, opened 2019-10-07
HGM improvements


edgarcosta:
- Replace the 'X' for Ordinary? with a No.
- Display the Hodge polygon (see abelian varieties code)
- the picture needs to be recentered and shown with the right aspect ratio, see:
  - http://beta.lmfdb.org/Motive/Hypergeometric/Q/A4.4_B2.2.2.1
  - http://beta.lmfdb.org/Motive/Hypergeometric/Q/A12.8.2.2_B10.6.3.1.1

tomgrubbmath:
Possibly something changed between when this issue was raised and now, but I'm not sure if the search functionality is broken. Following the two links above takes me to the search page for _individual motives_ with this data, not to the search page for _families_. Currently there are no individual motives with this data in the LMFDB. When searched over families I obtain the correct results: https://beta.lmfdb.org/Motive/Hypergeometric/Q/?famhodge=%5B1%2C5%2C1%5D&search_type=Family 

The search boxes on the HGM Home Page (https://beta.lmfdb.org/Motive/Hypergeometric/Q/) _do_ seem a bit confusing though; it took me a few minutes to figure out how to search for families vs how to search for individual motives. For example; suppose you enter Hodge Family = [1,1] and Root Number = 1 and then hit "Enter" on your keyboard. If your cursor was in the Root Number Box you get the correct search (over individual motives). If your cursor was in the Hodge Family Box you get the following server error: https://beta.lmfdb.org/Motive/Hypergeometric/Q/?famhodge=%5B1%2C1%5D&sign=1&search_type=Family . This could potentially be made clearer to the reader, perhaps by either improving the layout on the homepage, or by forcing them to go to a separate search page for families vs individual motives.

roed314:
I'm going to add some comments here about things to do:
* Knowls
* Distinguish between family and motive search.  Could be a switch between family and motives.  Could have a link like subgroups to one of the two
* Should be able to enter alpha-beta or cyclotomic into find box.

SamSchiavone:
I added knowls: https://github.com/LMFDB/lmfdb/pull/6001

PR #3923, opened 2020-07-21
Improved HGM Circles a little bit

tomgrubbmath:
Deals with part of Issue #3406 . Added a frame around HGM circles to prevent skewing of the aspect ratio, thickened the circle, made the background transparent, and deepened the colors. Two examples below, left is new, right is old. It is easy to change color schemes if anyone has preferences:

image

image

### DIFF
--- a/lmfdb/hypergm/plot.py
+++ b/lmfdb/hypergm/plot.py
@@ -48,8 +48,7 @@ def piecewise_linear_image(A,B):
 def circle_image(A,B):
     G = Graphics()
     G += circle((0,0), 1 , color = 'black',thickness = 3)
-    backgroundColor = (227/255,242/255,253/255) #This is the RGB color (in Sage format) of the properties bar (found in style.css). Making the framing circle this color makes it invisible on the page. 
-    G += circle((0,0), 1.4, color = backgroundColor) #This adds a framing circle to the plot, which protects the aspect ratio from being skewed. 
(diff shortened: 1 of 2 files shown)


lmfdb/hypergm/plot.py
```diff
@@ -47,7 +47,9 @@ def piecewise_linear_image(A,B):
     
 def circle_image(A,B):
     G = Graphics()
-    G += circle((0,0), 1 , color = 'grey')
+    G += circle((0,0), 1 , color = 'black',thickness = 3)
+    backgroundColor = (227/255,242/255,253/255) #This is the RGB color (in Sage format) of the properties bar (found in style.css). Making the framing circle this color makes it invisible on the page.
```
edgarcosta, line 51: `(227/255,242/255,253/255, 0)` might work better and work independently of the current theme

edgarcosta:
Overall this is a significant improvement and I'm happy to merge it.

lmfdb/hypergm/plot.py
tomgrubbmath: Good idea; I'll edit it in and test it out to make sure.

tomgrubbmath: Turns out there is a builtin transparency parameter ("alpha") for graphics that I missed. The frame should now be invisible, regardless of background theme.

edgarcosta:
Once the tests pass I will merge

Issue #3782, opened 2020-03-30
Mass Spectral parameter range input error


AndrewVSutherland:
If you enter 27.3..27.4 or 27.3-27.4 (indicating the interval [27.3,27.4]) into the Spectral parameter range input on https://www.lmfdb.org/ModularForm/GL2/Q/Maass/, the search works as expected.  but if you use more precision, eg. 27.32..27.34 or 27.32-27.34 it will display an error indicating that this is a invalid input (but it is a valid input).

davidfarmer:
Fixed in blue.  Close after Drew merges.

AndrewVSutherland:
Fixed via #3929

Issue #3781, opened 2020-03-30
Maass form server error and wrong data


jwj61:
From the feedback page:

trying to see data for the Maass Form:
level 1
weight: 0
character: 1 (trivial)
Spectral parameter: 27.333

Only 13 Fourier coefficients are shown, although the database claims to know 5000

Also, when going to the URL above, a server error is reported (no details given).
This server error also happens for smaller values of R, where 5000 coefficients are shown
https://www.lmfdb.org/ModularForm/GL2/Q/Maass/4cb8503a58bca9145800000a

See feedback spreadsheet lines 173 and 174 for whom to respond to.

-------------------------------------------------------
Here is another page reported by the same person through the feedback page:

The problem I reported yesterday is related, but different.  

We are looking at Maass Forms
Level: 1
Spectral Parameter Range 20..30
Weight: 0
Character: 1 (trivial)

The list off Fourier coefficients for the spectral value R = 27.333 starts with c[0]=1.  This doesn't make sense for a cusp form.  I think that the index of the coefficient must be shifted by 1, i.e., c[1]=1, c[2]=-0.209 etc.    

At least, with the shifted values, the Fourier series seems to be invariant under zz -> -1/z near i.  Other values cannot be checked because not enough coefficients are given.

The spectral values R=28.531 and 28.863 have the same problem.  These three eigenfunctions are even.  Even eigenfunctions above 30 look okay again.  (Odd eigenfunctions above 30 seem to be missing.  That's okay, as no claim of completeness is made.)

davidfarmer:
Seems to be the same as issue #2647 , which has been closed.

davidfarmer:
Fixed in blue.  Close after merging.

AndrewVSutherland:
Fixed via #3929 and now live on www.lmfdb.org.

Issue #1836, opened 2016-07-19
Add tests to Maass forms


AndrewVSutherland:
The test script https://github.com/LMFDB/lmfdb/blob/master/lmfdb/modular_forms/maass_forms/maass_waveforms/test_mwf.py for the Maass forms code has not been updated in 3 years and has only 2 tests in it.

Tests should be added that exercise all the functionality currently available at http://www.lmfdb.org/ModularForm/GL2/Q/Maass/

AndrewVSutherland:
This will be addressed when we add more maass forms (and uniformize the data formats), removing the 1.1. milesone

AndrewVSutherland:
Done in #3929.

PR #3929, opened 2020-07-23
New Maass forms data and UI

AndrewVSutherland:
This PR represents a re-implementation of the Maass form UI that uses a new table maass_newforms that contains 14995 Maass newforms of weight 0 and trivial character that has been sanity checked by David Farmer for correctness -- in particular this corrects several problems with coefficients being misaligned, old forms that were incorrectly included, or data that was missing or incorrect.

The UI has been re-implemented using the common search and download tools used elsewhere in the LMFDB to make the code easier to maintain and provide functionality like random searches that was previously unavailable.

This PR has already received some preliminary review by David Farmer and John Voight, but I would invite everyone to take a look to see if there is anything I might have missed (including checking the L-functions pages and degree 2 L-function browse pages -- all previous references to code in lmfdb/modular_forms/maass/... should now point to code in lmfdb/maass_forms/.).

The easiest way to do so is on https://blue.lmfdb.xyz/ModularForm/GL2/Q/Maass/

There is work in progress to substantially expand the database of Maass forms in a rigorous way and to compute the data to higher precision; we expect to eventually replace the current database with a rigorously computed data.  This PR will make that upgrade easier, as well as allowing us to close a several open issues related to Maass forms including #1836, #3781, and #3782.


AndrewVSutherland:
The Travis failures are timeouts but there is one actual failure I found by running the tests locally, I will fix this shortly.

PR #3950, opened 2020-07-31
Update main.py

kedlaya:
Correct spelling in "Labeling convention"


edgarcosta:
I will merge once tests pass.

JohnCremona:
I would write "labelling"...

kedlaya:
> I would write "labelling"...

Do we have a style sheet for spelling/typography? I just used an Oxford comma somewhere and wasn't sure whether we had a global convention.

Anyway, I'm sure you wouldn't write "labeleling", which is what we had before.

Issue #1912, opened 2016-08-14
Magma code on Dirichlet character pages


AndrewVSutherland:
This is listed in the roadmap and was recently requested at an LMFDB presentation made at the UConn number theory summer school on elliptic curves and modular forms.

AndrewVSutherland:
We should add code snippets for pari, but Magma doesn't have this functionality built-in.  Snippet might need to be long or could be a download.

davidlowryduda:
We currently have pari and sage code on these pages. There is a parallel between the magma code and the sage code.

The sage snippet relies on Bober's Conrey package, and we just assume that the user has this. From what I understand, Drew has a short magma package (or perhaps really a function) that could do this too in magma. One option would be to make that available.

In fact, I think we should also include a link to Bober's package in the comment saying it's a nonstandard package.

I can imagine another possibility. We could probably generate code on the fly using Bober's code and then translate this into sage code, and do something similar for magma code.

jvoight:
The best solution to this is to have Magma include our code for working with Dirichlet characters, including Conrey labels, minimality, etc.  Just wrote to John Cannon to see if this can be included in the near future.  Otherwise, we can try to be creative about code snippets, but I'm not sure it will be very appealing.

PR #3962, opened 2020-07-31
Add sage code for dirchars without conrey package

davidlowryduda:
This commit adds code for computing Dirichlet characters without using
the extra package conrey_dirichlet_characters from Bober's github. In
addition, this now also links to that package from the Dirichlet
character pages.

To see this code, go to a page such as

https://www.lmfdb.org/Character/Dirichlet/60/7

and click on the `Show commands for SageMath` option.

Right now, I present both the conrey-code and an independent sage code.
We could choose to only show one if desired.

There is a choice to be made for the code down the rest of the page.
Almost all of the code snippets work for both conrey_characters and for
sage_characters, except for the final couple (where we allow explicit
evaluation of Gauss sums, Jacobi sums, and Kloosterman sums for whatever
reason).

This is related to (but also slightly orthogonal to) #1912

### DIFF
--- a/lmfdb/characters/web_character.py
+++ b/lmfdb/characters/web_character.py
@@ -193,8 +193,6 @@ class WebDirichlet(WebCharObject):
 
     @lazy_attribute
     def generators(self):
-        #import pdb; pdb.set_trace()
-        #assert self.H.gens() is not None
         return self.textuple([str(k) for k in self.H.gens()])
 
     """ for Dirichlet over Z, everything is described using integers """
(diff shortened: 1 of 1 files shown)


AndrewVSutherland:
Given how simple the independent sage code is, I think it would be better to just show that pure sage code.

Including the conrey-code is going to make the code snippets more confusing, with no real benefit (if we wanted to add a mention of Bober's package in the Dirchlet character labels learnmoreabout box I think that would be fine).

davidlowryduda:
I agree. I removed the conrey-code. We still mention Bober's package in the description of the source of the data, and I'm content with that.

roed314:
Looks good, except you're missing an import:
```
sage: H = DirichletGroup(1473)
sage: M = H._module
sage: chi = DirichletCharacter(H, M([0,247]))
...
NameError: name 'DirichletCharacter' is not defined
```

davidlowryduda:
Aha. I was overzealous in removing the previous conrey-code. I've fixed this. Thanks for taking a look.

roed314:
No problem!

PR #3976, opened 2020-08-05
Isodeg

roed314:
This will make the following queries not time out: 
* https://beta.lmfdb.org/EllipticCurve/?isodeg=60-1000
* https://beta.lmfdb.org/EllipticCurve/Q/?isodeg=60-1000

Before merging it into web, we need to copy the ec_curves and ec_nfcurves databases to production.


edgarcosta:
@roed314, can you confirm that you just added a column?

roed314:
I added an `isodeg` column to `ec_curves` and to `ec_nfcurves`; that was the only change that I made.  I also added a `GaloisConjugatesSigns` column to `artin_reps`, but that change was unnecessary and the new column can be dropped (or we can keep it and update that column from `jsonb` to `integer[]` as well).

edgarcosta:
I'm copying the data.

edgarcosta:
done

PR #3982, opened 2020-08-14
Cluster pictures

rbommel:
Displays cluster pictures for genus 2 curves.


alexjbest:
That line is counting the number of test files, and you've added one so you should change the 36s in .travis.yml to 37s I assume?

Also it looks like there are some artifacts from a merge conflict in the files.

lmfdb/genus2_curves/web_g2c.py
```diff
@@ -727,7 +744,11 @@ def make_object(self, curve, endo, tama, ratpts, is_curve):
                 data['two_torsion_field_knowl'] = r"splitting field of \(%s\) with Galois group %s" % (intlist_to_poly(t[1]),group_display_knowl(t[2][0],t[2][1]))
 
             tamalist = [[item['p'],item['tamagawa_number']] for item in tama]
+<<<<<<< HEAD
```
alexjbest, line 747: Oh no!

rbommel:
> That line is counting the number of test files, and you've added one so you should change the 36s in .travis.yml to 37s I assume?

Thanks, I did not realise that I can change that. The file was hidden.

> Also it looks like there are some artifacts from a merge conflict in the files.

Yeah, there was one more merge conflict hidden somewhere. Should be resolved now, hopefully...

lmfdb/cluster_pictures/test_clusterpicture.py
```diff
@@ -0,0 +1,88 @@
+""" Still todo
+# -*- coding: utf-8 -*-
+from lmfdb.tests import LmfdbTest
+
+class NumberFieldTest(LmfdbTest):
```
roed314, line 5: It looks like `test_clusterpicture.py` got copied from number fields and not edited.

lmfdb/cluster_pictures/web_cluster_picture.py
```diff
@@ -0,0 +1,84 @@
...
+    out += '<br>Size: '
+    out += str(wcp.size())
+    out += '<br>Potential toric rank: '
+    out += str(wcp.potential_toric_rank())
+    out += '<br>Potential good reduction curve: '
+    out += str(wcp.potential_good_reduction())
+    out += '<br>Potential good reduction Jacobian: '
+    out += str(wcp.potential_good_jacobian_reduction())
+    out += '</div>'
+    #out += '<div align="right">'
```
roed314, line 37: These commented lines should be removed, or edited if you intend to eventually have cluster picture homepages and want to keep a placeholder.

lmfdb/cluster_pictures/web_cluster_picture.py
```diff
@@ -0,0 +1,84 @@
...
+
+class WebClusterPicture:
+    """
+     Class for retrieving cluster picture information from the database
+    """
+    def __init__(self, label, data=None):
+        self.label = label
+        if data is None:
+            self._data = self._get_dbdata()
+        else:
```
roed314, line 50: Writing a bunch of access methods is okay, but it might be easier to do something like
```python
if data is None:
    data = db.cluster_pictures.lookup(self.label)
self.__dict__.update(data)
```
Then you could just access the attributes directly.

lmfdb/cluster_pictures/web_cluster_picture.py
```diff
@@ -0,0 +1,84 @@
+# -*- coding: utf-8 -*-
+from sage.all import cached_function
+
+from lmfdb import db
+from lmfdb.logger import make_logger
+wcplog = make_logger("WCP")
+
+@cached_function
```
roed314, line 8: Why does this need to be cached?  You're not doing anything expensive here....

lmfdb/genus2_curves/web_g2c.py
```diff
@@ -634,9 +644,16 @@ def by_label(label):
...
             ratpts = {}
-        return WebG2C(curve, endo, tama, ratpts, is_curve=(len(slabel)==4))
-
-    def make_object(self, curve, endo, tama, ratpts, is_curve):
+        clus = []
+        for x in tama:
+            if x['p'] != 2:
+                clusentry = db.cluster_pictures.lucky({"label": x['cluster_label']})
+                #clusimg = clusentry['image']
+                clusthmb = clusentry['thumbnail']
```
roed314, line 652: If the data is missing, this `KeyError` will likely crash the server.  Is `cluster_label` ever not computed?  Should you check to see if `clusentry` is `None`?

roed314:
Also, should this be included for elliptic curves as well, or is it not worth the complexity there?

lmfdb/cluster_pictures/test_clusterpicture.py
rbommel: Yes, a lot has been copied from the number fields case. I was not sure about good tests yet, so I just left it like this. I will think again about some test to do.

lmfdb/cluster_pictures/web_cluster_picture.py
rbommel: Yeah, I think this is why I kept them, because I don't know what the final state will be. But I can also remove them, if you prefer.

lmfdb/cluster_pictures/web_cluster_picture.py
rbommel: This is again done in the same way as for number fields. But if we later decide to compute some light stuff on the spot, this method would be a bit inconvenient, no?

lmfdb/cluster_pictures/web_cluster_picture.py
rbommel: Don't know. This was just copied from the number fields. I thought it was required.

lmfdb/genus2_curves/web_g2c.py
rbommel: It is not supposed to be not computed, but I can't promise that future people will always compute it for new curves. So you seem to suggest that I should throw some other error in case the key does not exist, or in case clusentry is empty. I will see if I can find another place where this is done and copy that approach.

rbommel:
Thanks a lot for all your comments! :+1: 

> Also, should this be included for elliptic curves as well, or is it not worth the complexity there?

Yes, in the future this will happen. The plan is to first get it in for genus 2 curves, and if people are happy with the approach to also do it for elliptic curves (over Q and over number fields).

lmfdb/genus2_curves/web_g2c.py
rbommel: Okay, I modified this now to fit with the strategy used in the rest of web_g2c.py. If the code to find the cluster data fails, it will throw a KeyError with an explanation.

rbommel:
> The cluster picture knowl should probably include a link to the expository paper you guys wrote in June. Also, it would be nice if the labels for the attributes (Depth, Size, Potential toric rank, etc) could have knowls as well.

Knowls have been added and modified. Probably it is not perfect yet, but it should be better.

lmfdb/cluster_pictures/web_cluster_picture.py
roed314: You can easily mix the methods, but it's really just a matter of preference.  The way you've done it is perfectly fine.

PR #3985, opened 2020-08-14
Issue/3534

samuel-rice:
Addresses #3534

The number fields in the BMF and HMF search results should now all be knowls with pretty names where possible.

Note: I've had issues with `field_pretty` on my local build (I'm running in WSL at the moment and sage segfaults somewhere in there). However, replacing that function call with some other text displays the knowl and new text as expected, so this _should_ work in a better environment.

### DIFF
--- a/CONTRIBUTORS.yaml
+++ b/CONTRIBUTORS.yaml
@@ -425,5 +425,6 @@ affil: The University of North Carolina Greensboro
 name: David Yuen
 affil: Lake Forest College
 url: http://www.lakeforest.edu/academics/faculty/yuen/
-
+---
+name: Samuel Rice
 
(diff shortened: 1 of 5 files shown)


edgarcosta:
Looks good to me, merging once he adds himself to contributors

edgarcosta:
This resolves #3534

jwj61:
Unless you know the field is in the database, you should use formatfield from web_number_field.py.  It will try to lookup the polynomial.  If it is in the database, it will make a number field knowl (which calls field_pretty) and otherwise makes a knowl containing the polynomal.

edgarcosta:
We are using the field label here, so the fiel is in the db. 
```
lambda e: nf_display_knowl(e['field_label'], field_pretty(e['field_label']))
```
Should we have used something different?

jwj61:
In that instance, it seems right.  However, I think cross-references to number fields should be done via coefficients.  Then you can have number fields in other parts of the lmfdb without worrying about whether the fields are in the database.  If it is not there at first but gets entered at some point, you don't have to change what you have in the other table -- the cross-reference will work properly in both cases.

PR #3988, opened 2020-08-18
Remove hack of hecke_ring_power_basis

roed314:
Here are some modular forms to check that the change is correct:

* A cyclotomic field: 29.2.d.a
* Cyclotomic field in weight 1: 116.1.j.a
* Dimension 2 cyclotomic: 16.2.e.a, 13.2.e.a
* Dimension 2 non-cyclotomic, non-power basis: 28.2.d.a

### DIFF
--- a/lmfdb/classical_modular_forms/templates/cmf_newform_common.html
+++ b/lmfdb/classical_modular_forms/templates/cmf_newform_common.html
@@ -218,7 +218,7 @@ function show_qexp(qstyle) {
     </p>
   </form>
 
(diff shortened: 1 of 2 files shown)


JohnCremona:
If no-one else has checked this by tomorrow then I will .

JohnCremona:
To be sure about the purpose o this change: it is to keep the behaviour of the website unchanged by doing something in a different way:  specifically the attribute hecke_ring_power_basis is no longer ever set to True when the Hecke ring basis is not the power basis.

PR #4002, opened 2020-08-28
Add modulus to dirchar properties box

### LINKED ISSUE #3995: Dirichlet pages do not list their modulus in their properties or gaga box
The information is contained in the label which is displayed as $\chi_m (n, \cdot)$, but to make things easier for newcomers it would be better if there was an explicit field in the basic properties section and the properties box giving the modulus of the character.
On pages like Character/Dirichlet/5/1

We already list this information in the search results as a separate column so this seems to be a (minor) oversight than a deliberate choice not to include duplicate info.

davidlowryduda:
This fixes #3995 (not the picture part --- but that's being handled
elsewhere).

### DIFF
--- a/lmfdb/characters/templates/Character.html
+++ b/lmfdb/characters/templates/Character.html
@@ -28,6 +28,7 @@
 
 <table>
   <tr> <td colspan='3'> {{ place_code('cond') }} </td> </tr>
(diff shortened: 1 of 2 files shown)


AndrewVSutherland:
The modulus should also be listed under "Basic properties" on the main page (above the conductor, it is arguably the most basic property).

AndrewVSutherland:
Looks good, happy to commit once travis finishes

PR #4003, opened 2020-08-28
Remove block style for a tags in ntdata fixes #4001

### LINKED ISSUE #4001: citation in knowl attached to column heading displayed incorrectly
If you click on the "cluster picture" column heading in the local information table for a genus 2 curve, e.g.

   [ https://beta.lmfdb.org/Genus2Curve/Q/1116/a/214272/1](https://beta.lmfdb.org/Genus2Curve/Q/1116/a/214272/1)

the citation reference to arXiv is formatted incorrectly (unnecessary line breaks around the brackets).  It displays find when rendered in a page where it is not inside a column header, e.g.

   [https://beta.lmfdb.org/knowledge/show/ag.cluster_picture](https://beta.lmfdb.org/knowledge/show/ag.cluster_picture)

so there is some particular interaction between how \cite is being rendered when in a table column heading.

alexjbest:
Some more eyes on different pages to check that nothing looks weird with this change would be good
@edgarcosta can you make a colour branch?

### DIFF
--- a/lmfdb/templates/style.css
+++ b/lmfdb/templates/style.css
@@ -656,6 +656,10 @@ table a[knowl] {
   padding: 0;
 }
 
+table.ntdata>tr>td>a {
+  display: block;
+}
+
 table.chardata a {
   display: inline;
 }
--- a/lmfdb/templates/style.css
+++ b/lmfdb/templates/style.css
@@ -656,10 +656,6 @@ table a[knowl] {
   padding: 0;
 }
 
-table.ntdata a {
-  display: block;
-}
-
 table.chardata a {
   display: inline;
 }
--- a/lmfdb/templates/style.css
+++ b/lmfdb/templates/style.css
@@ -656,7 +656,7 @@ table a[knowl] {
   padding: 0;
 }
 
-table.ntdata>tr>td>a {
+table.ntdata td>a {
   display: block;
 }
 


edgarcosta:
blue.lmfdb.xyz

alexjbest:
https://blue.lmfdb.xyz/ArtinRepresentation/?search_type=List&all=1 looks bad now, so we should re-set the style properly on that page before this is merged.

AndrewVSutherland:
@alexjbest Are you planning to do more on this?

alexjbest:
I plan to fix the above issue by adding a local style to that page somehow, not 100% sure the exact approach I would take yet. I was also waiting to see if anyone would flag any other pages as broken after this page, I tried to do due diligence but of course I cannot be completely sure I didn't subtly change something on some page, as ntdata is used quite often across the lmfdb.

edgarcosta:
We need to do something about this, otherwise we need to accept broken
anchors in knowls inside of tables.

AndrewVSutherland:
@alexjbest OK, I'll let this PR sit for the moment.

alexjbest:
Ok this is probably a better fix, lets check how blue looks and if there are no issues merge in a couple of days.

alexjbest:
blue is updated to the correct branch if anyone wants to take a look, but I'm fairly happy with this now.

PR #4006, opened 2020-09-04
hot fix for knowl search

### LINKED ISSUE #3020: Filter after knowls search
If I search for all the knowls containing "Maass", and then want to filter the
results to see which are still beta, I can't do that because the filter options are
not available.

A slightly related issue is that I can't see a way to find all the knowls
with label beginning "mf.mwf".

This is in Chrome on a Macbook Pro.

edgarcosta:
This does not address #3020, but now using the search box will search on every knowl type.


roed314:
Looks good to me.  Feel free to merge once master->web and master->dev are done.

PR #4011, opened 2020-09-05
Fix issue 4010: use sage int for a big integer

jwj61:
The title pretty much describes it.  Some integer arithmetic used integers which were too large for python.

To test,

http://127.0.0.1:37777/ArtinRepresentation/16.74210483036805646041349463016441.24t2912.a.a
http://beta.lmfdb.org/ArtinRepresentation/16.74210483036805646041349463016441.24t2912.a.a


edgarcosta:
Feel free to merge once tests pass.
Thanks for the fix!

PR #4032, opened 2020-09-14
Interesting objects via knowls

roed314:
Generates lists of interesting objects using top and bottom knowls.

Each section of the LMFDB needs to update its annotations so that the list is populated with interesting objects.  This mainly means ensuring that the list is non-empty, that each knowl has a good title, and that the knowls all have content.

- [x] Classical [newforms](https://red.lmfdb.xyz/ModularForm/GL2/Q/holomorphic/interesting_newforms) (13) and [newspaces](https://red.lmfdb.xyz/ModularForm/GL2/Q/holomorphic/interesting_spaces) (2)
- [x] [Maass forms](https://red.lmfdb.xyz/ModularForm/GL2/Q/Maass/interesting) (7)
- [x] [Hilbert modular forms](https://red.lmfdb.xyz/ModularForm/GL2/TotallyReal/interesting) (4)
- [x] [Bianchi modular forms](https://red.lmfdb.xyz/ModularForm/GL2/ImaginaryQuadratic/interesting) (3)
- Siegel modular forms (skipped)
- [x] [Elliptic curves over Q](https://red.lmfdb.xyz/EllipticCurve/Q/interesting) (74)
- [x] [Elliptic curves over number fields](https://red.lmfdb.xyz/EllipticCurve/interesting) (5)
- [x] [Genus 2 curves](https://red.lmfdb.xyz/Genus2Curve/Q/interesting) (31)
- [ ] [Belyi maps](https://red.lmfdb.xyz/Belyi/interesting) (2)
- [x] [Higher genus families](https://red.lmfdb.xyz/HigherGenus/C/Aut/interesting) (4)
- [x] [Abelian varieties over Fq](https://red.lmfdb.xyz/Variety/Abelian/Fq/interesting) (15)
- [x] [Number fields](https://red.lmfdb.xyz/NumberField/interesting) (21)
- [x] [p-adic fields](https://red.lmfdb.xyz/LocalNumberField/interesting) (5)
- [x] [Dirichlet characters](https://red.lmfdb.xyz/Character/Dirichlet/interesting) (2)
- [x] [Artin representations](https://red.lmfdb.xyz/ArtinRepresentation/interesting) (3)
- [x] [Hypergeometric families](https://red.lmfdb.xyz/Motive/Hypergeometric/Q/interesting_families) (5)
- [x] [Galois groups](https://red.lmfdb.xyz/GaloisGroup/interesting) (7)
[...]

PR #4046, opened 2020-09-21
BMFs: pass KNOWL_ID to template for top/bot knowls

JohnCremona:
A small but necessary change so that BMF home pages can show top and bottom knowls.

So far these three forms have such knowls:
http://localhost:37777/ModularForm/GL2/ImaginaryQuadratic/2.0.7.1/10000.1/b/
http://localhost:37777/ModularForm/GL2/ImaginaryQuadratic/2.0.4.1/16384.1/d/
http://localhost:37777/ModularForm/GL2/ImaginaryQuadratic/2.0.4.1/6561.5/a/


roed314:
This change is included in #4032 already, with `mf.bianchi` rather than `bmf`.

PR #4047, opened 2020-09-22
BMF change display of a_P, and update import script

JohnCremona:
The home page for a Bianchi form now
   - lists the Atkin-Lehner eigenvalues first (otherwise they might not be noticed at the bottom of the page);
   - lists the Hecke eigenvalues a_P for good prime P only (previously the number listed as a_P was not actually the Hecke eigenvalue for bad primes, it was the Fourier coefficient);
   - says how many a_P are on file and defined what they are (with knowl links).

Visit the page af a random BMF to see the difference.

### DIFF
--- a/lmfdb/bianchi_modular_forms/templates/bmf-newform.html
+++ b/lmfdb/bianchi_modular_forms/templates/bmf-newform.html
@@ -39,39 +39,6 @@
 </table>
 </p>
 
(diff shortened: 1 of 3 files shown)


roed314:
I've updated the data from the file in your home directory on legendre.  This PR looks good.

JohnCremona:
Thanks -- I put the file where you could read it for that reason.  I thought there might be a different way to make the switch.  I'll check it  tomorrow, then we can copy the data to the prod db and then close the associated issue .

PR #4049, opened 2020-09-22
Fix hardcoded PostgresTable in reload

roed314:
This change just makes it so that you can immediately use a table after reloading (before this change some features would be broken since we were using the wrong class).


edgarcosta:
LGTM feel free to merge once tests pass

PR #4051, opened 2020-09-23
updated Bianchi scripts

JohnCremona:
This was supposed to be part of #4047, now merged.  It only updates the Bianchi forms upload scripts so can safely be merged -- once I have fixed the one pyflakes error I just noticed.

### DIFF
--- a/scripts/bianchi_modular_forms/import_bianchi_forms.py
+++ b/scripts/bianchi_modular_forms/import_bianchi_forms.py
@@ -2,31 +2,37 @@
 r""" Import newforms to Bianchi database of newforms
 
 Initial version (Warwick June 2017): John Cremona
+Revised for postgres (2020)
 
-The database 'bmfs' has two collections: "dimensions" and "forms".
-The "forms" collection entries have the following fields:
+The LMFDB database has two tables for Bianchi Modular Form data, "bmf_dims" and "bmf_forms".
 
-   - '_id':            (ObjectId): internal mongodb identifier
-   - 'field_label':    (string) label of the number field, e.g. '2.0.4.1'
-   - 'field_disc':     (int) absolute value of discriminant of the number field, e.g. 4
-   - 'field_deg':      (int) degree of the number field, e.g. 4
-   - 'level_label':    (string)  label of the level ideal, e.g. '[1,0,1]'
-   - 'level_norm':     (int)  norm of the level ideal, e.g. 1
-   - 'level_ideal':    (string)  defining data of the level ideal, e.g. [25,25,4*w-3]
-   - 'level_gen':      (string)  generator of the level ideal, e.g. 7+4i
-   - 'label_suffix':   (string) letter code for the newform at fixed level: a, b, c, ...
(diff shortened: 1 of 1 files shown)


JohnCremona:
OK now

JohnCremona:
Hmm, it seems that pyflakes on Travis is fussier than the one on legendre.  To be fixed...

edgarcosta:
Travis always runs the latest pyflakes, which will sometimes be more recent than the on legendre.
I should also note that the pyflakes of sage should mostly be always update, as it is the version from when I built sage,
```
 $ sage -python -m pyflakes -V
2.2.0 Python 3.7.3 on Linux
```
I have updated the system pyflakes, but pip itself doesn't provide an easy way to upgrade all its modules, so unless someone asks, I won't be trying to keep the modules up to date.

roed314:
Looks good.  A few suggestions for the future:

* It's good to get in the habit of using `os.path.join` rather than `"/".join` since it's platform independent and better handles paths that already start or end with a slash.
* When iterating over a file, you can just use `for line in formfile` rather than `for line in formfile.readlines()`.  It's more memory friendly (since it's a generator) and is shorter.
* It might be better to use `db.bmf_forms.search_cols` and `db.bmf_forms.col_type` rather than hard coding them in `COL_NAMES` and `COL_TYPES`; there's then one fewer place these need to be changed if the table is updated.  But perhaps this way has the benefit of reminding you that you need to change the code that actually generates the data.

JohnCremona:
Thanks for the suggestions.  I do use os.path.join sometimes but am lazy when I know that the code will only be run on a linux machine.  On the other hand, pyflakes got me to remove an "import os" which I think was there before precisely for this!

I have been using readlines() for so long, that must be what I learned first.  I'll stop right away!

Lastly, yes, I was planning to switch to non-hard-codedcolumn names and types, and was about to do just that for me new encf impot system -- the only drawback is that makes things only work on legendre and in ethe lmfdb directory, while all my ecnf scripts are in a separate repository.  So it seemed to make sense (for ecnf rather than bianchi) to have scripts in (repository) ecnf-data which output a file exactly in the format needed for update_from_file or copy_from instead of having some scripts there and some in the lmfdb scripts directory.  I think I worked out one way to do this: would this work:
(start sage anywhere, such as $HOME/ecnf-data/sage)
`sage: cd  /scratch/home/jcremona/lmfdb`
`sage: from lmfdb import db`
`sage: cd /scratch/home/jcremona/ecnf-data/sage`
(and then read in files from here, which import stuff from other files here).

JohnCremona:
This works:

def get_db():
    cwd = os.getcwd()
    os.chdir(os.path.join(HOME,'lmfdb'))
    from lmfdb import db
    os.chdir(cwd)
    return db

def get_colummn_names_and_types():
    return get_db().ec_nfcurves.col_type

jwj61:
Instead of changing directories, I think you can just use

sys.path.append(os.path.join(HOME, 'lmfdb'))

Then

from lmfdb import db

should work.

JohnCremona:
Thanks @jwj61, that's a lot simpler.

PR #4058, opened 2020-09-25
Add download links to top of search results

tomgrubbmath:
Addresses issue #3166. Main changes are: 

-Changed download_search_results to "span" instead of form to allow for inline formatting.
-Changed wording of matches/download_search_results for better spacing. 
-Manually styled a h2 header to allow for an inline header.

### DIFF
--- a/lmfdb/templates/download_search_results.html
+++ b/lmfdb/templates/download_search_results.html
@@ -8,9 +8,8 @@
   {% set download_limit = 100000 %}
 {% endif %}
 
-<br>
-<br>
-<div id="download-area">
+
+&nbsp;
 {% if info.number > download_limit %}
   <span id="download-msg">
     There are too many search results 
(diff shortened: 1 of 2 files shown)


roed314:
You need to update some of the tests to account for the change in text in the number of results.  If you click on "Details" to the right of the test failure section, then click on one of the builds with an X by it, then search for "FAILURES" you'll find a list of places that should be fixed.

tomgrubbmath:
@roed314 actually it seems like something I did broke the Sato Tate search page; I have no clue why since the changes I made aren't too drastic. I will have to look more closely at the Sato Tate page to see why this is, but this pr shouldn't be merged as is.

tomgrubbmath:
The issue seems to be that Sato Tate does not support downloading of search results. Should I make a separate matches file just for the ST page?

roed314:
Another possibility is to just use a variable in the matches template to determine whether to include the download template.  This can default to true (if the variable isn't set)

tomgrubbmath:
@roed314 I think your recommendation is a nice way of doing it. By searching through which files call "download_search_results" there are 17 search pages from which you should be able to download; you can identify the pages by either their "title" variable or their "body_class" variable, so I have included this conditional check into the code now. I will rewrite the test code next and hopefully everything will clear now!

tomgrubbmath:
@roed314 Looks like it is good to go now! I have checked searches on every category from the sidebar and it seems to be working as desired.

edgarcosta:
This is live on https://pink.lmfdb.xyz for people to check out

Issue #3975, opened 2020-08-04
Filter for unreferenced knowls


AndrewVSutherland:
It would be handy to have a way to filter for unreferenced knowls when browsing knowls.

roed314:
This was done in #4059.

PR #4059, opened 2020-09-25
Improvements to knowl index

roed314:
This PR addresses #3975 and #3020.  To review, the changes to the knowl index are visible [here](https://pink.lmfdb.xyz/knowledge) and the new orphans page is visible [here](https://pink.lmfdb.xyz/knowledge/orphans).

### DIFF
--- a/lmfdb/knowledge/knowl.py
+++ b/lmfdb/knowledge/knowl.py
@@ -41,7 +41,7 @@ url_from_knowl = [
 ]
 grep_extractor = re.compile(r'(.+?)([:|-])(\d+)([-|:])(.*)')
 # We need to convert knowl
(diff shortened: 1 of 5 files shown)


edgarcosta:
The code changes look good to me

edgarcosta:
There is an issue with the orphans' page, as is is not detecting that some knowls are in the code.
For example, https://pink.lmfdb.xyz/knowledge/show/ag.cluster_picture

roed314:
The code references for cluster picture are found because in [code_references](https://github.com/LMFDB/lmfdb/blob/master/lmfdb/knowledge/knowl.py#L429) we grep for the knowl id (for the orphans page we instead grep for `KNOWL(` and `kid=`).  This isn't really reasonable to do for the 1000 knowls in the database (though maybe we could do a second pass for the ones we think are orphans?).  Any suggestions?

Issue #3102, opened 2019-05-17
Random for the whole database


jvoight:
I really love hitting the random button on the browse/search pages.  Could we have a random button for the whole LMFDB?  :)

edgarcosta:
If you allow me to pick a <class> from the sidebar objects at random, then one could just redirect to <class>/random

AndrewVSutherland:
This would not be hard to implement; we might want to think a bit about how the top-level choice to be distributed (otherwise most random objects will be number fields, or L-functions if we decided to include them (there currently is no L/random link, but we could easily add one).

edgarcosta:
@AndrewVSutherland, that is what I meant, first we pick a sidebar element at random X, e.g. CMFs or G2C, and then use the already builtin route X/random to give a random element.

Where would we place this random button?

jvoight:
Yup, that's exactly what I had in mind.  I guess we'd want to put this on the main page (http://www.lmfdb.org/), maybe in the "Search and Browse" box?  If there's more interest, we could try to make a "I'm Feeling Lucky" button or link more prominent...

PR #4061, opened 2020-09-25
Global random object

SamSchiavone:
Addressing https://github.com/LMFDB/lmfdb/issues/3102

### DIFF
--- a/lmfdb/homepage/index_boxes.yaml
+++ b/lmfdb/homepage/index_boxes.yaml
@@ -22,7 +22,7 @@ title: Search and browse
 image: search2
 content: <p>Search for objects with specific properties,
     or browse categories.</p><p>Browse&#58; <a href="/L/">L-functions,</a> <a href="/ModularForm/GL2/Q/holomorphic/">Modular forms,</a> <a href="/EllipticCurve/Q/">Elliptic curves,</a>
-    <a href=" /NumberField/">Number fields</a>
+    <a href=" /NumberField/">Number fields</a><p><a href="/random">See a random object from the database</a></p>
 control: 0
 links:
   - [ 0,0 ]
--- a/lmfdb/homepage/random.py
+++ b/lmfdb/homepage/random.py
@@ -43,7 +43,21 @@ def random_url():
                     "SatoTateGroup/"
                 ]
     route = routes[randint(0,len(routes)-1)]
-    return route+"random"
+    if route == "ModularForm/GL2/Q/holomorphic/":
+        ind = randint(0,1)
+        if ind == 0:
+            route += "random"
+        else:
+            route += "random_space"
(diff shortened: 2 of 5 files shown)


davidfarmer:
Small comments:

"ModularForm/GSp/Q/" is only on Beta.

Some of those things have L-functions, which also should occasionally appear
as a random object.  It might be reasonable to make the L-function of an object
less likely than the object itself.

jwj61:
David's comment on GSp is right.

How hard is it to set up a route L/random to pull a random L-function from the database?

SamSchiavone:
Thanks for the comments! I removed ModularForm/GSp/Q/ from main list.

Drew said it shouldn't be hard to add L/random so I'll look into it.

SamSchiavone:
All right, I made a L/random and then added the route to the global random button. Let me know if you have any other suggestions!

SamSchiavone:
Now on https://purple.lmfdb.xyz/

davidfarmer:
I like it and probably clicked it more than 50 times.

The link to a random object is not obvious.  A natural place is
under "Introduction", below Universe and Knowledge, but I know
space in the nav bar is at a premium.

roed314:
Yeah, it's pretty fun!

I agree that it would be nice to have in a more visible place.  To be fair, it's probably more visible to new users; I think most of us have come to ignore the homepage since it doesn't really change.

davidfarmer:
Is it possible to get an individual modular form (what the modular forms people
call an "embedding")?  Those would be good, and not just spaces or galois orbits.
(It is possible to get a galois orbit?)

davidfarmer:
I think the "random newform" link always gives an orbit.

edgarcosta:
The webpage of an embedding is a reorganized version with less information than the orbit page, thus I don't think it makes much sense to return those on a random search, we also don't search on embeddings.

davidfarmer:
It would be good to have E/Q and also isogeny classes of E/Q as possible random objects.

davidfarmer:
Yet embeddings are the things that have primitive L-functions associated 
to them.

I understand there is a perspective that does not consider embeddings as
the important objects, and only focuses on orbits.  But not everyone
thinks that way.

edgarcosta:
Try purple.lmfdb.xyz/L/random

jwj61:
I think this can be committed and further refinements hashed out afterwards.

But until that happens, what do people think of increasing the probability of getting an L-function, e.g. 20% of the time?

edgarcosta:
We just need to add "L/" to routes a couple of more times

jwj61:
Right, or have a first call to random for L-functions if you want precise control over the probability.

SamSchiavone:
All right, I added a couple more "L/" to routes. I could change the random_url to take a list of weights if we'd like to have more precise control over the frequency with which each type of object appears.

PR #4071, opened 2020-09-30
Add statistics for Bianchi modular forms

roed314:


AndrewVSutherland:
The links in the 1st, 2nd, and 4th tables all result in search input errors (and its clear from the errors in the first that this really is a bug in how the stats are being setup, not a search error).

Also, I could not find a link to the statistics on the browse page (I had to create a /stats URL manually to test this).  We need to add one

AndrewVSutherland:
If I click in the upper left entry for the second table with count 61, I only see 3 results (the same as for the entry in the first table), if there isn't a corresponding search I think it would be better to not make it a link.

JohnCremona:
I will look at this soon.

JohnCremona:
Thanks for doing this, @roed314 !   It is very good to have this page.  Regarding Drew's comment, he is right -- we do have pages for newform spaces but do not have any searching for these, though one click from the top page takes you to (for example) http://localhost:37777/ModularForm/GL2/ImaginaryQuadratic/gl2dims/2.0.4.1 and I think that what we should be seeing when we click on that 61 is part of what you see at http://localhost:37777/ModularForm/GL2/ImaginaryQuadratic/gl2dims/2.0.3.1 -- but that *is* only 3 spaces, not 61 spaces!  So whatever counting is giving that 61 is not right.

The stats page reveals rather starkly something which I knew very well but which was rather hidden, namely that the data we have consists of 3 almost disjoint parts: (1) my data, for 5 fields, rational newforms only, GL2 only, large range; (2) some additional dimension data from Alexander Rham covering all 9 files *but* giving the SL2-dimension and not the GL2-dimension; (3) some data for dimension 2 newforms provided by Ciaran Schembri, only ove Q(sqrt(-1)).  Both (2) and (3) were done i na couple of days at the 2017 Warwick workshop and never expanded (though they added quite a lot of complications to the code).  There is, I think, a case to be made for only including (1) as long as the data for (2) and (3) is so limited.  For (3), I'm sure that Ciaran (or someone else using his Magma code) could compute more quite easily, but that would most sensibly be done alongside a similar exercise for HMFs (one day...)

roed314:
I've fixed the links by allowing the page that browses cusp space dimensions to restrict the level norm.

JohnCremona:
OK -- I will test some time today.

AndrewVSutherland:
@JohnCremona Did you get a chance to review this or would you like me to?

AndrewVSutherland:
LGTM, I'm going to go ahead and merge this.

Issue #3900, opened 2020-07-09
Add BSD data for EC/NF


JohnCremona:
For elliptic curves over number fields I have mostly completed the job of computing Mordell-Weil rank (or lower and upper bounds), generators, analytic ranks, special L-value and analytic order of Sha.  Currently this is complete for number fields of degree 2, 3, 4 and partially complete in degrees 5 and 6.  I plan to upload the data which I have (which wil involve adding several new columns to the table in the database) and then work on displaying this on individual curve pages and isogeny class pages.

Note that when I say that the data is complete I include the possibility that the lower and upper bounds on the rank are different, in which case the number of generators is equal to the lower bound.  If the analytic rank is the same as the lower bound then we can still compute analytic Sha, otherwise not. For example, here is the rank distribution (by isogeny class) for 2.0.4.1:

0: 13696
1: 22063
2: 4461
3: 17
[0,1]: 18 (all have a.r.=1)
[0,2]: 20 (all have a.r.=0)

So there are 18 classes where the rank must be 1 but we are missing a generator, and 20 for which the rank must be 0 but Magma could not prove it.  For the last set we compute analytic Sha on the assumption that the rank is 0, but for the first set we cannot (though we can guess that Sha is trivial and then deduce the height of the missing generator).

AndrewVSutherland:
@JohnCremona Any update on this?

JohnCremona:
I will be doing the upload early next week.

JohnCremona:
Somewhere @roed314 asked me about searching by rank but I cannot remember where.  This will make sense after I have updated the data here, after which we'll have ranks for the vast majority of curves and rank bounds for all.  We'll also have analytic ranks for: all curves over fields of degree <=4,  most over degree 5 (23 out of 34 fields are complete) but not over degree 6 (where Magma is taking days or weeks per curve).

Usually, lower bound = upper bound = analytic rank (= number of generators stored).  The exceptions are as follows (for degrees up to 5 incomplete in degree 5):

[lb,ub]  ar  ngens
[0,1]     1      0   (so the rank is expected to be 1 but we have no generator)
[0,2]     0      0   (so the rank is expected to be 0)
[0,3]      0     0   (" " ")  only in degree 4&5, 5 cases
[1,2]     2      1   (so the rank is expected to be 2 but we have only one generator) 3 cases, in degrees 4&5
[1,3]     1       1 (so the rank is expected to be 1)  4 cases, degrees 4&5

[...]

PR #4074, opened 2020-10-02
Add statistics page for ECNF, search by rank

roed314:


AndrewVSutherland:
The rank table looks good but the links in the torsion structure search table throw errors.

JohnCremona:
As of next week it will no longer be true to say that most ranks are missing.   It will still be true that for quite a few curves we will only have rank bounds, so I hope that the search facility on rank will explain how that is handled.  E. G. We mave bounds 0 and 2 with analytic rank 2.

roed314:
That's great!  Currently the search facility on rank only takes the "rank" column into account.  What behavior would you like to see?  Have curves show up if input for the rank box overlaps the range on their rank bounds?

Also, any thoughts on what we should show on the statistics?

JohnCremona:
I made some relevant comments at #3900  .  In particular, there will be a case for the rank search using only the analytic rank column since in all cases where the rank bounds are the same, the analytic rank is equal to the rank (you would not trust my data if that were not true!) but more importantly, when the lower bound is less than the upper bound, the expected value of the rank is the same as the analytic rank.

However I really think that progress on this issue should wait a few days as I am in the process of adding (and filling columns).

roed314:
Sounds good; I'll wait until you finish updating the data.

Issue #3683, opened 2020-01-28
Geometric endomorphism ring of elliptic curves


jvoight:
The knowl (ec.endomorphism_ring) defines the endomorphism ring of an elliptic curve over a number field to be the ring of endomorphisms defined over the algebraic closure.  It also says this is a special case of the endomorphism ring of an abelian variety, where the knowl (ag.endomorphism_ring) does not make it clear over what field the endomorphisms are defined over.  For genus 2 Jacobians (g2c.jac_endomorphisms), a distinction is made between the endomorphism ring (and algebra) and the *geometric* endomorphism ring.  

We need to be consistent, as this can and does frequently cause confusion.

The first minimal change I propose is to replace "endomorphism ring" by "geometric endomorphism ring" or if that's too long "QQbar-endomorphism ring" (which is slightly ambiguous, but easily clarified by looking at the knowl, and consistent with a choice made already for genus 2 (g2c.geom_end_alg).  
[It's much too common for folks to use the phrase "CM" to mean "geometric endomorphism algebra a CM field", so I do not think we should redefine this.  (In private, when necessary I find it helpful to distinguish between CM and potential CM, or PCM, but it seems unlikely that this will catch on.)]

Second, I propose that the knowl ag.endomorphism_ring makes it clear that these are defined over the base field of the abelian variety.  This is then consistent with what is done for genus 2 Jacobians and abelian varieties over finite fields.  

After that, it would be nice if for elliptic curves over number fields we listed both, which means just flagging those elliptic curves with CM by K by if K is a subfield of the field of definition of E (something we surely do not want to do on the fly).  We probably also want to search on this, with the options being "yes", "no", and "potentially" (in the case where the CM is defined over an extension).  Arguably we should do this for elliptic curves over QQ as well: although it would add some consistency, it will only ever be ZZ, so this would just involve adding a constant text value to the template--probably pointless.  For elliptic curves over number fields, it's still the case that most of the time the endomorphism ring will be ZZ, but not always--as for genus 2 Jacobians.

samuel-rice:
Hello, just commenting as a way to attach myself to this issue. Is the suggestion that we would prefer to search by D directly, or would we still be exposing the options as "yes"/"no"/"potentially"?

JohnCremona:
[...]

Issue #486, opened 2015-10-28
Consistency of math mode in property boxes


swisherh:
Someone needs to make sure we have consistency with math mode in properties boxes.  This should probably be discussed first.

EDIT: Here is a more precise description of the issue.  If you look carefully at the property box on
the page https://www.lmfdb.org/L/EllipticCurve/Q/11/a/ you will see that the degree, conductor and motivic weight display integers in text mode, but the sign is displayed in math mode, resulting in a slightly different font being used.

For a more obvious example see https://www.lmfdb.org/ModularForm/GL2/TotallyReal/2.2.5.1/holomorphic/2.2.5.1-36.1-a where the level and weight are both lists of integers but the weight is displayed in text mode while the level is displayed in math mode.  In this case the level really needs to be in math mode because it typically will not just be a list of integers, see https://www.lmfdb.org/ModularForm/GL2/TotallyReal/2.2.5.1/holomorphic/2.2.5.1-31.1-a for example.

AndrewVSutherland:
@swisherh Can you be more specific?  Math mode is used in property boxes in many places and I don't see any obvious problems.  What are some examples of the inconsistencies you are concerned about?

AndrewVSutherland:
I contacted @swisherh directly and she agreed that we can close this.

AndrewVSutherland:
This issue came up in Kiran's talk at UConn today, where it was pointed out that on the L-function pages one often sees a mixture of math mode and non math mode when displaying integer values in properties boxes.  See http://www.lmfdb.org/L/EllipticCurve/Q/11.a/, for example, where the sign and motivic weight are both equal to 1 but displaced in different fonts.

Elsewhere, the properties boxes for elliptic curves use math mode when displaying integers, but most others (e.g. classical modular forms, number fields, genus 2 curves, Hilbert modular forms, etc...) do not.

On a related note, we are inconsistent about capitalization of yes/no and true/false.

I would propose that numeric attribute values displayed in properties boxes should always be in math mode, and that text strings representing boolean values (yes/no and true/false) be displayed in lower case.

Rather than modifying the code everywhere, I suggest enforcing these conventions in the template that displays the properties box.

swisherh:
Thanks Drew, for reopening this.  I agree that a holistic property box
template solution would be ideal, since even finding where all of the
examples of the issue are can be tricky.

-Holly



On Sun, Aug 14, 2016 at 8:52 AM, Andrew Sutherland <notifications@github.com

[...]

PR #4079, opened 2020-10-02
ECs over Q: Show both endomorphism rings and introduce "potential Complex Multiplication"

samuel-rice:
This works towards finishing off issue #3683 

This builds on and includes the first commit of #4075. This is somewhat redundant information for ECs over Q, but if we like this presentation then I can also do this for ECs over number fields, where the data will be more interesting.

### DIFF
--- a/lmfdb/elliptic_curves/templates/ec-curve.html
+++ b/lmfdb/elliptic_curves/templates/ec-curve.html
@@ -154,15 +154,15 @@ white-space: pre in order to keep line breaks! #}
         <tr>
         <td>{{ KNOWL('ec.endomorphism_ring', title='Endomorphism Ring') }}:</td>
         <td>\(\Z\)</td>
-        <td rowspan=2>&nbsp;</td>
-        <td rowspan=2>({%if data.data.CMD %}potential {%else %}no {%endif %}
-	  {{ KNOWL('ec.complex_multiplication', title='Complex Multiplication') }})
-        </td>
         </tr>
 
         <tr>
         <td>{{ KNOWL('ec.geom_endomorphism_ring', title='Geometric Endomorphism Ring') }}:</td>
         <td>{{ data.data.EndE }}</td>
+        <td>&nbsp;</td>
+        <td>({%if data.data.CMD %}potential {%else %}no {%endif %}
+	  {{ KNOWL('ec.complex_multiplication', title='Complex Multiplication') }})
+        </td>
         </tr>
 
         <tr>
--- a/lmfdb/elliptic_curves/templates/ec-curve.html
+++ b/lmfdb/elliptic_curves/templates/ec-curve.html
@@ -152,7 +152,7 @@ white-space: pre in order to keep line breaks! #}
         </tr>
 
         <tr>
-        <td>{{ KNOWL('ec.endomorphism_ring', title='Endomorphism ring') }}:</td>
+        <td>{{ KNOWL('ec.endomorphism_ring', title='Geometric Endomorphism Ring') }}:</td>
         <td>{{ data.data.EndE }}</td>
         <td>&nbsp;</td>
         <td>({%if not data.data.CMD %}no {%endif %}
--- a/lmfdb/elliptic_curves/templates/ec-curve.html
+++ b/lmfdb/elliptic_curves/templates/ec-curve.html
@@ -152,12 +152,17 @@ white-space: pre in order to keep line breaks! #}
         </tr>
 
         <tr>
-        <td>{{ KNOWL('ec.endomorphism_ring', title='Geometric Endomorphism Ring') }}:</td>
-        <td>{{ data.data.EndE }}</td>
+        <td>{{ KNOWL('ec.endomorphism_ring', title='Endomorphism Ring') }}:</td>
+        <td>{{ \(\Z\) }}</td>
         <td>&nbsp;</td>
         <td>({%if not data.data.CMD %}no {%endif %}
 	  {{ KNOWL('ec.complex_multiplication', title='Complex Multiplication') }})
         </td>
+        </tr>
+
+        <tr>
+        <td>{{ KNOWL('ec.geometric_endomorphism_ring', title='Geometric Endomorphism Ring') }}:</td>
+        <td>{{ data.data.EndE }}</td>
         </tr>
 
         <tr>
--- a/lmfdb/elliptic_curves/templates/ec-curve.html
+++ b/lmfdb/elliptic_curves/templates/ec-curve.html
@@ -154,8 +154,8 @@ white-space: pre in order to keep line breaks! #}
         <tr>
         <td>{{ KNOWL('ec.endomorphism_ring', title='Endomorphism Ring') }}:</td>
         <td>\(\Z\)</td>
-        <td>&nbsp;</td>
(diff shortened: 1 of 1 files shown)


samuel-rice:
I apologise if the multiple PRs are confusing. Probably best to just run with this one.

samuel-rice:
Just pushed some build fixes

JohnCremona:
@samuel-rice It's good practice to give a couple or URLs so that reviewers can see what has changed.  Here there are two types of curves worth looking at:

    - http://localhost:37777/EllipticCurve/Q/11a1/ (no CM, not even potential)
    - http://localhost:37777/EllipticCurve/Q/27a1/ (potential CM)

I suggest that in the second case "(potential CM)" would look better on the second line (the Geometric one) than on the first of the two lines.

Secondly, on the search page and also the search results page, for consistency we should change the label of one search box from "CM" to "potential CM".  (The "CM discriminant" box can stay as it is.)

We also now need a knowl for "potential CM", or perhaps add to the existing knowl for CM the definition of potential CM.  Then, make the entire phrase "potential CM" into the knowl link on both the search box and on the curve's own page.

Lastly -- unless people think that this should be a separate issue, it seems to go naturally in with this change -- the should improce the way CM is indicated on an isogeny class page such as http://localhost:37777/EllipticCurve/Q/27a/.  The point here is that isogenous curves either all have CM or all do not, and the CM field is the same for all but the CM order (and discriminant) is not.  Currently the only menyion of CM on an isogeny lcass page is in the Properties box which shows one discriminant, probably that of curve #1 in the class.  Here is one suggestion for what we can do here:
   1. Add a section to the page with title "Complex Multiplication" which either says "The curve/curves in isogeny class * do not have Complex Multiplcation", or  "The curve/curves in isogeny class *  have Complex Multiplcation by an order / orders in the field *" where the * are filled in appropriately.
   2. In the properties box, replace the single discriminant by the field, e.g. $\Q(\sqrt{-3})$.
   3. In the table of curves, *if* the class is CM but not otherwise, add a column headed "CM discriminant" listing the CM discriminant of each curve.

I'm happy for this last point to be made into a separate issue.

samuel-rice:
@JohnCremona I shall remember to add example pages in future, thanks.

In the cases of "no CM" and "potential CM", it seems these are both statements about the Geometric Endomorphism Ring, whereas "CM" is a property of the Endomorphism Ring, so maybe should stay next to that. Does this seem reasonable?

For relabelling the search box/search results - was it not the plan to change the search box and search results to allow CM to take the values "yes/no/potential" instead? @AndrewVSutherland 

I don't have the ability to edit knowls, but my gut feeling is that it's simpler to explain potential CM in the knowl for CM anyway.

If we could, I'd prefer a separate issue for isogeny class CM, just so I can focus on fewer things at once.

JohnCremona:
I don't agree with your first statement.  In general (for any field of char.0) the endomorphism ring ER and  Geometric endomorphism ring GER determine one of three cases:

   1. Z=ER=GER:  not CM, not potential CM
   2. Z=ER<GER(= imag.quad.order): not CM, potential CM
   3. Z<ER=GER: genuine CM

The 3rd case cannot happen over Q, or over any field with a real embedding, since the ER always embeds into the base field.  So over Q (which is what this PR is about) we only have the first two cases.

Reading your comment again, perhaps that is what you meant and I misunderstood you.  But since the ER will always be Z, and we are adding a comment about whether or not we have potential CM, this depends on whether the GER is or is not Z.

I agree with the second point.  I'll edit the CM knowl so that we define potential CM too.

Fine to make a new issue for the isogeny class stuff.

jvoight:
This looks like it's going well, I think it's a good idea to add potential CM to the knowl.  

Everyone should have the power to edit knowls (that's supposed to be the easiest thing to do!).  You just need an account on beta--David, Edgar, or Drew can set you up. 

I think this PR covers elliptic curves over the rationals.  Don't we now need to do something similar for elliptic curves over number fields?  A handful of cases will now be in JC's 3rd case.

AndrewVSutherland:
@JohnCremona I think the confusion here might be that the most natural (and I think most common) definition of potential CM includes CM.  So in characteristic 0 an elliptic curve has potential CM if and only if its geometric endomorphism ring is larger than Z and it has CM if and only if its endomorphism ring is larger than Z.  I think these are the right definitions (they are, for example, invariant under base change).

JohnCremona:
@AndrewVSutherland I see.  So, rather than the three mutually exclusive cases in my list you would just say "CM iff ER!=Z and pot CM iff GER!=Z".  That is nice and simple.  See my comment on the knowl (which I edited, adding a definition of potential CM -- which you may well now wnat to re-redit).  The comment is basically that however we word this some people will be confused so we had better make sure that our definitions are easy to see.  (e.g. "no elliptic curve defined over Q has CM" is a statement which would take many people by surprise).

@jvoight For examples, search for field=2.0.4.1 and CM discriminant=-4 (or -16 or perhaps some more).

samuel-rice:
That is indeed the definition I was working with. Perhaps the line about complex multiplication ought to be in a merged cell to the right of both?

I was working under the assumption that it'd be simpler to add this for ECNFs once we've agreed on a layout.

samuel-rice:
What do we think of the merged cells in that last commit?

AndrewVSutherland:
@JohnCremona I think we can avoid jarring/confusing people by being judicious about what we say where.  My suggestion would be to avoid saying "no-CM" entirely.  When both End(E) and End(E_Qbar) are Z I'm not sure we need to say anything, since this is the generic case.  When End(E)=Z but End(E_Qbar)!=Z I would put (potential complex multiplication) in the geometric endomorphism ring line, and when End(E)!=Z I would put (complex multiplication) in the endomorphism ring line and not put anything in the geometric endomorphism line.

In the property box I might suggest just showing End(E_Qbar)\otimes Q, as we do for genus 2 curves (or maybe End(E_Qbar) if you want to be more precise),  but I'm happy to defer to you and/or JV on this.

AndrewVSutherland:
Actually, now that I think about it, I would be fine with saying (no complex multiplication) on the geometric endomorphism line in the case End(E) = End(E_Qbar) = Z.  This would then mean that we say something in all three cases, but that something appears on only one of the two lines.

samuel-rice:
I'm getting a lot of segfaults in PARI as I'm working in WSL, to the point where I can't view ECNF pages locally without omitting a lot of information. Would somebody else be able to modify ECNFs while I'm still diagnosing that problem?

JohnCremona:
+1 to the latest suggestions on what to put in brackets for the three cases.

Not sure about the property box, where conciseness (is that a word?) is important.

JohnCremona:
I like the way it is right now (!) so I will merge this.  I mean both the parenthetical info to the right of the geometric endo ring (present in all cases) and also the property box, which says either "no" or "yes (D=-3)" with the discriminant appropriate to the curve.

AndrewVSutherland:
@JohnCremona Thanks!  I notice there are some capitalization issues (e.g. "Complex Multiplication" should be "complex multiplication" and the captions should use sentence case), but I can fix these in a later PR (there are a bunch of issues like this scattered about on various object pages and I was planning to check them all at the same time I address #486 (which is going to involve looking at every object page anyway).

Issue #2755, opened 2019-03-15
Use new statistics framework across the LMFDB


roed314:
Classical modular forms use a new statistics framework, which combines statistics with counts for search results, and makes it easy to create statistics pages.  Statistics pages in other sections of the LMFDB should be rewritten to use the new framework.

The following tables use old statistics methods (grep for `get_oldstat`)

* Higher genus curves with automorphisms
* Hilbert modular forms
* Lattices
* Hecke algebras

The following tables already use the new methods:
* Classical modular forms
* Belyi maps
* Genus 2 curves
* Elliptic curves over Q
* Elliptic curves over number fields
* Number fields

roed314:
The first step is to improve the documentation of the new statistics functionality.  I'll be working on that over the next week.

roed314:
One improvement to the framework that might be good to do with this is to create links to dynamics stats from each entry in the curated stats

JohnCremona:
As far as I can see the new all-purpose stats functionality is not appropriate for some of the uses to which the old auxiliary collections, now names "oldstats", are used for.  In nfcurves and bmfs and hmfs we need to have a record of which base fields exist, per degree and per signature and for each of these subsets (per field, per degree, per signature) have number of curves, number of classes and maximum conductor/level norm.  These are more entries as well, for example in nfcurves we have list of all the torsion structures and that is used to populate the drop-down menu on the search page.  This last one could be done by having a stats entry on torsion which would also count how many times each structure appeared, so that would serve two purposes.

It is very easy to write a function for nfcurves which collects the data needed to be stored in a table just like the oldstats one -- I have such a function -- but as long as oldstats are treated as second-class objects which the new interface does not allow writing to except by raw SQL calls this will be unnecessarily hard to implement.  For that reason the ecnf web pages currently show the wrong number of curves since I recently added 4 but cannot yet update the oldstats table easily.

I look forward to being shown that I am wrong and that the rigid counts/stats system being proposed as one-size-fits-all can handle  such cases too.

JohnCremona:
I am still waiting for some documentation.  It is currently completely opaque how the functions add_stat, add_bucketed_count etc are supposed to be used.  Some of the parameters are not mentioned in the docstring.  Many examples are required.

I also think that for every table whose stats are present and necessary for the correct appearance of the website, there should be a script *in the repository* (and not just on someone's personal computer) which will recompute the relevant stats and counts from scratch.  These would also serve as models.

[...]

PR #4080, opened 2020-10-02
Fix HGCWA Stats

jenpaulhus:
This pull request is to update stats on the HGCWA pages and it takes care of part of #2755.  At the bottom of this explanation I mention two small issues we're having related to the stats data and which I hope someone who understands the stats tables in postgres better than me will immediately be able to tell me how to fix! 

Updated page:
http://localhost:37777/HigherGenus/C/Aut/stats
vs
https://beta.lmfdb.org/HigherGenus/C/Aut/stats

We took this opportunity to add stats for quotient genus >0 also and modified the list of groups per genus page to also reflect topological and braid equivalence up through genus 7 (where that data has been added so far).  
http://localhost:37777/HigherGenus/C/Aut/stats/groups_per_genus/4
vs
https://beta.lmfdb.org/HigherGenus/C/Aut/stats/groups_per_genus/4
(also quotient genus >0 data is added through genus 4 at the bottom of these pages)

The HGCWA data is a unique little snowflake and so we didn't just follow the given stats framework.  In particular, a lot of the stats data is relatively static on these pages, i.e. once a new genus is added to the database, many of its stats are fixed.  

So we created two new tables in the db: hgcwa_unique_groups and hgcwa_complete which store much of the data which is presented the new stats page. The schema for both of them is located here:
https://github.com/jenpaulhus/group-actions-RS/blob/master/Schema.md

The one exception is the dimension data which will continue to grow across all values as the genus increases.  So we used the stats_display code to make that table (found at the bottom of our stats page).

There are 2 data issues for dimension, however, that I am not exactly sure how to fix because I don't really understand the stats tables postgres stores.  One issue is that right now there are two entries for each of dimension 0 through 10. The other is that some of the values where there are no entries of that dimension do show up (e.g. dim=22) but there are others where it does not show up (e.g. dim=16, 18, etc). I'm not sure how I managed to do these two things.

Credits:
One of my students, @kwang23 did a large amount of the work for this PR. I've added his name to contributors (I think we are still listing people there, instead of on individual section pages).

@ngonhi  and @370417  coded some of the original MongoDB versions in Summer 2018, and some of their code is used here to.


AndrewVSutherland:
@roed314 Have you had a chance to look at this?

roed314:
I was busy preparing for my talk on Monday in STAGE, then working on some other stats PRs yesterday.  I'm looking at this now.

roed314:
@jenpaulhus Currently the summary has less information than the short summary (these are the snippets at the top of the homepage and the top of the stats page).  Should we just set `short_summary = summary + "Here are some further statistics"`?  Or do you want these to be different in this way?

roed314:
Overall it looks good!  Here are some comments, and I'll be submitting a PR to your PR with some fixes and improvements.

* I wouldn't include the text data files in the code repository.  I've deleted them in my PR, which I think is enough (I don't think they're big enough that we need to worry about excising them from the history).
* It would be good to include knowls for the terms in the headers (genus, dimension, etc).
* It would be nice if as many of the counts as possible could be links to search results (I think this is possible for the "distinct generating vectors" row of the first table for example).  It's nice for people to be able to get examples of the kind of objects that you're enumerating.  I know that for some of these counts, this would require additional search functionality which probably needs to wait for another PR.

PR #4083, opened 2020-10-05
fix pyflakes issue, add knowls

AndrewVSutherland:


AndrewVSutherland:
The failures are all timeouts, pyflakes is happy.

PR #4094, opened 2020-10-08
ecnf: code changes to display BSD data

JohnCremona:
After updating the ec_nfcurves table to include several extra columns, these code changes show full BSD data.  (Without this you can still see ranks a gens as those columns were there already).

This adresses Issue #3900 

There are many slightly different cases to be dealt with depending on whether we have missing data (e.g. over sextic fields) or when the lower and upper bounds on the rank are different. I tried to be very systematic about this, and display all useful information. Here are examples to look at:

http://localhost:37777/EllipticCurve/2.0.3.1/111972.3/d/3 rank 0 known
http://localhost:37777/EllipticCurve/3.3.1509.1/54.2/c/1 rank 1 known
http://localhost:37777/EllipticCurve/2.2.97.1/100.2/b/1 rank 2 known
http://localhost:37777/EllipticCurve/2.2.93.1/197.2/b/1 rank 3 known
http://localhost:37777/EllipticCurve/3.3.1101.1/64.4/c/1 bounds 0,2; analytic rank 0
http://localhost:37777/EllipticCurve/2.0.11.1/3100.5/c/1 bounds 0,1; analytic rank 1
http://localhost:37777/EllipticCurve/4.4.15317.1/94.3/c/1 bounds 0,3; analytic rank 1
http://localhost:37777/EllipticCurve/4.4.2225.1/1216.5/h/1 bounds 1,3; analytic rank 1
http://localhost:37777/EllipticCurve/6.6.592661.1/409.1/a/1 bounds 0,2; analytic rank 2
http://localhost:37777/EllipticCurve/6.6.453789.1/43.1/a/1 no rank bounds or analytic rank

### DIFF
--- a/lmfdb/ecnf/WebEllipticCurve.py
+++ b/lmfdb/ecnf/WebEllipticCurve.py
@@ -107,7 +107,7 @@ def ideal_from_string(K,s, IQF_format=False):
         return "wrong" ## caller must check
 
 def pretty_ideal(I):
-    easy = I.number_field().degree()==2 or I.norm()==1
+    easy = True#I.number_field().degree()==2 or I.norm()==1
     gens = I.gens_reduced() if easy else I.gens()
     return r"\((" + ",".join([latex(g) for g in gens]) + r")\)"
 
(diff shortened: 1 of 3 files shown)


JohnCremona:
Only 1 of the 5 Travis checks failed, I think it was a timeout.

JohnCremona:
I fixed the data error whcih caused test failures in an ecnf test.  I restarted that test and will also rerun the failing tests on other PRs.

JohnCremona:
@AndrewVSutherland I fixed all the things we decided on earlier, I think.   The one gap I noticed comparing with G2C is the statement "Hasse-Weil conjecture: not verified".  Certainly for these, I think there is thought needed about which base fields we actually know that the L-functions has the expected properties.  Presumably being modular is enough, which covers degrees 2 (real), 3 (totally real) and most of 4 (totally real) -- and on fact *all* our curves over totally real fields are *certainly* modular even in degrees 5 and 6 since Samir and I checked that they do not arise from exceptions to his theorem.

AndrewVSutherland:
@JohnCremona Great, I will go ahead and review this now.  Regarding the Hasse-Weil conjecture, the main motivation is that this is really needed in order for the analytic rank to be well-defined.  Certainly modularity is enough, but I don't think it is necessary (i.e. there may be cases where we know Hasse-Weil holds without knowing modularity).  Presumably it should only depend on the field and possibly whether the curve has CM or not.

JohnCremona:
Thanks. I'll need to add to the reliability knowl anyway regarding magma's analytic rank computation.

PR #4095, opened 2020-10-08
Artin stats

roed314:
Add search on the number of ramified primes, create statistics page.  Note that this PR contains #4089 so that should be reviewed first.


roed314:
I'm getting errors from pylint that `range built-in referenced when not iterating (range-builtin-not-iterating)`.  Are we enforcing this?

```python3
def intervals(start, end, step):
    return list(zip(range(start, end, step), range(start+step, end+step, step)))
```
seems like a pretty clean way to write this function to me, though I'm sure I could rewrite it without using range.

jwj61:
Will merge when tests pass

Issue #4093, opened 2020-10-08
Search on projective image type for Artin representations yields no results


roed314:
See https://beta.lmfdb.org/ArtinRepresentation/?projective_image_type=Dn&search_type=List

jwj61:
After fixing a small logic error in the code, it is having a database query error.  The issue is that it want to see if a field stored as integer[] is in a list, so it is generating
"Proj_GAP" = ANY(ARRAY[ARRAY[4,2],ARRAY[6,1],ARRAY[8,3],ARRAY[10,1],ARRAY[12,4],ARRAY[14,1],...

This does not seem to be legal SQL for postgres.  Thoughts on how to get it to make the right syntax?

roed314:
I think the right answer might be to use strings rather than arrays of length 2.  But I will try to see if I can get something to work using the current column type.

edgarcosta:
@jwj61, I tried hard and couldn't figure out a workaround.

Logically, I would just write it with a bunch of ORs

roed314:
Fixed by #4097.

PR #4097, opened 2020-10-09
Fix search for Artin reps of dihedral type

jwj61:
Fixes issue #4093 

To test, you can use

   https://beta.lmfdb.org/ArtinRepresentation/?projective_image_type=Dn&search_type=List
   https://http://127.0.0.1:37777//ArtinRepresentation/?projective_image_type=Dn&search_type=List

to test.  I switched it from using "any" to an or with a list of the options.


AndrewVSutherland:
Reviewing this now.

PR #4099, opened 2020-10-09
Add optional argument to formatfield

jwj61:
The function formatfield will look up a list of coefficients to see if the polynomial is in the number field database.  It then either produces a number field knowl if possible, or it produces something else.  By default, the something else is "Deg 24" for a degree 24 field as a knowl, and if you open the knowl, you get the polynomial.

This adds an option to the function show_poly.  If it is set to True and the field is not in the database, then just display the polynomial.

There is no place to see the new behavior without changing a current usage of formatfield.  It is added in anticipation of people using it.  You can see that the old behavior still works on say an S_5 field's page (lots of high degree siblings won't be in the database).

### DIFF
--- a/lmfdb/number_fields/web_number_field.py
+++ b/lmfdb/number_fields/web_number_field.py
@@ -249,13 +249,26 @@ def psum(val, li):
 def decodedisc(ads, s):
     return ZZ(ads[3:]) * s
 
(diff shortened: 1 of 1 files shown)


roed314:
Test failures were just from hgcwa problems that have been fixed.

PR #4105, opened 2020-10-09
ecnf: change one word on main search page

JohnCremona:
It's no longer true to say that the rank is not known for "many" curves!  Out of 661079 curves we now have rank (not just bounds) for all but 4647, and analytic ranks for all but 16413.  The first of these numbers will go down shortly as the last field's rank data is added.  Meanwhile this is a trivial. change.  to a template.

I see that genus 2 curves all search on analytic rank, but not on MW rank.  We could do the same too for ECNF, or have both.


roed314:
Looks good.

Issue #2783, opened 2019-03-17
Formulas knowls not appearing on MF home page


davidfarmer:
Several modular forms have "formulas" knowls, which generally record things
like eta product representations.

These used to have a place on the home page of those modular forms,
but now do not appear.

Search for "formulas" in a knowl label, such as:
mf.11.2.1.a.formulas

AndrewVSutherland:
Yes, these should be annotations (top or bottom knowls).  Something to work on this week...

davidfarmer:
Actually, they are not top or bottom: they go in the middle
of the page (at least, that is where they used to be).  After
the q-expansion if I recall, because they give an alternate
expression for the funciton.

alexjbest:
I talked with @roed314 about this a while ago.

I had some difficulty finding a nice reference for these, possibly you both know more than this off the top of your head?! But this is what I found last time I looked:

In fact the best reference I found was a project of Henri Cohen at a Luxembourg summer school, http://math.uni.lu/nt/summerschool2018/Cohen-MF-project5.pdf which comes with answers: http://math.uni.lu/nt/summerschool2018/projectanswer5.pdf . In exercise 2 he lists all possible eta *products* giving 28 in total (so this does not include quotients like http://www.lmfdb.org/knowledge/show/mf.80.2.1.b.formulas), this (I guess) matches up with the 21 *products* we have as knowls currently, plus 7 new ones for weight 1 that should be added now.

Finally in his answer for exercise 7 he has written a gp function `mfisetaquo` to test whether  a given form is an eta *quotient*, I guess we should be able to run this on some range to check for more quotient formulas.
For instance his example there is http://www.lmfdb.org/ModularForm/GL2/Q/holomorphic/16/4/a/a/ which does not appear to have ever had a knowl on the lmfdb (maybe it was outside the range before, now the old ones have gone I have forgotten what was there before). This also appears in Table 1 of <https://arxiv.org/pdf/1805.00544.pdf>.



There is a also a book called Eta Products and Theta Series Identities by Günter Köhler which looks like it may have things of interest in but to me it was quite difficult to read.

davidfarmer:
The point of the formulas knowls was to record interesting formulas.
It was meant to be edited by a person, not generated automatically.
Eta products were just one example of an interesting formula.

It would be good to discuss what to do in general, as one of the 
themes of this week's workshop is to add interesting anomalous
information.

In most cases (in my opinion) the top knowl is not the right place
[...]

PR #4106, opened 2020-10-09
Add eta quotients for classical modular forms

davidlowryduda:
This commit adds eta quotients to pages on classical modular newform
orbit pages, if they exist.

The commit message follows.

There are a few design decisions here, and I document my reasons for
them now. Firstly, this is a relatively rare property, and so it should
not be a column in the db associated to each modular form orbit --- it
would simply be "None" for most of them. Thus it should either be its
own table in the db, hardcoded in the file, or in a yaml. I have chosen
to put this in a yaml file, similar to how some parts of the LMFDB parse
a yaml file to generate their code segments.

This yaml file needs to be parsed sometime and somewhere. It could
either be parsed in web_newform.py or in main.py. I've chosen to parse
it as late as possible, exactly in the rendering function in main.py.
But this is more open to interpretation.

There remain a few things before fully live.
1. This makes the bottom_knowls for functions expressible as eta
   products obsolete. These should be removed so that the list of
   "interesting forms" isn't filled with them.
2.  This also reuses a knowl, mf.half_integral_weight.eta_quotients,
   which does not seem to have been revisited in a long time. It may
   need to be rewritten.

See also the discussion in #2783.


AndrewVSutherland:
Reviewing this now.

AndrewVSutherland:
Thanks for doing this!  I only have two cosmetic suggestions:

 - Display quotients as fractions.
 - Also show the corresponding q-product formula.

See https://beta.lmfdb.org/knowledge/show/cmf.80.2.a.b.bottom for an example of both.

lmfdb/classical_modular_forms/main.py
```diff
@@ -322,14 +323,38 @@ def parse_prec(info):
...
         return abort(404, err.args)
+
     info = to_dict(request.args)
     info['display_float'] = display_float
     info['format'] = info.get('format', 'embed')
+
+    _curdir = os.path.dirname(os.path.abspath(__file__))
+    etaquotients = yaml.load(
+            open(os.path.join(_curdir, "eta.yaml")),
+            Loader=yaml.FullLoader)
```
edgarcosta, line 354: what about making eta quotients a global for that file and you load the YAML file when it gets imported?

davidlowryduda: This makes sense. I'll do that.

davidlowryduda:
@AndrewVSutherland I can do that. I'll update this in the next couple of days.

PR #4112, opened 2020-10-10
Sato-Tate group statistics

roed314:
Also adds search functionality for maximality, second and fourth moments of trace and first moment of a_2.


AndrewVSutherland:
Thanks for this!  I noticed two things:
- There is no link to the statistics on the front page.
- The links to the counts in the first table don't give the correct number of results.  I'm guessing this is because the search facility does nont support searching by component group, only number of components.  We should either add a search on component group (which would be useful in general), or change the table to have rows for number of components rather than the component group.

AndrewVSutherland:
Also, pyflakes is not happy.

roed314:
I added search by component group.  The only wrinkle is that if you search by GAP id for a cyclic group of order larger than 511 and ask for irrational ST groups they won't show up since `db.gps_small` doesn't include larger groups and we don't know what the GAP id is.

AndrewVSutherland:
Looks good, thanks!

PR #4114, opened 2020-10-10
Removing redundant script files

JohnCremona:
Removing redundant stuff from scripts/ecnf/.   All data handline scripts for ecnf are now in another repository (ecnf-data).  A few (about 3) remaining functions here are still used but eventually they will be tidied up too.

Merge once tests pass, but the website code is not affected.


roed314:
There's a pyflakes error in `hmf_check_find`; it should be easy to fix.

JohnCremona:
That is weird since I thonk that all I did to that file was delete one function. I'll check.

roed314:
Right; pyflakes is complaining that an import is no longer used.

PR #4116, opened 2020-10-11
Add class number formula to number field pages

jwj61:
Just go to a number field page and scroll down below units.

Typical page (but no whining about Artin representation labels being long now that they are multiplied out):
http://127.0.0.1:37777/NumberField/7.1.193607.1
Assumes grh:
http://127.0.0.1:37777/NumberField/27.9.67585198634817523235520443624317923.1
Data not computed:
http://127.0.0.1:37777/NumberField/47.1.3847243917163654348435987752506690356998556730699127064362242968132131231881167.1 
Case where the first formula is exact (so there is an equals sign):
http://127.0.0.1:37777/NumberField/2.0.8.1
Only case where both values are exact (so no approximation symbols):
http://127.0.0.1:37777/NumberField/1.1.1.1

I chose to leave things in the form of the formula, including exponents of 0

### DIFF
--- a/lmfdb/number_fields/number_field.py
+++ b/lmfdb/number_fields/number_field.py
@@ -497,6 +497,7 @@ def render_field_webpage(args):
         'unit_rank': nf.unit_rank(),
         'root_of_unity': rootofunity,
         'fund_units': nf.units_safe(),
+        'cnf': nf.cnf(),
         'grh_label': grh_label,
         'loc_alg': loc_alg
     })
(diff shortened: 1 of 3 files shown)


JohnCremona:
I'll take a look at this as it was my suggestion.

JohnCremona:
Looks good!  For the last example (Q) can we display "1" instead of "1.00000"?

Knowl issue: the knowl forget to define the class number h.  I'll leave you to edit that in, then I can positively review it. (Do we have to call what I have always known as the "Dedekind zeta function", denoted $\zeta_K$, as the "Dedekind L-function"?   Will we have to start referring to the Riemann L-function next?!))

jwj61:
Good points.  Both the knowl and the code (for Q) have been changed.

JohnCremona:
> Good points. Both the knowl and the code (for Q) have been changed.

Good, I'll take another quick look.

PR #4118, opened 2020-10-12
Add searching by number of arithmetically equivalent fields.

jwj61:
As it says, if you go to the Galois groups index page, you can now search on the number of arithmetically equivalent fields a corresponding field would have.  It is also added to the stats page.  As a side note, the number of these which have not been computed yet is going down, computations are in progress.

### DIFF
--- a/lmfdb/galois_groups/main.py
+++ b/lmfdb/galois_groups/main.py
@@ -152,6 +152,7 @@ def galois_group_search(info, query):
     parse_ints(info,query,'n','degree')
     parse_ints(info,query,'t')
     parse_ints(info,query,'order')
+    parse_ints(info,query,'arith_equiv')
     parse_ints(info,query,'nilpotency')
     parse_galgrp(info, query, qfield=['label','n'], name='Galois group', field='gal')
     for param in ('cyc', 'solv', 'prim'):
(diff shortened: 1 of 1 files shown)


roed314:
I think it makes more sense to use `"proportioner": proportioners.per_col_total` so that you get the fraction of a given degree with each count, rather than the fraction of things with 7 arithmetically equivalent fields that are in degree 32.

Other than that, it looks good.

I noticed that search results are quite slow when running on localhost.  We're currently doing multiple database lookups for each search result when searching on Galois groups.  These are coming from `otherrep_list` or `subfields` on the `WebGaloisGroup` object, which then do a `group_display_knowl` for each sibling.  The lookup is required in order to determine the display string to use the knowl.

We're caching the results, and things aren't as slow on beta or prod so maybe this isn't a problem that we need to address.  But one way we could speed things up would be to store the "pretty" names for siblings in addition to their labels, which would obviate the need for the database lookup.  Maybe we can revisit this once this ticket is pushed to beta if it's actually slow there.

jwj61:
Thanks for the feedback.  I basically had no idea what I was doing with the stat part and will look into that.

jwj61:
I wanted to change the label on the row from -1 to not computed, which was successful.  But that means that clicking on something from that row produces a search looking for "not computed" as the value in the database.  How do I avoid that?

jwj61:
Should work now

roed314:
Yep, looks like you got it.  I've tried to document the functionality you've been using in `lmfdb/utils/display_stats.py`; if you noticed any part of it where you would like more explanation (or a FAQ) let me know.

Issue #4053, opened 2020-09-25
Download Bianchi modular form data


AndrewVSutherland:
We need to provide a download option for Bianchi modular forms.

AndrewVSutherland:
Hilbert modular forms already provides this capability and might be a good model to follow.  I think it would be enough to just provide a download eigenvalues option, since I don't think these is a built-in Bianchi modular form object in Magma or sage (but @JohnCremona should correct me if I am wrong about this).

BarinderBanwait:
@AndrewVSutherland It looks like there is a ['ModFrmBianchi'](https://magma.maths.usyd.edu.au/magma/handbook/modular_forms_over_imaginary_quadratic_fields) object in Magma. From the introduction there, "Technically, ModFrmBianchi is implemented as a subtype of ModFrmHil", so a 'Modular form to Magma' option may also be feasible. I couldn't find it in sage however. (If nobody else is looking into this, I'd be happy to take it on as my first issue.)

AndrewVSutherland:
@BarinderBanwait I think there has been some work started on implementing a download option for the eigenvalues, but I don't where that stands and it won't address the question of a download to magma option, so I would say please go ahead.  If you are available to join us for part of our LMFDB Friday session tomorrow we could coordinate there (I'll send you an email with the Zoom link and an invitation to the Zulip cha).

BarinderBanwait:
Sure, I'd be happy to join tomorrow (received the zulip+zoom links, thanks!)

BarinderBanwait:
PR opened - #4123

PR #4123, opened 2020-10-14
Feature/4053 download bmf data

BarinderBanwait:
This PR for Issue #4053  adds Download functionality - into Sage and Magma - to each BMF homepage, in essentially the same way as currently done for HMFs.

Unit tests have been added, and executed locally on legendre - all pass. (Don't know where/how to execute the tests properly.)

### DIFF
--- a/lmfdb/bianchi_modular_forms/test_bmf.py
+++ b/lmfdb/bianchi_modular_forms/test_bmf.py
@@ -1,12 +1,6 @@
 # -*- coding: utf-8 -*-
-# To run this, do the following:
-# sage -python lmfdb/bianchi_modular_forms/test_bmf.py BMFTest.test_jump
-import sys
-import unittest
-from sage.all import *
-sys.path.append('/home/barinder/lmfdb')
-
 from lmfdb.tests import LmfdbTest
+from sage.all import *  # needed for 'test_download_sage'
 
 base_url = '/ModularForm/GL2/ImaginaryQuadratic/'
 
@@ -122,7 +116,7 @@ class BMFTest(LmfdbTest):
 
     def check_compile_and_get_level(self, download_data):
         """Simulates a user downloading the sage code, and then loading it into
-        a sage session."""
+        a sage session. This requires the sage import at the top"""
         aux_file = 'testing.sage'
 
         with open(aux_file, 'w') as the_file:
@@ -153,7 +147,8 @@ class BMFTest(LmfdbTest):
         L = self.tc.get('/ModularForm/GL2/ImaginaryQuadratic/2.0.7.1/88.5/a/download/magma').get_data(as_text=True)
         assert 'NN := ideal<ZF | {44, 2*a + 30}>;' in L
         assert '[263, a + 123],' in L
-        #  assert 'heckeEigenvaluesArray := [-1, 0, 2, 4, -1, -6, 0, 0,' in L
+        assert 'heckeEigenvalues[primes[32]] := 18;' in L
+        assert 'heckeEigenvalues[primes[53]] := -12;' in L
 
         page = self.tc.get('ModularForm/GL2/ImaginaryQuadratic/159.0.7.1/30.5/a/download/magma').get_data(as_text=True)
         assert 'Bianchi newform not found' in page
@@ -162,32 +157,19 @@ class BMFTest(LmfdbTest):
         from sage.all import magma
         import sys
         for label, expected in [
-                #['2.0.4.1/100.2/a',
-                # 'heckeEigenvaluesArray := [0, -1, -1, -2, 2, 2, -6, -6, 6, 6, 2, 2, 6, 6']
-                ['2.0.4.1/1066.4/a2',
-                 'P<x> := PolynomialRing(Rationals());']
+                ['2.0.4.1/100.2/a',
+                 'ALEigenvalues[ideal<ZF | {i + 1}>] := -1;'],
+                ['2.0.11.1/933.1/a',
+                 'heckeEigenvalues[primes[193]] := -30;']
         ]:
             sys.stdout.write("{}...".format(label))
             sys.stdout.flush()
             page = self.tc.get('/ModularForm/GL2/ImaginaryQuadratic/{}/download/magma'.format(label)).get_data(as_text=True)
             assert expected in page
             assert  'make_newform'  in page
-            # try:
-            #     print("Yabba Dabba Do!")
-            #     magma_code = page + '\n';
-            #     magma_code += 'f, iso := Explode(make_newform());\n'
-            #     magma_code += 'assert(&and([iso(heckeEigenvalues[P]) eq HeckeEigenvalue(f,P): P in primes[1..10]]));\n'
-            #     magma_code += 'f;\n'
-            #     assert 'success' in magma.eval(magma_code)
-            # except RuntimeError:
-            #     pass
-            print("Yabba Dabba Do!")
+
             magma_code = page + '\n';
             magma_code += 'f, iso := Explode(make_newform());\n'
             magma_code += 'for P in primes[1..15] do;\n if Valuation(NN,P) eq 0 then;\n  assert iso(heckeEigenvalues[P]) eq HeckeEigenvalue(f,P);\n end if;\nend for;\n'
             magma_code += 'f;\n'
             assert 'success' in magma.eval(magma_code)
-
-
-if __name__ == "__main__":
-    unittest.main()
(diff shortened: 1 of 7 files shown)


lmfdb/hilbert_modular_forms/test_hmf.py
```diff
@@ -158,12 +158,9 @@ def test_download_magma(self):
...
             assert expected in page
             assert  'make_newform'  in page
-            try:
-                magma_code = page + '\n';
-                magma_code += 'f, iso := Explode(make_newform());\n'
-                magma_code += 'assert(&and([iso(heckeEigenvalues[P]) eq HeckeEigenvalue(f,P): P in primes[1..10]]));\n'
-                magma_code += 'f;\n'
-                assert 'success' in magma.eval(magma_code)
-            except RuntimeError:
-                pass
```
BarinderBanwait, line 168: Currently, this `try...except` block makes the test pass even if the magma code is populated with erroneous code which doesn't compile in magma.

edgarcosta:
you should fix these pyflakes errors:
```
./lmfdb/bianchi_modular_forms/test_bmf.py:3:1 'from sage.all import *' used; unable to detect undefined names
./lmfdb/bianchi_modular_forms/test_bmf.py:125:9 'load' may be undefined, or defined from star imports: sage.all
./lmfdb/bianchi_modular_forms/test_bmf.py:126:9 'os' may be undefined, or defined from star imports: sage.all
./lmfdb/bianchi_modular_forms/test_bmf.py:127:16 'NN' may be undefined, or defined from star imports: sage.all
```

lmfdb/hilbert_modular_forms/test_hmf.py
edgarcosta: A small suggestion, as the server where we ran automatic tests don't have magma:
```
magma_code = ....
try:
    assert 'success' in magma.eval(magma_code)
except RuntimError:
    try: 
        assert 'success' in magma_free(magma_code)
    except socket.timeout as err:
            print("Connecting with magma.maths.usyd.edu.au timed out")
            print(err)

**Comment by BarinderBanwait** [comment]
> you should fix these pyflakes errors:
> 
> ```
> ./lmfdb/bianchi_modular_forms/test_bmf.py:3:1 'from sage.all import *' used; unable to detect undefined names
> ./lmfdb/bianchi_modular_forms/test_bmf.py:125:9 'load' may be undefined, or defined from star imports: sage.all
> ./lmfdb/bianchi_modular_forms/test_bmf.py:126:9 'os' may be undefined, or defined from star imports: sage.all
> ./lmfdb/bianchi_modular_forms/test_bmf.py:127:16 'NN' may be undefined, or defined from star imports: sage.all
> ```

the latest commit fixes the first three of these; I'm unsure how to resolve the pyflakes error for the last one. I'll say more by commenting at that specific line in the code.

File: lmfdb/hilbert_modular_forms/test_hmf.py (hunk 1 above)
Comment (BarinderBanwait) [reply lmfdb/hilbert_modular_forms/test_hmf.py:168]: The problem with RuntimeError when calling `magma.eval(magma_code)` is that *all errors* manifest as RuntimeError; e.g.:

```
sage: magma.eval("assert 1 eq 2;")  # raises RuntimeError, not AssertionError
sage: magma.eval("boaty mcboatface rocks!")  # raises RuntimeError, not SyntaxError
```

However, in my latest commit, I've used your idea of using `magma_free` and catching the timeout, but replacing the outer `try ... except` with `if magma.is_local()`:

```
if magma.is_local():
    assert 'success' in magma.eval(magma_code)
else:
    try:
        from sage.all import magma_free
        assert 'success' in magma_free(magma_code)
    except socket.timeout as err:
        print("Connecting with magma.maths.usyd.edu.au timed out")
        print(err)
```

I don't know how Travis will react to that ... I'll wait and see; but it works locally on legendre

File: lmfdb/bianchi_modular_forms/test_bmf.py [hunk 2]
```diff
@@ -113,3 +114,62 @@ def test_friends(self):
...
+        """Simulates a user downloading the sage code, and then loading it into
+        a sage session. This requires the sage import at the top"""
+        aux_file = 'testing.sage'
+
+        with open(aux_file, 'w') as the_file:
+            the_file.write(download_data.get_data(as_text=True))
+
+        load(aux_file)  # if there's a problem with loading, this will make the test fail
+        os.remove(aux_file)
+        return NN
```
BarinderBanwait, line 129: This causes the one remaining pyflakes error, and I admit that the whole premise of this method might raise some eyebrows. Pyflakes complains because NN is seemingly not defined, but what it doesn't know is that the `load(aux_file)` defines it (it's the level of the BMF).

My intention with this method is to simulate a user downloading the sage code, and loading it in a sage session. This test will catch any errors in loading the sage code, which there are currently no checks for. It also returns the level ideal to the calling method, which then checks that it is equal to the ideal on the homepage.

Maybe there's another way of running this sage compilation check, one that in particular keeps pyflakes happy? I don't know what the LMFDB policy on declared exemptions to pyflakes/pylint etc are ...

JohnCremona:
You should not be importing * ever. Just import the names you need from
sage.all (or somewhere more specific if the name is not imported into
sage's global namespace).

lmfdb/bianchi_modular_forms/test_bmf.py
edgarcosta: You should have no statements like `from foo import *`
Also, why are you writing to a file to load it? You should be doing something similar to what you did in magma, evaling the string, see for example: 
https://github.com/LMFDB/lmfdb/blob/master/lmfdb/classical_modular_forms/test_cmf.py#L575

BarinderBanwait:
Thanks @JohnCremona, now only importing the 5 things I need from sage.all - this'll be the last time an `import *` will be seen in one of my PRs!

lmfdb/bianchi_modular_forms/test_bmf.py
BarinderBanwait: Aah cool, I didn't know about `exec`.

However, after trying it, there's an issue in that it's not finding the meaning of 'special' characters in sage, such as the `<` in `P.<x>` or `^`; a MWE is the single line in the sage console: `exec("print(x^2)")`. 

Do you know how I can get `exec` to work with this special sage notation?

(I've rewritten the sage code to not use `<` and `>`, but not using `^` makes things a bit more inconvenient, since it's entwined into the hecke poly.)

edgarcosta: I honestly recommend against all those shortcuts, as those are converted to valid python code with a preparser.

Here are various ways to create polynomials, without relying on the operator '^'

```
sage: R = PolynomialRing(ZZ, 'x')
sage: x = R.gen()
sage: f = 3*x**3 + 5
sage: R(list(f))
3*x^3 + 5
sage: list(f)
[5, 0, 0, 3]
sage: R("x^5 + 1")
x^5 + 1
```

BarinderBanwait: Ok thanks - I'll change the sage download code to output a list for the hecke polynomial, and then read that into the number field

lmfdb/hilbert_modular_forms/test_hmf.py
BarinderBanwait: OK so Travis didn't like that; apparently `magma.is_local()` was True even though it doesn't have magma. The latest commit tries the following:

```
try:
    assert 'success' in magma.eval(magma_code)
except RuntimeError as the_error:
    if str(the_error).startswith("unable to start magma"):
        try:
            from sage.all import magma_free
            assert 'success' in magma_free(magma_code)
        except socket.timeout as err:
            print("Connecting with magma.maths.usyd.edu.au timed out")
            print(err)
            pass
    else:
        raise
```

BarinderBanwait: So that didn't work either; the issue with the above is that `magma_code` is enormous, and waay bigger than the 50,000 bytes limit for the free online magma calculator, so the input is ignored, and returns an empty magma expression, so the assertion fails.

At this point I think all bets are off for executing the magma code on Travis, so I'll aim to check the magma compilation works only when full magma is able to start:

```
try:
    assert 'success' in magma.eval(magma_code)
except RuntimeError as the_error:
    if str(the_error).startswith("unable to start magma"):
        pass
    else:
        raise
```

edgarcosta: Sounds good!

edgarcosta:
Once the tests pass, this can be merged.

PR #4124, opened 2020-10-14
Fix problems with Bianchi stats

roed314:


lmfdb/backend/statstable.py
```diff
@@ -1537,16 +1536,28 @@ def _get_values_counts(
...
+                    if isinstance(cx, dict) and all(isinstance(k, str) and k and k[0] == "$" for k in cx):
                         # Have to handle some constraint parsing here
-                        from lmfdb.backend.utils import postgres_infix_ops
                         typ = self.table.col_type[x]
                         for k, v in cx.items():
-                            op = postgres_infix_ops.get(k)
-                            if not op:
+                            if k in ['$gte', '$gt']:
+                                oe = '>='
+                                ke = '$gte'
```
edgarcosta, line 1545: what is the purpose of `ke`? not used

roed314: Thanks; I deleted it.

PR #4100, opened 2020-10-09
Bianchi download skeleton

bebr117:
A tiny amount of skeleton code for downloading properties of Bianchi modular forms (Issue #4053 ). Based on the analogous code used for Hilbert modular forms.

Issue #4054, opened 2020-09-25
Provide option to display more a_p for Bianchi modular forms


AndrewVSutherland:
By default only 25 a_p are shown, but there are more in the database.  We should provide the user the option to see more (as in classical modular forms, for example).

AndrewVSutherland:
I believe Brian Reinhart is planning to implement this (but if not I will take care of it on Friday).

bebr117:
Now that my PR has been approved, this issue should be closed.

PR #4127, opened 2020-10-16
Bianchi eigenvalue display

bebr117:
Added the functionality for #4054 . I'm not very good at git, so the skeleton code from #4100 is in here as well. I can probably go in and manually delete that stuff if necessary.


bebr117:
This most recent commit cleans up the unnecessary code from #4100 that I mentioned earlier.

AndrewVSutherland:
Looks good!  I will merge this as soon as Travis finishes.

Issue #3994, opened 2020-08-21
Kernel fields for Dirichlet characters


AndrewVSutherland:
It would be useful (and consistent with the knowl heading for [related fields](https://www.lmfdb.org/knowledge/show/character.dirichlet.related_fields)) to store the kernel fields for Dirichlet characters of small order (say order up to 24, where we would expect them to be in the number fields database).

These fields need to be computed and stored for the Dirichlet characters of small order for modulus up to 10,000.  For Dirichlet chaacters of larger modulus that have very small order (e.g. 2), we might want to compute them on the fly.

alexjbest:
There is a 10 year old Sage ticket to add this functionality https://trac.sagemath.org/ticket/9407 which has some okay looking code. I'll try and brush that up and go from there.

jwj61:
I have not looked at the sage code.  In gp I would use polsubcyclo to get the candidates of a given degree in Q(zeta_n), then use primes where the character is trivial to eliminate all but one.

JohnCremona:
Alex, you should not hesitate to contact Chris Wuthrich about this, I'm sure he would be very happy to work with you.

alexjbest:
Thank you everyone for the suggestions (and code)! I'll try and compare the pari code with whats in that Sage ticket and see what works best on Friday.

jwj61:
Given the age of the sage ticket, I doubt it will be as good as Aurel's suggestion.

JohnCremona:
> Given the age of the sage ticket, I doubt it will be as good as Aurel's suggestion.

Agreed -- probably Withrich would never have written that code if the possibility of calling a libpari function had been available then.

AurelPage:
Yes, I doubt that an old Sage code beats Bill's implementation.

Here is better code:
> ? conreytofield(chi) = my(G,K); [G,chi] = znchar(chi); K=charker(G,chi); galoissubcyclo(G,K);

alexjbest:
Ok thanks everyone, I've changed the Sage ticket https://trac.sagemath.org/ticket/9407 to use Aurel's code, so thank you for that!

That is slightly orthogonal to getting these in lmfdb of course, but I think it would be sad if Sage didn't get this functionality as a by-product.

AndrewVSutherland:
[...]

PR #4130, opened 2020-10-16
Add fixed field knowl to Dirichlet character homepage

BarinderBanwait:
This PR addresses issue #3994 , to add a knowl for the fixed field to the Dirichlet character homepage.

Note that currently the value field is not a knowl; it was verbally decided to change it to a knowl in today's session, but i didn't get around to implementing that.

### DIFF
--- a/lmfdb/characters/templates/Character.html
+++ b/lmfdb/characters/templates/Character.html
@@ -137,15 +137,14 @@
     {% if vflabel %}
     <td> <a href="/NumberField/{{vflabel}}">{{ valuefield }}</a> </td>
     {% else %}
-    <td> {{ valuefield }} </td>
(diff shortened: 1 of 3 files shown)


jwj61:
Please do not change the default behavior of formatfield.  It will make current uses of it unweildy.

One option is to use the show_poly keyword argument so that it just displays the polynomial when it is not in the database.  Another would be to add another optional keyword argument of what string to use for the knowl title.  If it is not provided, the current behavior happens (so it doesn't change existing pages).

AndrewVSutherland:
@jwj61 In the interest of time I am going to merge this then implement the change you suggest (and one other fix).  We have a bunch of other PRs related to Dirichlet characters (including one I am working on but have not submitted) and I'd like to get them all merged today.

PR #4132, opened 2020-10-17
to_dict: do not overwrite keys passed as input

edgarcosta:
this prevents users (and bots) to overwrite variables in the info dictionary that were supposed to be set by the code, for example:
https://beta.lmfdb.org/ModularForm/GL2/Q/holomorphic/?search_array=foobar

@AndrewVSutherland and @roed314, this should stop most of flasklog errors coming from www.lmfdb.org


AndrewVSutherland:
Thanks for fixing this!

PR #4133, opened 2020-10-17
undo formatfield change, add keyword arg

AndrewVSutherland:


AndrewVSutherland:
The only test failure is the same pyflakes warning about unused imports which I will fix in the next PR cleaning up Dirichlet characters.

PR #4134, opened 2020-10-17
Dirichlet character cleanup

AndrewVSutherland:
Fixes pyflakes errors, minor formatting issues, value field number field display, and modifies search results to abbreviate rows for large Galois oribits.  

These are all straight forward fixes to minor problems caused by the last 3 PRs that were merged related ot Dirichlet characters.  I will merge this once tests pass then push to beta.

Download search results has been removed since we don't yet support downloads for Dirichlet characters in any form (there was code copied from abvars that I deleted).  Once we add downloads to Dirichlet character pages we can add them to search results (this mostly amounts to deciding how we want to handle Dirichlet characters in Sage and Magma where there is a dependency on code to handle Conrey characters that is not  built-in to Sage and Magma, Pari/GP does provide built-in support for Conrey characters.  In any case, I think it makes sense to defer dealing with downloads for Dirichlet characters to a separate PR, and I don't think we want to try and address this in 1.1.2 (which is almost done, only one more PR is pending!)

### DIFF
--- a/lmfdb/characters/main.py
+++ b/lmfdb/characters/main.py
@@ -69,7 +69,7 @@ def render_characterNavigation():
 
 class DirichSearchArray(SearchArray):
     jump_example = "13.2"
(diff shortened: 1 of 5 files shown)


alexjbest:
Sorry for missing these! Looks good to me

PR #4143, opened 2020-10-18
Uniformization of properties boxes, search results, handling large integers

AndrewVSutherland:
This PR addresses #486 to uniformize the appearance of quantities in property boxes, as well as cleaning up several other minor deviations from the [style sheet](https://github.com/LMFDB/lmfdb/blob/master/StyleSheet.md) that I noticed while addressing #486 (most notably, consistently putting integers listed in search results in math mode, we were split about 50/50 on this).

Possibly the most notable change to the properties boxes is to not show factorizations of integers (in all cases where this occurred the factorization is available on the main page), so that now the conductor of an elliptic curve is displayed in the same way as the conductor of its L-function.  Large integers in the property box are now shown in scientific notation to prevent them from making the property box to large to display them in a reasonable way (note that this problem arises whether you factor integers or not, see  [this number field](https://www.lmfdb.org/NumberField/47.1.551086183047315924646263001161663715403004994182429866636300214689149995499751082102635646125101986285966481127.1) for example).

Other changes include some further cleanup of Dirichlet characters, including the remove of the Hecke character routes (which were inaccessible except by direct URL and did not work correctly), some formatting changes to select boxes that appear as part of a label in search arrays to make the pages cleaner (especially the refine search pages), and some further changes to handle the very long labels that can appear in Artin representations and number fields.

This PR is up and running on https://blue.lmfdb.xyz.  A good way to test it is to repeatedly click on the "Random" link in the sidebar and watch the property boxes.  Here are examples of object pages where there are noticeable differences (primarily in the property box).

[EC L-function before](https://beta.lmfdb.org/L/EllipticCurve/Q/399977/a/)
[EC L-function after](https://blue.lmfdb.xyz/L/EllipticCurve/Q/399977/a/)

[NF before](https://beta.lmfdb.org/NumberField/46.46.36655895649209236301982551721202737575299278801903892310770876621683915109917518654573451777500377652400169.1)
[NF after](https://blue.lmfdb.xyz/NumberField/46.46.36655895649209236301982551721202737575299278801903892310770876621683915109917518654573451777500377652400169.1)

[Artin before](https://beta.lmfdb.org/ArtinRepresentation/12.1033895336359889350315432111578828457709236174913760903042486725925406772026569.55.a.b)
[Artin after](https://blue.lmfdb.xyz/ArtinRepresentation/12.1033895336359889350315432111578828457709236174913760903042486725925406772026569.55.a.b)

Below are some examples of search results pages where there are noticeable differences.

[EC/Q before](https://beta.lmfdb.org/EllipticCurve/Q/?search_type=List&all=1)
[EC/Q after](https://blue.lmfdb.xyz/EllipticCurve/Q/?search_type=List&all=1)

[ECNF before](https://beta.lmfdb.org/EllipticCurve/?search_type=List&all=1)
[ECNF after](https://blue.lmfdb.xyz/EllipticCurve/?search_type=List&all=1)

[HGCWA before](https://beta.lmfdb.org/HigherGenus/C/Aut/?search_type=List&all=1)
[HGCWA after](https://blue.lmfdb.xyz/HigherGenus/C/Aut/?search_type=List&all=1)

[Dirichlet before](https://beta.lmfdb.org/Character/Dirichlet/?all=1)
[Dirichlet after](https://blue.lmfdb.xyz/Character/Dirichlet/?all=1)

[Sato-Tate before](https://beta.lmfdb.org/SatoTateGroup/?search_type=List&all=1)
[Sato-Tate after](https://blue.lmfdb.xyz/SatoTateGroup/?search_type=List&all=1)

### DIFF
--- a/lmfdb/abvar/fq/isog_class.py
+++ b/lmfdb/abvar/fq/isog_class.py
@@ -407,7 +407,7 @@ class AbvarFq_isoclass(object):
                     bc = av_display_knowl(twist[1])
                 else:
                     bc = "(not in LMFDB)"
-                ans += "<tr><td>%s</td><td style="center">$%s$</td><td>%s</td></tr>\n" % (av_display_knowl(twist[0]), str(twist[2]), bc)
+                ans += "<tr><td>%s</td><td style='center'>$%s$</td><td>%s</td></tr>\n" % (av_display_knowl(twist[0]), str(twist[2]), bc)
                 i += 1
         ans += "</tbody></table>\n"
         return ans
--- a/lmfdb/abvar/fq/isog_class.py
+++ b/lmfdb/abvar/fq/isog_class.py
@@ -407,7 +407,7 @@ class AbvarFq_isoclass(object):
                     bc = av_display_knowl(twist[1])
                 else:
                     bc = "(not in LMFDB)"
-                ans += "<tr><td>%s</td><td>$%s$</td><td>%s</td></tr>\n" % (av_display_knowl(twist[0]), str(twist[2]), bc)
+                ans += "<tr><td>%s</td><td style="center">$%s$</td><td>%s</td></tr>\n" % (av_display_knowl(twist[0]), str(twist[2]), bc)
                 i += 1
         ans += "</tbody></table>\n"
         return ans
--- a/lmfdb/abvar/fq/isog_class.py
+++ b/lmfdb/abvar/fq/isog_class.py
@@ -399,7 +399,7 @@ class AbvarFq_isoclass(object):
         else:
             ans = "Below are some of the twists of this isogeny class."
         ans += '<table class = "ntdata">\n'
-        ans += "<tr><td>Twist</td><td>Extension degree</td><td>Common base change</td></tr>\n"
+        ans += "<thead><tr><th>Twist</th><th>Extension degree</th><th>Common base change</th></tr></thead><tbody>\n"
         i = 0
         for twist in self.twists:
             if twist[2] <= 3 or show_all or i < 3:
@@ -409,7 +409,7 @@ class AbvarFq_isoclass(object):
                     bc = "(not in LMFDB)"
                 ans += "<tr><td>%s</td><td>$%s$</td><td>%s</td></tr>\n" % (av_display_knowl(twist[0]), str(twist[2]), bc)
                 i += 1
-        ans += "</table>\n"
+        ans += "</tbody></table>\n"
         return ans
 
     def curve_display(self):
--- a/lmfdb/abvar/fq/templates/show-abvarfq.html
+++ b/lmfdb/abvar/fq/templates/show-abvarfq.html
@@ -78,17 +78,21 @@ This isogeny class does not contain a {{KNOWL('ag.jacobian',title='Jacobian')}},
 
 <p>{{ KNOWL('ag.fq.point_counts',title="Point counts of the abelian variety") }}
 
-<table class = "ntdata">
-    <tr><td>$r$</td>
-        {% for r in range(1, cl.length_A_counts()+1) %}
-        <td class="center"> {{ r }} </td>
+<table>
+  <thead>
+    <tr style="border-bottom: 1px solid #000;"><th>$r$</th>
+        {% for r in range(1, max(6,cl.length_A_counts()+1)) %}
+        <td class="center"> ${{ r }}$ </td>
         {% endfor %}
     </tr>
+  </thead>
+  <tbody>
     <tr><td>$A(\F_{q^r})$</td>
         {% for count in cl.abvar_counts %}
-        <td> {{ count }} </td>
+        <td> ${{ count }}$ </td>
         {% endfor %}
     </tr>
+  </tbody>
 </table>
 </p>
 
@@ -99,17 +103,20 @@ This isogeny class does not contain a {{KNOWL('ag.jacobian',title='Jacobian')}},
 {{ KNOWL('av.fq.curve_point_counts',title="Point counts of the (virtual) curve") }}
 {% endif %}
 
-<table class = "ntdata">
-    <tr><td>$r$</td>
+<table>
+  <thead>
+    <tr style="border-bottom: 1px solid #000;"><th>$r$</th>
         {% for r in range(1, cl.length_C_counts()+1) %}
-        <td class="center"> {{ r }} </td>
+        <th class="center">${{ r }}$</th>
         {% endfor %}
     </tr>
+  </thead><tbody>
     <tr><td>$C(\F_{q^r})$</td>
         {% for count in cl.curve_counts %}
-        <td>{{ count }}</td>
+        <td>${{ count }}$</td>
         {% endfor %}
     </tr>
+  </tbody>
 </table>
 
 </p>
--- a/lmfdb/abvar/fq/templates/show-abvarfq.html
+++ b/lmfdb/abvar/fq/templates/show-abvarfq.html
@@ -79,7 +79,7 @@ This isogeny class does not contain a {{KNOWL('ag.jacobian',title='Jacobian')}},
 <p>{{ KNOWL('ag.fq.point_counts',title="Point counts of the abelian variety") }}
 
 {% set n = cl.length_A_counts() if cl.length_A_counts() < 6 else 5 %}
-<table style="left-margin:10px;">
+<table style="left-margin:5%;">
   <thead>
     <tr style="border-bottom: 1px solid #000;"><th>$r$</th>
         {% for r in range(1, n+1) %}
@@ -104,7 +104,7 @@ This isogeny class does not contain a {{KNOWL('ag.jacobian',title='Jacobian')}},
 {{ KNOWL('av.fq.curve_point_counts',title="Point counts of the (virtual) curve") }}
 {% endif %}
 
-<table style="left-margin:10px;">
+<table style="left-margin:5%;">
   <thead>
     <tr style="border-bottom: 1px solid #000;"><th>$r$</th>
         {% for r in range(1, cl.length_C_counts()+1) %}
--- a/lmfdb/abvar/fq/templates/show-abvarfq.html
+++ b/lmfdb/abvar/fq/templates/show-abvarfq.html
@@ -78,18 +78,19 @@ This isogeny class does not contain a {{KNOWL('ag.jacobian',title='Jacobian')}},
 
 <p>{{ KNOWL('ag.fq.point_counts',title="Point counts of the abelian variety") }}
 
+{% set n = cl.length_A_counts() if cl.length_A_counts() < 6 else 5 %}
 <table>
   <thead>
     <tr style="border-bottom: 1px solid #000;"><th>$r$</th>
-        {% for r in range(1, max(6,cl.length_A_counts()+1)) %}
+        {% for r in range(1, n+1) %}
         <td class="center"> ${{ r }}$ </td>
         {% endfor %}
     </tr>
   </thead>
   <tbody>
     <tr><td>$A(\F_{q^r})$</td>
-        {% for count in cl.abvar_counts %}
-        <td> ${{ count }}$ </td>
+        {% for i in range(n) %}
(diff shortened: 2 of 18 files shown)


edgarcosta:
So far the changes look good to me.

I find odd the fact that we centralized some sections and not others for L-functions, but I don't think we need to decide/fix that now.

AndrewVSutherland:
@edgarcosta You are not the only one who finds it odd, John Jones and I were debating this earlier this morning.  Currently the we center some of the sections on the L-function pages, number field pages, and Artin rep pages, but nowhere else I believe, and we are not consistent about what we center (e.g. the function equation for the L-function is centered on the L-function pages but the class number formula is not centered on number field pages and L-polynomials are centered on L-function pages but not in some other places).  I think we should be consistent.

My current inclination is to remove centering everywhere.  If there is a consensus that certain types of content should be centered then we should make that part of the house style and apply it consistently.  I'll go ahead and make the change now so that people can see what it looks like.

AndrewVSutherland:
One reason not to center things is that it has different effects depending on where the content appears on the page (you can see this on the L-function pages where the functional equation and Euler product are centered differently because of the properties and learnmoreabout boxes.

roed314:
There's something wrong with the functional equation at https://blue.lmfdb.xyz/L/EllipticCurve/Q/399977/a/.  It looks like you added `\(` and `\)` but forgot to delete `\[` and `\]`.

roed314:
Similarly, there's a `\n` displayed in the Euler factor section.

roed314:
Perhaps I'm just used to centering display mode math in papers, but I don't know that I like changing all `\[ DISP \]` to `\( DISP \)`.  In the Euler factor section on the page above, I think that putting it all on one line gets pretty cluttered.  Using `\[ \]` is an easy way to get something in display more and get some vertical space.  Maybe what would be ideal would be the same amount of vertical offset but instead of centering we would indent a fixed amount (20 pixels or something).  Maybe write a macro to do that?

jwj61:
The Euler product seems to be indented less than the L-values.  Maybe try \displaystyle for the Euler product to give it more space, even when using inline math mode.

AndrewVSutherland:
@jwj61 @roed314  I'm still fiddling, you weren't meant to see these intermediate changes, but I have to commit to see what it looks like on blue, I'll close this PR for the moment and reopen it when I'm ready for people to look at it.

roed314:
Sounds good!

davidfarmer:
The Euler product should be display style do that the "p prime" goes under
the product sign.

I don't like all the flush left throughout the page,
but I will reserve judgment since right now all I see is that it is 
different.

AndrewVSutherland:
@jwj61 @roed314, @davidfarmer, @edgarcosta  Take a look at https://blue.lmfdb.xyz/L/Genus2Curve/Q/830/a/ and let me know what you think.  If there are sections you think should be centered, I think it would be good to say why in a way that is not specific to the L-function pages.  E.g. I could see an argument for "all equations expressing fundamental identities" should be centered (so the class number formula on the number field page would then also be centered).  Or "tables of local data (with rows per p) should be centered"  But I don't think the table of L-polys on the L-function page should be centered when tables of L-polys are not centered in many other places (e.g. isogeny class pages and CMF pages).

Also keep in mind that centering will format inconsistently depending on where the properties/related objects/learnmore boxes show up on the page.  So tables with the exact same dimensions may wind up getting aligned differently depending on where they show up vertically on the page.

An alternative to centering would be to use a fixed indentation, but it can't be too much of an indentation or it will cause problems on small screens.

AndrewVSutherland:
I could also see an argument for centering graphs/plots (currently isogeny graphs are centered by Z-plots are not).

davidfarmer:
I don't like a product (or a sum) with limits using inline display.
That is not a suggestion for centering it on the page:  it is a
suggestion for \displaystyle.  (This applies to the Euler product,
but not any place else I have seen.)

For the functional equation of the L-function, I prefer the
two-line display because the first line is the definition of
\Lambda(s), and the second line is the actual functional equation.
Not arguing for it to be centered, but for its layout wherever
it is located.

Is using ":=" to indicate "defined by" something that has been discussed?

Aesthetically I still don't like everything flush left.  It makes things
run together.  Centering also indicates that something is important
and unique, and not just one of several things that belong together.
Here is a pleasing looking page:

https://blue.lmfdb.xyz/ArtinRepresentation/1.712904.2t1.a.a

I would find it harder to parse if r_1 and r_2 were not centered.
I looked at several random Artin representation pages, some with
complicated data and some with simple, and that reinforced what I
wrote above.

So, the principle I suggest is to center items that are of
stand-alone interest.  So, maybe that often happens when a heading
only contains one thing?  (Not the list of BSD invariants, for example.)
I also think that partial series (with a ... or big-O at the end)
should not be centered, because they have varying width.
I still don't like the look of small tables shoved to the left,
and find those hard to parse.  (Example:  Atkin-Lehner signs table.)

On the holo mod forms pages, the last table is pretty wide always,
so centered or not does not make much difference.  But being flush
left still makes it harder for me to parse.

davidfarmer:
When I described the Artin pages as "pleasing looking", that was before the centering was removed!

I really hope we are not going to have an all-flush-left look.  It is true that my initial
reaction is that it looks ugly, but I also find that the column of information makes it
hard to separate items and also harder to parse each item.  Maybe having extra white
space above each heading would help, but that just adds space which isn't needed
when more things are centered.

AndrewVSutherland:
@davidfarmer I added displaystyle as you suggested, and I've reverted the change to the FE, which I think it does makes sense to center; I think the class number formula on the number field pages should also be centered..

I don't think it makes sense to center the table of Euler factors unless we think all similar tables should be centered (which I wouldn't necessarily object to).  I also don't like the fact that it centers differently from the FE.

Out of curiosity, on the page https://blue.lmfdb.xyz/Genus2Curve/Q/610627/a/610627/1, which sections, if any, would you like to see centered?

I'm happy to revert the change to the Artin rep page, but I'd like to give John Jones a chance to weigh in first (he previously said he was fine with not centering).

An alternative between flush left and centering would be a largish indent (e.g. 100px).

jwj61:
Looking at blue now, I think the roots of the Artin representations look fine in flush left.

jwj61:
Not to kibitz too much, but the display of the class number formula is more complicated since sometimes it is a formula, and sometimes it is text.

AndrewVSutherland:
@jwj61 Take a look at https://blue.lmfdb.xyz/NumberField/12.0.41223887921.1 and let me know what you think.  In cases where the class number formula is not computed, I don't think it's necessarily a problem to still show it centered (the lim_{s\to1}) is still tehre, but we could left-justify it in this case, or just omit the section entirely when it hasn't been computed (I think that would be fine).

davidfarmer:
> Out of curiosity, on the page
https://blue.lmfdb.xyz/Genus2Curve/Q/610627/a/610627/1, which sections, if any, would you
> like to see centered?

On that page, I probably can get used to it as-is.

The Euler factors table being flush left, for example

https://blue.lmfdb.xyz/L/EllipticCurve/Q/446/a/

still looks wrong to me.  A large indent is an alternative, and could be
applied generally as long as that does not introduce problems when tables
are very wide (making the forced indent larger than the margin if it was
centered).  By "looks wrong to me" I think I mean "feels like it is trying
latch on to other parts of the page instead of being its own thing."
I think I am referring to one of the Gestalt principles which I
internalized a long time ago and now have trouble articulating because I have
never taught the subject.

The L-function graph flush left does not look right, either.

I am fine with the formula for the Euler product, and the Particular
values, and everything else except the functional equation, on the left.

davidfarmer:
On the Artin pages, I agree that the roots should be flush left
(I suggested otherwise, but then later suggested that in general
partial series should be flush left).

Still would like to to see the two tables at the bottom centered,
for the same reasons as the Euler factors:

https://blue.lmfdb.xyz/ArtinRepresentation/2.4116.7t2.a.b

jwj61:
I am happy either way with the Artin pages.  But, if we are not centering tables elsewhere, then I don't see a good reason to do it here.

The class number formula looks good in all cases (at least that I looked at).

AndrewVSutherland:
@davidfarmer When you say the L-function graph do you mean the Z-plot?  I think I agree that it would be better centered (its currently flush left, I did not change it).

On the Artin rep pages I think I agree with John Jones that it is fine flush left, in fact I think I prefer it that way (but possibly that is because my screen is larger than yours -- my eye has to jump further).

@davidfarmer and @jwj61 what do you think of the class number formula on https://blue.lmfdb.xyz/NumberField/6.0.9747.1 or https://blue.lmfdb.xyz/NumberField/19.5.23920741637672928000872108534200.1, for example.  When the RHS has not been computed we could change the layout (or even omit it completely), but it doesn't actually bother me seeing it centered even in that case.

jwj61:
The two class number formulas you give are computed and look good to me.

In cases where it is not computed, it should probably follow the same convention as special values of L-functions on L-function pages.

AndrewVSutherland:
The Z-plot is now centered on the L-function pages, see https://blue.lmfdb.xyz/L/Genus2Curve/Q/714240/b/., for example.

davidfarmer:
> @davidfarmer and @jwj61 what do you think of the class number formula on https://blue.lmfdb.xyz/NumberField/6.0.9747.1 or
> https://blue.lmfdb.xyz/NumberField/19.5.23920741637672928000872108534200.1, for example. When the RHS has not been
> computed we could change the layout (or even omit it completely), but it doesn't actually bother me seeing it centered
> even in that case.

I like the class number formulas centered, and I am fine
with nothing else centered on that page.  (Those tables have
a different style and also are very wide.)

davidfarmer:
I clicked Random many many times, and the only thing that jumps
out at me is those blue-and-white striped tables that are flush left.

I have not see those with a large left margin (that squishes when
necessary), but that would alleviate some of my discomfort because
it is both the proximity to the TOC and the left alignment which
concern me.

AndrewVSutherland:
@davidfarmer I added a 5 percent indent to the ntdata tables, which I think adds enough white space to clearly separate them from the sidebar -- now that you point out the issue I can see the benefit of some more visual separation.  I think it looks pretty good, but let me know what you think.  We could easily go bigger than 5 percent, but at 10 percent I think it starts to hit an "uncanny" valley between indentation and a botched attempt at centering on a narrow screen.

We could add a similar indent to other table styles if we wanted to.

And let me add a big hat tip to @SamSchiavone, the global random link has been extremely useful today!

davidfarmer:
I think the 5 percent helps. I don't have any specific suggestions for
now.  I may fiddle with the layout some time in the future.

I think moving over all the similar tables would be good.

AndrewVSutherland:
@davidfarmer, @jwj61 I added 5 percent to the pages on the NF tables, e.g. https://blue.lmfdb.xyz/NumberField/3.1.509547.1.  I think it's an improvement.

davidfarmer:
Yes, I like the number field changes.  Breaking the left edge
with table margin really helps the eye carry some of the
cognitive burden.

jwj61:
Looks good

AndrewVSutherland:
I'm done fiddling with this.  Unless there are any further comments or concerns, I think it is ready to be merged.

roed314:
I only noticed one issue, which is pretty minor.  At a certain screen width I'm getting a closing curly brace on its own line for displaying the Dirichlet character group for a number field:
Screen Shot 2020-10-20 at 3 18 04 PM
I wasn't able to replicate this on beta.

davidfarmer:
The closing curly brace on its own can happen on beta:
it happens for me here (if I slowly shrink the window)

https://beta.lmfdb.org/NumberField/2.2.1081.1

Similar things were fixed recently in this pull request:

https://github.com/LMFDB/lmfdb/pull/4062

Otherwise, all looks good to me.

AndrewVSutherland:
@roed314 Nothing changed in the display of the character group, the fact that the property box size and layout changed means the html may be breaking differently (and as @davidfarmer just noted, this can happen on beta).

In any case I just added a nowrap span around the last element of the Dirichlet group and the close bracket which should prevent this from happening.

roed314:
It looks good to me.

PR #4149, opened 2020-10-21
Hmf stats fix

edgarcosta:
This enables us to run `db.reset_all_stats()` outside of app context.

@roed314 I will merge this asap, but I would appreciate a second set of eyes


roed314:
That looks reasonable.  I noticed that there was a problem with HMF, but only fixed it part way.  Thanks.

roed314:
The test failures should be resolved by #4154.

PR #4152, opened 2020-10-21
fixing configuration and command line arguments

edgarcosta:
- one can again use optional command-line arguments `--enable-profiler`
- these that precedence on whatever is read from the config file
- added the option `--disable-debug` to disable debug mode
- if there is no config file, we create one with all the default options, but the values passed via the command line optional take precedence over the default ones.
For example, running `sage -python start-lmfdb.py --disable-debug --port 5555 --color=17` for the first time will result in
```
[core]
debug = False
color = 17

[web]
port = 5555
...
```

### DIFF
--- a/lmfdb/backend/config.py
+++ b/lmfdb/backend/config.py
@@ -127,11 +127,7 @@ class Configuration(object):
             return sec, opt
 
         # 1: parsing command-line arguments
-        if  writeargstofile:
-            args = parser.parse_args()
-        else:
-            # only read config file
-            args = parser.parse_args([])
+        args = parser.parse_args()
         args_dict = vars(args)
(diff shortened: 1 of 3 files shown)


AndrewVSutherland:
If I have debug = true in config.ini using --disable-debug does not disable debug mode (but if debug = false using --debug does turn debug on as it should).  Otherwise looks good.

edgarcosta:
this should be fixed now

AndrewVSutherland:
disable-debug is still not working for me

edgarcosta:
I removed that option and just fixed a bug that happened when you import from the interactive shell.

jwj61:
This does not seem to be working.    Starting with sage -python start-lmfdb.py --debug does not put me in debug mode.

PR #4155, opened 2020-10-21
Add stale knowl review

roed314:
Stale knowls are those where the version on the production site is out of date (this happens when there is both a reviewed version and a newer beta version).


AndrewVSutherland:
This is awesome, thanks!  My only suggestion would be to sort the results by last modification date (I think it makes sense to put the earliest last modification date first, since it is the "stalest").

PR #4159, opened 2020-10-22
master --> web

AndrewVSutherland:
@edgarcosta @roed314 Are we ready to do this?  In particular, are all the stats now ready to go on prododb?

If so, please merge this PR (if not, please close it).


edgarcosta:
Finishing importing nf fields and triple checking that we have all the stats.

Issue #4165, opened 2020-10-22
Random link for Dirichlet characters is broken


AndrewVSutherland:
https://www.lmfdb.org/Character/Dirichlet/random gives an invalid label error (two in fact).

Also, three error messages are displayed when you enter the label "banana".    Only one should be displayed.

PR #4171, opened 2020-10-23
Fix the three error problem

BarinderBanwait:
cf #4165 - Still more to do regarding more informative error message, which i'm working on

### DIFF
--- a/lmfdb/characters/main.py
+++ b/lmfdb/characters/main.py
@@ -166,7 +166,9 @@ def jump(info):
     try:
         validate_label(jump_box)
     except ValueError as err:
-        flash_error("%s is not a valid label: %s.", jump_box, str(err))
+        tip_msg = "It must be of the form modulus.number, with modulus and number natural numbers"
+        flash_error("%s is not a valid label: %s.", jump_box, tip_msg)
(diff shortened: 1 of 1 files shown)


AndrewVSutherland:
There is  a minor pyflakes complaint that should be addressed:

```
./lmfdb/characters/main.py:168:5 local variable 'err' is assigned to but never used
```

AndrewVSutherland:
Looks good other than the pyflakes issue.  It might help to make the error message a little more specific (e.g. "coprime natural numbers") to explain why 4.2 and 4.5 are invalid labels.

BarinderBanwait:
i'll address the pyflakes issue as well as improving the error message; i wanted to use the regular expression model used in the number field case, which gives a dichotomy of invalid label vs label not in database, will do this today

BarinderBanwait:
The last commit makes some changes to `label_to_number`; instead of returning 0 whenever there is anything wrong with the label, it now returns a ValueError with the appropriate message for why that label is invalid. This then gets passed up the stack and flashed on the screen when the user tries to search for 'banana' or '4.2' or '4.6'.

AndrewVSutherland:
tests passed on grace

Issue #4166, opened 2020-10-22
No links in EC/Q stats tables


AndrewVSutherland:
Unlike all other stats pages in the LMFDB, the stats page

    https://beta.lmfdb.org/EllipticCurve/Q/stats

has no links to search results in any of the tables (even though it does appear to be using the new stats framework).

roed314:
It's using the count functionality from the backend, but it's not using any of the front end statistics infrastructure (the `StatsDisplay` object from `lmfdb.utils.display_stats` or the `display_stats.html` template).  We could convert it to do so (though the general infrastructure doesn't support the cool layout of the torsion structure table), or manually add the links to the `lmfdb/elliptic_curves/templates/ec-stats.html` template.

PR #4178, opened 2020-10-24
Add links in EC/Q stats table

BarinderBanwait:
This PR addresses issue #4166 ; it uses the `StatsDisplay` object suggested by @roed314. I've tried to stay true to the original stats page, though there are two differences:

1. The rank table does not contain stats on isogeny classes. Speaking with @edgarcosta yesterday, it seems this is best handled from the backend, so I'll leave this for David/Edgar.

2. The display of the torsion structures (displaying torsion groups of the same order vertically) has not been preserved. It is however now similar to tables in other stats pages.


AndrewVSutherland:
@BarinderBanwait There is a test failure at line 22 of lmfdb/elliptic_curves/test_browse_page.py that should be fixed (the error is complaining that "Distribution of rank not in the /EllipticCurve/Q/stats")

BarinderBanwait:
Yep, just looking at that now! Actually there are several other test failures as well, in `EllCurveTest`. 

But the one you mentioned, "Distribution of rank" not on the webpage: actually it is on the webpage, but it's just that now 'rank' is a knowl, so i guess it's not finding it. If I change the test to check for "Distribution of <a title="Rank of an elliptic curve over a number field [ec.rank]" knowl="ec.rank" kwargs="">rank</a>" then I imagine that will pass; is this an acceptable solution?

JohnCremona:
Yes, just change the expected string in the test to what is actually on the page now.  You can use View Source-- it has to be exact.

BarinderBanwait:
Right, thanks for confirming John, I'll change the expected string as above.

AndrewVSutherland:
The links in the torsion and sha tables don't work, I think you need to add query formatters for these (see G2C_stats in genus2_curves/main.py, for example).

BarinderBanwait:
Oh dear, I should have checked those :grimacing: taking a look now ...

BarinderBanwait:
Query formatters added, links now seem to work.

Are there unit tests to check that the links in the stats pages work? I couldn't find them anywhere. If not, it might be good to do this uniformly across all stats pages if possible

AndrewVSutherland:
Looks good, happy to merge this once tests pass.

I agree that it would be great to have a general method to test all the stats pages (maybe one link per table), perhaps something for @roed314 to think about.

PR #4180, opened 2020-10-26
to_dict: do not overwrite keys passed as input

edgarcosta:
@AndrewVSutherland, I believe you inadvertently removed this in https://github.com/LMFDB/lmfdb/commit/c12e5e3f97a1785daa06dfb304544734c1fd18a9


AndrewVSutherland:
Good catch, thanks.  Travis is on strike again but this PR is clearly safe so I will merge it.

PR #4198, opened 2020-10-29
fix db.copy_to_from_remote

edgarcosta:


AndrewVSutherland (approved):

AndrewVSutherland:
This seems to fix the problem.  I don't understand the travis failures (I just restarted the build).

edgarcosta:
The tests should pass now and copy from remote should still work

AndrewVSutherland:
This works, thanks!

Issue #4195, opened 2020-10-28
Another bogus URL causing a server error that should be a 404


AndrewVSutherland:
There are a bunch of errors in the flasklog from URL requests like

https://www.lmfdb.org/L/SymmetricPower/2/EllipticCurve/Q/21/asnjhl%27j%22yh/Reliability

which are throwing server errors.  They should return 404.

PR #4201, opened 2020-10-30
Resolves #4194

### LINKED ISSUE #4194: Limit download_qexp requests for elliptic curves
Requests like

https://www.lmfdb.org/EllipticCurve/Q/download_qexp/37.a1/9000000

(which showed up in the flasklog) put a heavy CPU load on the server and eventually lead to a 502.  I suggest capping these at 10^5.

AndrewVSutherland:


AndrewVSutherland:
Also fixes #4195. Tests pass locally and Travis is on strike so I will merge this.

Issue #4196, opened 2020-10-29
Add charpolys of Hecke operators


AndrewVSutherland:
The "Hecke characteristic polynomials" tables that appear on the newform and newspace pages currently display L-polynomials, but they should be displaying charpolys of T_p (acting on V_f), which is N_{K_f/\mathbb Q)(T-a_p), or equivalently, minpoly(a_p)^[K_f:Q(a_p)].  It probably makes sense to only do this for newforms of dimension <= 20 where we have exact a_p's.  

The charpolys could be computed on the fly, but (especially in the larger weight cases) it probably makes sense to precompute them.

BarinderBanwait:
so they can be computed from the existing a_p values, and added as a new column in the appropriate table (`db.mf_newforms` or `db.mf_hecke_nf`)? I haven't worked with the backend so far, so i'd be happy to try this

BarinderBanwait:
I've got the data file ready, but I'm having issues loading it.

I first tried reloading it to the table we created on Friday (`db.mf_hecke_charpolys`), but it complained that the types are invalid:

```
ValueError: Invalid types: p should be integer instead of smallint, charpoly_factorization should be numeric[] 
instead of jsonb
```

The first five lines of the file I have are:

```
hecke_orbit_code|p|charpoly_factorization
bigint|smallint|jsonb

201326593|2|{{{24,1},1}}
201326593|3|{{{-252,1},1}}
```

So I changed the second line of the file to `bigint|integer|numeric[], and tried reloading again, but this gave the following error:

```
DataError: malformed array literal: "{{{24,1},1}}"
DETAIL:  Unexpected array element.
CONTEXT:  COPY mf_hecke_charpolys_tmp, line 1, column charpoly_factorization: "{{{24,1},1}}"
```

At this point I created a new table with differently declared types, thinking this might be relevant:

```
db.create_table('mf_hecke_charpolys_again', search_columns={"bigint":"hecke_orbit_code", "smallint": "p",
"jsonb":"charpoly_factorization"},label_col=None)
```

However, reloading this new table with the original datafile also gave a DataError:

```
DataError: invalid input syntax for type json
DETAIL:  Expected string or "}", but found "{".
CONTEXT:  JSON data, line 1: {{...
COPY mf_hecke_charpolys_again_tmp, line 1, column charpoly_factorization: "{{{24,1},1}}"
```

[...]

Issue #4117, opened 2020-10-11
Extra parentheses


jvoight:
Trivial matter, but still: look at
  http://www.lmfdb.org/ModularForm/GL2/Q/holomorphic/81/2/c/
and scroll down to the Hecke polynomial at 7.  Why the extra parentheses?  (It's harder to read as written.)

tomgrubbmath:
It seems like this is maybe just an oversight in the interaction between the "display_hecke_polys" function in lmfdb/lmfdb/classical_modular_forms/web_space.py and the "list_factored_to_factored_poly_otherorder" function from lmfdb/lmfdb/utils/utilities.py .  I can try to fix it.

AndrewVSutherland:
This was addressed in #4217 (and we no longer show Hecke charpolys for spaces, only for newforms)

PR #4217, opened 2020-11-03
Display Hecke charpolys. Add test. PEP8 blank lines

BarinderBanwait:
This PR addresses #4196, to add Hecke characteristic polynomials to newform and newspace homepages. Until recently these were showing L-polys instead.

One test method has been added to `test_cmf.py`, with checks on 3 newforms and one newspace, all of varying dimensions. Test passes locally.

Also one minor typo on the homepage fixed.

### DIFF
--- a/lmfdb/classical_modular_forms/test_cmf.py
+++ b/lmfdb/classical_modular_forms/test_cmf.py
@@ -422,7 +422,6 @@ class CmfTest(LmfdbTest):
             assert r'0.990338' in page.get_data(as_text=True)
             assert r'0.550990' in page.get_data(as_text=True)
 
-
         for url in ['/ModularForm/GL2/Q/holomorphic/5/9/c/a/?n=2-10&m=1-6&prec=6&format=satake',
                 '/ModularForm/GL2/Q/holomorphic/5/9/c/a/2/1/',
                 '/ModularForm/GL2/Q/holomorphic/5/9/c/a/3/1/',
@@ -437,7 +436,6 @@ class CmfTest(LmfdbTest):
                 ]:
             page = self.tc.get(url)
             assert '0.00593626' in page.get_data(as_text=True)
-
 
         for url in ['/ModularForm/GL2/Q/holomorphic/31/2/c/a/?m=1-4&n=2-10&prec=6&format=satake',
                 '/ModularForm/GL2/Q/holomorphic/31/2/c/a/5/1/',
@@ -615,7 +613,6 @@ class CmfTest(LmfdbTest):
         assert '-0.498394' in page.get_data(as_text=True)
         assert '0.406839418685' in page.get_data(as_text=True)
         assert '1.2.3.c9' in page.get_data(as_text=True) # Sato-Tate group
-
 
         page = self.tc.get('/ModularForm/GL2/Q/holomorphic/download_full_space/20.5', follow_redirects = True)
         assert r"""["20.5.b.a", "20.5.d.a", "20.5.d.b", "20.5.d.c", "20.5.f.a"]""" in page.get_data(as_text=True)
@@ -891,7 +888,7 @@ class CmfTest(LmfdbTest):
 
     def test_Fricke_signs_search(self):
         r"""
-        Test that we display Fricke sings
+        Test that we display Fricke signs
         """
         page = self.tc.get('/ModularForm/GL2/Q/holomorphic/?level=15%2C20&weight=2&dim=1&search_type=List',  follow_redirects=True)
         assert 'Fricke sign' in page.get_data(as_text=True)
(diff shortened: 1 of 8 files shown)


AndrewVSutherland:
As discussed at LMFDB Friday, it would probably make sense to address #4117 in this PR as well if it is easy to do.

BarinderBanwait:
Latest commit addresses #4117

lmfdb/classical_modular_forms/test_cmf.py
```diff
@@ -937,4 +909,53 @@ def test_is_self_dual(self):
...
+                    # The following is a newspace with 9 newforms, but only 7 have been computed exactly.
+                    # This also checks correct parentheses
+
+                    '4866/2/a/' : {2: r'\( ( 1 + T )^{2} \)\( ( -1 + T )^{11} \)\( ( 1 + T )^{13} \)\( ( 1 + T )^{14} \)\( ( -1 + T )^{14} \)\( ( 1 + T )^{17} \)\( ( -1 + T )^{20} \)',
+                                   3: r'\( ( -1 + T )^{2} \)\( ( -1 + T )^{11} \)\( ( 1 + T )^{13} \)\( ( -1 + T )^{14} \)\( ( 1 + T )^{14} \)\( ( -1 + T )^{17} \)\( ( 1 + T )^{20} \)'},
+
+                    # The last one is the original issue highlighted in #4117
+
+                    '81/2/c/' : {# We want to put brackets around each monomial power for newspaces, to make it clear that we are taking a product over multiple distinct newforms
+                                 3: r'(\( T^{2} \))(\( T^{4} \))',
```
BarinderBanwait, line 950: I assumed that in this case, we *do* want parentheses around the $T^2$ and $T^4$, to make it clear that the hecke characteristic polynomial is a product coming from each of the newforms, but let me know if it should be otherwise

lmfdb/utils/utilities.py
```diff
@@ -87,7 +87,10 @@ def list_factored_to_factored_poly_otherorder(sfacts_fc_list, galois=False, vari
                     gtoprint[(val, i)] = elt/p**val
         glatex = latex(ZZpT(gtoprint))
         if  e > 1:
-            outstr += '( %s )^{%d}' % (glatex, e)
+            if len(glatex) != 1:
+                outstr += '( %s )^{%d}' % (glatex, e)
+            else:
+                outstr += '%s^{%d}' % (glatex, e)
```
BarinderBanwait, line 93: I'm not 100% sure that changing this method won't affect anything else in the codebase. It looks like it gets used in `galois_pretty_print` in the lfunctions folder, which only ends up getting used in `/TensorProducts/`, so I'm reasonably confident, but could do with a second opinion (especially since I know nothing about the `TensorProducts` area)

lmfdb/classical_modular_forms/test_cmf.py
```diff
@@ -937,4 +909,53 @@ def test_is_self_dual(self):
...
+
+                    # The following is a newspace with 5 newforms, hence the more verbose polynomials.
+                    # These newspace examples have historically also been prone to superfluous parentheses
+                    # (e.g. LMFDB Issue #4117)
+
+                    '255/1/h/' : {2: r'(\( 1 + T \))(\( 1 + T \))(\( -1 + T \))(\( -1 + T \))(\( T^{2} \))',
+                                  3: r'(\( 1 + T \))(\( -1 + T \))(\( 1 + T \))(\( -1 + T \))(\( 1 + T^{2} \))',
+                                  5: r'(\( 1 + T \))(\( -1 + T \))(\( -1 + T \))(\( 1 + T \))(\( 1 + T^{2} \))'},
+
+                    # The following is a newspace with 9 newforms, but only 7 have been computed exactly.
```
BarinderBanwait, line 941: Should we even be displaying Hecke Characteristic Polynomials in this case? There are 9 newforms, but only 7 are computed exactly; for the remaining two newforms, the characteristic polynomials are not in the database. So the displayed product of 7 polynomials is only a *factor* of the true Hecke Characteristic polynomial of this newspace, and not the whole thing? If so, either we make clear that "The following listed polynomials are only factors", or we say "The polynomials are not in the database", because some factors are missing?

lmfdb/classical_modular_forms/test_cmf.py
AndrewVSutherland: I think we should treat all factored polynomials the same and not try to show any extra structure by adding parentheses.

lmfdb/utils/utilities.py
AndrewVSutherland: I don't think TensorProducts is currently used.

lmfdb/classical_modular_forms/test_cmf.py
AndrewVSutherland: No, we should only show charpolys for spaces where we actually know what it is (which for newspaces means we need to know the charpolys for all the newforms in the space).

lmfdb/utils/utilities.py
edgarcosta: it looks good to me.

AndrewVSutherland:
This page takes a very long time to load (and does not on beta), and the resulting L-poly display looks like it is just concatenating the L-poly factorization, but it really should merge them as a multi-set (e.g. (T)...(T)(T^2)...(T^2)(T^3)...(T^3) should be T^114.

http://127.0.0.1:37777/ModularForm/GL2/Q/holomorphic/5550/2/a/

AndrewVSutherland:
I'm not convinced there is that much value to showing the char polys on the newspace pages.  Rather than figuring out how to make this work efficiently (which may require precomputing them and storing them in the database), I would suggest that only show them on the newform pages and remove the code to display them on the newspace pages. 

If someone has a compelling reason to want these charpolys on the newspace pages, we can address that in a separate PR (but I don't want to merge this PR until the code that is trying to display the charpolys on the newform pages is removed).

JohnCremona:
Sage factorization objects can be multiplied (efficiently by just adding exponents) in case that is relevant.    But I tend to agree with Drew.  The only factorization I can think of wanting to see on newspace pages would be with one factor per newform (orbit), no inner factorization.

lmfdb/classical_modular_forms/test_cmf.py
BarinderBanwait: The parentheses around T^2 and T^4 were arising because this is a newspace with two newform subspaces, and the present code was concatenating the factorisations rather than multiplying them and then factorising (as you noted below with 5550.2.a). As discussed below, this will be addressed by not showing these polynomials for newspaces.

lmfdb/utils/utilities.py
BarinderBanwait: Great, thanks!

lmfdb/classical_modular_forms/test_cmf.py
BarinderBanwait: This will be addressed by not showing Hecke charpolys for newspaces.

BarinderBanwait:
OK, I can remove the code that displays these characteristic polynomials on newspace pages, and update the tests accordingly.

BarinderBanwait:
The latest commit removes the code that displays Hecke charpolys for newspaces, so now they are only being displayed for newforms.

The original method for doing this was in `web_space.py`, where it now no longer belongs; so I ported it over to `web_newform.py`, and refactored it to reflect that we're now only calling it for one `form_label` (whereas previously we had several in the newspace case).

lmfdb/classical_modular_forms/web_newform.py
```diff
@@ -119,6 +120,77 @@ def td_wrapr(val):
...
+    The code for the table wrapping, scrolling etc. is common with many others
+    and should be eventually replaced by a call to a single class/function with
+    some parameters.
+
+    INPUT:
+
+    - ``form_label`` - a string, the label of the newform
+    - ``num_disp`` - an integer, the number of characteristic polynomials to display by default.
+    """
+    #from time import clock
```
BarinderBanwait, line 137: Does LMFDB have a position on commented out code? (Here and below.) I couldn't find this addressed in the Developer's guide. 

In general I feel that this should be avoided, although in this instance I can see how someone *could* make a case to keep this profiling code commented out

edgarcosta: I would just remove it

BarinderBanwait: Yep, that was also my inclination. They'll be gone in the next commit.

Issue #4205, opened 2020-10-31
CMF download fails for newform with very high dimension


AndrewVSutherland:
The download all data link for 983.2.c.a throws a server error (the other download links seem to work)

https://www.lmfdb.org/ModularForm/GL2/Q/holomorphic/download_newform/983.2.c.a

edgarcosta:
What should we do about it?

AndrewVSutherland:
We shouldn't offer download links that we know are not going to work, so if the issue is simply that there is too much data (as opposed to an actual bug) we should not offer that link.  I notice that the download all embeddings and Satake angle links do not throw server errors, but they also don't work in that they eventually hit a network failure (after about 100 MB).

I don't think we should be offering the option to download all the embeddings (or Satake angles) in one link, at least not when the dimension is large.  The user can always pick their favorite embedding and download the coefficient data from the embedded CMF page (but I notice there is not a Satake angle download on those pages, is that intentional?)

I think the download all stored data link on the Galois orbit pages should not include embedding data (including Satake angles), or at least not when the dimension is large.

AndrewVSutherland:
Here is another example of a download that is failing where the dimension is smaller ("only" 1360)

https://www.lmfdb.org/ModularForm/GL2/Q/holomorphic/83/6/c/a/

AndrewVSutherland:
Even after #4218 (which is on web), we seem to still be getting timeouts, e.g. on

    https://beta.lmfdb.org/ModularForm/GL2/Q/holomorphic/download_newform/101.5.i.a

I don't understand why, because the amount of data is quite small.  The download works on beta (although it took a suspiciously long time the first time I tried it).

edgarcosta:
I don't understand it also.  Maybe once in a while goes for a sequential search? When I run the `explain analyze` on proddb and on grace I get different plans to do the search, on grace it decides to do a sequential search. But also doing a sequential search on grace is much faster but in either case no reason for a timeout.

This is on proddb
```
lmfdb=# EXPLAIN ANALYZE SELECT "an", "hecke_ring_numerators", "hecke_ring_denominators", "hecke_ring_inverse_numerators", "hecke_ring_inverse_denominators", "hecke_ring_cyclotomic_generator", "hecke_ring_character_values", "hecke_ring_power_basis", "maxp" FROM "mf_hecke_nf" WHERE "hecke_orbit_code" = 549839700069 LIMIT 1;
                                                                       QUERY PLAN
[...]

PR #4218, opened 2020-11-06
don't include cc data on "all stored data to text"

edgarcosta:


edgarcosta:
This resolves #4205

edgarcosta:
The test 
```
 self.check_args("/Variety/Abelian/Fq/?g=4&p_rank=4&jacobian=not_yes&polarizable=yes", "4.2.ad_c_a_b")
```
fails. But this is not related to this PR.

@roed314, any ideas?

roed314:
Ha.  I just added data for $g=4$ and $q=2$ so we now know that `4.2.ad_c_a_b` does in fact contain a Jacobian.  I'll make a PR to fix the test.

PR #4228, opened 2020-11-11
better check for start-lmfdb.py

edgarcosta:
@jwj61, I think this should fix the issue.
Please let me know if it doesn't


jwj61:
Yes, that seems to work.

PR #4230, opened 2020-11-13
fully disable download_cc and download_satake_angles

edgarcosta:
This resolves the last bits of #4205


AndrewVSutherland:
Should the satake angles be included in the download coefficient data on the embedded newform pages (or another download option)?

edgarcosta:
"Coefficient data to text" gives you all that:
https://github.com/LMFDB/lmfdb/blob/939b2e69365d2a8599ef339ca3c32e701de9f9e9/lmfdb/classical_modular_forms/download.py#L280

Maybe we should rename the link to make it more obvious that includes everything

AndrewVSutherland:
Well I guess "Satake angles" could be considered part of "coefficient data".  I don't feel strongly about it.

Issue #4172, opened 2020-10-23
Galois orbits for Dirichlet characters


jvoight:
We should make pages for Galois orbits of Dirichlet characters!  This is how we think and use them for classical modular forms, and indeed the analogy is good: the first organizational principle is on Galois orbits, then we think about their complex embeddings.  (Anything we currently search on for Dirichlet characters is well-defined on the Galois orbit.)

BarinderBanwait:
I've got a prototype version of Galois orbits live at the pink branch; e.g. [Galois orbit 6500.el](https://pink.lmfdb.xyz/Character/Dirichlet/6500/el).

Before proceeding with a PR, could I ask for some community feedback from an end-user perspective? Is this the information we want to display on this page? Is there anything else we want to show?

It's very much a version 0.1 at the moment; I spent a few minutes clicking on random links between Dirichlet Group/Dirichlet Orbit/Dirichlet Character, and didn't find any breaks; but I can't guarantee that it's free of bugs! So please let me know of them as and when you encounter them!

Some remarks:

1. Dirichlet characters now have three parts to the url; Modulus, Orbit label, and Conrey number. However, the current functionality still works; that is, specifying the orbit label is not necessary. 

2. As discussed during one of the Friday sessions, there are two types of breadcrumbs, one for modulus > 10,000, where the orbit label isn't a crumb, and one for modulus <= 10,000, where it is.

davidfarmer:
This looks like a good start on the home page of a Galois orbit.
Some small comments:

The word "character" should be at the top of the page:
Dirichlet character orbit 6500.el

The link from a character to its orbit is in the page,
but should also be under "Related objects"

I am interested to read about the labeling.  Looking at a few
examples did not enable me to figure out how the orbits are
enumerated.

The two sentences above the table can be said more concisely.
Maybe nothing needs to be written at all?  Maybe that subtitle
should include "in the orbit"?

You said that the Dir char URL includes the orbit,
but I did not see that.  But why is that desirable?
If
Character/Dirichlet/6500/283
and
Character/Dirichlet/6500/el/283
are the same, the first seems better.  And I find the second
to be confusing, even though I understand it.

I did see the orbit in the breadcrumbs, which I thought
was good.

BarinderBanwait:
Great, thanks for your feedback @davidfarmer !

> The word "character" should be at the top of the page:
Dirichlet character orbit 6500.el

Done! Live on pink!

> The link from a character to its orbit is in the page,
[...]

PR #4231, opened 2020-11-13
Initial cleanup of web_character.py

BarinderBanwait:
First step towards LMFDB #4172


lmfdb/characters/test_characters.py
```diff
@@ -116,11 +106,11 @@ def test_dirichletcharbig(self):
             This test also makes sure the code scales a little bit.
         """
         W = self.tc.get('/Character/Dirichlet/40487/5')
-        assert '40486' in W.get_data(as_text=True)
-        assert '12409' in W.get_data(as_text=True), "log on generator"
```
BarinderBanwait, line 120: @AndrewVSutherland Since we're no longer using `WebDirichletCharacter` for modulus between 10,000 and 100,000, the `Values on generators` doesn't appear on this page, so this test failed. I therefore changed it to check the field of values; but do let me know if I misunderstood something, or if the behaviour should be otherwise.

lmfdb/characters/test_characters.py
```diff
@@ -169,5 +159,5 @@ def test_dirichletchar9999lfunc(self):
     def test_dirichletchar99999999999999999lfunc(self):
         """ Check Dirichlet character with very large modulus"""
         W = self.tc.get('/Character/Dirichlet/99999999999999999999/2')
-        assert parity_string(-1) in W.get_data(as_text=True) and '536870912' in W.get_data(as_text=True)
```
BarinderBanwait, line 172: Since we're removing the `Galois orbit` section in the large modulus case, this test failed, so as above I changed it to something that *is* there. Again do let me know if it should be otherwise.

lmfdb/characters/test_characters.py
AndrewVSutherland: Actually, I realized that we can keep the values on generators and the values, because Pari now supports Conrey characters and can compute these very efficiently.  To compute the value of chi_q(m,n) you can use:

```
 def conreyangle(q,m,n):
     return Pari("G=znstar(%d,1); chareval(G,znconreylog(G,%d),%d)"%(q,m,n))
```

which will return a rational number r such that chi_q(m,n) = exp(2*pi*r).  This is very fast even for 10 or 20 digit values of q.

To get generators you can use `Integers(q).unit_gens()`.

With this we can easily keep the values on generators and values sections for all Dirichlet characters, as well as the ability to compute values, just as we do for characters in the database.

lmfdb/characters/test_characters.py
AndrewVSutherland: I agree with removing the Galois orbit section for characters of modulus > 10000 which are not in the database, since we don't actually know the orbit label.

AndrewVSutherland:
One other comment, I think we should just use Pari to compute everything about Conrey characters (we only need to know how to compute the conductor, order, parity, isreal, and the inducting primitive character), all of which Pari can do quite efficiently).  This would let us remove the dependence on the dirichlet_conrey package, which has always been a headache.  See https://pari.math.u-bordeaux.fr/dochtml/html/Arithmetic_functions.html for documentation.

JohnCremona:
I agree that using pari for d.c. chars is a good idea. (I discovered recently that a trivial intervention is needed to pip install Bober's package in sage 9.2 anyway.)

To avoid multiple rather arcane calls as in Drew's earlier comment it would nice if someone wrote a few little wrapper functions for what we'll need.

lmfdb/characters/test_characters.py
BarinderBanwait: OK thanks for the feedback @AndrewVSutherland , I'll aim to get `Values on generators` as well as `First values` back onto the pages for large modulus via Pari.

BarinderBanwait:
OK sounds good - I'll aim to replace all the dirichlet_conrey stuff with direct calls to the `znconrey` methods, complete
with all the pari wrapper functions we'd need. :toolbox:

lmfdb/verify/mf_newspaces.py
```diff
@@ -330,7 +341,7 @@ def check_dims(self, rec, verbose=False):
         if rec['level'] < 3:
             dirchar = rec['level']
         else:
-            dirchar = DirichletGroup_conrey(rec['level'])[rec['conrey_indexes'][0]].sage_character()
+            dirchar = get_dirchar(rec['level'], rec['conrey_indexes'][0], rec['char_order'])
```
BarinderBanwait, line 344: test

lmfdb/verify/mf_newspaces.py
```diff
@@ -318,10 +329,10 @@ def check_Skchi_dim_formula(self, rec, verbose=False):
...
```
BarinderBanwait, line 332: The changes in this file - here and below  - are untested, since I didn't know how to test them. I'd be grateful if someone could show me how to run these verification scripts.

BarinderBanwait:
OK, so I've removed the dirichlet_conrey package as indicated above; most of the Pari calls went into the existing `ConreyCharacter` class in the `TinyConrey` module (a few are in a new `PariConreyGroup` class in the same module; this is a bare-bones analogue of the previous `DirichletGroup_conrey`).

This PR has snowballed somewhat, so I feel an overview of what it's now doing is in order.

The main user-facing changes are as follows:

1. Galois orbits have been removed for modulus >=10,000.
2.  Values on Generators, and the Values/First Values table have been added for modulus >=100,000.
3. The label of the induced character for modulus >= 100,000 is now being computed and is showing (previously it said 'None'; e.g. [here](https://www.lmfdb.org/Character/Dirichlet/100001/23)).
4. Compute buttons for Values/Gauss/Jacobi/Kloosterman sums now work (previously they did nothing).
5. The SageMath code stubs to compute the character now explicitly specifies the base ring; not doing this can sometimes yield a ValueError when running the previous code in sage (e.g. for 91.3; the database considers the multiplicative order as 6, but just calling DirichletGroup(91) yields base_ring as CyclotomicField(12), so the call to DirichletCharacter was failing; this was noticed when fixing unit test failures).

The other non-user-facing changes are that web_character has been cleaned up a bit (which was the original goal), and some tests have been added.

The `dirichlet_conrey` package was being used in the verify directory, which I don't know anything about. I've flagged these points as comments in this PR, since I haven't been able to test these changes.

AndrewVSutherland (approved):

AndrewVSutherland:
It looks like the 10000 cutoff is off by 1, we have characters of modulus <= 10,000 in the database, but code seems to think we only have them up to < 10,000.

AndrewVSutherland:
Also the values table is a bit wide for the characters of modulus > 10,000, e.g. 

http://127.0.0.1:37777/Character/Dirichlet/10001/7

Now the we have the ability to compute arbitrary values on these pages, I think just showing, say, the first 10 values in addition to the value at -1 would be enough.

AndrewVSutherland:
Looks good, happy to merge once tests pass.

BarinderBanwait:
Latest commit addresses the two issues raised in the last two comments.

@AndrewVSutherland David Roe showed me how to run the verification scripts, but that test I changed is not passing, and it looks like it might go back to an issue with constructing the sage character; so until I've got to the bottom of that it's probably best to hold-off merging this PR.

lmfdb/verify/mf_newspaces.py
BarinderBanwait: Thanks @roed314 for showing me how to run these (and fixing them along the way)! The verification is not passing right now, which I'm getting to the bottom of

BarinderBanwait: Latest commit c321e21 resolves the verification test failures. For details of the issue see my comment below that commit.

BarinderBanwait:
TL;DR: Verification tests now pass. @AndrewVSutherland  I believe this is good to be merged now.

More details:

There was an issue with the way I was constructing the sage character. I based it off the sage generating code which shows when you click "Show commands for: SageMath". The list of genvalues corresponds to the numerators of the fractions that appear in the values on generators. But these values can't be used verbatim when constructing the character in sage; they need to be adjusted to account for the denominator; in the db this is the displayed order, but in sage this is `H.zeta_order()`, and they can well be different. This meant that the gauss/jacobi/kloosterman sums might be wrong.

This adjustment was being done in Bober's package, which is how the verification tests were passing, so I took a look at his code and copied his logic into a new `get_sage_genvalues` method.

The two verification tests then passed; in the interest of time I set ratio as `float(0.0001)`, which corresponded to 127 test cases; `check_Skchi_dim_formula` passed after 43 seconds, and `check_dims` passed in about 900 seconds.

The various examples that were failing while debugging the verification test failures have now been encoded into a new unit test `test_sage_code_gens`, to ensure correct sage character in future.

Issue #4235, opened 2020-11-16
Malformed latex equations for elliptic curves over number fields


JohnCremona:
Example:  https://www.lmfdb.org/EllipticCurve/2.2.5.1/2581.2/a/1 has malformed latex equation. In this case it is because the generator \phi is concatenated with x, and this error is in the database entry so it is not a code problem. 

I think this was introduced during my most recent database upload.  To save space I removed spaces in the latex string (see line 925 of https://github.com/JohnCremona/ecnf-data/blob/master/sage/files.py) which in some cases causes problems.  In this case the stored string is

"y^2+xy+\\left(\\phi+1\\right)y=x^{3}+\\phix^{2}-9\\phix-4\\phi-10"

while Sage's latex(E) actually gives

y^2 + x y + \left(\phi + 1\right) y = x^{3} + \phi x^{2} - 9 \phi x - 4 \phi - 10

This will be tedious but trivial to fix (in the database).  Meanwhile I think that a code fix replacing each 'x' by '{x}' and similarly for 'y' should fix it at least temporarily, and probably faster.

JohnCremona:
OK, that does work.  Compare the above link with http://localhost:37777/EllipticCurve/2.2.5.1/2581.2/a/1.  For an example to show that braces around y are also needed, look at https://www.lmfdb.org/EllipticCurve/2.2.5.1/31.1/a/5 and http://localhost:37777/EllipticCurve/2.2.5.1/31.1/a/5 .  (This is not specific to field 2.2.5.1.)  PR coming up.

JohnCremona:
I have corrected the data one beta for 2.2.5.1 and also realised that this *is* the only field affected, since it's the only one for which the generator has a name which is a macro (\phi). Almost all other fields use a as generator (and i for 2.0.4.1) which don't cause problems when concatenated with x.

Please copy ec_nfcurves to prod, and them close this issue.  Thanks (and apologies for the error).

AndrewVSutherland:
Copying now.

AndrewVSutherland:
Done.

PR #4236, opened 2020-11-16
tweak latex string for ECNF equations to work around issue with db data

JohnCremona:
This is a small code change to fix issue #4235 -- see there for a coupld of cases to test.

I will also correct the stored latex strings.

### DIFF
--- a/lmfdb/ecnf/WebEllipticCurve.py
+++ b/lmfdb/ecnf/WebEllipticCurve.py
@@ -370,9 +370,9 @@ class ECNF(object):
 
         # The equation is stored in the database as a latex string.
         # Some of these have extraneous double quotes at beginning and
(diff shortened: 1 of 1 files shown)


JohnCremona:
I have just fixed the data on beta for field 2.2.5.1 only.  I am now going to apply the same fix to all the rest, after which  -- *and after the table ec_nfcurves has been copied to prod* -- this PR will be redundant.  So hold off.

JohnCremona:
I'm closing this as it is not needed: I have corrected the data one beta for 2.2.5.1 and also realised that this *is* the only field affected, since it's the only one for which the generator has a name qhich is a macro (\phi).  Almost all other field use a as generator (and i for 2.0.4.1) which don't cause problems when concatenated with x.

PR #4240, opened 2020-11-18
fix EC/Q completeness knowl display

### LINKED ISSUE #4239: Elliptic curve completeness information bug
From the feedback page:

I tried to go to the page 

https://www.lmfdb.org/EllipticCurve/Q/Completeness

 and it gave the error "This information can only be displayed with a more recent version of the LMFDB."

(This error comes up if the summary function is not available, so it looks like a coding bug.)

JohnCremona:
This fixes #4239.  Compare http://localhost:37777/EllipticCurve/Q/Completeness and http://localhost:37777/knowledge/show/dq.ec.extent with http://www.lmfdb.org/EllipticCurve/Q/Completeness and http://www.lmfdb.org/knowledge/show/dq.ec.extent


JohnCremona:
I forgot to add a test so this will not happen again.  Coming up...

JohnCremona:
I think @jwj61 merged this before I added the test.  Since tests pass, I am merging again.

Issue #4248, opened 2020-11-26
Explaining Q_p(\sqrt(*)), Q_p(\sqrt(p*)), etc.


edgarcosta:
I couldn't find an explantion for the notation `Q_p(\sqrt(*))`, `Q_p(\sqrt(p*))`, etc.
We should consider adding an explanation for this notation.

(This was brought to my attention by Daniel Hast, which was triggered by a question on [Math.SE](https://math.stackexchange.com/questions/3920934/understanding-some-square-root-notation-in-lmfdb-occuring-in-the-definition-of-a) expressing confusion over this notation.)

jwj61:
I think there is a typesetting issue with the issue, but it seems to be about $\Q_p(\sqrt{p*})$ and $\Q_p(\sqrt{*}$. 

The knowl for discriminant root field already explained the situation in detail, but used u to denote a non-square mod p instead of * (with an analogue for p=2).  That was an effort to make the knowl not be LMFDB specific.  Of course, the field itself is a knowl telling the user exactly what field it is.   

The use of * felt right when making the JR local database (where it is , and is used in the LMFDB for consistency.

I just added an explicit mention that * is a stand in for u in the knowl.  If this is a good idea, then someone can mark the knowl as reviewed.

edgarcosta:
I have reviewed it

rbommel:
The knowl looks good, but it seems to me that this doesn't solve the problem of the user. I can see that the * in question also occurs at the discriminant root field, but if I am reading the "Intermediate fields" section, how am I supposed to know that I should be looking at the discriminant root field to find out what the * stands for?

JohnCremona:
I don't like the use of * in the name of a field at all, it looks too much like a wildcard.  For example take the page https://www.lmfdb.org/LocalNumberField/2.2.0.1 where you see the * in the title.  Here the discriminant root field and the unramified subfield are the same field.  The knowl for one, but not the other, explains what * means, but I don't think that the user should have to click on that one specific knowl to get an explanation of *.  For the field's nickname cannot we just call it Q_2(\sqrt{-3})?

AndrewVSutherland:
I tend to agree with @JohnCremona; at a minimum I think it would be better to use u rather than *, but I see no harm and some benefit in picking a particular u (this will help the novice user and the expert will know that the choice was arbitrary).

jwj61:
In a way, * is a wildcard since it can be any non-square unit.

I will change it to pick particular values of *.

[...]

PR #4249, opened 2020-11-27
Fix issue #4248

jwj61:
On local field pages, this replaces the use of * with an integer when describing a quadratic extension of Qp.

From
  http://127.0.0.1:37777/LocalNumberField/?hst=List&n=2&p=2&search_type=List
and
  http://127.0.0.1:37777/LocalNumberField/?hst=List&n=2&p=7&search_type=List
you can click on the fields to see p=2 and p>2.

### DIFF
--- a/lmfdb/local_fields/main.py
+++ b/lmfdb/local_fields/main.py
@@ -4,7 +4,7 @@
 
 from flask import render_template, request, url_for, redirect
 from sage.all import (
-    PolynomialRing, QQ, RR, latex, cached_function, Primes, kronecker, ZZ)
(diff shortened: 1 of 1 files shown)


JohnCremona:
You could use the method quadratic_nonresidue, i.e. return int(Integers(p).quadratic_nonresidue()) for odd p.  But it would not be prime, it just returns the least nonsquare. Perhaps that must always be prime though...

jwj61:
Yes, it has to be prime.  I imagine it does the same algorithm, but of course, your suggestion is shorter.

PR #4252, opened 2020-12-02
307 for random pages

edgarcosta:
Various random routes are broken on beta, as apache caches the redirect, thus, you get always the same object.
I tried to make sure that every random route has a 307 as its redirect code.

### DIFF
--- a/lmfdb/abvar/fq/main.py
+++ b/lmfdb/abvar/fq/main.py
@@ -709,7 +709,7 @@ def by_label(label):
 def random_class():
     label = db.av_fq_isog.random()
     g, q, iso = split_label(label)
-    return redirect(url_for(".abelian_varieties_by_gqi", g=g, q=q, iso=iso))
+    return redirect(url_for(".abelian_varieties_by_gqi", g=g, q=q, iso=iso), 307)
 
 @abvarfq_page.route("/interesting")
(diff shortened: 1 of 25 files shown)


SamSchiavone (approved):

edgarcosta:
looks like we also need to disable cache:
```
    response = make_response(redirect(url_for(".by_label", label=label), 307))
    response.headers['Cache-Control'] = 'no-cache, no-store'
    return response
```

AndrewVSutherland:
I fixed the travis failures (which had nothing to do with this PR).  Is this PR ready to merge?  I wasn't clear whether your content from 9 days was suggesting another commit was forthcoming or not.

edgarcosta:
This is progress, but will still not fix the fact that on beta.lmfdb.org the random button will always give you the same object unless you set  `Cache-Control` to `no-cache, no-store`.

I will add a commit later on in the week to take care of that.

edgarcosta:
I used the wrapper that we can easily redirect with cache disabled.

Locally, you won't be able to test this, but the only effect will be on beta, where clicking the random button will start to give different objects.

Tests pass on grace.

AndrewVSutherland:
Can I merge this?

edgarcosta:
yes

PR #4254, opened 2020-12-03
Fix smallint typecasting

roed314:


JohnCremona:
Thanks -- I will test to see if this solves the issues I was having.

JohnCremona:
That works!  Please merge once tests pass.  But I still have an issue that entering a number in the "cyclic isogeny degree" box does not work (all curves are returned).  I was expecting this fix to sort that out too?

PR #4257, opened 2020-12-04
Fix python 2->3 problem in verify

roed314:


BarinderBanwait:
Aah, this change is also in my PR #4231

edgarcosta:
in light fo that comment, I will close this PR

PR #4259, opened 2020-12-05
use galt

edgarcosta:


roed314:
Looks good to me.  Feel free to merge once tests pass.

jwj61:
Is this fixing a bug?

It has been a while, but I think we may have been preparing to delete the galt column since it is redundant.  I don't know that it is currently used anywhere.

edgarcosta:
@jwj61, we are trying to avoid timeouts triggered by that query.

For certain groups, Postgres decides to do a sequential query, instead of using the query, and some group pages timeout at random on www.lmfdb.org. If we query on degree and galt, that doesn't seem the be the case anymore.
Furthermore, when both searches use the respective index, the one on degree and galt outperforms the one on the galois_label.

@AndrewVSutherland also suggested removing that if statement completely.

AndrewVSutherland:
@jwj61 On a fairly regular basis (every few days or so), we have been seeing timeouts in the flasklog on Galois group pages, e.g. this morning the page https://www.lmfdb.org/GaloisGroup/6T11 too more than 30 seconds to load (it is essentially instantaneous now because the data it needs has been cached).  After digging into the query timins, Edgar concluded that it was the search on nf_fields by galois_label that was causing the problem, apparently even though there is an index on this field, it isn't always being used in this query (possibly this is specific to the exists method, I don't know).

Another option would be to remove the query entirely and always show the "Number fields with this Galois group" link, leaving it to the user to discover whether there actually are any -- if the link leads to a search results page with no results I don't think that is a bad thing, in fact it is arguably a good think because it makes it more obvious that there actually are not any (if the link is not there the user isn't likely to conclude there are no number fields in the database with that Galois group, even though that is what the absence of the link means).

BTW, we are noticing issues like this (and other error-handling issues we have fixed recently) only because we recently added automatic alerts that send an email to me, Edgar, and David, every time the flasklog on prodweb1 or prodweb2 becomes non-empty -- we can add you to the list if you would like (but be prepared for a lot of emails).

jwj61:
OK.  Thanks for the explanation.

I get enough email right now, so I will hold off on getting more.

PR #4263, opened 2020-12-10
Change description of class group data

jwj61:
This just changes text on

  http://localhost:37777/NumberField/QuadraticImaginaryClassGroups
  http://beta.lmfdb.org/NumberField/QuadraticImaginaryClassGroups

Now throughout, d<0 is the discriminant of a field.  It is still a bit awkward because the filenames are based on -d.


JohnCremona:
In the new text can we always refer to |d| and not -d (since d is now negative)?

jwj61:
You are right.  I made the change.

PR #4265, opened 2020-12-11
remove plot route from EC/Q

AndrewVSutherland:
This will eliminate some flasklog errors that keep showing up (there is a bot somewhere that keeps hitting this obsolete route).


JohnCremona:
I had also deleted this obsolete route in my working branch. I didn't know it caused problems. It has been obsolete for a very long time.  I'll merge these changes.

PR #4269, opened 2020-12-11
being able to use any variable name in the search box

edgarcosta:
with this PR we may use any variable to specify our polynomial in the jump box for number fields, before we could only use `X` or `x`

### DIFF
--- a/lmfdb/backend/encoding.py
+++ b/lmfdb/backend/encoding.py
@@ -14,7 +14,11 @@ from psycopg2.extras import Json as pgJson
 from psycopg2.extensions import adapt, ISQLQuote
 try:
     from sage.all import ceil
-    from sage.rings.complex_number import ComplexNumber
+    try:
+        # this fails in sage 9.3
+        from sage.rings.complex_mpfr import ComplexNumber
+    except ImportError:
+        from sage.rings.complex_number import ComplexNumber
     from sage.rings.complex_field import ComplexField
     from sage.rings.real_mpfr import RealLiteral, RealField, RealNumber
     from sage.rings.integer import Integer
--- a/lmfdb/elliptic_curves/elliptic_curve.py
+++ b/lmfdb/elliptic_curves/elliptic_curve.py
@@ -1,9 +1,7 @@
 # -*- coding: utf-8 -*-
 import ast
-import os
 import re
 from six import BytesIO
-import tempfile
 import time
 
 from flask import render_template, url_for, request, redirect, make_response, send_file, abort
--- a/lmfdb/number_fields/number_field.py
+++ b/lmfdb/number_fields/number_field.py
@@ -120,7 +120,8 @@ def poly_to_field_label(pol):
     try:
         wnf = WebNumberField.from_polynomial(pol)
         return wnf.get_label()
-    except:
+    except Exception:
+        raise
         return None
 
 
--- a/lmfdb/number_fields/number_field.py
+++ b/lmfdb/number_fields/number_field.py
@@ -746,7 +746,7 @@ def download_search(info):
 def number_field_jump(info):
     query = {'label_orig': info['jump']}
     try:
-        parse_nf_string(info,query,'jump',name="Label",qfield='label')
+        parse_nf_string(info, query, 'jump',name="Label", qfield='label')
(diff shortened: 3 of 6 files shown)


jwj61:
Why bother with the change?  The example given for the search uses x, so why would someone think they should use something like k?

roed314:
You might be pasting in a polynomial from somewhere.  Manually changing all the occurrences of the variable to x is annoying.

edgarcosta:
That is indeed the use that I have in mind, I have too many times substituted the variable by hand.
The hint is a polynomial in the variable x, but in my mind, I simply see that I'm allowed to give polynomial.

jwj61:
I would have thought that nowadays, changing the variable in a polynomial was a basic skill (I would do it in gp).

The current code now rejects x (or 3x+2) for generating Q, and rejects polynomials in the variables q or Q.

edgarcosta:
@roed314 and I have updated the code to first try to convert it to a polynomial before anything else.

Has a side effect now `junk` or `banana` is interpreted as polynomial `x`. Is this okay?

We think that `banana^2 - 5` is okay, but perhaps we want to treat the cases above differently?

jwj61:
Looks ok to me (although it breaks the banana test by making banana a legal entry).

edgarcosta:
And are you okay with that behavior?

See what I said above.
> Has a side effect now `junk` or `banana` is interpreted as polynomial `x`. Is this okay?
> 
> We think that `banana^2 - 5` is okay, but perhaps we want to treat the cases above differently?

I updated the test, from testing `junk` to `junk.5`.

JohnCremona:
Is it too late to suggest restricting to a single letter (upper or lower case). I can well imagine people trying X.  Longer strings I cannot see much benefit in as surely mathematicians only ever use single characters?

JohnCremona:
I am reminded of a graffito on a desk when I was a student:

\int 1/cabin d cabin = log cabin +C = houseboat

jwj61:
@JohnCremona the original accepted x or X so would not have needed a change.

On one hand, I am not sure of the utility of some of the generalization: I think x_a is ok as a variable name, but not x1 or x_1.  I am not arguing for the latter two, but they would be more useful than x_a.

There are some edge cases which are questionable, but it is so unlikely that someone would hit them that I am reluctant to point them out (qi^2-2 goes to Q(sqrt(2)) but qi goes to Q(i), so qi is sometimes a variable and sometimes a nickname; similarly 3x and x3 do the same thing, but qsqrt3 and 3qsqrt do not).

Having all of that in mind, as I said above, I think it is ok as is.  While @AndrewVSutherland said yesterday that all search boxes should fail on banana, I take that as a joke, so it is ok for this search box to have banana take one to Q.

roed314:
The code in `coeff_to_poly` doesn't currently accept `x_1` as a variable name, but I think it should.  A compromise if we want to rule out banana is to prohibit multiletter variable names only when the polynomial is just the generator of the polynomial ring (so there are no arithmetic operations involved).  If someone types in `tau^2 +1` it seems pretty clear what they're trying to do.  Similarly, I think if someone types in `qi^2-5` they're using a strange variable number but they probably want `Qsqrt5` rather than `Qi`.

jwj61:
The more I have thought about this, the more I think you should just allow single character names for variables.  It is starting to sound complicated, which means we would need a knowl to explain the nuances of the search box which just seems worse all around.

Allowing any single letter may not cover every instance which will ever come up for people, but it should cover almost all of them and they should be thankful we allow polynomials in this search box at all.

AndrewVSutherland:
I agree with @jwj61 that we should stick to single letter variable names; in particular, I don't think "banana" should be a valid way to specify the field Q.

JohnCremona:
This was my suggestion a few comments back so +1 from me

AndrewVSutherland:
Sorry for the misattribution @JohnCremona, I must have been too distracted by your houseboat joke to notice that you already suggested this.

edgarcosta:
I can content myself with single-letter variable names.

edgarcosta:
I have updated the PR accordingly

jwj61:
Now this has conflicts

edgarcosta:
fixed

PR #4272, opened 2020-12-12
avoid displaying "There are no characteristic polynomials of Hecke operators in the database"

edgarcosta:
At the moment in CMF's when we don't have exact data we say at the end of the page:
"There are no characteristic polynomials of Hecke operators in the database", see for example, https://beta.lmfdb.org/ModularForm/GL2/Q/holomorphic/671/2/i/a/

This PR removes that whole section if we don't have any data to display, as we do for Hecke cutters and twists.

I also moved the function inside the class


roed314:
```
>   {% if newform.display_hecke_char_polys() is not None %}
E   jinja2.exceptions.TemplateAssertionError: no test named 'None'
```
I think you're supposed to use `none` in jinja.

edgarcosta:
The tests pass now :)

Issue #4275, opened 2020-12-13
Some Dirichlet character pages are failing to load


AndrewVSutherland:
The page https://www.lmfdb.org/Character/Dirichlet/947/934 is timing out.  Some similar pages such as https://www.lmfdb.org/Character/Dirichlet/947/933 and https://www.lmfdb.org/Character/Dirichlet/947/935 work but are slow too load and the knowl for the fixed field does not work.  I believe this is due to some of the recent changes that were made #4231 -- @BarinderBanwait can you take a look at this?

Below is the trace back from the page that is failing to load.  the failure is inside the call to "zeta_order" on line 156 of https://github.com/LMFDB/lmfdb/blob/master/lmfdb/characters/TinyConrey.py.  I don't think that call should be taking any time, but if Sage is doing something silly we should just compute zeta_order directly.  I confess it's not clear to me why we are using Sage DirichletGroup and Sage characters at all (it appears they are being used in just 2 places).

```
Traceback (most recent call last):
  File "/home/sage/sage-9.1/local/lib/python3.7/site-packages/flask/app.py", line 2447, in wsgi_app
    response = self.full_dispatch_request()
  File "/home/sage/sage-9.1/local/lib/python3.7/site-packages/flask/app.py", line 1952, in full_dispatch_request
    rv = self.handle_user_exception(e)
  File "/home/sage/sage-9.1/local/lib/python3.7/site-packages/flask/app.py", line 1821, in handle_user_exception
    reraise(exc_type, exc_value, tb)
  File "/home/sage/sage-9.1/local/lib/python3.7/site-packages/flask/_compat.py", line 39, in reraise
    raise value
  File "/home/sage/sage-9.1/local/lib/python3.7/site-packages/flask/app.py", line 1950, in full_dispatch_request
    rv = self.dispatch_request()
  File "/home/sage/sage-9.1/local/lib/python3.7/site-packages/flask/app.py", line 1936, in dispatch_request
    return self.view_functions[rule.endpoint](**req.view_args)
  File "/home/lmfdb/lmfdb-git-web/lmfdb/characters/main.py", line 367, in render_Dirichletwebpage
    webchar = make_webchar(args)
  File "/home/lmfdb/lmfdb-git-web/lmfdb/characters/main.py", line 313, in make_webchar
    return WebDBDirichletCharacter(**args)
  File "/home/lmfdb/lmfdb-git-web/lmfdb/characters/web_character.py", line 925, in __init__
[...]

PR #4283, opened 2020-12-14
Change base_ring parameter to always be even (or 1)

BarinderBanwait:
This PR addresses #4275

### DIFF
--- a/lmfdb/characters/TinyConrey.py
+++ b/lmfdb/characters/TinyConrey.py
@@ -156,7 +156,7 @@ class ConreyCharacter(object):
         return 1 if self.modulus <= 2 else lcm(2,order)
 
     def sage_character(self, order, genvalues):
-        H = DirichletGroup(self.modulus, base_ring=CyclotomicField(order))
+        H = DirichletGroup(self.modulus, base_ring=CyclotomicField(self.sage_zeta_order(order)))
         M = H._module
         order_corrected_genvalues = get_sage_genvalues(self.modulus, order, genvalues, self.sage_zeta_order(order))
         return DirichletCharacter(H,M(order_corrected_genvalues))
--- a/lmfdb/characters/TinyConrey.py
+++ b/lmfdb/characters/TinyConrey.py
@@ -150,7 +150,15 @@ class ConreyCharacter(object):
         return Rational(pari("chareval(%s,znconreylog(%s,%d),%d)"%(self.G,self.G,self.number,x)))
 
     def gauss_sum_numerical(self, a):
-        return pari("znchargauss(%s,%s,a=%d)"%(self.G,self.chi_pari,a))
+        # There seems to be a bug in pari when a is a multiple of the modulus,
+        # so we deal with that separately
+        if self.modulus.divides(a):
+            if self.conductor() == 1:
+                return euler_phi(self.modulus)
+            else:
+                return Integer(0)
+        else:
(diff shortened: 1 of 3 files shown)


lmfdb/characters/TinyConrey.py
```diff
@@ -150,13 +150,21 @@ def conreyangle(self,x):
         return Rational(pari("chareval(%s,znconreylog(%s,%d),%d)"%(self.G,self.G,self.number,x)))
 
     def gauss_sum_numerical(self, a):
-        return pari("znchargauss(%s,%s,a=%d)"%(self.G,self.chi_pari,a))
+        # There seems to be a bug in pari when a is a multiple of the modulus,
+        # so we deal with that separately
```
BarinderBanwait, line 154: Currently, none of the four compute buttons for Values/Gauss/Jacobi/Kloosterman sums are working on lmfdb.org, but they are working on beta.lmfdb.org; compare e.g. [live](https://www.lmfdb.org/Character/Dirichlet/16/3) to [beta](https://beta.lmfdb.org/Character/Dirichlet/16/3)

I noticed that I could similarly 'crash' my local lmfdb by going to 16.15 and asking for the Gauss sum at 16. This should return 0, but instead just hangs, and thereafter none of the buttons work. Digging a bit further, it looks like it might be an issue with pari:

```
? [g,chi] = znchar(Mod(15,16));
? znchargauss(g,chi,16)
  ***   at top-level: znchargauss(g,chi,16)
  ***                 ^---------------------
  *** znchargauss: the PARI stack overflows !
  current stack size: 8000000 (7.629 Mbytes)
  [hint] set 'parisizemax' to a non-zero value in your GPRC

  ***   Break loop: type 'break' to go back to GP prompt
```
This latest commit gets around this pari stack overflow

lmfdb/characters/TinyConrey.py
```diff
@@ -150,7 +150,15 @@ def conreyangle(self,x):
...
+        if self.modulus.divides(a):
+            if self.conductor() == 1:
+                return euler_phi(self.modulus)
+            else:
+                return Integer(0)
+        else:
+            return pari("znchargauss(%s,%s,a=%d)"%(self.G,self.chi_pari,a))
```
BarinderBanwait, line 161: Currently, none of the four compute buttons for Values/Gauss/Jacobi/Kloosterman sums are working on lmfdb.org, but they are working on beta.lmfdb.org; compare e.g. [live](https://www.lmfdb.org/Character/Dirichlet/16/3) to [beta](https://beta.lmfdb.org/Character/Dirichlet/16/3)

I noticed that I could similarly 'crash' my local lmfdb by going to 16.15 and asking for the Gauss sum at 16. This should return 0, but instead just hangs, and thereafter none of the buttons work. Digging a bit further, it looks like it might be an issue with pari:

```
? [g,chi] = znchar(Mod(15,16));
? znchargauss(g,chi,16)
  ***   at top-level: znchargauss(g,chi,16)
  ***                 ^---------------------
  *** znchargauss: the PARI stack overflows !
  current stack size: 8000000 (7.629 Mbytes)
  [hint] set 'parisizemax' to a non-zero value in your GPRC

  ***   Break loop: type 'break' to go back to GP prompt
```
This latest commit gets around this pari stack overflow

edgarcosta:
Curious....both servers are running the same sage version.

Issue #4268, opened 2020-12-11
Allow multiplication and exponentiation in input boxes


roed314:
For integer search boxes, it would be nice if we allowed expressions like `2^5*17` in integer inputs.  This can be achieved by editing line `344` of `lmfdb/utils/search_parsing.py` to change
```
def parse_ints(inp, query, qfield, parse_singleton=int):
```
to 
```
def parse_ints(inp, query, qfield, parse_singleton=my_parser):
```
and write a function that can parse multiplication and exponentiation.  If you use `sage_eval`, make sure you run the input through a regular expression (eg. only allow digits, * and ^) first so that it's safe.

jwj61:
Making that one change should come with a basic arithmetic parser, or regex test with sage_eval (which I assume is essentially the same as taking ZZ(input)?).  We certainly don't want a bunch of people writing their own parsers.

AndrewVSutherland:
@jwj61 Agreed, the implementation should do a regex match that only allows digits and multiplicative operators and then call sage eval (we don't want to support additive operators to avoid bad interactions with ranges, but almost all of our integer inputs are multiplicative (conductor, discriminant, order, ....)

AndrewVSutherland:
Addressed by #4268 and now live.

AndrewVSutherland:
Reopening this issue because there is a problem in the way malformed expressions are handled, typing "3*" into an integer input now throws a server error, but it should be handled the same way as "banana" or any other invalid input.  I'm guessing the regex match is not requiring a full match (which it should).

roed314:
I think it's easier to just catch the error than to write a regex that handles all the right syntax.

edgarcosta:
Yeah, I would also go for the catch syntax error

PR #4288, opened 2020-12-16
Allow multiplication and exponentiation in input boxes, 4268

tahulse:
Modified search_parsing.py to allow inputs involving multiplication and exponentiation. This uses regex to verify the inputted text only uses [0-9], (),^,* symbols and then attempts a sage_eval(). If sage_eval() succeeds, the search is run with that, otherwise an error is displayed.

See issue #4268

### DIFF
--- a/lmfdb/classical_modular_forms/test_cmf.py
+++ b/lmfdb/classical_modular_forms/test_cmf.py
@@ -661,6 +661,10 @@ class CmfTest(LmfdbTest):
             print("Connecting with magma.maths.usyd.edu.au timed out")
             print(err)
 
+    def test_expression_level(self):
+        # checks we can search on 2*7^2
+        page = self.check('/ModularForm/GL2/Q/holomorphic/?hst=List&level=2*7%5E2&search_type=List', '98.2.a.a')
+
     def test_download_search(self):
         page = self.tc.get('/ModularForm/GL2/Q/holomorphic/?Submit=sage&download=1&query=%7B%27level_radical%27%3A+5%2C+%27dim%27%3A+%7B%27%24lte%27%3A+10%2C+%27%24gte%27%3A+1%7D%2C+%27weight%27%3A+10%7D&search_type=Traces', follow_redirects = True)
         assert '5.10.a.a' in page.get_data(as_text=True)
--- a/lmfdb/classical_modular_forms/test_cmf.py
+++ b/lmfdb/classical_modular_forms/test_cmf.py
@@ -663,7 +663,7 @@ class CmfTest(LmfdbTest):
 
     def test_expression_level(self):
         # checks we can search on 2*7^2
-        self.check('/ModularForm/GL2/Q/holomorphic/?hst=List&level=2*7%5E2&search_type=List', '98.2.a.a')
+        self.check_args('/ModularForm/GL2/Q/holomorphic/?hst=List&level=2*7%5E2&search_type=List', '98.2.a.a')
 
     def test_download_search(self):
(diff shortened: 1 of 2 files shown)


JohnCremona:
I am very nervous about applyong sage_eval to input text.  It feels dangerous.  But while Sage's Integer() and ZZ() do not allow expressions, gp does:

sage: ZZ(gp('2^3 * 3^2'))                                                                                                                             
72

and you can catch non-integral values as they will raise a type error (as you see if you replace * by / above). So perhaps that is safer.

AndrewVSutherland:
I like @JohnCremona's suggestion.  While the regex check makes me comfortable that calling sage_eval is safe in this instance, if there is a simple alternative that avoids sage_eval entirely (which it a appears there is) that would be better (otherwise in the future anytime someone gets paranoid and does a git grep for "sage_eval" they have to go through the exercise of convincing themselves that this particular usage is safe).

But rather than worrying about catching non-integral values, I would just keep the regex check in and change the call to sage_eval to ZZ(gp(...)) (belts and braces).

edgarcosta:
isn't calling `gp` like calling `sage-eval` but for gp inside sage?

```
sage: gp('a = 3 + 5;')
8
sage: gp('a')
8
```

tahulse:
I can use whatever function you'd like me to use.

tahulse:
I am pretty new to this so I don't know if the second commit I just submitted was done correctly. Please let me know if I should do this differently. Thanks.

JohnCremona:
> isn't calling `gp` like calling `sage-eval` but for gp inside sage?
> 
> ```
> sage: gp('a = 3 + 5;')
> 8
> sage: gp('a')
> 8
> ```

This is a  good point.  If we were to pass the string directly to gp with no parsing then we would be asking for trouble since someone might type factor(nextprime(10^100)*nextprime(10^101)) in the box.  But the current code only sends the strong to gp if it matches the MULT_PARSE regex which only allows digits parentheses, * and ^ (not even + and -).  gp() will still fail on an input like 2^(2^100) and will raise a TypeError.  

I think the try/except block should catch the TyppeError explicitly -- it is bad practice to have "except:" on its own.  Although lmfdb/ does have 146 instances of this, many of which were put there by people who should know better....

edgarcosta:
Here is my suggestion:
```
sage: import ast
....: class PowMulNodeVisitor(ast.NodeTransformer):
....:     def visit_BinOp(self, node):
....:         if isinstance(node.op, ast.Pow):
....:             return  self.visit(node.left) ** self.visit(node.right)
....:         elif isinstance(node.op, ast.Mult):
....:             return  self.visit(node.left) * self.visit(node.right)
....:     def visit_Constant(self, node):
....:         return ast.literal_eval(node)
....:
....: foo = ast.parse('2^3 * 3^2'.replace('^', '**'), mode='eval')
....: bar = PowMulNodeVisitor().visit(foo).body
....: print(type(bar) , bar)
<class 'int'> 72
sage: foo = ast.parse('5*(2^(2^3))'.replace('^', '**'), mode='eval')
....: bar = PowMulNodeVisitor().visit(foo).body
....: print(type(bar) , bar)
<class 'int'> 1280
```

this will certainly choke at 2^(2^100), without modifying `visit_BinOp`

edgarcosta:
I have implemented what I proposed, and but a limit on not powering things that will lead to more than 3000bits.
You can try it out here:
http://localhost:37777/ModularForm/GL2/Q/holomorphic/?hst=List&level=2*7%5E2&search_type=List

I have also added a test.

AndrewVSutherland:
@edgarcosta The test you added is causing a failure

`_______________________________________________________________________ CmfTest.test_expression_level _______________________________________________________________________

self = <lmfdb.classical_modular_forms.test_cmf.CmfTest testMethod=test_expression_level>

    def test_expression_level(self):
        # checks we can search on 2*7^2
>       self.check('/ModularForm/GL2/Q/holomorphic/?hst=List&level=2*7%5E2&search_type=List', '98.2.a.a')
E       TypeError: check() missing 1 required positional argument: 'text'

lmfdb/classical_modular_forms/test_cmf.py:666: TypeError
`

PR #4298, opened 2020-12-19
Fix some minor issues in ECQ URL/label validation

AndrewVSutherland:
We are currently getting server errors in the flasklog arising from URLs like

https://www.lmfdb.org/EllipticCurve/Q/11/banana/

and inappropriate behavior from URLs like

 https://www.lmfdb.org/EllipticCurve/Q/11/a666banana/

This PR fixes these problems by adding some more careful validation of isogeny class labels and adds some tests to check them.

All tests pass on grace.


JohnCremona:
These changes will need to be ported over to my ECQ branch too. I should probably take the plunge and make a PR for that after the weekend.

AndrewVSutherland:
@JohnCremona If it would be easier for you I would be happy to wait until you submit your PR and then make these changes, they I don't think they are urgent, just part of the ongoing effort to reduce the number of server errors showing up in the flasklog.

JohnCremona:
Thanks I think that makes sense.

PR #4302, opened 2020-12-23
Remove lingering reference to ec_curves caused by merge conflict

AndrewVSutherland:
Small cleanup to resolve a problem caused by a bad interaction between the two most recent PRs and a minor formatting tweak.  I will merge this as soon as tests pass.

### DIFF
--- a/Postgres_FAQ.md
+++ b/Postgres_FAQ.md
@@ -412,7 +412,7 @@ Note that you need editor priviledges to add, delete or modify data.
 
 1. What is an `extra_table`?
 
-   A few large tables (e.g. `nf_fields` and `ec_curves`) have been
+   A few large tables (e.g. `nf_fields`) have been
    split in two.  The columns in the search table can be used in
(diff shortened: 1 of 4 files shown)


JohnCremona:
There is one thing here which needs fixing, in the find_firends which I had not noticed: j-invariants are now stored as a list of 2 ints [num,den] not a string.  Clearly we don't have a test for that...

AndrewVSutherland:
@JohnCremona Are you fixing this or would you like me to?

JohnCremona:
> @JohnCremona Are you fixing this or would you like me to?

I just did -- see new PR

JohnCremona:
> > @JohnCremona Are you fixing this or would you like me to?
> 
> I will!

Done,  see PR #4305 .  It's only in a script so I don't know whether it matters as much as I had thought.  There should not be any more of these, as I had checked all occurrences of "ec_" at some point, but I missed this one even though in the same line as I just corrected I had changed curves to curvedata.

PR #4303, opened 2020-12-23
Two small fixes

AndrewVSutherland:
This PR fixes two small bugs that were causing server errors to show up in the flasklog.  The first is to correctly handle the Dirichlet group of modulus 1 and the second is to handle CMF L-function pages with invalid character orbit labels.  I'll merge this as soon as tests pass.


AndrewVSutherland:
Test pass on grace, merging.

PR #4304, opened 2020-12-23
removed last mentions of ec_curves

JohnCremona:
Deleted 2 lines in script files which referred to the old table ec_curves, and corrected one comment (trust me, what the comment now says is right...).


AndrewVSutherland:
Looks good and tests are not relevant (beyond pyflakes, which passes) so merging.

PR #4306, opened 2020-12-23
master -> dev

AndrewVSutherland:
pushing the recent round of changes to beta, but let's hold off pushing to web for a few days.


JohnCremona:
I agree that waiting before pushing to prod is sensible.   I will of course fix issues which arise but do not want to over the next 4 days.

AndrewVSutherland:
tests pass on grace, merging

PR #4309, opened 2020-12-28
Cnumber -> lmfdb_number

edgarcosta:


JohnCremona:
Sorry about this.  I thought I had very thoroughly looks at every line of code referring to the now nonexistent column names.

PR #4313, opened 2020-12-31
Minor updates to Sato-Tate groups

AndrewVSutherland:
This PR makes some minor formatting changes, adds subgroup multiplicities, generators, and mixed moments for the Sato-Tate groups of degree 4.  This is all just preparation for adding genus 3 Sato-Tate groups.

Tests pass on grace.


AndrewVSutherland:
Travis is going to fail because it cannot connect to devmirror right now (because of DB upgrade), but tests pass on grace so I am going to merge this.

PR #4322, opened 2021-01-01
Fix bug in Kodaira symbol

roed314:
This works around a bug in Sage ([#31147](https://trac.sagemath.org/ticket/31147)).  See https://www.lmfdb.org/EllipticCurve/Q/164032/cd/1 for an example in practice.


JohnCremona:
I'll look at this -- @roed314 will you submit a patch to Sage or would you like me to?

JohnCremona:
OK, I see that you have already dealt with this upstream at https://trac.sagemath.org/ticket/31147.  Meanwhile let's merge this.

PR #4325, opened 2021-01-07
big bunch of typos found by codespell

fchapoton:
should be not controversial


davidfarmer:
I looked through the changes and all are okay except possibly for:

ommitted --> omitted
  in
sato_tate_groups/main.py 

because that is a variable name, not a word in a comment.  So one needs to
check that the code is still okay,

and maybe not everyone agrees that "Focussed" was spelled wrong.

AndrewVSutherland:
I checked the changes to the variable name "ommitted" in sate_tate_groups/main.py and they look good.

JohnCremona:
The changes are acceptable to me.  I agree with @AndrewVSutherland -- a variable name need not have been changed (but pyflakes shows that the change is OK), and in British English the past tense of "focus" is indeed spelled (or is that spelt?) "focussed" as I originally had it, but I am not fussy.

Issue #4327, opened 2021-01-08
get rid of Deprecation warnings


JohnCremona:
When running tests I see several warnings like this:

lmfdb/characters/main.py:70
  /scratch/home/jcremona/lmfdb/lmfdb/characters/main.py:70: DeprecationWarning: invalid escape sequence \(
    jump_egspan = "e.g. 13.2 for the Dirichlet character \(\displaystyle\chi_{13}(2,·)\),or 13.f for its Galois orbit."

wihc are harmless now but one day will be fatal.  All that is needed, I think, is to put an r infront of the opening quote for such strings.

JohnCremona:
I'm fixing these where possible.  Some appear to be in python packages, e.g.
lmfdb/siegel_modular_forms/test_siegel_modular_forms.py::HomePageTest::test_browse_page
  /scratch/sage/sage-9.2/local/lib/python3.6/site-packages/jinja2/lexer.py:648: DeprecationWarning: invalid escape sequence '\m'
    .decode("unicode-escape")
and perhaps can be fixed by upgrading these (using sage -pip perhaps).

edgarcosta:
Is this on Legendre? If so the packages were freshly installed a couple of weeks ago.

JohnCremona:
> Is this on Legendre? If so the packages were freshly installed a couple of weeks ago.

Yes, on legendre.   See the PR#4328.

We still see Deprecation Warnings from /scratch/sage/sage-9.2/local/lib/python3.6/site-packages/jinja2/lexer.py:648 and /scratch/sage/sage-9.2/local/lib/python3.6/site-packages/markdown/treeprocessors.py:285
both triggered (I think) by lmfdb/lfunctions/test_lfunction_degree_2.py::LfunctionTest

edgarcosta:
The packages are up to date, I think the issues might be in our templates.

Here are the remaining warnings:
```
  /home/edgarcosta/LMFDB/lmfdb/lmfdb/knowledge/__init__.py:10: DeprecationWarning: 'etree' is deprecated. Use 'xml.etree.ElementTree' instead.
    from . import main
--
  /home/sage/sage-9.2/local/lib/python3.6/site-packages/jinja2/lexer.py:648: DeprecationWarning: invalid escape sequence '\('
    .decode("unicode-escape")
--
  /home/sage/sage-9.2/local/lib/python3.6/site-packages/jinja2/lexer.py:648: DeprecationWarning: invalid escape sequence '\c'
[...]

PR #4328, opened 2021-01-08
get rid of several string escape deprecation warnings

JohnCremona:
This fixes most of the deprecaton warnings mentioned at #4327, and a few unnecessary print statements.  In one file I changed the way some Sage things were being imported as one of those also raised a warning.


edgarcosta:
all the changes look good to me, and the tests pass.

I modified one import statement, to be backward compatible with sage 9.1.
I can't upgrade Travis right now to sage 9.2

JohnCremona:
Ah, I had not realised that Travis was on 9.1.  Thanks for the extra fix.

JohnCremona:
I restarted the Travis build anyway.

JohnCremona:
It's a pyflakes failure but I don't see the output -- where's pyflakes log file on Travis?

JohnCremona:
The problem is that it does not know what ModuleNotFoundError is.  @edgarcosta !

JohnCremona:
This makes no sense to me -- I ran these import lines on sage 9.1 on another machine and it works fine.  I see that ModuleNotFoundError is new to Python 3.6 though.  Which python version was used in the Travis build of Sage 9.1?  Can you run that and see what happens when you enter "from sage.banana import banana"?

JohnCremona:
I think I should revert the change of import of verbose in that file and we can ignore the deprecation warning until Travis gets Sage 9.2.   @edgarcosta you did not make sorting this out any easier by making a code change directly on github!  I think I will have to close this PR and make a new one.

edgarcosta:
@JohnCremona Thanks.

You can always edit other people PRs, and I was planning to fix this 👍

Issue #4323, opened 2021-01-06
Download ECNF, sage issue


edgarcosta:
About once a day, a bot manages to trigger a weird sage bug (traceback below), which I can't replicate.
Does anyone know the root of this issue? It would be nice to report this upstream.

 
```
Exception on /EllipticCurve/6.6.371293.1/27.1/a/2/download/gp [GET]
Traceback (most recent call last):
  File "sage/structure/category_object.pyx", line 839, in sage.structure.category_object.CategoryObject.getattr_from_category (build/cythonized/sage/structure/category_object.c:7075)
    return self.__cached_methods[name]
KeyError: '_pari_bnf'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/home/sage/sage-9.2/local/lib/python3.8/site-packages/sage/rings/number_field/number_field.py", line 4239, in pari_bnf
    bnf = self._pari_bnf
  File "sage/structure/category_object.pyx", line 833, in sage.structure.category_object.CategoryObject.__getattr__ (build/cythonized/sage/structure/category_object.c:6994)
    return self.getattr_from_category(name)
  File "sage/structure/category_object.pyx", line 848, in sage.structure.category_object.CategoryObject.getattr_from_category (build/cythonized/sage/structure/category_object.c:7160)
    attr = getattr_from_other_class(self, cls, name)
  File "sage/cpython/getattr.pyx", line 367, in sage.cpython.getattr.getattr_from_other_class (build/cythonized/sage/cpython/getattr.c:2553)
    raise AttributeError(dummy_error_message)
AttributeError: 'NumberField_absolute_with_category' object has no attribute '_pari_bnf'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/home/sage/sage-9.2/local/lib/python3.8/site-packages/flask/app.py", line 2447, in wsgi_app
    response = self.full_dispatch_request()
  File "/home/sage/sage-9.2/local/lib/python3.8/site-packages/flask/app.py", line 1952, in full_dispatch_request
    rv = self.handle_user_exception(e)
  File "/home/sage/sage-9.2/local/lib/python3.8/site-packages/flask/app.py", line 1821, in handle_user_exception
    reraise(exc_type, exc_value, tb)
  File "/home/sage/sage-9.2/local/lib/python3.8/site-packages/flask/_compat.py", line 39, in reraise
    raise value
[...]

PR #4329, opened 2021-01-08
simplify ECNF constructor to do minimal arithmetic on ideals

JohnCremona:
This deals with more of the issues mentioned at Issue #4323.  The ECNF constructor now does much less work regarding constructing ideals from their string representation in the database, going directly from those strings to latex for display, even for the factorizations of the conductor and discriminant.

After this, the following will also be done, involving data (schema) changes which this code will handle without change:  (1) all ideal strings stored (for conductor, minimal discriminant, and bad primes) will be replaced by minimal generators (e.g. one generator for principal ideals, otherwise two) delimited by "(",")", whereas at present they are of the form "[N,a,alpha]" where N is the norm and a the smallest positive integer in the ideal.  The code here can tell when this has been done by seeing if the string started with "(" or "[".  (2) We now store the minimal discriminant in the database, and usually that is the same as the ideal generated by the discriminant of the model stored, but for cases where that is not the case I will add a column to hold the model discriminant; right now this is computed on the fly -- but when it is not the same as the minimal discriminant.

This code now also allows for computing reduced gens on the fly, as a temporary measure, and that is currently done when the degree is 2 or 3 *or* the class number is 1.  Reviewers are welcome to suggest a different condition, but this is only temporary -- users will not like to see ideals presented with 2 generators over fields like Q(i).  I'll work on the data changes next week.


JohnCremona:
I restarted the failing check after the "escapes" branch was merged and all tests now pass -- as you can see if you look at the Travis log, even though the red cross has not been updated on this page.

JohnCremona:
> I restarted the failing check after the "escapes" branch was merged and all tests now pass -- as you can see if you look at the Travis log, even though the red cross has not been updated on this page.

My mistake, it is green now -- but I am about to update the PR with some more simplifications so no hurry.

JohnCremona:
This PR is again ready.  It contains all code changes in ECNF which have been made to (1) be more efficient already and (2) work when I have made some changes to the table ec_nfcurves.  This will add columns disc, normdisc (for the ideal discriminant of the model);  ensure all ideal strings (conductor, minimal discriminant ideal, bad primes) are in simplified form, e.g. show a single generator for principal ideals, and replace Kodaira symbol strings with pari encodings.

I am also considering deleting the column which holds the latex equation, since it can be quite along string.  This code version will create that string on the fly of the column does not exist.

In all respects, this code is supposed to work perfectly now and after the data changes.

AndrewVSutherland:
@JohnCremona It looks like there are some merge conflicts caused by your earlier PR which I just merged -- can you take a look at these?

JohnCremona:
> @JohnCremona It looks like there are some merge conflicts caused by your earlier PR which I just merged -- can you take a look at these?

OK, I'm on it

JohnCremona:
Is more information needed to help review this?  I'll soon be ready to make some data changes which can only be done once this is merged -- and that includes correcting data as in issue #4332.

JohnCremona:
On second thoughts I can update the data for those RQFs without waiting, but would still like this to progress

AndrewVSutherland:
@JohnCremona Our plan is to wait and review and merge this next week (on Wednesday).

JohnCremona:
That's fine.  I have some data uploads which can wait.

PR #4336, opened 2021-01-14
Minor change to Sato-Tate group label format

AndrewVSutherland:
This PR implements a minor (backward compatible) change to the Sato-Tate group label format that is needed before we add Sato-Tate groups in USp(6).  The tables gps_sato_tate and gps_sato_tate0 are replaced by gps_st and gps_st0 so that the current code running on www.lmfdb.org is unaffected by the data changes (after this PR is in production I can change the names back, but there is no rush to do this).  I also need to update the top and bottom knowl names, but I will wait until this PR is in production before doing so (they will still work as is on both the old and new code).

Tests pass on grace.


AndrewVSutherland:
The label knowl will also need to be updated, I'll do that as soon as this PR is on beta.

AndrewVSutherland:
Tests all pass on grace, merging.

PR #4344, opened 2021-01-15
A few more minor ST updates

AndrewVSutherland:
Adds symplectic form and Hodge circle


AndrewVSutherland:
Tests all pass on grace and Travis appears to be in his usual Friday slump...merging.

PR #4350, opened 2021-01-16
Minor rearrangement of L-function search inputs

AndrewVSutherland:
This PR makes some minor tweaks the layout of the browse and refine search options to (1) remove the arithmetic selector from the rational search page (where it makes no sense), (2) combine rational/arithmetic into a single selector with rationa/irrational/algebraic/transcendental options, (3) fill in an unnecessary gap, (4) keep the refine search array 5 boxes wide (as elsewhere in the LMFDB) to make it work better on narrower screens and avoid colliding with the learnmore box.


AndrewVSutherland:
All tests pass on grace.

edgarcosta:
I will merge this once tests pass

PR #4361, opened 2021-01-24
More Sato-Tate group UI tweaks

AndrewVSutherland:
Tweak aspect ratio of histograms and correctly handle joint probabilities, showing them in lowest terms


AndrewVSutherland:
Tests pass on grace, merging.

PR #4363, opened 2021-01-24
fixing various typos (found by codespell)

fchapoton:
just a little bouquet of typos


jwj61:
I suspect it is innocuous, but changing variable names seems like a bad idea.

edgarcosta:
The only one with variable name changes is `scripts/local_fields/import_fields.py `

PR #4364, opened 2021-01-24
various cosmetic changes (one import per line, etc)

fchapoton:
as suggested by flake8

This may in particular help for #4327


AndrewVSutherland:
@fchapoton thanks, these all look good to me.  Happy to merge once tests pass (I'll run them on grace if Travis doesn't start soon).

PR #4365, opened 2021-01-24
More UI tweaks to Sato-Tate groups

AndrewVSutherland:
Show z-vector probability data in a table to make presentation less cluttered.


AndrewVSutherland:
tests pass on grace, merging.

PR #4372, opened 2021-01-26
EC/Q fix code use of labels

### LINKED ISSUE #4371: EC/Q: code snippets should not use labels for conductors over 500000
Another piece of fallout from the addition of some curves with conductor > 500000 and no Cremona labels, but also an error in the Sage, Magma and Pari/GP scripts (all three!) when displaying a curve with LMFDB labels -- which is the only option for conductor >500000.

In all three languages the users is given two constructions for the curve, one using the a-invariants (which is fine) and one using the label.  This *does not work* for LMFDB labels (none of the three languages can decode them).

I'll probably fix this by not giving the alternative.  A more complicated solution would be, for conductors<500000 when displaying the LMFDB label version of a page,  substitute the Cremona label into the template, and for large conductors, delete the second line of the template.

JohnCremona:
This fixes #4371, correcting the Sage/Magma/GP code for constructing curves.  First, the second line showing how to construct a curve from a label, uses the Cremona label (which is the only labelling which any of the three recognise).  Second, this second line is omitted for conductors over 500000.  The change works both in the code snippets on a home page and on the compete file of code to download.

To test, take  random curves both with conductor <500000 and >500000 and look at the code snippets on the home page (all three languages) and the code download.

### DIFF
--- a/lmfdb/elliptic_curves/code.yaml
+++ b/lmfdb/elliptic_curves/code.yaml
@@ -17,15 +17,9 @@ not-implemented:
     // (not yet implemented)
 
 curve:
-  sage: |
-    E = EllipticCurve(%s)    # or
-    E = EllipticCurve("%s")
-  pari: |
-    E = ellinit(%s)       \\ or
-    E = ellinit("%s")
-  magma: |
-    E := EllipticCurve(%s); // or
-    E := EllipticCurve("%s");
+  sage:  E = EllipticCurve(%s)
+  pari:  E = ellinit(%s)
+  magma: E := EllipticCurve(%s);
 
 gens:
   sage:  E.gens()
@@ -121,5 +115,5 @@ galrep:
 
 padicreg:
   sage: |
-    [E.padic_regulator(p) for p in primes(3,20) if E.conductor().valuation(p)<2]
+    [E.padic_regulator(p) for p in primes(5,20) if E.conductor().valuation(p)<2]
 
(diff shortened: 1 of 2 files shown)


JohnCremona:
Travis started before I fixed the typo.   Advice: never make trivial changes to the code after local testing and before uploading...

AndrewVSutherland:
@JohnCremona I'm running tests on grace now, assuming they pass, I will merge this.

AndrewVSutherland:
Actually I think you should make another change, which is to either remove or comment out the line that attempts to create the elliptic curve using the Cremona label when the conductor is below 500,000.  Using this makes us overly dependent on the particular version of Sage/Magma/Pari-GP the user is running, and it also assumes the user has installed the full Cremona database, and that they have updated that database to include all curves up to conductor 500,000.

Currently all three code snippets fall on my local machine for the curve 401002a1, and Sage and Pari both fail for the curve 1001c1 because I don't have the full Cremona database installed.

JohnCremona:
OK, I will do that (as I suggested on the Issue).  I will not just comment it out as that would require adding in more text telling people how to install optional stuff.

I'll update the PR soon.

JohnCremona:
As requested.  I also simplified the code, and I checked the full code set in all 3 languages on legendre where the optional databases for Sage and Pari/GP are not installed, for one (small) curve.

AndrewVSutherland:
Looks good.  Tests are running on grace now, I'll merge this and close #4371 as soon as they do.

PR #4379, opened 2021-01-29
more Sato-Tate tweaks

AndrewVSutherland:
Switches G_{3,3} -> SU(2)xSU(2) and G_{1,3} -> U(1)xSU(2) (the old names will still work) and adds a bunch more aliases (e.g. you can now type any identity component into the jump box and it will take you to the corresponding Sato-Tate group even if its name is not the same as that of the identity component).


AndrewVSutherland:
Tests now pass on grace, merging.

PR #4400, opened 2021-02-01
L-function fixes

### LINKED ISSUE #4327: get rid of Deprecation warnings

### LINKED ISSUE #4398: L-function links on www redirecting to beta and download_zeros links causing server errors.
Currently L-function links like https://www.lmfdb.org/L/EllipticCurve/Q/11/a/ are redirecting to beta pages like https://beta.lmfdb.org/L/2/11/1.1/c1/0/0.  We need to update the whitelist so this does not happen.

In addition, some of the old style download_zeros L-function links like

https://www.lmfdb.org/L/download_zeros/ModularForm/GL2/TotallyReal/2.2.12.1/holomorphic/2.2.12.1-47.1-b/0/0/

are throwing server errors when they should really redirect to https://www.lmfdb.org/L/download_zeros/4/6768/1.1/c1e2/0/0

edgarcosta:
Resolves #4398 by:
- editing the whitelist
- modifying the function `white_listed` function to allow `L/` followed by a digit.
- only allowing `L/download*/<label>/` and failing gracefully if we are not given an existing label

Also resolves #4327 by replacing the `import markdown.util.etree` import with `from xml.etree import ElementTree`


AndrewVSutherland (approved):

AndrewVSutherland:
Tests pass on grace.

PR #4413, opened 2021-02-03
fix downloads for Dirichlet L-fcns

edgarcosta:


edgarcosta:
This is an internal fix of Dirichlet specific L-function downloads.  
I will merge as tests pass on grace

PR #4422, opened 2021-02-04
EC/Q: update completeness info to include primes<10^6

JohnCremona:
This is in 3 parts, after uploading all elliptic curves with prime conductor between 500000 and 10^6:

- the source and reliability knowls have been updated, and will need reviewing
- the completeness knowl's content is set in the code and this PR updates that
- the 6 tables db.ec_curvedata, db.ec_localdata, db.ec_mwbsd,  db.ec_classdata, db.ec_2adic, db.ec_galrep need to be copied to prod

To see the 3775 new curves on beta, put in conductor>500000 and exclude 2,3,5,7 from the bad primes:  https://beta.lmfdb.org/EllipticCurve/Q/?hst=List&conductor=500000-&bad_quantifier=exclude&bad_primes=2%2C3%2C5%2C7&search_type=List

### DIFF
--- a/lmfdb/elliptic_curves/elliptic_curve.py
+++ b/lmfdb/elliptic_curves/elliptic_curve.py
@@ -159,9 +159,11 @@ class ECstats(StatsDisplay):
         self.nclasses = db.ec_classdata.count()
         self.nclasses_c = comma(self.nclasses)
         self.max_N_Cremona = 500000
-        self.max_N_Cremona_c = comma(500000)
+        self.max_N_Cremona_c = comma(self.max_N_Cremona)
         self.max_N = db.ec_curvedata.max('conductor')
         self.max_N_c = comma(self.max_N)
+        self.max_N_prime = 1000000
+        self.max_N_prime_c = comma(self.max_N_prime)
         self.max_rank = db.ec_curvedata.max('rank')
         self.max_rank_c = comma(self.max_rank)
         self.cond_knowl = display_knowl('ec.q.conductor', title = "conductor")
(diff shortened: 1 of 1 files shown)


AndrewVSutherland:
The 5 tables you mentioned have been copied to prod.

JohnCremona:
> The 5 tables you mentioned have been copied to prod.

Thanks!

jwj61:
Putting completeness information in code rather than in a knowl is a shift.  Do we want to do that?  Certainly, reviewing the knowl becomes irrelevant.

JohnCremona:
Well, not really a shift since it has been done this way for ages, but you may be right that it is undesirable.  The content of the knowl extracts from the database the total number of curves and isogeny classes, and the maximum conductor, which is why it is dynamic.  I did it this way when the ECQ database was expanding regularly.  Now I could just put in the numbers as of today -- would you prefer that?  Then, when I do add data (and I am planning on adding more curves of prime conductor fairly soon) I would just edit the knowl itself, as I did for the source and reliability knowls today.

If that's what is wanted, I can easily edit the content of dq.ec.extent right now, and remove the redundant bit of code after.

jwj61:
This knowl is doing 2 things, extent and completeness.

For extent, I think it is fine to dynamically check the database to see what is in there, e.g., maximum conductor and number of curves.

Completeness, saying that we have all curves of a certain type, is different and more care should be taken with that.  So, if you throw in some additional curves, say to get examples with higher ranks, then it makes sense for the extent part to automatically update.  But I think the part on completeness, which is now in the code should probably be text in the knowl.

JohnCremona:
OK, I will try to sort this out.  By the way, as I was looking to see how you dealt with similar issues for number fields, I noticed that the bullet points on https://www.lmfdb.org/NumberField/Source are badly aligned, making it looks as if everything is from Bordeaux! I'll let you edit it @jwj61

JohnCremona:
I see from the dq knowl list https://beta.lmfdb.org/knowledge/?all=1&category=dq&search=&reviewed=on&beta=on&normal=on that the knowls whose id ends in 'extent' are sometimes "Extent of ..." and sometimes "Completeness of..." so no wonder I was getting the two mixed up.

jwj61:
Thanks for fixing the elliptic curve knowl, and pointing out the problem with the formatting on the number field source knowl (now fixed).

To make matters more confusing, at one point we had decided to replace all dq with rcs, and then later decided not to enforce that.  So now, some of them are dq and some are rcs.

AndrewVSutherland:
I'm fine with switching dq to rcs everywhere (I actually thought we had, I guess I don't remember why we decided not to).

jwj61:
It was just that some people hadn't and we decided not to force them.

There should be a simple mapping from one to the other, so it could be a task for someone to get their feet wet in code development (manly renaming knowls and their references in templates and possibly code).

PR #4425, opened 2021-02-04
EC/Q: separate out completeness and extent info

JohnCremona:
This is a simple follow-up to #4422.   The extent part and completeness part of the extent/completeness knowl are separated, so the static part includes the statement that we have *all* curves in three sets, while the dynamic part just gives total numbers and largest conductor.

Until this is merged the extent knowl on bet looks stupid as some info is included twice!  So please merge ASAP.


JohnCremona:
Thanks -- that was quick!

Issue #4326, opened 2021-01-07
Don't use magma_free in tests


JohnCremona:
The test_download_magma() test in classical_modular_forms/test_cmf.py uses Sage's magma_free() function to run the code on the free online Magma service, but we should not be doing this.  Instead we shold do something closer to what similar test functions do in  test_bmf and test_hmf.

Secondly, rather than having several almost identical test functions which fetch some Magma code (or Sage of gp) from a URL, and run them to make sure that they run without errors and that some string appears i the output, it would be better to have a single function for each language which these test functions would just call.  Then we can easily add similar tests for other code download URLs -- being careful, of course, that the examples used i nthe test all run quickly.

JohnCremona:
@BarinderBanwait do you have time to deal with the first part of this (at least)?

edgarcosta:
Here is the place where one can add such functions: https://github.com/LMFDB/lmfdb/blob/master/lmfdb/tests/__init__.py

Calling magma_free is quite useful when magma is not available.
For example, we can't have magma on Travis, and not everyone has access to magma when submitting PRs.


In general, automatic testing would be much better if we replaced Travis by doing testing on our own virtual servers.
For example, with a webhook system (https://docs.github.com/en/free-pro-team@latest/developers/webhooks-and-events/webhooks) or via Google Cloud Build (https://github.com/marketplace/google-cloud-build).
At the moment, the time investment setup for these things is too high in comparison to its benefits.

AndrewVSutherland:
@edgarcosta As we discussed, while I agree that Magma's online calculator is a very handy tool that is a great way to try out Magma or do a quick calculation, it really isn't appropriate for us to take advantage of Magma's generosity by incorporating it into our automated CI process.

Any LMFDB developers who do not have access to Magma and need it for work they are doing on the LMFDB should feel free to contact me and we can setup an account for them on one of our development machines (all of which have Magma installed).

BarinderBanwait:
right, so if i understand correctly, the tasks here are:

1. Deprecate magma_free (and replace with what `test_download_magma` in test_bmf is doing);

2. Create `{x}_code_getter_and_executor` for x in {Sage, Magma, PARI/GP} functions, place them in the file @edgarcosta  points to, and have the various `test_download_{x}` tests call the respective getter_and_executor?

sure, i think i can take that on

edgarcosta:
Yup and yup.

Thanks

JohnCremona:
[...]

PR #4430, opened 2021-02-06
Lmfdb#4326 Code download and compile helpers

BarinderBanwait:
This PR addresses Part 2 of #4326, to create helper functions when creating unit tests for downloading code, checking it compiles, and running assertion checks on some variables in that code.

The previous PR addressing Part 1 of this ticket effectively added a magma helper function which replaced all calls to magma in the rest of the testing code.

This PR has implemented a similar helper function for sage, replacing all calls to `exec(sage_code, globals())` in the testing code.

The individual tests affected by this refactor have been run locally on legendre, and they still pass.

### DIFF
--- a/lmfdb/bianchi_modular_forms/test_bmf.py
+++ b/lmfdb/bianchi_modular_forms/test_bmf.py
@@ -115,23 +115,15 @@ class BMFTest(LmfdbTest):
             # but the goal is to test that itself doesn't show in the friends list
             assert notitself not in L.get_data(as_text=True)
 
-    def check_compile_and_get_level(self, download_data):
-        """Simulates a user downloading the sage code, and then loading it into
-        a sage session. This requires the sage import at the top"""
-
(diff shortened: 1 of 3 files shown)


BarinderBanwait:
Hmmm, so pylint fails because it thinks that the sage imports `QQ, PolynomialRing, NumberField` are not being used; but they are being used within the `exec(sage_code, globals())` call (for certain values of `sage_code`). I experimented with other places to put these imports, notably in the test files that call this new helper method, but that kept running into import errors. I'm unsure how to resolve this - Has anyone dealt with a similar situation elsewhere?

JohnCremona:
After lines saying import X,Y etc put lines saying assert X and assert Y. Does nothing but keeps pyflakes happy

BarinderBanwait:
Thanks @JohnCremona ! Hopefully passes all tests now.

edgarcosta:
Looks good to me!

Issue #1483, opened 2016-06-01
EC/ECNF: search by size of isogeny class


AurelPage:
**_1 Upvote**_ It would be nice to be able to search elliptic curves by size of the isogeny class (<=, ==, >=). This should be possible with the existing data using the fact that the elliptic curves in an isogeny class are labelled 1, ... , size of isogeny class.

AndrewVSutherland:
+1 I would like to have this feature too, and I agree that an search would be very easy to implement (just query on the "number" attribute).

Doing an interval search would require a little more thought (you would want to project down to one record per isogeny class).  The non-trivial part of this would be implementing browse/search pages for isogney classes, which do not currently exist.  On the other hand, implementing a search for curves **contained** in an isogeny class of a given size would be trivial (just a few lines of code and one new input box on the search pages).

JohnCremona:
I think there may be a small misunderstanding here: the 'number' field is the index of a curve in its class, so searching for {'number':5} will find a curve in every class whose size it at least 5, not exactly 5.  To avoide awkward code we could store the size ofthe isogeny class with every curve, or with one curve per class.  So far we have managed to avoid having a separate collection for isogeny classes, and hope that we can continue to avoind that.  There are some data fields which only exists for one curve per class already, including the isogeny matrix which is a list of lists of ints, so if one can search for objects for which a list field has a certain length then Aurel's request would be easy (if not necessarily efficient).

It is possible:

nfcurves.find_one({'isogeny_matrix':{'$size':int(18)}})

AndrewVSutherland:
@JohnCremona You are right of course, I responded to quickly.

JohnCremona:
Over Q  each curve now has fields 'class_size' and 'class_deg' being the size of the isogeny class and the maximal degree of a cycylic isogeny between curves in the class.  This is displayed on each curve's home page.  We just implemented search by isogeny degree (returning all curves with a cyclic isogeny (not an endomorphism) of the given degree(s)) and it would be simple to add searches by these other two parameters.
Over number fields I also just implemented search isogeny degree but forgot to add the two additional fields to each curve, so that will need doing before implementing the above for ecnf, unless it is done using code like I suggested earlier.

AndrewVSutherland:
It probably makes sense to do this for number curves as well.

JohnCremona:
[...]

PR #4434, opened 2021-02-09
Resolve issue 4055: ECNF search options semistable, potential good reduction

### LINKED ISSUE #4055: Add search option for elliptic curves with potentially good reduction everywhere
It would be nice to have a search option for elliptic curves over number fields (including Q) that have potentially good reduction everywhere (equivalently, integral j-invariants).  The easiest way to accomplish this would be to add a boolean flag which could easily be populated using the local data we already store that includes the valuation of the denominator of the j-invariant.

One could also ask for the same thing for genus 2 curves over Q and/or their Jacobians.

JohnCremona:
This resolves #4055 as stated, allowing both 'semistable' and 'potential good reduction' in EC/NF searches, as they are already for EC/Q, using the two associated boolean columns created and filled for this purpose.

### DIFF
--- a/lmfdb/ecnf/main.py
+++ b/lmfdb/ecnf/main.py
@@ -16,8 +16,8 @@ from lmfdb.backend.encoding import Json
 from lmfdb.utils import (
     to_dict, flash_error,
     parse_ints, parse_noop, nf_string_to_label, parse_element_of,
-    parse_nf_string, parse_nf_elt, parse_bracketed_posints,
-    SearchArray, TextBox, ExcludeOnlyBox, SelectBox, CountBox,
+    parse_nf_string, parse_nf_elt, parse_bracketed_posints, parse_bool, 
+    SearchArray, TextBox, ExcludeOnlyBox, SelectBox, CountBox, YesNoBox,
     search_wrap, parse_rational,
     redirect_no_cache
     )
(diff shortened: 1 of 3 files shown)


JohnCremona:
I am going to add another commit here to implement #1483 too.

JohnCremona:
In addition to the above (semistable and pot. good reduction) I have added: analytic Sha, regulator, isogeny class size and degree.  This needed a bit of rearranging for both the main search page and the search results page.  I tried to put similar optoins next to each other.  I could not include the "primes dividing |Sha|" query as we have for EC/Q since that is not (yet) a column in the ec_nfcurves table.   Since Regulator and Sha are not always known, just as (and i the same case as) the Rank is not known, I expanded the footnote on the search page accordingly and added an asterisk to Regulator and Analytic order captions (I put the * on "order" rather than on the russian symbol as that did not look right).  I also put this footnote on the esrach results page, where it should have been already (look on beta and you'll see the caption Rank* but no footnote).
Lastly, in the rearrangement of the search boxes I put the j-invariant box, which takes up 4 columns and is surely rarely used, at the bottom instead of the top!

To test just go the the main ECNF search page and try out the new options, checking that the search results page looks OK too.

This also resolves #1483.

edgarcosta:
It looks good to me, others can try it out here: https://purple.lmfdb.xyz/EllipticCurve/

Issue #4436, opened 2021-02-11
parse_bracketed_posints raises ZeroDivisionError


edgarcosta:
```
Exception on /NumberField/ [GET]
Traceback (most recent call last):
  File "/home/sage/sage-9.2/local/lib/python3.6/site-packages/flask/app.py", line 2447, in wsgi_app
    response = self.full_dispatch_request()
  File "/home/sage/sage-9.2/local/lib/python3.6/site-packages/flask/app.py", line 1952, in full_dispatch_request
    rv = self.handle_user_exception(e)
  File "/home/sage/sage-9.2/local/lib/python3.6/site-packages/flask/app.py", line 1821, in handle_user_exception
    reraise(exc_type, exc_value, tb)
  File "/home/sage/sage-9.2/local/lib/python3.6/site-packages/flask/_compat.py", line 39, in reraise
    raise value
  File "/home/sage/sage-9.2/local/lib/python3.6/site-packages/flask/app.py", line 1950, in full_dispatch_request
    rv = self.dispatch_request()
  File "/home/sage/sage-9.2/local/lib/python3.6/site-packages/flask/app.py", line 1936, in dispatch_request
    return self.view_functions[rule.endpoint](**req.view_args)
  File "/home/lmfdb/lmfdb-git-dev/lmfdb/number_fields/number_field.py", line 359, in number_field_render_webpage
    return number_field_search(info)
  File "/home/lmfdb/lmfdb-git-dev/lmfdb/utils/search_wrapper.py", line 140, in __call__
    data = self.make_query(info, random)
  File "/home/lmfdb/lmfdb-git-dev/lmfdb/utils/search_wrapper.py", line 48, in make_query
    errpage = self.f(info, query)
  File "/home/lmfdb/lmfdb-git-dev/lmfdb/number_fields/number_field.py", line 788, in number_field_search
    parse_bracketed_posints(info,query,'class_group',check_divisibility='increasing',process=int)
  File "/home/lmfdb/lmfdb-git-dev/lmfdb/utils/search_parsing.py", line 95, in __call__
    rval = self.f(inp, query, qfield, *args, **kwds)
  File "/home/lmfdb/lmfdb-git-dev/lmfdb/utils/search_parsing.py", line 752, in parse_bracketed_posints
    if L[i+1] % L[i] != 0:
ZeroDivisionError: integer division or modulo by zero
[...]

PR #4437, opened 2021-02-11
Enforce that positive integers must be positive

jwj61:
Addresses issue #4436 by enforcing that the invariant factors of the class group need to be non-zero.

https://beta.lmfdb.org/NumberField/?hst=List&galois_group=D10&class_number=5&class_group=%5B0%2C1%2C5%5D&search_type=List
http://127.0.0.1:37777/NumberField/?hst=List&galois_group=D10&class_number=5&class_group=%5B0%2C1%2C5%5D&search_type=List

### DIFF
--- a/lmfdb/higher_genus_w_automorphisms/main.py
+++ b/lmfdb/higher_genus_w_automorphisms/main.py
@@ -617,7 +617,7 @@ def higher_genus_w_automorphisms_search(info, query):
     if info.get('signature'):
         # allow for ; in signature
         info['signature'] = info['signature'].replace(';',',')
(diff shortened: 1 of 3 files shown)


roed314:
There's a test failure: `self.check_args('/NumberField/?start=0&degree=6&signature=%5B0%2C3%5D&count=100', '6.0.61131.1')`

jwj61:
I think tests will pass this time.  I added a flag to the parsing function as to whether to allow 0 to be a positive integer.  I also grepped for other places this function was called.  Most did not want to allow 0 (usually torsion subgroups), but at least one other did want to allow 0 (higher genus curves), so I adjusted that too.

Issue #4435, opened 2021-02-11
Artin number field, Galois closure


davidfarmer:
On an Artin representation page, under Basic invariants the
Artin number field is always listed as the Galois closure of some field.
Even when that field is Galois.

But when the field is not Galois, the link to "Artin field" under Related
Objects still links to that field, not to its Galois closure.

The Projective field is sometimes listed as "Galois closure of..." and
sometimes not.   But on this one, it is Galois but doesn't realize that:
https://beta.lmfdb.org/ArtinRepresentation/2.2323.4t3.a.a
But when it is not Galois, the Related objects link goes to the non-Galois
field, for example:
https://beta.lmfdb.org/ArtinRepresentation/2.280853.3t2.a.a

It also is confusing (to me) that what are called "Artin number field" and
"Projective field" under Basic invariants, are called "Artin field" and
"Projective Artin field" under Related objects.

jwj61:
I will leave it to @davidfarmer to close if the new version is ok

davidfarmer:
I was going to close, because all items I mentioned seem to be addressed.  But I have a comment and a question.

It seems like something maybe has been lost.  Previously it might say "the Galois closure of X",
but now it says "Y", where Y is the Galois closure of X .  In the case where X = Y , i.e., X already
was Galois, then everything is fine.  But if X != Y, then X is no longer mentioned on the page. 
Is X still interesting in some way?  If it is, then should it have a name and get a mention?
(I don't know the answer to those questions, because I do not have a good enough understanding
of this topic.)

If I go here ( I just chose a non-Galois field, thank you for making that
easy with the new non-Galois search!)
    https://beta.lmfdb.org/NumberField/3.1.59.1
And then I follow the friend link to this Artin representation
    https://beta.lmfdb.org/ArtinRepresentation/2.59.3t2.a.a
then the links to the Artin stem field and Projective stem field are to the non-Galois field
I started with.  If I go to pink:
    https://pink.lmfdb.xyz/ArtinRepresentation/2.59.3t2.a.a
then it refers to the Galois closure of that field.

So, I am asking whether the version on beta is the intended result, or if those
links should be to the Galois closure.  (And if to the Galois closure, then my
questions in the 2nd paragraph above apply.)

One more thing:  it only says "Field" in the related objects box, and only has one line, when the
Artin stem field and Projective stem field are the same.  But it says "Field" twice
when they are different.  I realize there is not much space, but it does look odd.  For example:
[...]

PR #4446, opened 2021-02-13
Address issue #4435 on Artin representation pages

jwj61:
When giving the Artin field and Projective field of a representation, we now only say "Galois closure of" when the field is not already Galois closed (although this was not technically wrong before).  The related objects are just labeled "Field blah", which is in keeping of how things are labeled elsewhere as related objects, and avoids having to say something like "Stem field for the Artin field" in a small space.  If the Artin field is the same as the projective field, it is only listed once.

http://127.0.0.1:37777/ArtinRepresentation/2.39.4t3.a.a
http://127.0.0.1:37777/ArtinRepresentation/4.1609.5t5.a.a
http://127.0.0.1:37777/ArtinRepresentation/1.4.2t1.a.a

vs

http://beta.lmfdb.org/ArtinRepresentation/2.39.4t3.a.a
http://beta.lmfdb.org/ArtinRepresentation/4.1609.5t5.a.a
http://beta.lmfdb.org/ArtinRepresentation/1.4.2t1.a.a

In addition there is a change at the bottom of number field pages.  In the table of Artin representations, instead of giving polynomials, it gives labels (or "pretty names" for the field).  To test, click on the link to see the corresponding field for a representation above.  In particular, you get pretty names in the last case.

### DIFF
--- a/lmfdb/artin_representations/main.py
+++ b/lmfdb/artin_representations/main.py
@@ -354,7 +354,7 @@ def render_artin_representation_webpage(label):
                 friends.append(("Dirichlet character "+cc_name, url_for("characters.render_Dirichletwebpage", modulus=cc.modulus, number=cc.number)))
             else:
                 detrep = the_rep.central_character_as_artin_rep()
-                friends.append(("Determinant representation "+detrep.label(), detrep.url_for()))
+                friends.append(("Determinant "+detrep.label(), detrep.url_for()))
         add_lfunction_friends(friends,label)
 
         # once the L-functions are in the database, the link can always be shown
--- a/lmfdb/artin_representations/main.py
+++ b/lmfdb/artin_representations/main.py
@@ -334,12 +334,14 @@ def render_artin_representation_webpage(label):
     wnf = None
     nf_url = the_nf.url_for()
     if nf_url:
-        friends.append(("Artin field", nf_url))
+        friends.append(("Field {}".format(the_nf.label()), nf_url))
         wnf = the_nf.wnf()
     proj_nf = WebNumberField.from_coeffs(the_rep._data['Proj_Polynomial'])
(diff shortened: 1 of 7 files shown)


jwj61:
Making more changes

jwj61:
On an Artin rep page, the Artin field and Projective field labels now include the word stem if they are not Galois over Q, so no more "Galois closure".  Also the search results and number field pages now use the new knowl for Artin stem field.  Same test pages as above.

JohnCremona:
I haven't tested everything but the new pages show "True" or "False" at the top of the first list of stuff, perhaps something which should be in a td block is not?

jwj61:
Thanks for catching that.  It should now be fixed.

JohnCremona:
> Thanks for catching that. It should now be fixed.

No problem.  After noticing it, I understood the cause after doing exactly the same thing myself in #4455.  I'll finish checking now.

JohnCremona:
Looks good.  One small thing I noticed on http://127.0.0.1:37777/ArtinRepresentation/2.39.4t3.a.a:  the projective field gets a pretty name in the main page but just a label 4.0.1521.1 i the Related Objects list.  WOuld it be better to be consistent by showing the pretty name in that list?

While on the list of related objects, I can see why you might not want to unclude the name of the L-function ater "L-function" since it is rather long.  Can we abbreviate "determinant representation" to just "determinant"?  That matches what's on the main page (where the knowl for "determinant" makes it completely clear what it is).

jwj61:
On pretty versions of a field in related objects, that seems like a bigger changes since there are several places where fields may appear as a related object.

I made the suggested change on determinant representation.

JohnCremona:
OK!  Can be merged as soon as tests pass.

Issue #4439, opened 2021-02-12
Truncate root display on Artin representations


edgarcosta:
If you go to the Artin representation (but you shouldn't!) with label
```
44.2161362748276787342400000000000000000000000000000000000000000000000000000000000000000000000000000000000000.55.a.a
```
your browser will have a hard time latex rendering that page, as we are displaying the 11 roots with 1373 p-adic digits...

We should truncate it or display it in a more limited fashion, perhaps using #4394

PR #4447, opened 2021-02-16
Truncate roots

jwj61:
Fixes issue #4439.  We lower the p-adic precision to 10 (or what is stored if that is smaller), and then only increase if it is needed to distinguish the roots (this may never happen).

You can see the page from the issue number:

  http://127.0.0.1:37777/ArtinRepresentation/44.2161362748276787342400000000000000000000000000000000000000000000000000000000000000000000000000000000000000.55.a.a

now loads (don't look at the original, it takes too long to load).

### DIFF
--- a/lmfdb/artin_representations/math_classes.py
+++ b/lmfdb/artin_representations/math_classes.py
@@ -687,6 +687,7 @@ class NumberFieldGaloisGroup(object):
             if self._data is None:
                 # This should probably be a ValueError, but we use an AttributeError for backward compatibility
                 raise AttributeError("No Galois group data for polynonial %s"%(coeffs))
+        self.lowered = False
 
     @classmethod
     def search(cls, query={}, projection=1, limit=None, offset=0, sort=None, info=None):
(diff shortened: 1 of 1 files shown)


JohnCremona:
@jwj61 Are you sure you want to raise an error in database code?  I expect that you do not expect it to happen ever becuase you trust your data, but...  Perhaps that function could return False (otherwise its only return value is True I think) and this could be handled some other way?

jwj61:
First, I would be pretty shocked if it ever triggered.

Returning false would mean the display is as it currently is, but it would also mean that there is an error and no one would know to report it.  As written,  the user gets the "server error" page which seems like a reasonable way to get them to report an issue.

JohnCremona:
Fair enough.  (There are many other instacles of "raise" in the website code, though some of those might be caught and handled more gracefully).

In this example -- and I did display the page on beta too just for fun, it is still rendering -- it now says "The roots of $f$ are computed in an extension of $\Q_{ 13 }$ to precision 10."  where it used to say "...1373".  Would users want to know that we do have roots to higher precision, i.e. they actually have been computed to precision 1373 but are only displayed to precision 10?  At some point we may implement data download for these (I don't think we do yet) and users might like to know that the downloaded data will be to higher precision.  On the other hand, once you know the roots to the lower precisino shown it is presumably easy to Hensel lift them to greater precision, which begs the question - why do you compute and store them to higher precision?  (You can see that I am ignorant, perhaps there is a simple answer to this).

jwj61:
The choice of p-adic precision is done by magma when you ask for Artin representations.  I suspect it is based on the p-adic valuation of the discriminant to guarantee enough precision to distinguish the roots.

I see the point of giving the p-adic roots is just to nail down specific elements of the Galois group, and for that, you just need them to be different.  So, I don't see it as important to make available precision available.  That said, I think it is nice to provide several p-adic digits even if 1 or 2 would suffice, which is how I settled on 10 digits, when possible.

Issue #4449, opened 2021-02-17
literal_eval in higher genus  raising IndexError


edgarcosta:
Here is the traceback:

```
Exception on /HigherGenus/C/Aut/2.8-1.0.2-8-8/T.2.1 [GET]
Traceback (most recent call last):
  File "/home/sage/sage-9.2/local/lib/python3.8/site-packages/flask/app.py", line 2447, in wsgi_app
    response = self.full_dispatch_request()
  File "/home/sage/sage-9.2/local/lib/python3.8/site-packages/flask/app.py", line 1952, in full_dispatch_request
    rv = self.handle_user_exception(e)
  File "/home/sage/sage-9.2/local/lib/python3.8/site-packages/flask/app.py", line 1821, in handle_user_exception
    reraise(exc_type, exc_value, tb)
  File "/home/sage/sage-9.2/local/lib/python3.8/site-packages/flask/_compat.py", line 39, in reraise
    raise value
  File "/home/sage/sage-9.2/local/lib/python3.8/site-packages/flask/app.py", line 1950, in full_dispatch_request
    rv = self.dispatch_request()
  File "/home/sage/sage-9.2/local/lib/python3.8/site-packages/flask/app.py", line 1936, in dispatch_request
    return self.view_functions[rule.endpoint](**req.view_args)
  File "/home/lmfdb/lmfdb-git-web/lmfdb/higher_genus_w_automorphisms/main.py", line 994, in topological_action
    GG = ast.literal_eval(topo_class[0]['group'])
IndexError: list index out of range
[2021-02-17 15:50:32 UTC] 500 error on URL https://www.lmfdb.org/HigherGenus/C/Aut/2.8-1.0.2-8-8/T.2.1 ()
```

PR #4450, opened 2021-02-19
LMFDB#4449 higher genus url breaks

BarinderBanwait:
This PR resolves the issue of #4449 . 

The issue is just that a search in `db.hgcwa_passports` might turn up empty. This is dealt with in several other places in the file in question (`/home/lmfdb/lmfdb-git-web/lmfdb/higher_genus_w_automorphisms/main.py`) by flashing an error and redirecting to the index page, so that seems like a reasonable approach here also.

Some unit tests have been added.

(My IDE also autodeleted trailing whitespace, apologies if this makes the PR appear messy.)


edgarcosta:
looks good to me. I will merge once tests pass

PR #4451, opened 2021-02-19
ECQ: handle curves with no modular degree

JohnCremona:
Simple code change so that elliptic curves with no modular degree (null in the 'degree' column of ec_curvedata) do not crash.

I wrongly thought that self.degree would be None, but it does not exist at all.


JohnCremona:
To test, go to the home page of a curve with prime conductor > 1e7.  For example, a random curve of rank 5 (of which there are now several).

JohnCremona:
Thanks!  These curves are now fine on beta

PR #4455, opened 2021-02-22
EC/Q: fix display for incomplete MW/BSD data

JohnCremona:
Until today all elliptic curves over Q had complete data computed concerning their Mordell-Weil group, but  right now there is exaclt one curve where this is not (yet) true, namely label 289788419.a1 .  This will crash on beta but after this PR it is OK.

I cannot be sure that this PR will display well all possible combination of missing data, but this case is fairly simple: analytic rank 1 (hence algebraic rank 1 by a Theorem), but we have no generator.  Consequently: the integral points shown (in this case none) may not be complete (though in this case the chances of the huge generator being integral are zero);   we do not know the regulator or Analytic Sha; however we do know the product of the last two (by BSD, but the definition of Analytic Sha means that we do know this product! by definition it is (L'(E,1)*|torsion|^2)/(tamagawa product*period), and it is expected to be the height of the generator, tnough it might be n^2 times that height if |Sha|=n^2, so I thought it worth showing.

Please take a look at http://localhost:37777/EllipticCurve/Q/289788419/a/1 and tell me what you think.

No other curve displays should be affected by this, but looking at a few random ones would be a good idea, especially those of conductor >200000000 (that's 2*10^8) which were only uploaded today.

### DIFF
--- a/lmfdb/elliptic_curves/templates/ec-curve.html
+++ b/lmfdb/elliptic_curves/templates/ec-curve.html
@@ -65,7 +65,7 @@ white-space: pre in order to keep line breaks! #}
     <h2> {{ KNOWL('ec.mordell_weil_group', title='Mordell-Weil group') }} structure</h2>
     <p>
     {% if data.mwbsd.rank == '?' %}
-    Not known
+    Not computed
     {% if data.mwbsd.rank_bounds.0>0 %}
     (\( {{data.mwbsd.rank_bounds.0}} \le r \le {{data.mwbsd.rank_bounds.1}} \))
     {% else %}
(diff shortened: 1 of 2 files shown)


JohnCremona:
I would like this to be looked at so that any new curves we wish to include in future will display OK even if some gens are missing.   For this specific curve 289788419a1, I do now have the generator but will not upload it yet!

jwj61:
It looks ok, but I would modify the wording.

I think we more often say "not computed" when something hasn't been computed.  "Not known" (or "No generator known") implies that no one knows it.  Some time in the future, someone may compute the value in question and until it is added to the database, the statement carries the wrong implication.

JohnCremona:
Good point -- especially since I now *have* computed the missing data for this curve,  I'll fix it

jwj61:
For more consistency, the modular degree is now "not available", but presumably that is because it has not been computed.

JohnCremona:
Yes. Modular degree not computed for large conductors.

JohnCremona:
As this is now merged, I'll not add another commit for the modular degree -- that suggestion came after I had fallen asleep.

Issue #3889, opened 2020-06-26
When simplifying the model of a genus 2 curve, also change points on the page


alexjbest:
I think it would be more convenient and cause less surprise for users if when a genus 2 curve is switched to a simplified model (completing a square) using the button on the page, the <s>Weierstrass points and</s> rational points also changed to reflect this new model.

<s>(Mumford reps representing the gens of the Jacobian would also be cool. but I think the utility of doing those too compared to the annoyingness to do it on the fly is such that I don't propose doing that here) </s>
<b>Edit</b>: Mordell-Weil generators should also change.

JohnCremona:
This is the main reason I have resisted suggestions to include short Weierstrass equations for elliptic curves (with y^2=cubic or y^2=x^3+ax+b).  Though I have not prevented others from doing so, provided that they make *all* the other changes to the page contents!  As well as explicit points, there is also the discriminant (unless we change its description to "minimal discriminant" to be invariant, and the real period (similarly).

AndrewVSutherland:
@alexjbest I'm not sure what you are referring to when you mention Weierstrass points.  The coordinates of Weierstrass points are not displayed, only the number of Weierstrass points is (this number does not depend on the model).

I don't think it makes sense to change the rational points and not change the MW generators, since these often include rational points.

@JohnCremona  Whenever possible, anything that can be defined as an invariant of the (isomorphism class of) curve should be defined that way; so the discriminant of a curve should always mean the minimal discriminant (which makes sense even in cases of curves over number fields where there may be no model that achieves it).  The genus 2 discriminant knowl already makes this clear.  It is the curve's home page after all, not the model's.  

One other thing to bear in mind in this discussion is that for genus 3 curves we want to define these invariants consistently in situations where the models may be more complicated (e.g. degree-2 covers of pointless conics which cannot be defined by a single equation over Q); this is another reason not to make them model independent whenever possible.

AndrewVSutherland:
I meant to add that the genus 2 knowls for rational points and MW generators explicitly refer to the minimal equation.  I realize users may not necessarily click on the knowl, but if they do the situation will be clear.

AndrewVSutherland:
We just received another request for this feature.  I think we should do this, but when doing so change both the rational points and th Mordell-Weil generators.

AndrewVSutherland:
Closed via #4468

PR #4468, opened 2021-02-28
Points change when model is changed

rbommel:
This is an attempt to resolve issue #3889.


AndrewVSutherland:
I like it, thanks @rbommel !

Issue #4478, opened 2021-03-05
Decomposition of the Jacobian of a genus 2 curve as product of elliptic curves


edgarcosta:
For some genus 2 curves, whose Jacobian factors as a product of elliptic curves we go further than specifying isogeny class, as we specify the curves of the elliptic curve, for example:
https://www.lmfdb.org/Genus2Curve/Q/994009/a/994009/1
but we still say
> Decomposes up to isogeny as the product of the non-isogenous elliptic curves

and we also don't say how we picked each curve from its isogeny class.

edgarcosta:
@JRSijsling do you recall how we picked a curve in each isogeny class?

JRSijsling:
Yeah, it is a bit weird to give specific curves here. In this particular case, where there are two non-isogenous elliptic curves, there are distinguished central idempotents. In other words, the endomorphism algebra is QQ x QQ, and the elements (1, 0) and (0, 1) in that algebra are truly distinguished and uniquely determined. These are applied to the lattice, and the resulting image furnishes the curves that are indicated.

Things are different when the algebra is, for instance, M_2 (QQ): Then there are no central idempotents, uniqueness is lost, and the result counts for less because just "some" idempotent is taken.

So frankly, in general it only makes sense to indicate the isogeny classes.

rbommel:
See also pull suggest #4489.

rbommel:
[![](https://camo.githubusercontent.com/fcaf500cf6cd78ea7e41a56c61c688c9f2e4f2873113bffc8873c9dd4903dfa2/68747470733a2f2f692e696d67666c69702e636f6d2f3539686732312e6a7067)](https://imgflip.com/i/59hg21)
[from Imgflip Meme Generator](https://imgflip.com/memegenerator)

PR #4489, opened 2021-03-19
Show isogeny classes instead of curves in 'Jacobian is isogenous to...'

rbommel:
This should resolve issue #4478. You can see the changes for example on:
http://localhost:37777/Genus2Curve/Q/196/a/21952/1 and http://localhost:37777/Genus2Curve/Q/196/a/

### DIFF
--- a/lmfdb/genus2_curves/templates/g2c_isogeny_class.html
+++ b/lmfdb/genus2_curves/templates/g2c_isogeny_class.html
@@ -77,6 +77,14 @@ text-align: center;
 {% endif %}
 </p>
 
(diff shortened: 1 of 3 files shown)


AndrewVSutherland:
There is a test failure caused by this change in "test_by_curve_label".  Presumably it is looking for a complete elliptic curve label where there is no only an isogeny class label.

AndrewVSutherland:
lmfdb/genus2_curves/test_genus2_curves.py:23: AssertionError

rbommel:
Okay, I changed this now. Hopefully it works, but I have no way to test it myself as the test script doesn't run anymore in my Legendre account. It says "no module named 'lmfdb.utils'".

AndrewVSutherland:
Looks good.

PR #4508, opened 2021-04-07
Added Base Field search

limarta:
Added feature to search by moduli fields for Belyi maps.

### DIFF
--- a/lmfdb/belyi/main.py
+++ b/lmfdb/belyi/main.py
@@ -16,6 +16,7 @@ from lmfdb.utils import (
     display_knowl,
     parse_ints,
     parse_bracketed_posints,
+    parse_nf_string,
     redirect_no_cache,
     search_wrap,
     Downloader,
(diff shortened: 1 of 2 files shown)


lmfdb/belyi/main.py
```diff
@@ -764,8 +766,14 @@ def __init__(self):
             label="Geometry type",
             knowl="belyi.geometry_type",
             options=[("", "")] + list(geometry_types_dict.items()))
+        field = TextBox(
+            name="field",
+            label="Moduli field",
+            knowl="ag.base_field",
```
edgarcosta, line 772: This points to the wrong knowl

edgarcosta: The knowl should be: belyi.moduli_field

edgarcosta:
@jvoight and @SamSchiavone  are figuring out what is the right knowl to use here and in the column headings.
Given that belyi is still restricted to beta I would propose to go ahead and merge this.

SamSchiavone:
Yeah, there's some field of definition vs. field of moduli and pointed vs. unpointed variants of these that needs to get sorted out. For the moment, maybe it's best to just change the knowl to `belyi.base_field`, which already exists.

edgarcosta:
Superseded by #4534

PR #4515, opened 2021-04-14
pass_size search

aahhoey:
I added a search feature for passport size to the Belyi database.


edgarcosta:
@aahhoey feel free to add yourself to the contributor's list: https://github.com/LMFDB/lmfdb/blob/master/CONTRIBUTORS.yaml

PR #4526, opened 2021-04-16
Footer tweak

AndrewVSutherland:
This PR reduces the height of the footer to make it more consistent with the thinner header we now use.


AndrewVSutherland:
Travis error is due to a bug in the travis script (which has been fixed).

Issue #3771, opened 2020-03-25
Wishlist: search option "disjoint from" a set of primes for bad primes


AurelPage:
It would often be useful to be able to specify a list of primes that you don't want to be bad. This probably applies to many objects, but I am mainly thinking about classical modular forms ("Bad p" field); note that it could also be a separate search field like for global number fields ("Unramified primes" field).

The automatic "similar issues" feature of github suggests #3160, which makes me realise that specifying the number of bad primes also makes sense and could also be useful for modular forms.

edgarcosta:
Would be enough to have an option of "exclude" as in
https://www.lmfdb.org/EllipticCurve/Q/
?

AurelPage:
Yes, "exclude" is the same as "disjoint from".

edgarcosta:
I'm aware that "exclude" = "disjoint from", but the question would it be enough to have a single box, or are you asking for two boxes, as in number fields https://www.lmfdb.org/NumberField/?

AurelPage:
Either would be fine, but of course two boxes allow for finer searches. This could also be an opportunity to uniformise the way it is done in various objects: several search fields vs only one, formulation ("include" vs "superset") etc.

edgarcosta:
We have been trying to uniformize the search. 
Take a look at: https://olive.lmfdb.xyz/NumberField/
At the moment, seems that only place we are missing the exclude option is classical modular forms.

AurelPage:
I see, thanks! It seems that only number fields has a separate search box, so maybe simply adding an "exclude" option to the "Bad p" box of modular forms is reasonable.

JohnCremona:
I think we should be consistent across all objects we have which have a notion of bad or ramified primes, which is a lot of them. We should have the same search options for all these using as similar wording as possible

rbommel:
See also PR #3776.

roed314:
And PR #3779.

rbommel:
Is this issue resolved now?

AndrewVSutherland:
None of HMF, BMF, ECNF provide an option to constrain primes dividing the conductor norm.  Just adding a bad primes search option should automatically provide include/exclude/exactly/subset options.

JohnCremona:
What these three have in common is that the primes in question are number field primes so it is not at all clear how a user might input them.  Easier, though less good, would be for the database to include a list of prime numbers dividing the *norm* of the level / conductor.  Then one could easily constrain these in a search.

[...]

PR #4527, opened 2021-04-16
Adding search option for bad primes 

rbommel:
Adding search option for bad primes  for elliptic curves over number fields, Bianchi modular forms, and Hilbert modular forms, as requested in issue #3771.


AndrewVSutherland:
The search box layout on http://127.0.0.1:37777/EllipticCurve/ is pretty ugly.  One possible fix is to change "Primes dividing the conductor" to "Bad primes", which might be short enough to get the j-invariant box to align properly (or just change the length of the j-invariant input).

Neither caption is strictly correct (these are rational primes dividing the norm of the conductor), but "Bad primes" is shorter, more consistent with usage elsewhere, and arguably less confusing than "primes dividing the conductor" (these primes are "bad" for every reasonable definition of "bad" but they do not necessarily "divide the conductor" under the most natural definition of "dividing the conductor").

rbommel:
Changed "Primes dividing the conductor" into "Bad primes".

PR #4530, opened 2021-04-17
Source and acknowledgments

AndrewVSutherland:
This PR completes the source/acknowledgment knowl changes for all production sections of the LMFDB and renames all of the old dq.* knowls in production sections to the corresponding rcs.* knowl names where appropriate.  Credit strings have been copied into acknowledgments knowls for the moment, but these will require further editing (both expanding them to say more and removing information that is already contained in the source knowl).

There are also now generic and specific source and acknowledgment knowls for L-function data (specific knowls that depend on the origin are stored in the lfucn_rs_knowl table as before).

Credit strings no longer appear in the footer on any of the production pages (but they are still present for the beta sections Belyi/Siegel/HGM/Lattices -- these will need to be updated to use the new framework before we go into production).

There are also some minor page layout tweaks (footer is thinner and positioned more consistently, show commands and the  x -> x button now float right to avoid pushing the first header down the page, title bar now has a rounded lower right corner like it is supposed to, property box spacing has been lightly tweaked, header/footer box shadows have been removed, etc... -- most of these will not be noticeable unless you compare the pages side-by-side).

I am going to merge this as soon as Travis is happy. This PR touches almost every section of the LMFDB and I want to avoid potential merge conflicts.  I will then push to beta but then wait a few days before pushing to production to give people a change to catch anything I might have missed.

### DIFF
--- a/lmfdb/abvar/fq/main.py
+++ b/lmfdb/abvar/fq/main.py
@@ -39,12 +39,10 @@ def get_bread(*breads):
         bc.append(z)
     return bc
 
-abvarfq_credit = "Taylor Dupuy, Kiran Kedlaya, David Roe, Christelle Vincent"
-
 def learnmore_list():
     return [
+        ("Source and acknowledgments", url_for(".how_computed_page")),
         ("Completeness of the data", url_for(".completeness_page")),
-        ("Source of the data", url_for(".how_computed_page")),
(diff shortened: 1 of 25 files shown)


AndrewVSutherland:
Travis doesn't seem to want to run, but tests pass on grace.

PR #4533, opened 2021-04-21
updated various entries in the Contributors list

JohnCremona:
Updated several affiliations and URLs


JohnCremona:
It would be good to have someone else take a quick look at these, especially as some are MIT and Dartmouth people (looking at you @edgarcosta )

PR #4534, opened 2021-04-21
Base Field Search Fixed

kgravel:
supersedes https://github.com/LMFDB/lmfdb/pull/4508

added number field search, fixed knowls, and resolved conflicts

### DIFF
--- a/lmfdb/belyi/main.py
+++ b/lmfdb/belyi/main.py
@@ -758,7 +758,7 @@ class BelyiSearchArray(SearchArray):
         pass_size = TextBox(
             name="pass_size",
             label="Passport size",
-            knowl="belyi.passport",
+            knowl="belyi.pass_size",
             example="2",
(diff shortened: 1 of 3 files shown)


AndrewVSutherland:
This seems to work so I am going to merge it, but I think it should give an error message rather than no search results when the user enters an invalid number field label such as 1.2.3.4.5.6.7 or one that doesn't exist such as 1.1.1.2.

SamSchiavone:
@AndrewVSutherland Ah, that's good to know. We copied the base field search code from elliptic curves over number fields, which seems to have the same problem:

https://beta.lmfdb.org/EllipticCurve/?field=1.1.1.1.1.1&search_type=List

But the jump box on the number fields browse page will throw
> Error: 1.1.1.1.1 is not a valid label for a number field.
if given an invalid label like 1.1.1.1.1.

PR #4540, opened 2021-04-27
Hopefully final LocalNumberField -> padicField changes

jwj61:
Entering a url using LocalNumberField now redirects to padicField.

Also, links in the search table were still pointing to LocalNumberField, so that is now fixed.

A few other places were not using url_for (or some variation), and those are now fixed.  (There is no visible change to the user on those.)

Pages to check are anything in the padicField area, such as

http://127.0.0.1:37777/padicField/?p=3&n=6  (search results)
http://127.0.0.1:37777/LocalNumberField/3.6.6.4  (both that it redirects, and that the root field and unramified subfield friends still go to the correct places)

### DIFF
--- a/lmfdb/local_fields/__init__.py
+++ b/lmfdb/local_fields/__init__.py
@@ -27,6 +27,7 @@ def redirect_local():
 
 
 app.register_blueprint(local_fields_page, url_prefix="/padicField")
+app.register_blueprint(local_fields_page, url_prefix="/LocalNumberField")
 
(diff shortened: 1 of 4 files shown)


AndrewVSutherland:
I seem to get a 404 on http://127.0.0.1:37777/LocalNumberField/3.6.6.4

jwj61:
Should be fixed now

edgarcosta:
looks good to me

Issue #4512, opened 2021-04-10
Root angle L-function search option needs better knowl and to not throw server errors


AndrewVSutherland:
The root angle input on https://beta.lmfdb.org/L/ currently links to the root number knowl, but it really needs its own knowl which explains what the root angle is and how the input will be interpreted.

Similarly the Arg(epsilon) column in the search results needs a knowl that explains that it isn't actually displaying Arg(epsilon), it's displaying Arg(epsilon)/(2pi) mod 1 in (-0.5,0.5] (at least that's what I think it is displaying).

Also, it should not throw a server error on the input 0.1..0.6:

https://beta.lmfdb.org/L/?root_angle=0.1..0.6&search_type=List

AndrewVSutherland:
Resolved by #4547

PR #4547, opened 2021-04-30
Root angle search

roed314:
Addresses #4512.  Things to look at:

* Make sure that inputting `0.1-0.6` no longer gives a server error, and that various search inputs work on root angle.
* Check the two knowls that I've created

### DIFF
--- a/lmfdb/classical_modular_forms/main.py
+++ b/lmfdb/classical_modular_forms/main.py
@@ -712,7 +712,7 @@ def common_parse(info, query, na_check=False):
                 query['level'] = {'$in': ZZ(query['level']).divisors()}
         else:
             query['level_is_' + info['level_type']] = True
-    parse_floats(info, query, 'analytic_conductor', name="Analytic conductor", allow_singletons=True)
+    parse_floats(info, query, 'analytic_conductor', name="Analytic conductor")
     parse_ints(info, query, 'Nk2', name=r"\(Nk^2\)")
     parse_ints(info, query, 'char_order', name="Character order")
     parse_primes(info, query, 'level_primes', name='Primes dividing level', mode=info.get('prime_quantifier'), radical='level_radical')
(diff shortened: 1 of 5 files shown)


AndrewVSutherland:
Still seems a little wonky:

- 0-1 results in 0.00 not -0.5-0.5, which I think is a bug
- 0-0.6 gives 0.0-0.5,-0.5--0.4 as expected, but 0-0.7 gives 0.0-0.5,-0.5--0.30000000000000004, which is correct (within epsilon) but a little startling

The knowls look good.

roed314:
Do you also want `0-0` to give just the point `0.00`?  With the current strategy it's hard to distinguish between `0-0` and `0-1` since the first thing that's done is reduce everything mod 1.

AndrewVSutherland:
It would probably make sense to convert a-a to a before reducing mod 1.

roed314:
Are you okay with `0.4-0.2` being the same as `0.4-1.2` (namely the complement of `0.2-0.4`)?

AndrewVSutherland:
I have no problem with this, it is what I would expect.  But I also would have no problem with displaying an error for intervals a-b with a>b if you are concerned that 0.4-1.2 would be confusing.

roed314:
Ok, I basically rewrote `parse_floats` and `parse_mod1` so that they could both use an alternative to `parse_range2`, rather than trying to work around the limitations of that function.

Some consequences:

* Across the LMFDB, we now accept singletons for floats, which are converted into a small range (width `10^-10`)
* If you enter `1-` or `0-1` or something that covers more than a unit interval, your input is removed since it provides no constraint.
* Backward inputs like `0.4-0.2` raise an error.
* It does a better job of not changing inputs if they don't need to be changed, and not printing a silly number of digits.
* Small denominators (those dividing 4) are now privileged, and converted from a float into a fraction.  Type `0.25` for example.  Under the hood, it's still searching for these as a float range.

AndrewVSutherland:
@roed314 Thanks for this, it is much improved!  The only issue I found with the new version is that the input 1-1.1 produces a strange error:

    http://127.0.0.1:37777/L/?hst=List&root_angle=1-1.1&search_type=List

whereas 1.0-1.1 works as expected.

roed314:
It should be fixed now.

PR #4550, opened 2021-04-30
Update number field search regex

SamSchiavone:
As noted by @AndrewVSutherland in this PR (https://github.com/LMFDB/lmfdb/pull/4534), giving an invalid number field label when searching for elliptic curves, abelian varieties, or Belyi maps of a given base field would return no results rather than raising an error. This was due to an incorrect regex, which I've replaced.


AndrewVSutherland:
This is an improvement, so I will merge it, but it still doesn't alert the users to labels that don't match any number field, like 0.1.2.3 or 3.0.23.2 (doing so would help catch typos).

Issue #4509, opened 2021-04-08
Request for additional p-adic field search options


AndrewVSutherland:
It would be nice to be able to search on more invariants of p-adic fields, especially of their Galois closures, since the Galois closures themselves are very often not present.  In particularly, I'd like to be able to search on the inertia group by GAP id and the tame degree.  It would also be nice to be able to search on the wild inertia group by GAP id.

PR #4551, opened 2021-04-30
added search fields for unramified degree and tame degree

jakobderaaij:
Addressing a part of issue #4509 that was asking for the possibility to search for p-adic fields by their unramified degree and tame degree. I added these two search boxes to the p-adic field start page and to the p-adic field search page. They work exactly like the other search boxes on these pages.  

Before: http://www.lmfdb.org/padicField/ and http://www.lmfdb.org/padicField/?search_type=List&all=1
After: http://localhost:37777/padicField/ and http://localhost:37777/padicField/?u=4&t=2&search_type=List

### DIFF
--- a/lmfdb/local_fields/main.py
+++ b/lmfdb/local_fields/main.py
@@ -481,22 +481,22 @@ class LFSearchArray(SearchArray):
                 display_knowl('gg.label', 'transitive group labels')))
         u = TextBox(
             name='u',
-            label='Unramified degree',
+            label='Galois unramified degree',
             knowl='lf.unramified_degree',
             example='3',
-            example_span='4, or a range like 1..4'
+            example_span='3, or a range like 1..4'
             )
         t = TextBox(
             name='t',
-            label='Tame degree',
+            label='Galois tame degree',
             knowl='lf.tame_degree',
-            example='3',
+            example='2',
             example_span='2, or a range like 2..3'
             )
         results = CountBox()
 
         self.browse_array = [[degree], [qp], [c], [e], [topslope], [u], [t], [gal], [results]]
-        self.refine_array = [[degree, c, gal], [qp, e, topslope], [u, t]]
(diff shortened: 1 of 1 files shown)


AndrewVSutherland:
@jwj61 Can you review this when you get a chance (it will only take a moment) just to be sure you are happy with it?  I think Jakob is planning to do some more work related to #4509 next week, but I suggested doing these first.

jwj61:
Will do.

jwj61:
On the search results page, I don't we can add a column instead of another line.

The knowls need to be adjusted.  On the page of a field, they are in a section about the Galois closure which I think is enough to indicate that they are the tame and unramified degrees of the Galois closure.  But, the notions make sense for any p-adic field, so a user might think that the search boxes are referring the the fields in the search results as opposed to their Galois closures.  In short, they need to be labelled differently in this context.

AndrewVSutherland:
@jwj61 -- thanks for the feedback, I agree with both the points you made and will hold off merging until they have been addressed. The layout changes are easy and I can help Jakob with the know edits (but feel free to make them yourself if you prefer).

jwj61:
Maybe the labels for them should be "Galois unramified degree" and "Galois tame degree".

I suppose the knowls could each have an added paragraph that just said, e.g., The Galois unramified degree of a field $K$ is the unramified degree of its Galois closure.

Thinking about it now, it might also make sense to make the connection in the knowl between the unramified degree and the residue field degree.

jakobderaaij:
I changed the layout of the search result page such that the search boxes are now all in two rows and have clearer labels.
I edited the knows of the unramified and tame degree to contain the information about Galois unramified/tame degree.

edgarcosta:
Great work @jakobderaaij feel free to add yourself to the contributors list https://github.com/LMFDB/lmfdb/blob/master/CONTRIBUTORS.yaml

PR #4554, opened 2021-05-01
Replacing travis with github actions

edgarcosta:
In this PR @alexjbest  and I replaced @travis-ci, which no longer supports open source projects in practice, with GitHub actions.

At the moment we run 10 independent jobs, the first one is just linting, and for the rest, we tried to divide them uniformly among the various sections of the lmfdb so that each job would take more or less the same time.
The big outliers were classical modular forms and L-functions.
For some runs `test_cmf.py` was taking more than 15 minutes, thus as a simple fix, we divided the file in half.

Github actions support 20 simultaneous jobs, so we are hoping that two test sets can run simultaneously and each take less than 10 minutes.

Here is a breakdown of the current running times:
Screen Shot 2021-04-30 at 21 29 59


AndrewVSutherland:
LGTM.

PR #4555, opened 2021-05-04
Newpolyfix

edgarcosta:
I did some refactoring so we can display the message:
```
Note: The requested field Q[x]/f(x) is isomorphic to the field below, but uses a different defining polynomial.
```
- with polynomial properly formatted in latex, e.g., https://beta.lmfdb.org/NumberField/?jump=x^10+%2B+2048 vs http://localhost:37777/NumberField/?jump=x^10+%2B+2048
Screen Shot 2021-05-04 at 11 58 51

- display info message even if one didn't use x as the defining variable, e.g., https://beta.lmfdb.org/NumberField/?jump=w^10+%2B+2048 vs http://localhost:37777/NumberField/?jump=w^10+%2B+2048
Screen Shot 2021-05-04 at 11 59 07

### DIFF
--- a/.environment.yml
+++ b/.environment.yml
@@ -10,8 +10,6 @@ channel_priority: strict
 safety_checks: disabled
 channels:
   - conda-forge
-channels:
-  - conda-forge
 dependencies:
   - python=3.9
   - sage=9.2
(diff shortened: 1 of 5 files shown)


roed314:
LGTM

PR #4558, opened 2021-05-07
fix typo in allowed_annotation_id

AndrewVSutherland:
Hotfix to a bug caused by a typo in lmfdb/knowledge/main.py.

@edgarcosta @alexjbest according to the details on PR #4557, all 20 checks passed.  But pyflakes gives an error on this typo that was apparently not part of any of the checks.   Do we need to add pyflakes to the list of tests we run via GitHub actions?


edgarcosta:
I'm looking into it.

PR #4560, opened 2021-05-07
Explicit pyflakes path

edgarcosta:


AndrewVSutherland:
LGTM (and the only real way to test it is to merge it so I will).

PR #4561, opened 2021-05-07
fix stgroup g2 zvector display

AndrewVSutherland:


AndrewVSutherland:
This is a minor formatting fix to restore the zvector entry for a2=-2 for degree 4 Sato-tate groups which was inadvertently removed in the process of changes made to support degree 6 groups.

I'll merge this as soon as tests pass.

PR #4570, opened 2021-05-16
Add inertia and wild group search

jwj61:
Addresses the rest of issue #4509 by allowing one to search by gap id for the inertia and wild inertia subgroups.

There are two new columns in the database for these two search options.  Currently there are no indexes for them, but it already works and they can be added.

We were not displaying the wild inertia group up to isomorphism before, so this is added to the field's home pages.

Some items to be improved upon later:

- there are currently two inertia subgroups with orders which are not indexed by magma, 1536 and 2430, so you cannot search for those groups
- there is one wild ramification group of order 512.  If someone wants to compute its gap id, we can add that information.  It is the Sylow 2-subgroup of 12T222 (a.k.a. (D_4)^3).  There are roughly 3 million groups to check with the same rank and p-class (gap id numbers 7532393 to 10481221).
- you can only search by gap id.  An improvement would be to allow searching by nickname, and by T-notation.  The latter opens questions though -- if someone asks for inertia subgroup in T-notation, do we include places where the group is right as an abstract group, but maybe naturally sits only as an intransitive subgroup for the given polynomial?

To test, just search on the local fields page.  An example where data is not computed for the wild ramification group is

  http://127.0.0.1:37777/padicField/2.12.29.1

In almost every other case, it is computed.

### DIFF
--- a/lmfdb/galois_groups/transitive_group.py
+++ b/lmfdb/galois_groups/transitive_group.py
@@ -52,7 +52,7 @@ def smallgroup_cache(gap_labels, cache=None, reverse=None):
 def small_group_display_knowl(n, k, name=None, cache=None):
     label = '%s.%s' % (n, k)
     if not name:
-        myname = '$[%d, %d]$'%(n,k)
+        myname = 'GAP id $[%d, %d]$'%(n,k)
     else:
         myname = name
     if cache:
--- a/lmfdb/galois_groups/transitive_group.py
+++ b/lmfdb/galois_groups/transitive_group.py
@@ -644,6 +644,11 @@ def group_alias_table():
     ans += r'</tbody></table>'
     return ans
 
+def nt2gap(n, t):
+    res = db.gps_transitive.lookup('{}T{}'.format(n,t))
+    if res and 'gapid' in res:
+        return [res['order'], res['gapid']]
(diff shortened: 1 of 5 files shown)


AndrewVSutherland:
The GAP ID of the 2-Sylow subgroup of 12T222 is <512,7606661>.

jwj61:
That was fast!

jwj61:
The data has been inserted.  On the page above, you just get the gap id and not a knowl since the group is not in the old small groups part of the lmfdb.

jwj61:
The inputs for inertia and wild inertia should now take nicknames and nTj notation.  There will be one more code change to change the knowls so that the ones for searching explain the details/distinctions being made (e.g., why C3=[3,1] always and [3,1]=3T1 for wild inertia, but [3,1] != 3T1 for inertia).

jwj61:
Hopefully now done.

Along the way, I added search on the residue field degree.  In testing, I wanted to specify that (e.g., searching for inertia 3T1 should be the same as f=1 and group C3).

AndrewVSutherland:
@jwj61 Are you planning to address the merge conflict?  It looks easy to fix.

jwj61:
Yes, I hadn't noticed it was there.

Issue #4571, opened 2021-05-19
Deprecation warning from sage-9.3


JohnCremona:
I noticed that one line in lmfdb/nfutils/psort.py (which is a copy of a file in https://github.com/JohnCremona/sorting) needs a minor change to avoid a DeprecationWarning from the recently released Sage-9.3: just changed ``K.galois_group('gap').group()`` to ``K.galois_group()``.  But the changed code would not work with 9.2, so I'll fix this in a way which checks the current Sage version and adjusts if necessary -- unless there's a better way?

@roed314 will recognise this, as the author of https://trac.sagemath.org/28782.

roed314:
I don't think there's a way that will work well in both 9.2 and 9.3.  I would use `K.galois_group(algorithm='gap')` in 9.3 if the field could ever have degree at least 12 (when Pari stops being able to compute Galois groups).

JohnCremona:
Definitely not.  In fact right now it's only used for degree 2 where the line which uses this is completely trivial, but the code in psort is more general.

The use for BMFs is (1) to generate a list of prime ideals for displaying eigenvalues; (2) to construct an ideal from its label and vice versa for display.  This is only for imaginary quadratic fields of class number 1.

I'll do a special case for degree 2 for the line I just changed (which computes the set {1,2}).

PR #4572, opened 2021-05-19
avoid DeprecationWarning in Sage 9.3

JohnCremona:
I made a minor change in nfutils/psort so that the code works both with Sage versions <=9.2 and with 9.3: small change in interface to the galois_group mehotd for number fields.  See #4571.  

I also changed one line in logger/start which gives a warning if users start the website with a version of Sage which is "too old" -- that used to be 7.2 and is now 9.2.  This is probably redundant now that (almost?) no development happens outside legendre, but is surely harmless.

### DIFF
--- a/lmfdb/logger/start.py
+++ b/lmfdb/logger/start.py
@@ -17,7 +17,7 @@ def logger_file_handler():
     # set by start_logging
     return file_handler
 
-LMFDB_SAGE_VERSION = '7.1'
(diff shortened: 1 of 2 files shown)


roed314:
See my comment on the issue.  Will the degree of the field ever be larger than 11?

JohnCremona:
Hold off on this, I'll update it tomorrow.

Issue #4576, opened 2021-05-21
BMF broken 2.0.19.1-19.1-a


edgarcosta:
https://beta.lmfdb.org/ModularForm/GL2/ImaginaryQuadratic/2.0.19.1/19.1/a/

```
Exception on /ModularForm/GL2/ImaginaryQuadratic/2.0.19.1/19.1/a/ [GET]
Traceback (most recent call last):
  File "/home/sage/sage-9.2/local/lib/python3.6/site-packages/flask/app.py", line 2447, in wsgi_app
    response = self.full_dispatch_request()
  File "/home/sage/sage-9.2/local/lib/python3.6/site-packages/flask/app.py", line 1952, in full_dispatch_request
    rv = self.handle_user_exception(e)
  File "/home/sage/sage-9.2/local/lib/python3.6/site-packages/flask/app.py", line 1821, in handle_user_exception
    reraise(exc_type, exc_value, tb)
  File "/home/sage/sage-9.2/local/lib/python3.6/site-packages/flask/_compat.py", line 39, in reraise
    raise value
  File "/home/sage/sage-9.2/local/lib/python3.6/site-packages/flask/app.py", line 1950, in full_dispatch_request
    rv = self.dispatch_request()
  File "/home/sage/sage-9.2/local/lib/python3.6/site-packages/flask/app.py", line 1936, in dispatch_request
    return self.view_functions[rule.endpoint](**req.view_args)
  File "/home/lmfdb/lmfdb-git-dev/lmfdb/bianchi_modular_forms/bianchi_modular_form.py", line 567, in render_bmf_webpage
    data = WebBMF.by_label(label, max_eigs=numeigs)
  File "/home/lmfdb/lmfdb-git-dev/lmfdb/bianchi_modular_forms/web_BMF.py", line 68, in by_label
    return WebBMF(data, max_eigs)
  File "/home/lmfdb/lmfdb-git-dev/lmfdb/bianchi_modular_forms/web_BMF.py", line 57, in __init__
    self.make_form(max_eigs)
  File "/home/lmfdb/lmfdb-git-dev/lmfdb/bianchi_modular_forms/web_BMF.py", line 190, in make_form
    curve_bc_parts = [split_lmfdb_label(lab) for lab in curve_bc]
  File "/home/lmfdb/lmfdb-git-dev/lmfdb/bianchi_modular_forms/web_BMF.py", line 190, in <listcomp>
[...]

Issue #4573, opened 2021-05-19
Add BMFs for non-Euclidean imagainary quadratic fields of class number 1


JohnCremona:
We currently have Bianchi modular forms of dimension 1 only for the five Euclidean IQFs (for which modular symbol computations are simpler).  I will extend the data aset to include the four non-Euclidean fields of class number 1 (discriminants -19, -43, -67, -163) using a re-implementation of the theory developed in Elise Whitley's 1990 thesis being developed at https://github.com/JohnCremona/bianchi-progs with data in https://github.com/JohnCremona/bianchi-data.

Right now I had 2597 newforms over Q(sqrt(-19)) with level norm up to 10000, with associated elliptic curves, and will upload these soon, once I have modernised by Bianchi data upload scripts.  43 should not be far behind.

JohnCremona:
Update 2021-06-02:  all ecnf and bmf data for 2.0.19.1 is uploaded and copied to production together with small code changes.  Now all ecnf and bmf data for 2.0.43.1 is also uploaded, so the tables ec_nfcurves, bmf_forms and bmf_dims also need copying.  There will also be another minor code change (PR imminent) but nothing breaks with the new data and current beta/prod branches.

JohnCremona:
Update 2021-06-16: 2.0.19.1 now uploaded (both BMFs and elliptic curves) up to level/conductor norm 15000, 43 ditto up to 10000, 67 is still running and will go up to 10000 when finished, and 2.0.163.1 is finished and uploaded up to 1000 (that's only 10^3, not 10^4: this field is slower!).  I'll close this issue when I have uploaded data for 2.0.67.1 as then we'll have something for all the class number one fields.

JohnCremona:
OK, there is now data on beta for all four of these fields, both rational BMFs of weight 2 and elliptic curves, with a norm bound of 15000, 10000, 10000, 1000.    I have updated the knowls regarding extent and origin, and tomorrow I'll add the remaining two fields to the search pages https://beta.lmfdb.org/ModularForm/GL2/ImaginaryQuadratic/ and https://beta.lmfdb.org/EllipticCurve/.

JohnCremona:
I just uploaded more data for -163 (i.e. 2.0.163.1) for level norms 1001-5000, also elliptic curves to match.

Please could the tables ec_nfcurves, bmf_forms, bmf_dims be copied to prod?  Then this can be closed.  No code changes needed.

[...]

PR #4577, opened 2021-05-21
BMF & ECNF: minor updates for new data for 2.0.19.1

JohnCremona:
I have added BMF and ECNF data for a new field, Q(sqrt(-19)).  Some minor code changes were needed.  This PR also clears out old redundant upload scripts.

When these changes are pushed to prod, the ec_nfcurves tabel should also be copied.  By that time I will have also updated the relevant RCS knowls.

This PR relates to #4573, and the code update fixes something noticed in #4576.


AndrewVSutherland:
ec_nfcurves has been copied to production

JohnCremona:
> ec_nfcurves has been copied to production

Thanks!  2.0.43.0 should not be much longer, I have almost got to level norm 10000 with the BMFs.  Meanwhile if you could also copy bmf_forms and bmf_dims then these new curves should have matching forms on prod (except for one conjugate pair which does not, but this should be indicated properly).

AndrewVSutherland:
Does bmf_forms also need to be copied over?

JohnCremona:
Yes please (see above)

AndrewVSutherland:
Done!  See https://www.lmfdb.org/ModularForm/GL2/ImaginaryQuadratic/2.0.19.1/19.1/a/.

PR #4580, opened 2021-05-24
CM discriminant search fix

SamSchiavone:
@AndrewVSutherland This should address the issue you brought up with the CM discriminant search. Please let me know if there are still problems.

### DIFF
--- a/lmfdb/ecnf/main.py
+++ b/lmfdb/ecnf/main.py
@@ -15,12 +15,14 @@ from lmfdb import db
 from lmfdb.backend.encoding import Json
 from lmfdb.utils import (
     to_dict, flash_error,
-    parse_ints, parse_noop, nf_string_to_label, parse_element_of,
+    parse_ints, parse_ints_to_list_flash, parse_list, parse_noop, nf_string_to_label, parse_element_of,
     parse_nf_string, parse_nf_elt, parse_bracketed_posints, parse_bool, parse_floats, parse_primes,
     SearchArray, TextBox, ExcludeOnlyBox, SelectBox, CountBox, YesNoBox, SubsetBox, TextBoxWithSelect,
     search_wrap, parse_rational,
     redirect_no_cache
     )
+from lmfdb.utils.search_parsing import search_parser
+
 from lmfdb.utils.interesting import interesting_knowls
 from lmfdb.number_fields.number_field import field_pretty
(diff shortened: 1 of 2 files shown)


JohnCremona:
Thanks @SamSchiavone -- I am testing this now.

JohnCremona:
The PR makes progress but does not yet address the issue.  After parsing the discriminant list (checking that every D is negative and 0 or 1 mod 4) the code now appends the absolute value in the list before searching.  OK so far but the 4-way switch (CM, potential Cm etc) needs to be taken into account.  If the user  wants actual CM (not just potential) then *only* the absolute values should be searched for, if "potential CM" then both signs, if "potential CM but no CM" then negative only, and lastly if "no potential CM" then some negation is needed to exclude from the results and curve with either +D or -D in the column.  The logic of this when more than one discriminant is supplied is not very clear, but I suppose must be "or" rather than "and" since no curve can have any kind of CM with more than one discriminant, so if I enter -3,-4 and select CM then I should get curves wth actual CM by discriminant either -3 or -4, which will all be defined over either 2.0.3.1 or 2.0.4.1 respectively.

Note that right now we will only ever see actual CM over one of the imaginary quadratic fields (so far 2.0.d.1 for d=1,2,3,7,11,19, 43 will be soon) but it would not be good to assume this in the code since one day we may have more curves over fields properly containing an imaginary quadratic field.

SamSchiavone:
@JohnCremona Thanks for feedback. I think I’ve made my CM discriminant parser work with the 4-way switch now. If I search for CM discriminant -11 and PCM but no CM, the first result is /EllipticCurve/2.0.3.1/14641.1/a/1; if I search for the same CM discriminant but with CM, the first result is /EllipticCurve/2.0.11.1/9.1/CMa/1; and if I search for the CM same discriminant but no CM, I get no results. So hopefully that’s fixed now, but please let me know if you find more problems.

> Note that right now we will only ever see actual CM over one of the imaginary quadratic fields (so far 2.0.d.1 for d=1,2,3,7,11,19, 43 will be soon) but it would not be good to assume this in the code since one day we may have more curves over fields properly containing an imaginary quadratic field.

Have I made this assumption somewhere? I don't think I have, but if there's something that needs to change in the code, please let me know.

JohnCremona:
Sorry for the delay, I will test this very soon!

JohnCremona:
Looks good!

PR #4588, opened 2021-06-01
Improvements to storing table descriptions in meta_tables

roed314:
I fixed bugs in create_table_like and create_table, updated how descriptions are stored, and improved some front-end presentation of schemas in the API.

I've tested that `create_table_like` works.  For the front-end changes, check out the schema table in http://localhost:37777/api/ec_curves/ vs https://beta.lmfdb.org/api/ec_curves/

### DIFF
--- a/lmfdb/api/api.py
+++ b/lmfdb/api/api.py
@@ -343,11 +343,14 @@ def api_query(table, id = None):
         description = coll.description()
         if description:
             title += " (%s)" % description
-        schema = [(col, coll.col_type[col], coll.column_description(col)) for col in
-                  sorted(coll.search_cols) + sorted(coll.extra_cols)]
+        search_schema = [(col, coll.col_type[col], coll.column_description(col))
+                         for col in sorted(coll.search_cols)]
+        extra_schema = [(col, coll.col_type[col], coll.column_description(col))
+                        for col in sorted(coll.extra_cols)]
         return render_template("collection.html",
                                title=title,
-                               schema=schema,
+                               search_schema=search_schema,
+                               extra_schema=extra_schema,
                                single_object=single_object,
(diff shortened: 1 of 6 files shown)


edgarcosta:
I will merge this for the sake of the bug fixing.

edgarcosta:
@AndrewVSutherland you might want to check out the new layout of the schema in the API

AndrewVSutherland:
@edgarcosta Nice!  Does anything need to be done before pushing this change to web?  I'm also wondering about how we want metadata to get propagated -- I assume at the moment the meta data changes won't show up on the production site until/unless we copy the whole table over, which is not ideal.

edgarcosta:
I don't think so, but @roed314  has the ultimate answer.

We could try to do it as we do with the knowls, having some script that does it for us on a regular basis.
But now is a bit more odd, as we only want to update a column.

roed314:
There shouldn't be any problems pushing to production.

As for updating metadata, I think a script updating just `table_description` and `col_description` should be reasonable.  Where is the script that's updating the knowls?

edgarcosta:
https://github.com/edgarcosta/lmfdb-gce/blob/master/server_scripts/update_knowls_and_userdb.sh

I don't think we will be able to use pg_dump and pg_restore

roed314:
What about doing essentially the same thing, but creating a temporary table that we then execute
```
UPDATE meta_tables SET (table_description, col_description) = (tmp.table_description, tmp.col_description) FROM tmp WHERE name = tmp.name
```
on?

edgarcosta:
I agree. I will add something like that to the scripts next week

edgarcosta:
We are now updating the descriptions automatically.

PR #4589, opened 2021-06-02
BMFs and ECNFs: minor changes to show data for field 2.0.43.1

JohnCremona:
See #4573.  This adds Q(sqrt(-43)) to the browse lists for BMF and ECNF now that the data for this field has been uploaded.  It also updates a hard-wired list of special BMFs to include 31 new exceptional ones.  (There is an interesting story behind these, which this PR is too narrow to include details of.)

Before this is pushed to production it would be better for the three tables bmf_forms, bmf_dims, ec_nfcurves to be copied over.  (Otherwise users may click on Q(sqrt(-43)) on a brows page and get no forms of curves -- nothing worse than that).

I will deal separately with the issue that, now that the list of BMFs with no curves is longer (49 so far after just adding 31 more), that information needs to be handled differently.  I will do that in due course.

### DIFF
--- a/lmfdb/bianchi_modular_forms/bianchi_modular_form.py
+++ b/lmfdb/bianchi_modular_forms/bianchi_modular_form.py
@@ -58,10 +58,10 @@ def index():
     """
     info = to_dict(request.args, search_array=BMFSearchArray(), stats=BianchiStats())
     if not request.args:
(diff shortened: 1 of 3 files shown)


AndrewVSutherland:
The three tables have been copied over.

PR #4591, opened 2021-06-02
Sorting for g2c

alexjbest:
This PR addresses #3109 for genus  2 curves by providing several sort options.
To keep this list under control I tried to only include things I think people will actually want to sort by, this ended up being:
absolute discriminant, num. rational points (ascending and descending), num. weierstrass points (ascending and descending), torsion order (ascending and descending), analytic order sha (ascending and descending)

Some justification: Discrete invariants like Sato-Tate I figure people likely just have one group they are looking at at a time and dont actually want to sort on. Queries for the largest absolute discriminant in the database seem a bit unnatural / random (its also clearly stated what the max is already), nevertheless examples with large/small Sha or number of rational points may be of interest to people, despite the fact that the maximum isn't obtained in our db for some of these quantities (or that there isn't a maximum at all!).


AndrewVSutherland:
@alexjbest What is the status of this PR, is it ready for review yet?

alexjbest:
@AndrewVSutherland i was waiting till I got the chance to speak to @roed314 about it. Specifically i noticed that for a default search with no options or sort order chosen for g2c that before this patch the total count of number of results is shown, but after this patch it isn't. And I'd like to understand why and be sure I didn't break some sort of caching before proceeding.

roed314:
The missing count is a separate issue and is already visible on https://beta.lmfdb.org/Genus2Curve/Q/?search_type=List&all=1.  I'll make an issue for it, but in the mean time I think this PR is good to go from my perspective.

alexjbest:
Thanks @roed314 ! This is now ready for review then.

roed314:
The labels don't fit in the drop down box (for example, selecting "num. weierstrass points (ascending)" and "num. weierstrass points (descending)" look the same).  But I don't see a great solution, so I'm going to merge it.  If someone has a suggestion for making it better, make a followup PR.

PR #4594, opened 2021-06-03
Fix flask werkzeug

### LINKED ISSUE #4592: Tests broken with flask 2
https://github.com/pallets/werkzeug/issues/2148 seems to cause issues with global random search, the double redirect causes the test https://github.com/LMFDB/lmfdb/blob/master/lmfdb/tests/test_homepage.py#L64 to fail.
Currently CI is working fine on lmfdb as the conda environment is cached and still using flask 1.1.4, but on forks which didn't have an existing cache the latest pypi release >2.0 is installed (https://pypi.org/project/Flask/#history) and the tests fail e.g. https://github.com/alexjbest/lmfdb/runs/2732245524.
For now we can just wait and see if the flask issue gets resolved soon, otherwise we can specify the flask version in the environment.yml or requirements.txt ?
I guess if someone ends up with flask version 2 locally the tests will also fail for them which will be confusing, but I can't think of any workaround, maybe updating requirements is the way to go?.

alexjbest:
This hopefully resolves  #4592 while https://github.com/pallets/werkzeug/issues/2148 is open, and supercedes #4593 which might not be sufficient.

### DIFF
--- a/.environment.yml
+++ b/.environment.yml
@@ -14,4 +14,5 @@ dependencies:
   - python=3.9
   - sage=9.2
   - tox
+  - flask<=1.1.4
   - parallel
--- a/.environment.yml
+++ b/.environment.yml
@@ -15,5 +15,5 @@ dependencies:
   - sage=9.2
   - tox
   - flask<=1.1.4
-  - werkzeug<=1
+  - werkzeug<=1.0.1
   - parallel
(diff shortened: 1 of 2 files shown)


edgarcosta:
looks good to me.

ready to merge once the tests pass

Issue #4537, opened 2021-04-23
Search for number fields with a given completion


roed314:
It would be nice to search for number fields that had a specific local field as one of their completions.

Similarly, on a local field, it would be nice to have a link to this search results page in their related objects.

jwj61:
This looks doable.  The local algebras would have to move from the extras table to the main one.  In doing that, the format used for storing the data can change to facilitate the search.  For a new format, I am thinking of a list of p-adic field labels (so an array of strings).  Then the search can be a test of whether the given label is in the array.  Being an array, some labels may appear more than once to deal with multiplicity, but that's ok.

For a user search, we can start with a search by label; it is the easiest to implement.  Later we can think about allowing polynomials and testing them (does sage have Panayi's algorithm yet?).

JohnCremona:
@jwj61 Woring with elliptic curves was made a lot simpler when I dwitched from the olf ec_curves with separate "extra" table to the collection of individual tables, and recommend that you do the same.  You still need to have in one main table the columns you want to include i the mani browse/search, but when filling the data for one number field's web page it is easy to get other information from other tables.

Right now the onlt tables with extrs are
sage: [t for t in db.tablenames if db[t].extra_table]                      
['ec_curves',
 'ec_temp',
 'maass_newforms',
 'nf_fields',
 'nf_fields_test',
 'test_xyz']
of which ec_curves is obsolete and can be deleted (along with its extra table), ec_temp and I presume nf_fields_test and test_xyz are temporary things, so if you are able to do this for nf_fields then that will only leave maass_newforms.

jwj61:
FYI, this is underway.  I found that I need to recompute local algebra data for some fields, so it will take a while.

jwj61:
First, please upload nf_fields to the cloud.  I think you can just upload the main table and not the extras (if doing a rewrite preserves the links between them).  Then restart prod since this table has a new column.

Data has been computed and coding is done.  I just need to polish up the new knowls before making the pull request.

In doing this, I decided that the local algebras should be improved.  Currently, if a local algebra contains a field not in the local database, we just say "data not computed" for that prime.  That will make some search results incomplete (e.g., if Q_p is one of the completions, but we don't have all of them).

Next steps will be to recompute all of the local algebra data which preserves local factors we have, and provide partial information (e.g., degree, e, f, c) for the ones we don't have in the padic database.

AndrewVSutherland:
[...]

PR #4601, opened 2021-06-07
Search number fields by p-adic completions

jwj61:
This allows users to search for number fields based on p-adic completions (see main number field search page for testing) resolving issue #4537 .  As the knowl there explains, one can enter a single label of a p-adic field, or a list.  In the latter case, the fields in the search results will match all of p-adic fields listed (they can be for different p or some with the same p).

On the page for a p-adic field, if there are any number fields which have it as a completion, there is a friends link to search for them.  Any low degree p-adic field will have some.  An example where there is **no** friends link is

http://127.0.0.1:37777/padicField/19.15.14.2

The search parsing checks that the labels match the right general format, but does not check to see if they are labels of local fields in the database (if it is not in the database, the search results will be empty).

For the knowl for the number field search page, it currently uses

https://beta.lmfdb.org/knowledge/show/nf.place

but before noticing this, I wrote the more detailed

https://beta.lmfdb.org/knowledge/show/nf.padic_completion

We can switch it (which only involves editing a knowl), or delete the new knowl.

### DIFF
--- a/lmfdb/local_fields/main.py
+++ b/lmfdb/local_fields/main.py
@@ -36,7 +36,7 @@ def learnmore_list():
     return [('Source and acknowledgments', url_for(".source")),
             ('Completeness of the data', url_for(".cande")),
             ('Reliability of the data', url_for(".reliability")),
-            ('Local field labels', url_for(".labels_page"))]
+            ('$p$-adic field labels', url_for(".labels_page"))]
 
 # Return the learnmore list with the matchstring entry removed
 def learnmore_list_remove(matchstring):
@@ -321,6 +321,9 @@ def render_field_webpage(args):
             friends.append(('Unramified subfield', unramfriend))
         if rffriend != '':
             friends.append(('Discriminant root field', rffriend))
+        if db.nf_fields.exists({'local_algs': {'$contains': label}}):
+            friends.append(('Number fields with this completion', 
+                url_for('number_fields.number_field_render_webpage')+"?completions={}".format(label) ))
 
         bread = get_bread([(label, ' ')])
         return render_template(
--- a/lmfdb/number_fields/number_field.py
+++ b/lmfdb/number_fields/number_field.py
@@ -1081,6 +1081,12 @@ class NFSearchArray(SearchArray):
             example_span="2.2.5.1 or x^2-5 or a "+
                 display_knowl("nf.nickname", "field nickname"),
             example="x^2-5")
+        completion = TextBox(
+            name="completion",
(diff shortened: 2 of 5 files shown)


JohnCremona:
I think this is going to be very useful.  I suspect that few users will know the labels of p-adic fields so the links will be used more than the search itself (I guess).  I suspect that further feature requests will follow, e.g. on a number field home page, links to its completions (at small primes -- this exists for bad primes in the "local algebras for ramified primes" section, but could be extended) and something like a link to "other fields with the same completion at (some) primes(s).  But that's no reason to delay this.

I can see a case for having both knowls (one for a place and one for a completion), each referencing the others.  On the search page, a knowl for what a completion is seems more appropriate.

Thinking about the Archimdean analogue, of course the search on Signature does wverything but requires a second's thought e.g. to remember that real quadratic is signature [2,0] not [0,2].  Is there a simple way to allow "totally real" and "totally complex" as options for the signature?  -- OK, this is a separate issue!

jwj61:
If the signature box was replaced with a pair for the number of real and number of complex places, one could easily search for totally real and totally complex fields.

JohnCremona:
> If the signature box was replaced with a pair for the number of real and number of complex places, one could easily search for totally real and totally complex fields.

You mean to replace the single Signature box with two?   That sounds ok to me, especially if we could fit them into the same space -- not easy.  This gives greater flexibility thatn what we now have, as you suggest, since you could specify 0 for one of the two and leave the other open.  (Next someone will ask to be able tosearch on r1+r2-1, the unit rank...)

Anyway, that's a separate issue / feature request.  Ready to merge this one?

AndrewVSutherland:
The p-adic field pages on https://www.lmfdb.org are currently broken because this PR was pushed to web earlier by @edgarcosta and it appears there are schema changes to lf_fields and nf_fields that this PR depends on that have not been made to the production database.  It looks like I need to update both of these tables?  Is there anything else that needs to be done?

edgarcosta:
I'm fixing this now

edgarcosta:
Both www.lmfdb.org is now serving 053b73bbb90483333fc6e7053e4811676dfa4bf3.

Let me know when it is okay to reenable the automatic fetching and restart

AndrewVSutherland:
lf_fields has been updated and nf_fields is in progress.  I'll let you know when it finishes.

AndrewVSutherland:
@edgarcosta lf_fields and nf_fields have both been copied over, we should be able to switch www.lmfdb.org back to web.

Issue #4613, opened 2021-06-14
Error in group table for dirichlet characters


roed314:
http://localhost:37777/Character/Dirichlet/grouptable?modulus=2089&char_number_list=1%2C2088&poly=%5C%28+x%5E%7B2%7D+-+x+-+522+%5C%29 for example. 

```
Exception on /Character/Dirichlet/grouptable [GET]
Traceback (most recent call last):
  File "/home/sage/sage-9.2/local/lib/python3.8/site-packages/flask/app.py", line 2447, in wsgi_app
    response = self.full_dispatch_request()
  File "/home/sage/sage-9.2/local/lib/python3.8/site-packages/flask/app.py", line 1952, in full_dispatch_request
    rv = self.handle_user_exception(e)
  File "/home/sage/sage-9.2/local/lib/python3.8/site-packages/flask/app.py", line 1821, in handle_user_exception
    reraise(exc_type, exc_value, tb)
  File "/home/sage/sage-9.2/local/lib/python3.8/site-packages/flask/_compat.py", line 39, in reraise
    raise value
  File "/home/sage/sage-9.2/local/lib/python3.8/site-packages/flask/app.py", line 1950, in full_dispatch_request
    rv = self.dispatch_request()
  File "/home/sage/sage-9.2/local/lib/python3.8/site-packages/flask/app.py", line 1936, in dispatch_request
    return self.view_functions[rule.endpoint](**req.view_args)
  File "/home/lmfdb/lmfdb-git-web/lmfdb/characters/main.py", line 597, in dirichlet_group_table
    info['poly'] = PolynomialRing(QQ, 'x')(info['poly'])
  File "sage/structure/parent.pyx", line 900, in sage.structure.parent.Parent.__call__ (build/cythonized/sage/structure/parent.c:9336)
    return mor._call_(x)
  File "sage/structure/coerce_maps.pyx", line 161, in sage.structure.coerce_maps.DefaultConvertMap_unique._call_ (build/cythonized/sage/structure/coerce_maps.c:4623)
    raise
  File "sage/structure/coerce_maps.pyx", line 156, in sage.structure.coerce_maps.DefaultConvertMap_unique._call_ (build/cythonized/sage/structure/coerce_maps.c:4515)
    return C._element_constructor(x)
  File "/home/sage/sage-9.2/local/lib/python3.8/site-packages/sage/rings/polynomial/polynomial_ring.py", line 452, in _element_constructor_
    return self(p.parse(x))
[...]

PR #4614, opened 2021-06-14
Fix bug 4613

jwj61:
As mentioned in the issue #4613 , check 

http://localhost:37777/Character/Dirichlet/grouptable?modulus=2089&char_number_list=1%2C2088&poly=%5C%28+x%5E%7B2%7D+-+x+-+522+%5C%29

More important is where the link to this page was.  That should be corrected to send the raw polynomial rather than the typeset one.

### DIFF
--- a/lmfdb/characters/main.py
+++ b/lmfdb/characters/main.py
@@ -597,7 +597,7 @@ def dirichlet_group_table(**args):
         try:
             info['poly'] = PolynomialRing(QQ, 'x')(info['poly'])
             info['poly'] = raw_typeset(info['poly'])
-        except:
+        except Exception:
             pass
     return render_template("CharacterGroupTable.html", **info)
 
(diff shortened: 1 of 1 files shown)


lmfdb/characters/main.py
```diff
@@ -594,8 +594,11 @@ def dirichlet_group_table(**args):
     info['contents'] = c
     info['title'] = 'Group of Dirichlet characters'
     if info['poly'] != '???':
-        info['poly'] = PolynomialRing(QQ, 'x')(info['poly'])
-        info['poly'] = raw_typeset(info['poly'])
+        try:
+            info['poly'] = PolynomialRing(QQ, 'x')(info['poly'])
+            info['poly'] = raw_typeset(info['poly'])
+        except:
```
roed314, line 600: It's better to use `except Exception:` so that you don't catch things like `KeyboardInterrupts`.

roed314:
Looks good to me.

PR #4615, opened 2021-06-14
Search completions (part 2)

jwj61:
I decided on a format for the partial information for a completion.  There are only a couple of number fields where this data has been uploaded so most pages will look the same, but

  http://127.0.0.1:37777/NumberField/2.2.1641.1

shows an example of the partial data at a large prime (this page currently throws an error on beta).  The difference is for the local algebra of the large prime.

Fields which don't have the new data (almost every other one of the 21 million number field pages) will display the same.

### DIFF
--- a/lmfdb/number_fields/number_field.py
+++ b/lmfdb/number_fields/number_field.py
@@ -460,21 +460,25 @@ def render_field_webpage(args):
                 loc_alg += '<tr><td>%s<td colspan="7">Data not computed'%str(ram_primes[j]).rstrip('L')
             else:
                 from lmfdb.local_fields.main import show_slope_content
+                primefirstline=True
                 mydat = ramified_algebras_data[j]
                 p = ram_primes[j]
                 loc_alg += '<tr><td rowspan="%d">$%s$</td>'%(len(mydat),str(p))
-                mm = mydat[0]
-                myurl = url_for('local_fields.by_label', label=mm[0])
-                lab = mm[0]
-                if mm[3]*mm[2] == 1:
-                    lab = r'$\Q_{%s}$'%str(p)
-                loc_alg += '<td><a href="%s">%s</a><td>$%s$<td>$%d$<td>$%d$<td>$%d$<td>%s<td>$%s$'%(myurl,lab,mm[1],mm[2],mm[3],mm[4],mm[5],show_slope_content(mm[8],mm[6],mm[7]))
-                for mm in mydat[1:]:
-                    lab = mm[0]
-                    myurl = url_for('local_fields.by_label', label=lab)
-                    if mm[3]*mm[2] == 1:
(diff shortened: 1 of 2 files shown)


roed314:
This looks good to me, but I submitted a PR to your PR for adding closing tags and math mode around the prime when the data is not computed.

jwj61:
I think this incorporates the wanted changes.  It also fixes a bug I noticed along the way when there is more than one completion for a field at a given prime.

An example where p, e, f, c, degree are all different (so we can see that they are in the right places):

http://127.0.0.1:37777/NumberField/6.0.418227202051.1

and one where some primes have multiple lines:

http://127.0.0.1:37777/NumberField/10.0.209352647.1

lmfdb/number_fields/number_field.py
```diff
@@ -457,24 +457,28 @@ def render_field_webpage(args):
         loc_alg = ''
         for j in range(npr):
             if ramified_algebras_data[j] is None:
-                loc_alg += '<tr><td>%s<td colspan="7">Data not computed'%str(ram_primes[j]).rstrip('L')
+                loc_alg += '<tr><td>$%s$<td colspan="7">Data not computed'%str(ram_primes[j]).rstrip('L')
```
roed314, line 460: Add `</td>` and `</tr>` tags here; I'm also not sure why the `rstrip('L')` is here.

lmfdb/number_fields/number_field.py
```diff
@@ -457,24 +457,28 @@ def render_field_webpage(args):
...
+                    if len(mm)==4:
+                        loc_alg += '<td></td><td>Deg {}</td><td>{}</td><td>{}</td><td>{}</td><td></td><td></td>'.format(
+                            mm[1]*mm[2], mm[1], mm[2], mm[3])
+                    else:
+                        lab = mm[0]
+                        myurl = url_for('local_fields.by_label', label=lab)
+                        if mm[3]*mm[2] == 1:
+                            lab = r'$\Q_{%s}$'%str(p)
+                        loc_alg += '<td><a href="%s">%s</a></td><td>$%s$</td><td>$%d$</td><td>$%d$</td><td>$%d$</td><td>%s</td><td>$%s$</td>'%(myurl,lab,mm[1],mm[2],mm[3],mm[4],mm[5],show_slope_content(mm[8],mm[6],mm[7]))
+            loc_alg += '</tr>'
```
roed314, line 481: I would add a `\n` after the `</tr>`, but it's not a big deal.

lmfdb/number_fields/number_field.py
jwj61: The rstrip is there to remove an L if it appears when a large integer is converted to a string by python.

jwj61:
Added more closing tags (even though they are optional in html and just make things more cluttered and harder to read).

PR #4617, opened 2021-06-14
delete extraneous bracket

alexjbest:
tiny copy-paste error for values of dirichlet characters
see https://beta.lmfdb.org/Character/Dirichlet/14/d/3 under values we have `χ14​(3,a))  at a = ` with two right parens


AndrewVSutherland:
Thanks @alexjbest, I'll let tests run and then merge.

PR #4626, opened 2021-06-17
omit unknown base-change labels for ecnf

JohnCremona:
Some of the curves just uploaded over 2.0.163.1 are base-change of a pair of curves over Q which are -163-twists of each other, where one of the pair has conductor too large to be in the EC/Q database, so does not have a complete label and cannot be linked to.  In the ec_nfcurves table these are stored with dummy labels of the form N.?.? where N is the large conductor.

This small change stops these non-labels being shown (with links which obviously do not work).  For example http://localhost:37777/EllipticCurve/2.0.163.1/576.1/a/5

### DIFF
--- a/lmfdb/ecnf/WebEllipticCurve.py
+++ b/lmfdb/ecnf/WebEllipticCurve.py
@@ -651,6 +651,7 @@ class ECNF(object):
             ('CM', self.cm_bool)]
 
         if self.base_change:
+            self.base_change = [lab for lab in self.base_change if '?' not in lab]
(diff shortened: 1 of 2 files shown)


JohnCremona:
Two remarks: (1) for all the base-change ECNFs in the database before this, all curves/Q of which tey are base-change have been in the database, so this issue has not arisen before. (2) At some point we may have base-change curves over NFs none of whose "origin" curves over Q are in the database, and at that point we'll need to show the information differently.  Right now it appear in two places (see above link): in the Base Change section at the bottom of the curve's home page, and in the Related Objects section on the right.

roed314:
looks good.

PR #4631, opened 2021-06-18
Fix BMF scrolling

roed314:
To check, go to http://localhost:37777/ModularForm/GL2/ImaginaryQuadratic/gl2dims/2.0.11.1 and click on the Next button.  Addresses #2854.

### DIFF
--- a/lmfdb/bianchi_modular_forms/bianchi_modular_form.py
+++ b/lmfdb/bianchi_modular_forms/bianchi_modular_form.py
@@ -200,6 +200,7 @@ def bmf_field_dim_table(**args):
     field_label=argsdict['field_label']
     field_label = nf_string_to_label(field_label)
 
+    count = parse_count(argsdict, 50)
     start = parse_start(argsdict)
 
     info={}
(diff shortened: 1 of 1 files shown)


edgarcosta:
LGTM

JohnCremona:
Thanks for fixing this, I failed to find the bug -- and I still cannot see how defining count 10 lines earlier makes any difference?!

roed314:
In most places where previous/back buttons are used (in refine search pages), there's a hidden count field in the html.  That's not being set here, so when you call `parse_start` and it tries to figure out what the count should be (in order to be able to deal with the case that `start` is negative) it fails and just sets start back to the default value of `0`.  If you call `parse_count` before `parse_start`, then it detects that `count` is not in `info` and adds in the default value, at which point `parse_start` succeeds.

Now that the defaults have been uniformized across the LMFDB, it might be better to combine `parse_count` and `parse_start` into one function so that they're always called in the right order.  But I thought for this ticket I would just stick with a minimal change.

JohnCremona:
Thanks for the explanation

PR #4635, opened 2021-06-18
check single-letter variable names

edgarcosta:
This fixes the info banner for example where we don't use the polynomial, e.g.: 
https://beta.lmfdb.org/NumberField/?jump=Qsqrt-11


roed314:
Looks good.

PR #4652, opened 2021-06-27
this enables tests against proddb.

edgarcosta:
I also had to do a minor edit on `lmfdb/utils/config.py` to avoid running `lmfdb/__init__.py` when  importing `lmfdb.backend.config`.

### DIFF
--- a/.github/workflows/generate-matrix.py
+++ b/.github/workflows/generate-matrix.py
@@ -17,9 +17,13 @@ import os
 workflows_dir = os.path.dirname(os.path.abspath(__file__))
 assert workflows_dir.endswith('workflows')
 r = []
+strip = lambda elt: elt if not elt.startswith('lmfdb') else '/'.join(elt.split('/')[1:-1])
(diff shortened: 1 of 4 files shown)


edgarcosta:
@AndrewVSutherland, could you double-check that the name of the secret is LMFDB_CI_ACCESS?

Comparing to my fork, it seems that this PR doesn't see it.

Screen Shot 2021-06-27 at 14 16 33

AndrewVSutherland:
@edgarcosta I can confirm that it LMFDB_CI_ACCESS is a "repository" secret (but not an "environment" or "organization" secret, let me know if I need to change this).

edgarcosta:
@AndrewVSutherland, the best I can do at the moment is the following:
- on pull requests tests against devmirror
- after we merge the PR, it tests against devmirror and proddb

edgarcosta:
I will go ahead and merge this to be sure that I'm not wrong regarding the tests after merge.

PR #4653, opened 2021-06-29
added 4 more BMFs with no elliptic curve

JohnCremona:
I just added BMF and ECNF data for levels/conductors of norm 10k-15k.  This includes 4 more BMFs with no elliptic curve, whose labels are hard-wired in the code.

Compare https://beta.lmfdb.org/ModularForm/GL2/ImaginaryQuadratic/2.0.43.1/10609.1/a/ which says (under Related Objects) "Isogeny class 2.0.43.1-10609.1-a missing"; after this it will say "No elliptic curve" as it should.

Can be merged as soon as tests pass, which they should (though on my github form I always see failing tests, I do not know why).


AndrewVSutherland:
@JohnCremona I presume this means we need to update the bmf and ecnf files on the production database?

JohnCremona:
Yes please, and also I updated the ranges in the two relevant knowls, which I'll mark as reviewed when the data has been copied.

I am only planning one more such data update: -163 is getting close to level norm 4000 and I'll let it run to 5000 which will take a while yet.  And then I'll stop for these four fields.

Do you know why the tests pass here but not on my own fork?  Tests are triggered whenever I push to a branch on my fork, which is fine, or it would be if I could trust the results.

AndrewVSutherland:
The tables have been copied.  I'm not sure why tests are failing on your fork (but this was happening to me as well and I just turned off CI on my local fork).

JohnCremona:
Thanks!

edgarcosta:
You need to merge your branch with upstream/master, to address #4592

PR #4660, opened 2021-07-02
testing against proddb iff one has the secret

edgarcosta:


edgarcosta:
@JohnCremona @AndrewVSutherland this should make tests in forks work again :)

PR #4663, opened 2021-07-05
Local algebra display tweaks

jwj61:
Now that we show partial information about factors of local algebras on number field pages, there are places where we can easily display more information.  This adds that, and displays some numbers in that table in math mode for consistency (before, if we knew the local field, e, f, c were in math mode but not if we don't have the field).

http://127.0.0.1:37777/NumberField/6.2.873625.1
http://beta.lmfdb.org/NumberField/6.2.873625.1

shows a degree 1 factor, now labelled as Q_p, a ramified degree 2 factor with Galois group  and tame/unramified info, and a higher degree field with the current level of information

http://127.0.0.1:37777/NumberField/4.0.6025.1
http://beta.lmfdb.org/NumberField/4.0.6025.1

has two quadratic fields for a big prime, one ramified and one unramified to show that the tame/unramified info is displayed correctly.

### DIFF
--- a/lmfdb/number_fields/number_field.py
+++ b/lmfdb/number_fields/number_field.py
@@ -23,7 +23,7 @@ from lmfdb.utils import (
 from lmfdb.utils.interesting import interesting_knowls
 from lmfdb.galois_groups.transitive_group import (
     cclasses_display_knowl,character_table_display_knowl,
-    group_phrase, galois_group_data,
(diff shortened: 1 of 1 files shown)


roed314:
Looks good.  We could also give the Galois group and slope content in the case that `e=1`.

jwj61:
I thought of that after submitting the pull request.  Maybe I will add that.

PR #4669, opened 2021-07-09
Congruent numbers

JohnCremona:
This adds a page called Congruent numbers and congruent number curves (http://localhost:37777/EllipticCurve/Q/CongruentNumbers) with basic information about congruent numbers and the associated elliptic curves, with a basic ability to download Randall Rathbun's raw data files for all integers up to 1 million.  I also added the facility to search those files for the data one specific numbers, for example 
   - http://localhost:37777/EllipticCurve/Q/CongruentNumber/6 (6 is congruent and the curve is in the database)
   - http://localhost:37777/EllipticCurve/Q/CongruentNumber/10 (10 is not, the curve is in the db)
   - http://localhost:37777/EllipticCurve/Q/CongruentNumber/52014 (a large congruent number, curve not in db, with large generator)

The main page is linked from a new entry in the LearnMore box on the right, just as for the imaginary quadratic class number data.

Comments and suggestions gratefully received.  There are some issues arounf displaying huge rational points (e.g. http://localhost:37777/EllipticCurve/Q/CongruentNumber/894097).


edgarcosta:
This branch is up and running on purple.lmfdb.xyz.

https://purple.lmfdb.xyz/EllipticCurve/Q/CongruentNumbers

JohnCremona:
Thanks @edgarcosta !

JohnCremona:
I should add some citationis for some of the facts I quote about congruent numbers and curves, e.g. Koblitz's book.

AndrewVSutherland:
I like it!  Here are a few UI suggestions (not all of which necessarily should be implemented, I'm just doing a brain dump):

- I might change the caption in the Learn more box to just "Congruent number curves".
- I think the content for "Congruent number", "Congruent number problem", and "Congruent number curve" could would be better placed in knowls.
- I think you could probably condense most of the rest of the page into a single table with each row containing (1) download button (2) name (conductor / rank / gens / height / regulator / root number / analytic sha / class count), with the name linked to a knowl that explains what it is (it could be the standard knowl or one specific to this page if it needs to explain the content, e.g. unknown generators), and (3) just one interesting sample line.  This would make everything that is available visible at a glance.
- I would put all the stats / summaries at the bottom (or maybe in a separate page as we do elsewherre).  I could see adding a few more tidbits (e.g. what does the distribution of analytic order of Sha look like?  In particular, what is the largest that occurs?)

JohnCremona:
Thanks for great constructive feedback- I'll work on implementing some or all soon.

pascalmolin:
This beautiful topic is a nice addition!

I suggest to state the criterion of Tunnell
http://www.digizeitschriften.de/dms/img/?PID=GDZPPN002099403
Theorem 3 there can be LMFDB-ied with a link to the modular form g =
128.d.a.1
and eventually to the weight 3/2 products  g*θ_2 and g*θ_4 giving the
values L(En,1) for squarefree n.

Pascal

Le mer. 14 juil. 2021 à 09:57, John Cremona ***@***.***> a
écrit :

JohnCremona:
Thanks Pascal -- when I have made the text blocks into knowls, perhaps you would like to edit them to include this?  One day of course we may have the half integral weight forms themselved to link to.

JohnCremona:
I have done most of what @AndrewVSutherland suggested.  Three new knowls provide the main blocks of text, hopefully reading OK one after the other on the page while also being self-contained, with their own cross-references.  The downloads are in a much more compact format as suggested, with a few notes on some of them, and some stats at the bottom.

If we put the data into an SQL table we could more easily provide more comprehensive stats but I suggest that can wait.

I moved the form where you can enter an integer and find out whether or not it is congruent to (almost) the top, to be more prominent.  The top short paragraph and/or the bottom Acknowledgements could be moved to separate rcs knowls, or added to the elliptic curve rcs knowls instead, as long as the main supplier of te data has his name prominent.

More suggestions welcome -- including edits to the knowls.  It is better but I'm sure can be improved further.

AndrewVSutherland:
@JohnCremona I notice that the three knowls are still shown in full on the page, is that intentional?  If you removed those people could still see the content by clicking on the relevant knowl, and it would prevent the download buttons from being scrolled off the screen.

Also, I think "Congruent Number Elliptic Curves" should be in lower case and perhaps just "congruent number curves" for consistency?

JohnCremona:
> @JohnCremona I notice that the three knowls are still shown in full on the page, is that intentional? If you removed those people could still see the content by clicking on the relevant knowl, and it would prevent the download buttons from being scrolled off the screen.

It was intentional, but I'll try it out the other way to see how it looks (once legendre is back online).

> 
> Also, I think "Congruent Number Elliptic Curves" should be in lower case and perhaps just "congruent number curves" for consistency?

OK

JohnCremona:
I made the additional changes suggested, most obviously the content of the three knowls is no longer displayed on the page, just available by opening the knowls.

I made some other minor changes, to the Notes column and also deleted the last sentence from the acknowledgements at the bottom.

AndrewVSutherland:
Looks great!

JohnCremona:
This is now on web but broken -- does web have the data files?!

edgarcosta:
yup. Let me look at the code and see what went wrong

edgarcosta:
fixed

Issue #4661, opened 2021-07-02
Update schemas in API


roed314:
The API can now show schemas, removing the need for the inventory.  In preparation for switching over, we should check the quality of the descriptions of the column data.

If you notice an issue, here's how to update the description of a column while logged in as editor (needs a fairly recent version of the LMFDB code).

```
sage: db.av_fq_isog.column_description("abvar_count", "the number of points on this abelian variety over the base field")
```

Alternatively, you can provide a dictionary:

```
sage: db.av_fq_isog.column_description(description={"abvar_count": "the number of points on this abelian variety over the base field", "q": "the cardinality of the base field"})
```

Once this is done, we can change the sidebar to link to the API rather than the inventory.

- [ ] [artin_field_data](https://beta.lmfdb.org/api/artin_field_data/)
- [ ] [artin_old2new_labels](https://beta.lmfdb.org/api/artin_old2new_labels/)
- [ ] [artin_reps](https://beta.lmfdb.org/api/artin_reps/)
- [ ] [av_fq_endalg_data](https://beta.lmfdb.org/api/av_fq_endalg_data/)
- [ ] [av_fq_endalg_factors](https://beta.lmfdb.org/api/av_fq_endalg_factors/)
- [ ] [av_fq_isog](https://beta.lmfdb.org/api/av_fq_isog/)
- [ ] [belyi_galmap_portraits](https://beta.lmfdb.org/api/belyi_galmap_portraits/)
- [ ] [belyi_galmaps](https://beta.lmfdb.org/api/belyi_galmaps/)
- [ ] [belyi_galmaps_prim](https://beta.lmfdb.org/api/belyi_galmaps_prim/)
- [ ] [belyi_passports](https://beta.lmfdb.org/api/belyi_passports/)
- [ ] [bmf_dims](https://beta.lmfdb.org/api/bmf_dims/)
- [ ] [bmf_forms](https://beta.lmfdb.org/api/bmf_forms/)
- [ ] [char_dir_orbits](https://beta.lmfdb.org/api/char_dir_orbits/)
- [ ] [char_dir_values](https://beta.lmfdb.org/api/char_dir_values/)
- [ ] [cluster_pictures](https://beta.lmfdb.org/api/cluster_pictures/)
- [ ] [ec_2adic](https://beta.lmfdb.org/api/ec_2adic/)
- [ ] [ec_classdata](https://beta.lmfdb.org/api/ec_classdata/)
[...]

PR #4670, opened 2021-07-09
eliminate redundant use of convert_IQF_label

JohnCremona:
As discussed on #4661


JohnCremona:
I have made a more drastic commit which gets rid of the function entirely, which I am now checking locally.  Assuming that the unit tests have at least one example where it would have been used, then once that test has passed I can push the new commit and update this PR.

JohnCremona:
I am tracking down the error now in the genus 2 curve friends test...

PR #4677, opened 2021-07-17
ell-adic Galois image data for non-CM E/Q

AndrewVSutherland:
This PR adds information about the ell-adic Galois image for all non-CM elliptic curves over Q and all primes ell.  Previously only 2-adic and mod-ell image data was available, and this PR adds a lot of additional information about the image even in cases where it was already known.  There is no change to the data display for curves with potential CM (I plan to add ell-adic image data for these curves at a later date), except that "p" had been changed to "\ell" in the Galois representations section for consistency (and I think is better in any case  --- every other p on the elliptic curve page refers to a prime of reduction, but here one really wants to think of Frob_p as input to rho_{E,\ell} and writing p instead of ell is potentially confusing)

In addition, this PR fixes a bug that was causing the integral points section to be blank when there are no integral points, rather than saying "None" as it is meant to, and I moved the special L-value into the BSD invariants section where I think it more naturally belongs (this is consistent with what is done on the genus 2 pages).

In the Galois representations section for non-CM E/Q the table now has three columns instead of two: the prime ell, the mod-ell image, and the ell-adic image (which of course determines the mod-ell image but uses a different labeling scheme).  The ell-adic image label is a dynamic knowl that you can open to view the following information:

- generators for the group
- the level ell^e of the group
- the index of the group in GL(2,Z_ell)
- the genus of the group (and of the corresponding modular curve)
- the number of cusps and rational cusps
- whether the group contains -1 or not
- the Cummins & Pauli label of the corresponding congruence subgroup
- for groups of genus 0 or genus 1 and positive rank, the Sutherland & Zywina label
- for 2-adic groups the Rouse & Zureick-Brown label
- the minimal degree extension over which E with this image admit a rational ell^n isogeny for n=1,...,e
- the minimal degree extension over which E with this image have a rational point of order ell^n for n=1,...,e
- the degree of the ell^n-torsion field of E with this image for n=1,...,e
- for groups of positive genus, the labels of the newforms whose associated modular abelian varieteis are isogeny factors of the Jacobian of the corresponding modular curve
- for groups of positive genus, the analytic rank of the Jacobian of the modular curve

Eventually all this data will also be available on a separate home page for the modular curve (which might also include a model for the curve, map to the j-line, known rational points, etc...) which will be linked to in the knowl, but that will come later.

This PR is up and running on https://blue.lmfdb.xyz and can be reviewed there.  Some interesting pages to look at include:

- https://blue.lmfdb.xyz/EllipticCurve/Q/14/a/1 (2-adic image has level 8 and 3-adic image has level 9)
- https://blue.lmfdb.xyz/EllipticCurve/Q/17918/c/2 (2-adic image has level 32 and genus 3)
- https://blue.lmfdb.xyz/EllipticCurve/Q/1225/b/2 (37-adic image has level 37 and genus 4)
- https://blue.lmfdb.xyz/EllipticCurve/Q/1944c1/ (9-adic image has level 27 but mod-3 image is surjective)
- https://blue.lmfdb.xyz/EllipticCurve/Q/162b4/ (nonsurjective 2-adic,3-adic-7-adic but surjective mod 2)
- https://blue.lmfdb.xyz/EllipticCurve/Q/37/a/1 (maximal ell-adic image everywhere)
- https://blue.lmfdb.xyz/EllipticCurve/Q/27a1/ (CM curve with no real changes other than the general ones noted above)

Be sure to click on the ell-adic image knowls in the Galois representation section in each case.  The data used to generate these knowls is stored in the table gps_gl2zhat, which has already been copied to proddb (https://www.lmfdb.org/api/gps_gl2zhat/).  I will be adding more data to this table later, for the moment it just contains the groups that actually arise for come E/Q in the LMFDB.


JohnCremona:
The "maximal image" knowl link goes to the knowl for maximal mod-ell image, so perhaps we need anaother knowl for maximal ell-adic image.

I like the dynamic knowls attached to the ell-adic image labels, though at some point  these will have their own home pages, I suppose, which will be welcome as there are quite a lot of terms in these knowls which need explanation (e.g. what is meant by the genus of an ell-adic image).  I know what that means, but it needs a knowl of its own!  This can wait.

Regarding https://blue.lmfdb.xyz/EllipticCurve/Q/1944c1/ (surjective mod 3 but not 3-adic) it would be nice if that fact was visible on the home page.  Even a bottom knowl perhaps (for the first such example, anyway).  

I'm quite happy with moving the special L-value into the BSD invariant section: I don't know why it was not there already.

The above comments are minor and any changes can be in a follow-up PR, so I'll merge this now.

AndrewVSutherland:
@JohnCremona I added a bottom knowl to 1944c1 with a link to [Elkies paper](https://arxiv.org/abs/math/0612734) which mentions this curve.  Regarding the "maximal image" knowl, my plan is to eventually allow searching on maximal/non-maximal mod ell and ell-adic, with knowls explaining both.  But I want to add the ell-adic images for all the CM elliptic curves first.

JohnCremona:
Excellent -- and isn't it great that when you created that knowl it just appears on the page with no code changes needed!

PR #4687, opened 2021-07-21
congruent-number-curves -> congruent_number_curves

edgarcosta:


JohnCremona:
So the directory name on web is different from than on beta?  Perhaps we should then change the name on beta or put in a symlink so both work.  Or do that on web instead of changing this code?

edgarcosta:
Yeah, when I was showing Drew how to add files we realized that we had two copies, and we kept the wrong version.
Changing filenames on the web is quite hard, it is a read-only filesystem, and would take me quite a lot of time to work around that.

JohnCremona:
Anyway, it now works one web and beta-- thanks.

AndrewVSutherland:
This is still not working for me at https://www.lmfdb.org/EllipticCurve/Q/CongruentNumbers.

JohnCremona:
> This is still not working for me at https://www.lmfdb.org/EllipticCurve/Q/CongruentNumbers.

It's working ok for me!

edgarcosta:
Drew, what isn't working for you?

AndrewVSutherland:
Clearing my browser cache for the last 24 hours fixed it.

Issue #4678, opened 2021-07-18
One curve per isogeny class filter on E/Q search not working in combination with others


AndrewVSutherland:
If I search for E/Q with j-invariant -4096/11 I get 156 results:

 http://www.lmfdb.org/EllipticCurve/Q/?hst=List&jinv=-4096%2F11&search_type=List

which I note all lie in distinct isogeny classes.  But if I set the one curve per isogeny class I get 0 results:

 http://www.lmfdb.org/EllipticCurve/Q/?hst=List&jinv=-4096%2F11&optimal=on&search_type=List

The same thing happens for j=-3375

 http://www.lmfdb.org/EllipticCurve/Q/?hst=List&jinv=-3375&optimal=on&search_type=List

and it also happens if I don't set the j-invariant but instead set the CM discriminant to -7:

  http://www.lmfdb.org/EllipticCurve/Q/?hst=List&jinv=-3375&cm_disc=-7&optimal=on&search_type=List

but I notice that it works as expected for j=0 or 1728:

  http://www.lmfdb.org/EllipticCurve/Q/?hst=List&jinv=0&optimal=on&search_type=List
  http://www.lmfdb.org/EllipticCurve/Q/?hst=List&jinv=1728&optimal=on&search_type=List

JohnCremona:
I'll look into this.  I'm sure it used to be OK...

JohnCremona:
When you select just one curve per class it restricts to curves whose lmfdb_number is 1.  The curves in a non-CM isogeny class all have distinct j-invariants, so if you also specify a j-invariant is's chance whether or not that curve has number 1.  In CM-classes, since the CM is potential, there will be two curves with each j-invariant, but of course only one has number 1.

Because of the above I could say that specifying one curve per class together with specifying some other criterion which is not isogeny-invariant is not something you should ask for (!).  It would be similar to combine a one curve per class search with (say) torsion order, torsion structure, regulator, CM discriminant (not just CM field), cyclic isogeny degree, and more.

Otherwise, we would have to imnplement the one curve per class in a different way:  ignore it for the database query, and then process the search results to pick only one curve with each class label 'lmfdb_iso'.

Is there an efficient way to do that?

Meanwhile we could perhaps divide the search properties into those which are and those which are not isogeny-invariant, and say that the one-curve-per-class option is only enforced when nonne of the latter are specified.

roed314:
[...]

PR #4690, opened 2021-07-22
Search with distinct

roed314:
Add a one_per option to searching, use for isogeny classes of elliptic curves


edgarcosta:
looks good to me, but we should add a test to elliptic curves, e.g., what was listed in #4678

roed314:
I added the test; will merge once tests pass.

roed314:
I fixed the abvar stats test by adding statistics back in locally; I'm not sure why they were missing.

PR #4711, opened 2021-08-19
Fix bug in dropping column description

roed314:
This fixes a bug that @SamSchiavone observed, where the column description gets out of date when a column is dropped.


edgarcosta:
Will merge once tests pass

PR #4721, opened 2021-08-27
another little pep8 code-cleaning pass

fchapoton:
this one about space before and after operators

### DIFF
--- a/lmfdb/bianchi_modular_forms/test_bmf.py
+++ b/lmfdb/bianchi_modular_forms/test_bmf.py
@@ -167,7 +167,7 @@ class BMFTest(LmfdbTest):
         ]:
             page = self.tc.get('/ModularForm/GL2/ImaginaryQuadratic/{}/download/magma'.format(label)).get_data(as_text=True)
             assert expected in page
-            assert  'make_newform'  in page
+            assert 'make_newform' in page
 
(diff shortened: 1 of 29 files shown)


JohnCremona:
Sorry not to have come back earlier on this -- of course I am happy with the sort of changes that @AndrewVSutherland lists.  All I really wanted to check was that if we are running any standrda tests as a necessary condition for being merged, then these are clearly stated somewhere along with whihc subtests are enabled/disabled, so that people can run them locally o a regular basis, and before making a PR.

edgarcosta:
So @fchapoton original change `pycodestyle --select=E703,E271,E272 lmfdb/` would check for:
- E703 | statement ends with a semicolon
- E271 | multiple spaces after keyword
- E272 | multiple spaces before keyword

I'm more than happy with these.
So, are we okay with automatically testing for this as long as we also incorporate them in `test.sh`?

jwj61:
I don't think we should block merging because of things which are completely innocuous.  Most of our contributors, past, present, and future, are mathematicians who write some code, not professional programmers.  I have seen enough people trying and struggling with things as they are.

edgarcosta:
Sounds good. We will keep welcoming such uniformization of the code, but we will not enforce them at the PR time.

PR #4723, opened 2021-08-30
fix #4722

alexjbest:
the issue was clashing class names introduced in #4608

PR #4725, opened 2021-08-30
avoid output classname

edgarcosta:
To prevent another #4723, we should the output class name, as it is used in several pages, but not styled in the css, so one person tweaking their page will unwillingly affect the others


roed314:
Looks good.

PR #4727, opened 2021-08-31
avoiding bare exceptions

edgarcosta:
I have replaced `except:` with `except Exception:`, as we don't want to be catching `KeyboardInterrupt`, `SystemExit`.
I have also updated `tox.ini` to check for these, as in our base code, we don't want these around.

Also updated the `codestyle.sh` with some of the args that @fchapoton already managed to address.

### DIFF
--- a/codestyle.sh
+++ b/codestyle.sh
@@ -7,14 +7,22 @@
 # might introduce a lot of changes. That could be hard to merge!
 
 # increased line length. might get shorter in time ... for now even ignored
-ARGS='--max-line-length=120 --ignore=E501'
+#ARGS='--max-line-length=120 --ignore=E501'
+ARGS='--select=E703,E271,E272,E722'
+
+SAGE_COMMAND=$SAGE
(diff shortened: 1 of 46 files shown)


edgarcosta:
Just to be clear, `codestyle.sh` is a wrapper to pycodestyle and autopep8 written by Harald Schilly, which has not been run in a while.
The previous version checked for [everything](https://pycodestyle.pycqa.org/en/latest/intro.html#error-codes), except long lines, the new version, only checks for:
- E271 	multiple spaces after keyword
- E272 	multiple spaces before keyword
- E703 	statement ends with a semicolon
- E714      test for object identity should be ‘is not’
- E722 	do not use bare except, specify exception instead

Here are the stats of the current warnings that you would get if you ran it with the old version
https://gist.github.com/edgarcosta/799ef013cd9ece0d1c9378d29cee992f
Thus, whenever one of those many warnings gets fixed, one can then add it to the list

PR #4737, opened 2021-09-07
deleting unused code

edgarcosta:


JohnCremona:
I would have asked to keep the file WebHMF.py as I do use that.  It was written by Aurel and me to aid in woking on the HMF databases, as it allows a reasonable interface to the HMF collection's own number field database, complete with labelled lists of primes.  It would still be beeded, for example, in any attempt to find elliptic curves for the remaining dimension 1 HMFs over sextic fields.  It could also be used in a reasonable way in website code for HMFs, but no-one has ever done that.

If possible I would at least like to move the file into the scripts/HMFs directory instead of deleting it altogether.

I would have responded earlier to the PR, but was on holiday.

edgarcosta:
I will do a PR adding that file back.

PR #4738, opened 2021-09-07
Updating force beta

### LINKED ISSUE #4700: Page Not Found redirects to beta
It is possible this is working as intended, but if I go to www.lmfdb.org and try to go to a link that doesn't exist like www.lmfdb.org/junk,  it redirects me to beta to tell me that the page was not found.  This doesn't seem to happen if I type a nonsense urls for the individual sections (like EC or HGCWA).

edgarcosta:
This should resolve #4700

### DIFF
--- a/lmfdb/app.py
+++ b/lmfdb/app.py
@@ -1,4 +1,3 @@
-
 # -*- coding: utf-8 -*-
 from __future__ import absolute_import
 import os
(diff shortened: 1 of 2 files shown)


roed314:
It's probably good to remove the debugging print statements in `NotWhiteListedBreads`.

edgarcosta:
touche!

PR #4743, opened 2021-09-10
Update Lfunctionutilities.py

fchapoton:
is there a reason to use Laurent Polynomials  here ?


edgarcosta:
Not really as we are only inverting invertible power series.

PR #4745, opened 2021-09-10
Update g2LocSolv.py

fchapoton:
use the standard True and False


roed314:
Thanks; there are some very silly things in the LMFDB codebase apparently. :-)

nmascot:
I plead guilty :S

PR #4746, opened 2021-09-10
get rid of some six

fchapoton:
assuming that you are ready to drop python 2.7 support ?

only a partial cure, will need other pull requests


edgarcosta:
Yes, we are ready to drop python 2 support.

PR #4752, opened 2021-09-12
final removal of six imports

fchapoton:
next step will be removal of all  `from __future__`

### DIFF
--- a/lmfdb/api/api.py
+++ b/lmfdb/api/api.py
@@ -1,5 +1,5 @@
 # -*- coding: utf-8 -*-
-from six.moves.urllib_parse import unquote
+from urllib.parse import unquote
 import re
 import yaml
 import json
--- a/lmfdb/app.py
+++ b/lmfdb/app.py
@@ -1,10 +1,8 @@
 # -*- coding: utf-8 -*-
-from __future__ import absolute_import
 from .utils.config import get_secret_key
 import os
 from socket import gethostname
 import time
-import six
 from urllib.parse import urlparse, urlunparse
 
 from flask import (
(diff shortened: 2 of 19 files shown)


fchapoton:
hmm, we need to fix the pylint check. What should be checked by pylint ?

fchapoton:
now at least one of these wrong asserts apparently fails in Sato-Tate. Should I just remove it ?

fchapoton:
probably not a good idea to remove without a mathematical understanding of why it fails

fchapoton:
should it be "No matches" instead of "0 matches" ?

fchapoton:
ok, ready to go !

roed314:
Looks good.  My only question is about the line in `tox.ini`, which now reads
```
pylint --score=no -d C,R,E,W -e W0129 start-lmfdb.py user-manager.py lmfdb/
```
What are `C,R,E,W` and `W0129`?

fchapoton:
The previously used --py3k option was only checking compatibility with python3. Now that it is removed, by default pylint would run its full check suite. And this fails in **many many ways**.

C,R,E,W are four large categories of checks (E for errors, W for warnings, etc). None of these sets of checks currently pass, and by far. So we disactivate them all.

W0129 is checking that there is no wrong assert. I have chosen to re-activate this one as it was simple to fix, in order to keep at least one active check.

One could try to find other specific checks that do pass. I think this could be left to another merge request.

fchapoton:
a reference for W0129 : https://pycodequ.al/docs/pylint-messages/w0129-assert-on-string-literal.html

roed314:
Thanks for the explanation!

Issue #4755, opened 2021-09-16
Wrong counts for elliptic curves over Q


JohnCremona:
On www.lmfdb.org: if you put 1-500000 into the conductor box and ask for a list you see the number of curves is 3064705.  If you now change the "number of curves per isogeny class" from "all" to "one" and search again, the display does not change for a while, and when it eventually does,  the count does not change.  If you add that option to the original search, again you get the same count as without it.

AndrewVSutherland:
There is something very funny going on here.  Comparing

https://www.lmfdb.org/EllipticCurve/Q/?hst=List&conductor=15&optimal=on&search_type=List
https://www.lmfdb.org/EllipticCurve/Q/?hst=List&conductor=11&optimal=on&search_type=List

I see the same bug when restricting to curves of conductor 15, but not when restricting to curves of conductor 11.  Any ideas @roed314 ?

AndrewVSutherland:
FYI, I refreshed stats, but this had no impact (and the problem occurs on both beta and www)

AndrewVSutherland:
this is fixed on beta and will be fixed on production shortly

PR #4776, opened 2021-09-24
Fix bug in one_per search that caused the wrong counts

roed314:


roed314:
Addresses #4755.

AndrewVSutherland:
Looks good, happy to merge once tests pass.

PR #4786, opened 2021-09-29
BMF fix display of non-principal primes and update browse pages for bmf and ecnf

JohnCremona:
I have uploaded BMF and ECNF data for the imaginary quadratic fields of discriminants -23 and -31, the first two fields of class number 3.  This includes dimension data and rational newforms for all levels of norm up to 1000 in each case, and elliptic curves attached to each newform, with complete data for these and isogenous curves.

These are the first Bianchi newforms over fields of class number >1 in the database.  Minor tweaks are needed for the display of prime ideals in tables of Atkin-Lehner and Hecke eigenvalues, since the existing code assumed these were principal (so without these changes only the first generator is shown, see the AL table at https://beta.lmfdb.org/ModularForm/GL2/ImaginaryQuadratic/2.0.31.1/20.3/a/ for an example).

I also updated the browse pages for BMF and ECNF to include these two extra fields, and I updated the relevant rcs knowls, including a citation to Lingham's 2005 thesis (he first worked out how to handle these fields, using methods which apply in principal to all IQFs of odd class number).

Behind this is a major extension of my bianchi-progs C++ code to handle arbitrary class number, after some precomputation for each field (which the code only partially automates at present).  I may now do these precomputations for the 5 remaining fields of odd class number and discriminant > -100 (-59 and -83 have h=3, -47 and -79 have h=5, and -71 has h=7), and I may also extend the code to handle even class numbers, starting with -5 (h=2) as in Bygott's thesis, for which most of the extra coding has already been done.

As well as merging these code updates, please could someone copy bmf_forms, bmf_dims and ec_nfcurves to prod, and also review the 4 knowls I just updated?  Thanks.

### DIFF
--- a/lmfdb/bianchi_modular_forms/bianchi_modular_form.py
+++ b/lmfdb/bianchi_modular_forms/bianchi_modular_form.py
@@ -57,10 +57,10 @@ def index():
     """
     info = to_dict(request.args, search_array=BMFSearchArray(), stats=BianchiStats())
     if not request.args:
-        gl2_fields = ["2.0.{}.1".format(d) for d in [4,8,3,7,11,19,43,67,163]]
+        gl2_fields = ["2.0.{}.1".format(d) for d in [4,8,3,7,11,19,43,67,163, 23,31]]
         sl2_fields = gl2_fields + ["2.0.{}.1".format(d) for d in [20]]
-        gl2_names = [r"\(\Q(\sqrt{-%s})\)" % d for d in [1,2,3,7,11,19,43,67,163]]
-        sl2_names = gl2_names + [r"\(\Q(\sqrt{-%s})\)" % d for d in [5]]
+        gl2_names = [r"\(\Q(\sqrt{-%s})\)" % d for d in [1,2,3,7,11,19,43,67,163, 23,31]]
+        sl2_names = [r"\(\Q(\sqrt{-%s})\)" % d for d in [4,8,3,7,11,19,43,67,163,5]]
         info['gl2_field_list'] = [{'url':url_for("bmf.render_bmf_field_dim_table_gl2", field_label=f), 'name':n} for f,n in zip(gl2_fields,gl2_names)]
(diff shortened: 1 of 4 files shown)


JohnCremona:
If more information is needed, please ask.  This is mainly an addition of extra data, there are minimal code changes.

AndrewVSutherland:
@roed314 @edgarcosta I hit a weird error while trying to copy the tables over 
```
InvalidTableDefinition: multiple primary keys for table "bmf_dims_tmp" are not allowed
 happens while executing b'ALTER TABLE "bmf_dims_tmp" ADD CONSTRAINT "bmf_dims_tmp_pkey" PRIMARY KEY (id)'
```

edgarcosta:
`Successfully reloaded bmf_dims, bmf_forms, ec_nfcurves`

AndrewVSutherland:
Looks good!  The updated knowls have been reviewed.

JohnCremona:
Thanks for copying the data and reviewing the knowls.  I just upload some more data for these two fields, so -23 goes up to 2000 and -31 to 5000 (I do not know why Bianchi modular symbols are slower for -23 than for -31: the relation matrices are about the same sie and sparseness, but reducing them takes longer).

So please can we again copy bmf_forms, bmf_dims and ecnf_curves.  There are also 2 completeness knowls which have had the bound increased.

I am unlikely to go further with these two fields (though -23 has almost reached 2500).  There may be more.  Once there is a code push from master to web these will be visible on the main website, which will be nice, as being the first ever (to my knowledge) tables of elliptic curves (and Bianchi newforms) over any imaginary quadratic fields of class number >1!

JohnCremona:
Thanks, but I am still only seeing 2553 curves in 930 classes for 2.0.23.1, i.e. all the ones which were already there (norm up to 1000) but not the ones I uploaded recently (up to 2000), which I can see on beta. I have refreshed the display.

PR #4795, opened 2021-10-11
Group names in jump box

SamSchiavone:
This allows searching by family name for groups that occur in families, listed in the table `db.gps_families`. For instance, you can now enter things like `C6` or `GL(2,3)` and be taken to the homepage for the group. It also allows values from the `name` column in `db.gps_groups`.


roed314:
Sam said he was good with my changes, so I'm merging.

PR #4798, opened 2021-10-12
Adding a tester to run through all groups pages

roed314:
As suggested by @edgarcosta.


roed314:
This should be fixed by #4800.

PR #4807, opened 2021-10-14
Use new groups in the rest of the LMFDB

roed314:
Switch from `gps_small` to `gps_groups`, change from `small_group_display_knowl` to `abstract_group_display_knowl`, change name of `group_display_knowl` to `transitive_group_display_knowl` and its input from `(n, k)` to `("nTk")`, switch from `aut_grp_id` to `aut_grp_label` etc in `g2c`.


AndrewVSutherland:
I'm happy with this, but I think it would be good for @jwj61 to take a quick look, there are quite a few small changes in Galois groups, local fields, and number fields.

Issue #4806, opened 2021-10-14
Normalize g2c invariants on search


alexjbest:
Currently we have the following behaviour:
- Go on the page of a genus 2 curve
- copy paste the code to make it in magma
- call IgusaClebschInvariants(C) in magma
- Copy the invariants back in to genus 2 curve search page
- "No results"

This seems a bit inconvenient to me (for example it just happened that I wanted to identify some curves i copied from the lmfdb to a magma session a week ago and forgot to note which curves they were, and it was harder than I hoped!)
Slightly ironically this also means that the greyed out example input for searching by Igusa-Clebsch invariants produces no results whereas it should give https://www.lmfdb.org/Genus2Curve/Q/169/a/169/1 .

I'm not sure what the status is for Igusa and G2 invariants as I didnt think about how these are normalized but if thats an issue too hopefully we can also address it.

See also #4067 for discussion on how we normalize these invariants (and where @edgarcosta also suggests we normalize on search).

P.s. I also noticed we inconsistently spell Clebsh (/Clebsch) but I doubt that's worth fixing except as a "good first issue" for someone new who wants to get involved.

PR #4823, opened 2021-10-21
Normalize g2c invariants on search

edgarcosta:
This is live on https://red.lmfdb.xyz/Genus2Curve/Q/

The relevant commit is the first one, a7fad78745ddcfe399836df825fb69753337f24e, the second and third are formatting and cleaning up the tests.

This addresses #4806, but a knowl edit might be necessary for https://beta.lmfdb.org/knowledge/show/g2c.geometric_invariants
to clarify the assumption that the type implies the length.

For consistency, I have also added geometric invariants to the refine page.

### DIFF
--- a/lmfdb/genus2_curves/main.py
+++ b/lmfdb/genus2_curves/main.py
@@ -263,8 +263,8 @@ def by_url_isogeny_class_discriminant(cond, alpha, disc):
             ))])
     if len(request.args) > 0:
         # if conductor or discriminant changed, fall back to a general search
-        if ("cond" in request.args and request.args["cond"] != str(cond)) or
-            ("abs_disc" in request.args and request.args["abs_disc"] != str(disc)):
+        if ("cond" in request.args and request.args["cond"] != str(cond)) or (
+            "abs_disc" in request.args and request.args["abs_disc"] != str(disc)):
             return redirect(url_for(".index", **request.args), 307)
         data["title"] += " Search results"
         data["bread"].append(("Search results", ""))


lmfdb/genus2_curves/main.py
```diff
@@ -290,6 +289,45 @@ def genus2_lookup_equation(f):
...
 
+
+def geom_inv_to_G2(inv):
+    def igusa_clebsch_to_G2(Ilist):
+        # first to Igusa
+        I2, I4, I6, I10 = Ilist
+        J2 = I2/8
+        J4 = (4*J2**2 - I4)/96
+        J6 = (8*J2**3 - 160*J2*J4 - I6)/576
+        J8 = (J2*J6 - J4*2)/4 # we won't use this one at all
```
alexjbest: Doesn't match the knowl, should be J4**2 right? Did you test this code against the db, it should be possible to run through a bunch of test cases there?

alexjbest: Oh lol this is the unused one, so you probably did test it :D

lmfdb/genus2_curves/main.py
```diff
@@ -290,6 +289,45 @@ def genus2_lookup_equation(f):
...
+        J10 = I10/4096
+        return igusa_to_G2([J2, J4, J6, J8, J10])
+
+    def igusa_to_G2(Jlist):
+        monomials = [
```
alexjbest, line 425: This code is pretty unreadable, and I don't see how it is really any better than a direct definition of the G2 invariants in terms of the Ji

roed314:
I was looking at our [knowls](https://www.lmfdb.org/knowledge/show/g2c.geometric_invariants), which I assume is why Edgar was passing through the Igusa invariants when doing Igusa-Clebsch -> G2.

lmfdb/genus2_curves/main.py
alexjbest: Maybe I should have been more clear, the monomials part is ok, but slicing the list based on certain terms being zero because they only contained numerators already checked to be zero makes my head hurt.

alexjbest: Ok looks good now :)

PR #4843, opened 2021-10-29
Fix a bug in copy_from/copy_to

roed314:
The error message complained about extra quotes, and appeared when we updated to a new version of psycopg2.


edgarcosta:
In copy_to, cols_wquotes is only defined if the version is older than 2.9.

PR #4852, opened 2021-11-05
fix typo in CM menu sqrt(-7)

JohnCremona:
In the CM drop-down menu at http://www.lmfdb.org/EllipticCurve/Q/ one option for the CM field is "Q(sqrt(7))".  This PR just inserts the missing minus sign. (As reported by Enrique Gonzalez Jimenez.)

I think I would prefer the other two fields in the list to be shows as Q(sqrt(-1)) instead of Q(i) and Q(sqrt(-3)) instead of Q(zeta_3), for consistency.  (If you really wanted to stress the fact that one of those is a cyclotomic field then make the other one Q(zeta_4)!).  

It would also be nice to explain (perhaps in the knowl) why we list 3 fields and 13 discriminants (the reason of course is that only in these fields is there more than one order of class number 1).

If a reviewer agrees with my suggestion in the second parapraph, I'll update the PR beore it is merged.

### DIFF
--- a/lmfdb/elliptic_curves/elliptic_curve.py
+++ b/lmfdb/elliptic_curves/elliptic_curve.py
@@ -943,7 +943,7 @@ class ECSearchArray(SearchArray):
             example="2,3",
             select_box=nonmax_quant)
         cm_opts = ([('', ''), ('noCM', 'no potential CM'), ('CM', 'potential CM')] +
-                   [('-3,-12,-27', 'CM field Q(zeta_3)'), ('-4,-16', 'CM field Q(i)'), ('-7,-28', 'CM field Q(sqrt(-7))')] +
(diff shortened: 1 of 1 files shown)


edgarcosta:
I agree with your suggestion regarding Q(sqrt(-3)), Q(i) vs Q(sqrt(-1)) is a bit more debatable, for consistency, I think I agree with you.

JohnCremona:
Thanks.  I made the changes to Q(sqrt(-1)) and Q(sqrt(-3)), reordering them to fo -1, -3 -7.  About not using Q(i), it's not just about one's favourite name for these fields, but to see the negative integers explicitly for comparison with the list of discriminants which follows.  It would be nice to be able to see that Q(sqrt(-1)) means discriminants -4,-16 (and similar) in the menu but tht would clutter the menu up rather.

PR #4855, opened 2021-11-05
remove degree 3 link for interesting rational Lfun

### LINKED ISSUE #4847: Some lists of interesting objects are empty
For example on the browse page for p-adic fields there's a link to "some interesting p adic fields" but the user clicking on that link will be disappointed.

Pending someone coming up with any p adic fields interesting enough to be mentioned (by having a top or bottom knowl, I know) perhaps we can only display that option when there are any?

edgarcosta:
resolves #4847


AndrewVSutherland:
If you are going to remove the interest link you may as well also remove the random link, which currently takes you to an empty search results page.

edgarcosta:
I have also removed that link

PR #4857, opened 2021-11-06
Subgroup diagram improvements

jwj61:
This fixes:
  - dragging of subgroups is now reasonably fast for both the regular diagram and the up-to-aut diagram
  - both diagrams have links to their own full page versions
  - the diagram on the group home page will enlarge itself vertically if there are many orders dividing the order of the group.  This affected things like C_256 where the edges were not showing because it was all crammed together before.

Some samples to look at with the 3 variations of the layout

Basic example:
http://127.0.0.1:37777/Groups/Abstract/8.3

Has both diagrams, but CC one is wide, so the subgroup info is put below it
http://127.0.0.1:37777/Groups/Abstract/16.14

Too many subgroups for CC diagram, but there is an aut diagram with many levels:
http://127.0.0.1:37777/Groups/Abstract/256.56092

### DIFF
--- a/lmfdb/groups/abstract/main.py
+++ b/lmfdb/groups/abstract/main.py
@@ -709,12 +709,16 @@ def sub_diagram(label):
     if gp.is_null():
         flash_error("No group with label %s was found in the database.", label)
         return redirect(url_for(".index"))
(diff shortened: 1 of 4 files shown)


roed314:
Looks good to me.  I'll make another PR to merge to beta.

PR #4865, opened 2021-11-12
Various fixes and improvements to accompany new data

roed314:
Major changes:
* Support the lack of character tables for some groups
* Change subgroup lattice layouts to use data generated by graphviz, and to sort by number of primes dividing the order rather than the order itself.  Also note that we no longer center the whole group and the trivial group.
* Adds latex macros to `base.html` to support things like `\SD` and `\OD`.


AndrewVSutherland:
pyflakes isn't happy

roed314:
Tests pass now.

AndrewVSutherland:
I'm not sure if this was caused by this PR, but there is group data getting dumped to the console every time an abstract group home page is rendered, I think there is a print statement somewhere that should be removed.

roed314:
Thanks for catching that!

AndrewVSutherland:
test_subgroup_perfect_search failed

roed314:
Tests pass now.

jwj61:
This included a change which was pointed out on zulip.  I objected there, but it was merged without getting feedback from a broader collection of people.

AndrewVSutherland:
Apologies, I missed that.  I can revert, but maybe simpler to hold off pushing to beta until the issue is resolved.  This PR does fix a couple of unrelated problems that would be good to fix soon (e.g. latex not rendering in groups search results).

I guess the question is whether to level subgroups by cardinality, e.g. as on

    https://beta.lmfdb.org/Groups/Abstract/36.4

of by the number of prime divisors of the cardinality (counted with multiplicity), e.g. as on

    https://groups.lmfdb.xyz/Groups/Abstract/36.4
    https://people.maths.bris.ac.uk/~matyd/GroupNames/1/D18.html

Ordering by number of prime divisors seems like a pretty common choice.  In addition to GroupNames there is also  https://demonstrations.wolfram.com/SubgroupLatticesOfFiniteCyclicGroups/, and a Google image search for "subgroup lattice" seems to mostly show images that use the prime divisor convention in cases where there is a difference.

What do other people think?  Does it make more sense to layer subgroup lattices by number of prime divisors or by cardinality?

jwj61:
That is not the defining condition on GroupNames (e.g., see A_4).

To rephrase my position here, what is the more natural invariant of a subgroup, its order, or the number of prime factors in its order.  If it is the former, maybe we should be listing that instead of the order of the subgroup on subgroup pages.

roed314:
I do think we should make the change implemented on this PR, but I didn't intend to try to short circuit discussion.  I think Zulip is probably a more natural place to continue it than a closed PR.

I'll outline my reasoning on Zulip, but may not be able to do so tonight or tomorrow.

PR #4875, opened 2021-11-19
fix display short

### LINKED ISSUE #4874: P-adic Galois group in gaga box incorrectly formatted
I haven't looked into this at all, but I see on pages like

https://www.lmfdb.org/padicField/2.12.30.284

that the gaga box "Galois group" property misrenders `galois_group[tex_name]`.

edgarcosta:
This should resolve #4874, but please check it.

my sage is not working at the moment


davidlowryduda:
This seems right to me.

Issue #4867, opened 2021-11-16
Consistency in indicating cardinality


AndrewVSutherland:
In some places we use #S to denote the cardinality of S (e.g. on the abstract groups pages), and in other places we use |S| (e.g. on number field and p-adic field pages).  We  should be consistent.

AndrewVSutherland:
@roed314 I think that makes sense, as far as I can tell we use |G| everywhere else (Galois groups, local fields, and number fields).  If we decide want to switch to #G we should update the style guide and then make the change everywhere.

davidfarmer:
Can we use `\card{G}` everywhere so that the source is semantic and the macro definition
encodes the choice you are making?

JohnCremona:
On Wed, 17 Nov 2021, 12:35 David W. Farmer, ***@***.***>
wrote:

> Can we use \card{G} everywhere so that the source is semantic and the
> macro definition
> encodes the choice you are making?
>

That exactly what I would suggest though it would be more work to change
all the old occurrences.

PR #4876, opened 2021-11-19
Use \card{...} instead of |...| or #...

davidlowryduda:
I think I found them all.

For Issue #4867


AndrewVSutherland:
I think you missed a few?  A git grep -F "$|" and git grep -F "$\#" finds matches in 
 
lmfdb/galois_groups/templates/gg-show-group.html (uses | |)
lmfdb/sato_tate_groups/templates/st_results.html (uses #, see http://127.0.0.1:37777/SatoTateGroup/?search_type=List&all=1)

PR #4878, opened 2021-11-20
Groups

roed314:
This PR adds pages that are dynamically generated for orders not in the database (e.g. [512](https://groups.lmfdb.xyz/Groups/Abstract/512.12345), [1536](https://groups.lmfdb.xyz/Groups/Abstract/1536.123514), [768](https://groups.lmfdb.xyz/Groups/Abstract/768.23514)), and large [abelian groups](https://groups.lmfdb.xyz/Groups/Abstract/ab/2.3.4.5.6.7.8.9.10.11.12).

I still need to add some tests, but I was already getting strange test failures on the groups branch and wanted to see if they persisted with the PR (and to be able to look at the diff).

### DIFF
--- a/.environment.yml
+++ b/.environment.yml
@@ -12,7 +12,7 @@ channels:
   - conda-forge
 dependencies:
   - python=3.9
-  - sage=9.2
+  - sage=9.3
   - tox
   - flask<=1.1.4
   - werkzeug<=1.0.1
--- a/.environment.yml
+++ b/.environment.yml
@@ -12,7 +12,7 @@ channels:
   - conda-forge
 dependencies:
   - python=3.9
-  - sage=9.3
+  - sage=9.4
   - tox
   - flask<=1.1.4
   - werkzeug<=1.0.1
--- a/GettingStarted.md
+++ b/GettingStarted.md
@@ -52,10 +52,10 @@ Installation
      ```
      sage -pip install -r requirements.txt --upgrade
      ```
-   - For versions of macOSX after 10.15 (Catalina), Sage 9.2 is unable to insall gap_packages 
+   - For versions of macOSX after 10.15 (Catalina), Sage 9.2 is unable to insall gap_packages
      (see [this](https://ask.sagemath.org/question/54252/sage92-install-gap_packages-on-macos-1015-fails/) post for some documentation on
-     that). This problem does seem to be resolved for Sage 9.3beta, and it is currently 
-     unclear if the problem exists for Sage 9.0 or 9.1. 
+     that). This problem does seem to be resolved for Sage 9.3beta, and it is currently
+     unclear if the problem exists for Sage 9.0 or 9.1.
 
    - In case the last step fails by is Mac OSX with the error
 
--- a/lmfdb/app.py
+++ b/lmfdb/app.py
@@ -757,7 +757,7 @@ def WhiteListedRoutes():
         'Field',
         'GaloisGroup',
         'Genus2Curve/Q',
-        'Group',
+        'Group/foo', # allows /Group but not /Groups/*
         'HigherGenus/C/Aut',
         'L/Completeness',
         'L/CuspForms',
--- a/lmfdb/artin_representations/math_classes.py
+++ b/lmfdb/artin_representations/math_classes.py
@@ -218,7 +218,7 @@ class ArtinRepresentation(object):
             label = f"{gapid[0]}.{gapid[1]}"
             name = db.gps_groups.lookup(label, "tex_name")
             if name:
-                return abstract_group_display_knowl(label, name)
+                return abstract_group_display_knowl(label, f"${name}$")
         ntj = self._data['Proj_nTj']
         if ntj[1]:
             return transitive_group_display_knowl(f"{ntj[0]}T{ntj[1]}")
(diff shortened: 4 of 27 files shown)


edgarcosta:
I tracked it down to the choice of generators for ideals.
I will fix I try to fix the tests on a different PR.

roed314:
Thanks Edgar.  I'm going to close this for a day or two while I add some tests and fix a few problems (latex for metabelian groups, add maximal subgroups).

roed314:
Ok, latex is better and there are now maximal subgroups.  Some pages are a bit slow (e.g. [1536.1231247](https://groups.lmfdb.xyz/Groups/Abstract/1536.1231247)), so I'm curious if people think it's too much.  The parts that can be slow are computing maximal subgroups (and latex for them) and computing automorphism groups.

roed314:
Here are some other sample pages to look at:
* [ab/2.3.4.5.6.7.8.9.10.11.12](https://groups.lmfdb.xyz/Groups/Abstract/ab/2.3.4.5.6.7.8.9.10.11.12) - an abelian group with interesting Smith normal form
* [ab/2_50](https://groups.lmfdb.xyz/Groups/Abstract/ab/2_50) - a big elementary abelian 2-group
* [1920.240463](https://groups.lmfdb.xyz/Groups/Abstract/1920.240463) - a nonsolvable example
* [ab/3000](https://groups.lmfdb.xyz/Groups/Abstract/ab/3000) - a large cyclic group showing that we compute the structure of the automorphism group in this case.

roed314:
An example of a group that times out (of order `11^7`): [19487171.700000](https://groups.lmfdb.xyz/Groups/Abstract/19487171.700000).  I think the solution eventually should be to add some capability for timeouts, but neither of the two ways of using [timeout_decorator](https://pypi.org/project/timeout-decorator/) seem to work for us: we can't use the default threading implementation because of flask, and the backup multiprocessing implementation fails because there are things that can't be pickled.  Since we're just using this internally, we could maybe pass function source code to be executed (bypassing the need for pickling things like GAP groups), but I'd want to be careful about not introducing security problems.

roed314:
I also found and fixed a bug in the statistics for irreducible characters: currently the number of irreducible characters of [1800.328](https://beta.lmfdb.org/Groups/Abstract/1800.328) doesn't match the number of conjugacy classes.

AndrewVSutherland:
I think this is looking pretty good.  One question I have is for some of the larger groups (e.g. of order 3^7 or 5^5, 5^6) there are not that many groups and I think it would be easy to just include them and all their subgroups, is there a reason not to do that?

roed314:
The costs of including them are:
* a more complicated completeness knowl.  Currently we include all groups of orders up to `2000` except those larger than `500` divisible by `128`.
* computation time for these groups (not a big deal at all) and space for keeping the data (this is more annoying, since larger tables can slow searches down to some extent).

Here are some statistics (just in terms of number of groups, since counting subgroups and characters is more difficult).
* There are currently 257,936 abstract groups in the database.
* There are 1,396,077 groups of order `3^8=6561`, so we shouldn't include them all.
* There are 9,310 groups of order `3^7=2187`; 34,297 groups of order `5^7=78125`; 113,147 groups of order `7^7=823543`; 750,735 groups of order `11^7=19487171`.
* There are 684 groups of order `5^6=15625`; 860 groups of order `7^6=117649`;...; 32,472 groups of order `97^6=832972004929`.
* There are 77 groups of order `5^5=3125`; 83 groups of order `7^5=16807`;...; 265 groups of order `97^5=8587340257`.
* There are 15 groups of order `7^4=2401`; 15 groups of order `11^4=14641`;...; 15 groups of order `97^4=88529281`.
* There are 5 groups of order `13^3=2197`; 5 groups of order `17^3=4913`;...; 5 groups of order `97^3=912673`.
* There are 2 groups of order `47^2=2209`; 2 groups of order `53^2=2809`;...; 2 groups of order `97^2=9409`.
* I certainly think we don't want to include groups of prime order above `2000`.
* For groups with order a product of three distinct primes, there are 622 between `2000` and `3000`; 4,331 between `3000` and `10000`; 55,301 between `10000` and `100000`.
* For groups with order a product of two distinct primes, there are 370 between `2000` and `3000`; 2,460 between `3000` and `10000`; 27,808 between `10000` and `100000`.
* For groups of the form `2^k q` with `2<=k<=8`, there are 6,953 groups between `2000` and `3000` if you exclude the top four (19,324 each of orders `2^7*19` and `2^7*23`; 21,808 of order `2^7*17`; 1,083,472 of order `2^8*11`)
* For groups of the form `3^k q` with `2<=k<=6`, there are 327 groups between `2000` and `3000`; 12,052 between `3000` and `10000`; 84,627 between `10000` and `100000`.
* For groups of the form `5^k q` with `2<=k<=5`, there are 34 groups between `2000` and `3000`; 701 between `3000` and `10000`; 4,329 between `10000` and `100000`.
* For groups of the form `7^k q` with `2<=k<=4`, there are 14 groups between `2000` and `3000`; 232 between `3000` and `10000`; 1,172 between `10000` and `100000`.

I think the current description of which orders are complete is nice and clean, and I don't think it's worth the complexity to add just a few more.  I'd rather focus on extending our reach in other ways (adding all permutation groups up to certain bounds, adding matrix groups).  But I'm open to a concrete proposal: are you suggesting just `3^7`, `5^6` and `5^5`?  Or more than that?

AndrewVSutherland:
@roed314 Thanks for the explanation, and I agree there are other higher priorities. I'll respond in more detail on the Zulip.

Issue #4887, opened 2021-12-01
Euler factors of Dirichlet L-functions are listed as F_p(-T)


edgarcosta:
Go to an L-function https://www.lmfdb.org/L/?origin=DIR that is not the Riemann zeta function
and check that `F_p(T) = 1 + \chi(p) T` instead of `1 - \chi(p) T`.

If you want a concrete example:

```
_<T> := PolynomialRing(Integers());
G := DirichletGroup(5);
chi := G.1;
print chi;
L := LSeries(chi);
print LGetCoefficients(L, 10);
for p in PrimesUpTo(100) do
    print p,":", EulerFactor(L, p);
end for;

Kronecker character 5
[* 1, -1, -1, 1, 0, 1, -1, -1, 1, 0 *]
2 : T + 1
3 : T + 1
5 : 1
7 : T + 1
```
compare with 
https://www.lmfdb.org/L/1/5/5.4/r0/0/0
and 
https://beta.lmfdb.org/api/lfunc_lfunctions/?label=1-5-5.4-r0-0-0&_fields=label,euler_factors

edgarcosta:
Here is a screenshot before I push my fix all the way to web:
<img width="709" alt="Screen Shot 2021-12-03 at 12 45 50" src="https://user-images.githubusercontent.com/322822/144604824-6e7c216e-f4d4-4d22-bfad-fb440c3447ba.png">

PR #4888, opened 2021-12-01
hot fix for #4887

edgarcosta:


edgarcosta:
I'm merging this one and pushing it all the way to the web.
We can discuss a better fix later on, but this will prevent me from presenting wrong data later today.

edgarcosta:
I have not checked all of them, but checked plenty with:
https://beta.lmfdb.org/L/?hst=List&origin=DIR&search_type=Random
In all the examples is easy to spot that the first coefficient in the Dirichlet series always has the same sign and the linear term of the Euler factor at the first good prime.

PR #4891, opened 2021-12-04
adding some assurance on the convert_euler_Lpoly for Dirichlet L-fcns

edgarcosta:
This resolves completely #4887, as now we check if the fix needs to be applied.

### DIFF
--- a/lmfdb/lfunctions/Lfunction.py
+++ b/lmfdb/lfunctions/Lfunction.py
@@ -400,16 +400,23 @@ def apply_coeff_info(L, coeff_info):
         return res[0], CDF(res[1])
 
     base_power_int = int(coeff_info[0][2:-3])
+    fix = False
     for n, an in enumerate(L.dirichlet_coefficients_arithmetic):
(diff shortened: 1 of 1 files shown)


lmfdb/lfunctions/Lfunction.py
```diff
@@ -400,16 +400,23 @@ def convert_coefficient(an, base_power_int):
         return res[0], CDF(res[1])
 
     base_power_int = int(coeff_info[0][2:-3])
+    fix = False
     for n, an in enumerate(L.dirichlet_coefficients_arithmetic):
         L.dirichlet_coefficients_arithmetic[n] , L.dirichlet_coefficients[n] =  convert_coefficient(an, base_power_int)
-
+        # checks if we need to fix the Euler factors
+        if is_prime(n) and L.dirichlet_coefficients_arithmetic[n] != 0:
+            if fix:
```
roed314, line 408: I'm confused: this branch of the if statement won't ever be triggered....

edgarcosta: It will be triggered when the else changes the value of the fix variable

roed314: Ok, I was being stupid.

PR #4894, opened 2021-12-08
Add label to dirchar values tables

davidlowryduda:
This is a small change that adds a first column to the values table
shown on Dirichlet character pages. This came out of a conversation
where we went to the LMFDB to see a Dirichlet character and its
L-function, and we went to

https://www.lmfdb.org/Character/Dirichlet/4/3

The values table makes much less obvious sense when there are only two
values, and it was confusing. This adds a first column containing
labels.


roed314:
Looks good.

PR #4908, opened 2021-12-17
Belyi primitivization  search

nb4296:
Added search feature and test file for primitivization


lmfdb/number_fields/web_number_field.py
```diff
@@ -201,7 +201,8 @@ def string2list(s):
 def is_fundamental_discriminant(d):
     if d in [0, 1]:
         return False
-    if d.is_squarefree():
+    #if d.is_squarefree():
```
edgarcosta, line 204: ???

PR #4917, opened 2022-01-18
Fix shadows/jumping on initial display of the subgroup diagram.

jwj61:
Sometimes the initial display of the subgroup diagram looks like there are shadows (or double images) of some of the subgroups.  When you move the mouse over it or click, it fixes itself (sometimes involving subgroups moving a little).  This fixes that problem.  

Detailed explaination (if you care):  I think it was caused by us having the code draw a diagram, and then maybe redrawing with the correct one.  But, images used for subgroups fetched asynchronously and I think the redisplay call may happen before all of the first ones get drawn leading to double versions being drawn on the canvas.  This eliminates the initial draw so that the diagram is just drawn once to start with.

To see the problem, compare

http://beta.lmfdb.org/Groups/Abstract/16.5

http://127.0.0.1:37777/Groups/Abstract/16.5

### DIFF
--- a/lmfdb/groups/abstract/templates/abstract-show-group.html
+++ b/lmfdb/groups/abstract/templates/abstract-show-group.html
@@ -440,6 +440,7 @@ Order {{ordline[0]}}: {{ordline[1] | safe}} <br />
   <script>
     show_info("diagram");
     sdiagram.setSize();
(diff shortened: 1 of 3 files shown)


AndrewVSutherland:
Looks great, thank you!

Issue #4921, opened 2022-01-19
Column selector in search results


AndrewVSutherland:
Now that we have a standard search results template that is used throughout the LMFDB we are in a position to implement column selection generically.  A crude first draft of this can be seen at https://purple.lmfdb.xyz/EllipticCurve/Q/?search_type=List&all=1 (you can also see it in action on any of the other search results pages but we still need to add more column options elsewhere).

roed314:
With #4927 merged, let's use this issue to track requests for columns to be added to the various search results tables across the LMFDB.  As a first approximation, anything that you can enter in the search boxes is a good candidate for adding as a column.  If you request a column, make sure to specify whether you think it should be shown by default (it's possible for this to depend on the results: for example, the projective image of classical modular forms is shown by default only when there's some weight 1 form listed).

roed314:
In #3388, there were requests for additional columns on elliptic curves over Q.  Discriminant (which is now an optional column, but should maybe be shown factored and we should perhaps take the absolute value) and the curve equation (rather than a list of Weierstrass coefficients).

roed314:
For hgcwa, it's easy to add Quotient genus and three check columns for the three select boxes.

rbommel:
Maybe I will try adding some columns to number fields now.

rbommel:
Done, see #4951.

AndrewVSutherland:
I'm working on elliptic curves (over Q and number fields).

PR #4927, opened 2022-01-20
Column modification

roed314:
This PR adds the ability to control which columns are shown in search results and adds a few new optional columns (along with a few new default ones in elliptic curves over number fields).  It also fixes a few bugs introduced in the last PR that introduced `SearchColumns`.

Demo available at https://purple.lmfdb.xyz.


JohnCremona:
This is amazing and will be very popular and useful!  I have already had some fun playing with the EC/Q and EC/NF pages.

If this went live I think we would immediately get rquests to add columns to the list of options. (For example, I listed curves of rank 3 but wanted to see what their (analytic) Shas were, but that was not available.)  So should be come up with a way to collect suggestions for what the optional columns should be for each type of object?  Limited to columns in the relevant table, I assume.

AndrewVSutherland:
This PR is the first step in addressing #4921.  Perhaps that issue would be a good place to post suggestions for columns to add.  Note that you can display columns that are not a 1-to-1 match to columns in the relevant tables using the `ProcessedCol` and `MultiProcessedCol` methods ([click here](https://github.com/LMFDB/lmfdb/blob/9fe86042734ac23289a2cccadaefd76ddf168097/lmfdb/elliptic_curves/elliptic_curve.py#L320) to see some examples for elliptic curves over Q), anything that can be derived from the columns in the table is fair game (but we don't want to trigger any new database lookups inside these methods).

This is something I hope people will be motivated to work on during the coding sprint tomorrow and Saturday -- with the infrastructure @roed314 has put in place, it is very easy to add new column options to the selector (it is literally just one or two lines of code).

AndrewVSutherland:
Regarding this PR, I think the UI needs a minor tweak.  Having the column selector push the table down (and change the location of the search again and random buttons) when it opens is jarring.  I think it would be better if it overlaid, as the other selection buttons do (to see what I mean, compare what happens when you click on the torsion selector vs the column selector on https://purple.lmfdb.xyz/EllipticCurve/Q/?search_type=List&all=1).

JohnCremona:
Thanks for explaining about processed columns, that is excellent.

I agree about the column selector box overlaying as being better.  It's very good that the effect of de/selecting columns is instantaneous!

It was not immediately obvious to me at first that the word "done" at the bottom of the selector box was something you needed to click on when finished to make the box go away.  At least the explanatory knowl could mention this, but I am wondering if this can be implemented or shown differently.  Perhaps having "[click here when done]" across the bottom would be clearer?

AndrewVSutherland:
@JohnCremona You don't actually "need" to click done, it's provided as a way to close the selector, but you can also close it by  clicking the "columns to display" option you clicked to open, so maybe it isn't needed?  Or perhaps"close" would be better than "done"?

I think my preference would be to remove the "done" line entirely (we actually discussed doing this before David submitted the PR but decided to leave it in so that others could give feedback), but let me know what you think.

JohnCremona:
> @JohnCremona You don't actually "need" to click done, it's provided as a way to close the selector, but you can also close it by clicking the "columns to display" option you clicked to open, so maybe it isn't needed? Or perhaps"close" would be better than "done"?
> 
> I think my preference would be to remove the "done" line entirely (we actually discussed doing this before David submitted the PR but decided to leave it in so that others could give feedback), but let me know what you think.

OK, I see now that clicking on "columns to display" also closes the pop-up, but I would not have guessed that.  One thing I tried was to click on the "Select" which opens up the explanatory knowl.

It seems to me that if I have changed the optional columns at all then clicking outside the box closes the opo-up, but not when I have made no changes.  Is that what you see -- if so, can we make it so that clicking outside thet pop-up box always closes the box?  Then I certainly agree that the "Done" (or "Close") lines are not needed.

AndrewVSutherland:
@JohnCremona Ah, I see what you mean. I'll see if there is a way to fix that (and have the drop down overlay rather than push content down).  I agree that if we can get it to always close when you click anywhere outside the selection menu then the "Done" option is superfluous.

roed314:
@edgarcosta was taking a look at how to not make the select box push content down.  I agree that the current behavior isn't great, but I'm not sure how to change it.  Allowing a click outside to close the select together with removing done sounds like a good plan as well.

Finally, I'll note that it is possible to use another database lookup to get more information; this is done to look up group names in several places (e.g. [local fields postprocess](https://github.com/roed314/lmfdb/blob/column_modification/lmfdb/local_fields/main.py#L218) uses searches of [gps_transitive](https://github.com/roed314/lmfdb/blob/column_modification/lmfdb/galois_groups/transitive_group.py#L34) and [gps_groups](https://github.com/roed314/lmfdb/blob/9fe86042734ac23289a2cccadaefd76ddf168097/lmfdb/groups/abstract/main.py#L1762)).  This isn't ideal (at the least we should probably update `gps_transitive` with the better latex names from `gps_groups`), but at least we're not doing one database lookup per row of search results (which was done in several places before the last PR: artin representations, galois groups, number fields and local fields).

AndrewVSutherland:
@roed314 I believe I just fixed the click outside closing and removing done, can you push my latest commit to purple so people can see it?

roed314:
Done.  It should update in a few minutes.

I'm also taking a look at the test failures.

roed314:
The test failure looks spurious (missing secret key information?).  Hopefully it will go away on its own in a future commit.

edgarcosta:
I restarted the tests. That error happens once in a blue moon at GitHub actions, but I never saw it happening in real life.

edgarcosta:
I have pushed to purple

AndrewVSutherland:
I made one more fix to make the last column in the selector clickable now that "done" is gone (so maybe push to purple again).  Once tests pass I'm happy to merge if everyone else is.

roed314:
I just pushed to purple; should update in a couple minutes.

roed314:
For addressing the shifting, I assume that it's from this line:
```
onmousedown="$(this).attr('size',this.length); this.value='';"
```
Drew, where did you find the suggestion for using this to prevent the select from closing when you choose an option?

Edgar, did you have an idea for how to address this or should we merge?

edgarcosta:
I have fixed the movement of the lower blocks. Check it out.

roed314:
Awesome!  I'll merge once tests pass.

PR #4935, opened 2022-01-21
Removing has_magma in main.py and "Genus 2 curve -> Isogeny class" under friends

edgarcosta:


AndrewVSutherland:
jump prompt knowl needs to be updated, I'll merge this and do that

Issue #4564, opened 2021-05-08
Precision of special L-values of elliptic curves/Q


bmatschke:
Some special L-values of elliptic curves/Q seem to have several false digits at the end, e.g. https://www.lmfdb.org/EllipticCurve/Q/37a1 currently lists
0.30599977383405230182048368332169132525,
whereas sage's E.lseries().taylor_series() using large precision gives:
0.305999773834052301820483683321676474452637774590771998534541832481016050469290169911495257337795897...
I don't know how numerically stable the latter is, I simply increased precision until the first 100 digits seem to stabilize.

edgarcosta:
from https://beta.lmfdb.org/api/ec_mwbsd/?lmfdb_label=37.a1 we see that the 
the value stored is 
`0.30599977383405230182048368332169132525`
while the correct value is
`0.305999773834052301820483683321676474452637774590772?`

JohnCremona:
This must be my fault since I recently recomputed all these values using libpari from Sage like this:
`ar, sv = E.pari_curve().ellanalyticrank(precision=PRECISION).sage()`
with PRECISION=100, for example
`sage: E = EllipticCurve('37a1') `                                                                                                                      
`sage: E.pari_curve().ellanalyticrank(precision=100).sage() `                                                                                           
`[1, 0.30599977383405230182048368332169132525]`
while in gp itself we see
`? ellanalyticrank(ellinit("37a1"))`
`%4 = [1, 0.30599977383405230182048368332167647445]`
So something is going wrong in converting the real number from libpari to Sage.

Someone will need to work out why this happens;  submit a bugfix to Sage if it is a Sage issue; and in any case recompute all these values and re-upload them.  That person will not be me any time soon!

bmatschke:
Thanks, I didn't know about this bug. Similarly:
    
    sage: gp.set_real_precision(100)
    38
    sage: a = gp.sqrt(2)
    sage: a
    1.414213562373095048801688724209698078569671875376948073176679737990732478462107038850387534327641573
    sage: a.sage()
    1.414213562373095048801688724209698078569671875376948073176679737990732478462107038850387534327641573000000000000000

JohnCremona:
Sone of the reason for what you observe is that pari bit precision goes up in jumps of 32 or 64 while sage allows any bit precision.  This is supposed to be handled properly in the interface.

bmatschke:
Then maybe it's not even a bug. There are two opposite goals in such a conversion: 

[...]

PR #4936, opened 2022-01-21
do not display more digits than are correct in EC/Q

JohnCremona:
This fixes issue #4564 which was simply that we had stored and displayed more digits than had been computed.  If this is acceptable I'll do the same for ECNF.  Eventually the data stored will be changed: this fix hard-wires in the fact that the numbers were computed to 100 bits, so coercing into RealField(100) makes sense.


AndrewVSutherland:
You can view this on https://pink.lmfdb.xyz/EllipticCurve/Q/

PR #4937, opened 2022-01-21
avoid calling pari when it is easy to do so

AndrewVSutherland:
This PR implements LMFDB specific versions of the integer functions `is_squarefree`, `squarefree_part`, `divisors`, and `prime_divisors` using Sage's factor method to avoid calling pari, making it possible to run Sage in debug mode more easily.

There are still problematic cases where we are calling pari to compute Galois groups on the fly, or to manipulate polynomials, that are not addressed by this PR.


JohnCremona:
I'm happy with this once tests pass.  I even checked that the function integer_divisors() worked :)

JohnCremona:
I cannot see what the failing test was -- do we have to wait until all the other parts of the test finish?

AndrewVSutherland:
The test failure is a CMF test that is displaying the coefficient field using its LMFDB label rather than Q(sqrt(10)), I haven't figured out why this is happening, but we shouldn't merge this until I do.

JohnCremona:
> The test failure is a CMF test that is displaying the coefficient field using its LMFDB label rather than Q(sqrt(10)), I haven't figured out why this is happening, but we shouldn't merge this until I do.

Agreed

Issue #4170, opened 2020-10-23
BMF data - Mismatch between LMFDB and Magma for dimension 2 forms


BarinderBanwait:
There appears to be a discrepancy between the BMF eigenvalues for dimension 2 forms in the database, and what Magma computes.

E.g. Consider form [1066.4-a2](https://www.lmfdb.org/ModularForm/GL2/ImaginaryQuadratic/2.0.4.1/1066.4/a2/). The eigenvalues at the two primes above 5 are ostensibly the same, but the following Magma code shows that there are no BMF newforms of this level with the property that its two eigenvalues above 5 are the same:

```
P<x> := PolynomialRing(Rationals());
g := P![1, 0, 1];
F<i> := NumberField(g);
ZF := Integers(F);

NN := ideal<ZF | {-15*i -29}>;

M := BianchiCuspForms(F, NN);
S := NewSubspace(M);
NFD := NewformDecomposition(S);
newforms := [* Eigenform(U) : U in NFD *];

print "Are the Hecke eigenvalues at the two primes above 5 equal, for any newforms of this level?";

P1, P2 := Explode(Decomposition(ZF,5));

for f in newforms do;
 print HeckeEigenvalue(f,P1[1]) eq HeckeEigenvalue(f,P2[1]);
end for;
```

If this discrepancy indeed exists, then it could well be an issue with Magma. Either way, this ought to be investigated.

This was noticed in testing for #4053.

JohnCremona:
The original data for these (few) dim 2 bmfs were computed in magma in the first place. I suggest consulting Ciaran S who provided that data.

JohnCremona:
I ran this level on my C++ code (bianchi-progs on github).  At this level there's a rational newform and dimenesion 2, no oldforms.  The eigenvalues of the dim 2 newforms at the primes above 5 are negative-conjugate to each other, so if you call one 1+z (where z^2=-7) then the other is -1+z.    

So the data displayed for the dim 2 newforms is wrong.  *But*  the magma code listed here does not show that. since newforms[1] is the rational eigenform (who eigenvalues at these two primes are 0 and -2) while newforms[2] is the dim 2 one, and its eigenvalues are listed as $.1 and $.1-2 which is correct if you take $.1 to be a root of x^2+2x-6.

[...]

PR #4939, opened 2022-01-21
Hide dimension 2 BMFs

JohnCremona:
This addresses issue #4170.  I have not deleted the (small number of) dimension 2 BMFs from the table bmf_forms, but I have 
   - removed Dimension from the search options
   - adjusted the stats page (the last two tables are now deleted)
   - adjusted the wording in the description of newform spaces.
I just realised that after removing the Dimension from the search options I should still add the constraint dimension=1 to the search.  I will fix that and update the PR shortly

### DIFF
--- a/lmfdb/bianchi_modular_forms/bianchi_modular_form.py
+++ b/lmfdb/bianchi_modular_forms/bianchi_modular_form.py
@@ -172,6 +172,7 @@ def bianchi_modular_form_search(info, query):
     parse_nf_string(info, query, 'field_label', name='base number field')
     parse_noop(info, query, 'label')
     #parse_ints(info, query, 'dimension')
+    query['dimension'] = 1
     parse_ints(info, query, 'level_norm')
     parse_primes(info, query, 'field_bad_primes', name='field bad primes',
          qfield='field_bad_primes',mode=info.get('field_bad_quantifier'))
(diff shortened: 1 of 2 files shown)


AndrewVSutherland:
You can still access the dimension 2 forms via the jump box, e.g. http://127.0.0.1:37777/ModularForm/GL2/ImaginaryQuadratic/2.0.4.1/1066.4/a2/

This doesn't necessarily mean the PR needs to change, I think we want to remove these entries from the table, right?

JohnCremona:
> You can still access the dimension 2 forms via the jump box, e.g. http://127.0.0.1:37777/ModularForm/GL2/ImaginaryQuadratic/2.0.4.1/1066.4/a2/

True

> 
> This doesn't necessarily mean the PR needs to change, I think we want to remove these entries from the table, right?

OK -- I was going to ait but could do it now.  We would then need to synchronise the pushing of this to the web, copying the data to web, and positively reviewing the knowls.  I'll edit the knowls now in preparation, but I think it would be better not to delete the dim 2 forms from the database until this is on beta.  Is that sensible?

JohnCremona:
I have just edited the two relevant rcs knowls, by commenting out the sentence which mentions the dim.2 newforms.

AndrewVSutherland:
okay, pushing to beta now.

JohnCremona:
> okay, pushing to beta now.

Thanks, now I'll delete the (ir)relevant data...done.  So what's left is: at next code merge to prod, copy bmf_forms (only) and positively review the two rcs knowls.

Issue #4919, opened 2022-01-19
Projective field is Galois closure


AndrewVSutherland:
On Artin rep pages where we show the projective field the template should say "Galois closure of ...", as it does on the weight-1 CMF pages, e.g. compare

https://www.lmfdb.org/ArtinRepresentation/2.3_19.6t5.1
https://www.lmfdb.org/ModularForm/GL2/Q/holomorphic/57/1/h/a/

jwj61:
This is for a Galois orbit of an Artin representation.  Another option is to make it like the page for a single Artin representation where it says "stem field" in the label for the line instead of field when the field is not Galois.

AndrewVSutherland:
Unless the stem field is somehow distinguished from its siblings (in way that is intrinsic to the Artin representation), I think it makes more sense to say Galois closure (in both cases).

jwj61:
To me it is six of one ... .  I will leave it since it is marked as a good first issue.  For whoever does it, the individual artin rep page shows the test of whether the field is Galois or not.

AndrewVSutherland:
@jwj61 I don't think this behavior is new, see the Artin field on https://www.lmfdb.org/ArtinRepresentation/1.1.1t1.a for example (which is still running a version that does not include #4940) which refers to the Galois closure of Q.

But I agree with you that for both the projective and Artin fields we should show the prefix "Galois closure of" only in cases where the field is not Galois, and this should be an easy change.

jwj61:
It is new (for pages for reps).  On beta, it says "Projective field" if the field is Galois over Q, and "Projective stem field" if it is not.  There is a similar switch made on beta for Artin field and Artin stem field.

AndrewVSutherland:
But are you saying the Artin field should say "Galois closure over Q"?  Shouldn't they both just say Q?

jwj61:
They should both say Q.  
On the rep pages, the toggle before was on the knowl.  Now it will be on whether or not to add the phrase "Galois closure of".

AndrewVSutherland:
Got it, either I or @BarinderBanwait will implement this tomorrow (before we push to web).

BarinderBanwait:
I'll do it :hammer_and_pick:

PR #4940, opened 2022-01-21
add "galois closure of" to artin reps projective fields

BarinderBanwait:
addresses #4919 i think?


AndrewVSutherland:
http://127.0.0.1:37777/ArtinRepresentation/2.57.6t5.a.a hasn't changed, I think it should be consistent with http://127.0.0.1:37777/ArtinRepresentation/2.57.6t5.a

BarinderBanwait:
oh right, it had to be done on both the individual rep pages as well as galois orbits; hopefully the last commit does that

PR #4941, opened 2022-01-21
Improving genus 2 lookup

edgarcosta:


AndrewVSutherland:
The search -x^5+x+1 works on beta but not in this PR.

[[1,1,0,0,0,1],[]] does not work nor does [1,1,0,0,0,1] or X^5+X+1 or [x^5+x+1,0] or x^5+x+1,0.  I'm not sure which of these are intended to work but I think at least the first one should.

roed314:
There's also an extraneous `print(c)` added here.

AndrewVSutherland:
[[12, 0, -3, 0, -1],[0, 0, 0, 1]] doesn't work, even though -x^4-3x^2+12,x^3 and [-x^4-3x^2+12,x^3] do, as does [[0, 0, 0, 0, 1, 1],[1, 1, 0, 1]].

edgarcosta:
note to self, it is the negative numbers. add a a possibility of a minus to
the regex

edgarcosta:
it works now

Issue #4467, opened 2021-02-28
Speed up slow NF searches on ramified primes by converting include to exact when possible


a234:
When searching number fields that ramify only at 17, this search (i.e. one ramified prime including 17) returns an error:
https://www.lmfdb.org/NumberField/?num_ram=1&ram_primes=17&search_type=List
while this search (i.e ramified exactly at 17) gives the correct return.
https://www.lmfdb.org/NumberField/?hst=List&ram_quantifier=exactly&ram_primes=17&search_type=List

Would it be possible to redirect searches of the former kind to that of the later kind?

jwj61:
I ran the search several times with the given link and on beta.  None gave an error.

edgarcosta:
@a234, do you recall what the error message said?

edgarcosta:
I looked into this a bit further and it most likely was a timeout.
There are 5 timeout errors logged for that search query.

Here is a longer list of the latest timeouts on nf_fields for the last days: https://gist.github.com/edgarcosta/71641943b0941adb6dc3eaa3e1c4a624

AndrewVSutherland:
Independent of whether the query times out or not, the key point is that one of these equivalent searches is much faster than the other (on my machine, even after running both queries repeatedly to get as much cache benefit as possible, the first search takes 8.5 seconds while the second on takes 350 ms, so a factor of about 25).

It seems like it would be worth switching include to exact whenever we know it won't change the semantics of the search (e.g. count=1 include=blah is the same as count=1 exact=blah for any non-blank blah).  If you wanted to get fancy you could also change count=n include=blah to count=n exact=blah whenever blah is a list of n primes, but I don't think that's necessary.

rbommel:
I did not like Drew's solution that only works in very specific cases so much, so I went experimenting a bit myself. When I execute the query ``db.nf_fields.search({'num_ram':` {'$lt' : 2}, 'disc_rad' : {'$lt': 17}, 'ramps': {'$containedin': [17]}})``, it is already much faster. This is something we can already do in general right now. A similar query for the primes 2 and 17 then only takes 1.77 seconds instead of 10.8 seconds. Another idea that I had was to add the maximum ramified prime as a column in the table. That should allow us to filter more things out for these kind of queries.

[...]

PR #4944, opened 2022-01-21
Faster search for number fields with restricted ramification

rbommel:
This pull request implements a trick meant to mitigate #4467. It might not solve the problem completely, but hopefully speeds up a lot of queries.


AndrewVSutherland:
This is a very nice improvement, thanks @rbommel !

Issue #4393, opened 2021-01-30
Add comments to the top of the CMF Magma download file


AndrewVSutherland:
The Magma download files for most CMFs are sufficiently long that it can be quite hard for the uninformed user to figure out what function they are supposed to call.  We should put a comment that the very top of the file saying "Use qexpCoeffs()" to get the q-expansion of this modular form, and then also mention the functions MakeNewformModFrm_xxx and MakeNewformModSym_xxx.

I think these comments were present at one point, but as we have added more code I think they stayed with their functions and they really need to be write at the top of the file -- we shouldn't expect users to read through what may be hundreds or even thousands of lines of code to find the function they are supposed to call (and they may well miss the fact that they can very quickly get the first 1000 coefficients of the q-expansion without doing anything complicated).

The Download q-expansion to Sage files are fine, there are comments write at the top telling you exactly how to create the q-expansion.

assaferan:
Hi Drew, I can take this one.

PR #4945, opened 2022-01-21
4393

assaferan:
Solving issue #4393 - all comments are also added at the top of the magma file.


AndrewVSutherland:
Looks good!

Issue #4923, opened 2022-01-20
For abelian varieties, not all invariants in the properties box are listed under invariants.


edgarcosta:
At the moment we have sentences listing some of the positive properties, but not the negative ones.

For example, if you land here:
https://beta.lmfdb.org/Variety/Abelian/Fq/4/4/ae_h_aj_s
from the properties box, you will know that is neither Ordinary nor Supersingular, but you will not have a knowl to describe those.

This could be a potential first issue after some guidance.

davidfarmer:
I will do this, and follow the model on "group" pages where one long sentence lists all the yes/no properties:

https://beta.lmfdb.org/Groups/Abstract/256.15467

AndrewVSutherland:
Can we close this now?

PR #4947, opened 2022-01-21
Add boolean properties to main abvar page

davidfarmer:
Addresses
https://github.com/LMFDB/lmfdb/issues/4923

Every AbVar isogeny class page should now have a longer sentence at the end
of the Invariant table.


AndrewVSutherland:
There are several tests that need to be modified (is_simple, is_ordinary, is_supersingular)/

PR #4951, opened 2022-01-21
Adding some search page columns for number fields

rbommel:
Adding the following columns to the search result pages for number fields: degree, signature, is CM, is Galois, torsion order, unit rank, regulator.


AndrewVSutherland:
I think is_galois and is_cm should both use CheckCol

rbommel:
Changed.

PR #4956, opened 2022-01-22
only say Galois closure if not Galois

BarinderBanwait:
addresses latest discussion on #4919

### DIFF
--- a/lmfdb/artin_representations/templates/artin-representation-galois-orbit.html
+++ b/lmfdb/artin_representations/templates/artin-representation-galois-orbit.html
@@ -13,7 +13,7 @@
         {% if object.sign() %}
         <tr><td class="nowrap">{{ KNOWL('artin.root_number', title='Root number') }}: </td><td>${{ object.root_number()}}$</td></tr>
         {% endif %}
-        <tr><td class="nowrap"> {{ KNOWL('artin.number_field', title='Artin number field') }}: </td><td>
+        <tr><td class="nowrap"> {{ KNOWL('artin.number_field', title='Artin field') }}: </td><td>
         {% if wnf %}
             {% if wnf.is_galois() %}
                 {{wnf.knowl()|safe}}
--- a/lmfdb/artin_representations/templates/artin-representation-galois-orbit.html
+++ b/lmfdb/artin_representations/templates/artin-representation-galois-orbit.html
@@ -15,15 +15,24 @@
         {% endif %}
         <tr><td class="nowrap"> {{ KNOWL('artin.number_field', title='Artin number field') }}: </td><td>
         {% if wnf %}
-        Galois closure of {{wnf.knowl()|safe}}
+            {% if wnf.is_galois() %}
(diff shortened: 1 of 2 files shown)


BarinderBanwait:
the test failure here is unrelated and is apparently already in LMFDB:master

AndrewVSutherland:
Test failure is caused by BMF test failing because dimension 2 forms are no longer present, which is unrelated to this PR and fixed in #4957

AndrewVSutherland:
"Artin stem field" should be changed to "Artin field" on http://127.0.0.1:37777/ArtinRepresentation/2.39.4t3.a.a (the knowl link also needs to change), and I would change "Artin number field" to "Artin field" on http://127.0.0.1:37777/ArtinRepresentation/2.39.4t3.a (knowl link is fine here).

JohnCremona:
> Test failure is caused by BMF test failing because dimension 2 forms are no longer present, which is unrelated to this PR and fixed in #4957

Apologies for that: not caused by my BMF PR itself, which I tested, but by the subsequent data change.

BarinderBanwait:
> "Artin stem field" should be changed to "Artin field" on http://127.0.0.1:37777/ArtinRepresentation/2.39.4t3.a.a (the knowl link also needs to change), and I would change "Artin number field" to "Artin field" on http://127.0.0.1:37777/ArtinRepresentation/2.39.4t3.a (knowl link is fine here).

I thought that @jwj61 was saying that we use "Artin stem field" if the field is *not* Galois, which is the case for the first link you sent, so that "Artin stem field" should *not* be changed to "Artin field" in that first case?

On the other hand, in the second link to the Galois orbit, it *should* be changed from "Artin number field" to "Artin stem field"?

In all cases, all instances of "Artin number field" should be changed to "Artin field"?

jwj61:
I will see if I can clarify this.

With issue #4435, the page for Artin reps was changed so that it would say "stem field" or "field" depending on whether or not the field is Galois.  This. was for both the Artin field and the projective field.  The orbit pages were not updated to match at that time.

The current issue points out 2 problems.  One is that rep pages are not in sync with orbit pages.  The other is the preference for "Galois closure of" over use of the term "stem field".  So, no more use of the term stem field, and only say "Galois closure of" when the field is not Galois.

So, there are two types of pages (orbit and rep) and changes to be made in two places on each (Artin field and projective field), and they all need to be able to switch back and forth depending on whether or not the field is Galois.

BarinderBanwait:
Thanks everyone for the clarifications, hopefully the latest version is correct now

AndrewVSutherland:
Looks good to me!  But I'll give @jwj61 a chance to comment before merging.

jwj61:
Looks good to me.  If the test failure is inconsequential, I would merge.

AndrewVSutherland:
The test failure is unrelated (caused by data change to BMFs that invalidated a test that is fixed in another PR).  I'll merge.

Issue #4707, opened 2021-08-13
Inconsistency when displaying related Sato-Tate groups in Sato-Tate groups pages 


edgarcosta:
At the moment we are inconsistent with the display of related Sato-Tate groups.
For example, the connected component of https://www.lmfdb.org/SatoTateGroup/1.6.G.1.1a is displayed as `U(1)^2 x SU(2)` with no link.
But the maximal subgroup, which is also the identity component, is displayed as `F x SU(2)` and links to https://www.lmfdb.org/SatoTateGroup/1.6.G.1.1a.

This inconsistency also shows up for connected groups where the name of the identity component doesn't match the name of itself (in the properties and in the page).

Perhaps the easiest solution would be to have a display knowl for Sato-Tate groups? Just like we do for number fields.

edgarcosta:
I realized that you also do the same in lower genus: https://www.lmfdb.org/SatoTateGroup/?hst=List&components=1&search_type=List
Is this by design? If so, is it because you want to only consider the identity component as an abstract group?

AndrewVSutherland:
This is by design, and consistent with the notation used in the papers that define these groups and is motivated by the fact that there is often a natural way to describe a set of Sato-Tate groups, e.g. E_n and J(E_n), that would be awkward to formulate in a way that made the connected groups have names that match the identity component.  I think the right way to think about it is that the groups ST(A) and ST^0(A) live in different name spaces.

I agree that it it can be confusing, but I think the alternatives are all likely to be more confusing (especially if they disagree with the notation used in the papers).  On the plus side, the labels are consistent: the first three components specify the identity component (e.g. 1.6.G) while the full label specifies the Sato-Tate group (1.6.G.1.1a).

I like the idea of a Sato-Tate group knowl (that could include the label, the name, the name of the identity component, as well as the GAP id of the component group, the weight, the real dimension, and the ambient space).  Maybe change this request toask for  an ST group knowl?

I should note that the identity component is not quite an abstract group, for example, the groups U(1)_2 x SU(2) and U(1) x SU(2)_2 are both abstractly isomorphic to U(1) x SU(2) but denote different embeddings of this group in USp(6).

edgarcosta:
I agree that having a display knowl is a good first step.
However, not sure if this helps with the confusion when it comes to connected Sato-Tate groups.
I also don't have a good solution...
Perhaps, we should separate ST(A) and ST^0(A), e.g., the labels of ST^0(A) should only have 3 components:
[...]

Issue #294, opened 2015-07-30
Searching elliptic curves by mod-p Galois image


AndrewVSutherland:
In addition to being able to specify primes p at which the mod-p image is non-surjective, it would be very useful to be able to specify a list of mod-p Galois image group labels (e.g. 3B.1.2 5Nn) and get a list of all elliptic curves with these mod p images, both for elliptic curves over Q and for elliptic curves over number fields.

AndrewVSutherland:
#4957 adds this for number fields

JohnCremona:
I would like to be able to have something in between "non-maximal" and a specific rather technical image code name.  I mean to be able to search for "contained in non-split Cartan at 3" or similar.  For each prime, such terms would translate into a finite (short) list of image labels, so would be easy to do.  The UI would be harder if it was allowed to specify such things at more than one prime -- though these combinations occur rather rarely, when they do they are interesting!  So probably this is a "feature request".

AndrewVSutherland:
@JohnCremona It would be easy to add a list of named keywords (e.g. "contained in non-split Cartan at 3") that translates to a list of subgroup labels (this could even be stored in a table).  We would then need to add a selector that lets the user choose from the list of keywords.  Feel free to raise an issue (which I agree would be a "feature request").  If you give me a list of desired selection criteria I can easily generate a list of corresponding subgroup labels.

AndrewVSutherland:
One other thought (which might be both more general and easier to implement), we could allow strings like "<= 3Cn" or "> 5Cs" in the list of mod-ell image labels.  I think the mod-ell image labels are pretty intuitive, at least at the level of detail people are likely to be interested in (Borel, Cartan (split/non-split), Normalizer of Cartan, S4, ...) and explained in the knowl https://www.lmfdb.org/knowledge/show/ec.galois_rep_modell_image (we could also create another knowl more specific to searching that would explain things in more detail and give instructions on how to search for things like "contained in non-split Cartan at 3", with examples).

PR #4957, opened 2022-01-22
More search columns for elliptic curves (over Q and over number fields)

AndrewVSutherland:
I also changed the way torsion subgroups are displayed on EC/ECNF/G2C, see

https://blue.lmfdb.xyz/EllipticCurve/Q/?search_type=List&all=1
https://blue.lmfdb.xyz/EllipticCurve/?search_type=List&all=1
https://blue.lmfdb.xyz/Genus2Curve/Q/?search_type=List&all=1

Click columns to display to see all the new display options.


JohnCremona:
Looks great, even testing on a phone the column selection works fine. Though I didn't test everything.

I might suggest not showing Weierstrass coefficients by default, they are really not very illuminating and in the same space we could turn on several more interesting columns. Thoughts?

AndrewVSutherland:
Regarding omitting Weierstrass coeffs, this is an interesting suggestion and  I'm honestly not sure -- I like seeing them, but I also like the idea of freeing up the space.  I'll ask for opinions at our show-and-tell at noon EST.

AndrewVSutherland:
Also addresses #294 by adding mod-ell image search option for ecnf

JohnCremona:
> Regarding omitting Weierstrass coeffs, this is an interesting suggestion and I'm honestly not sure -- I like seeing them, but I also like the idea of freeing up the space. I'll ask for opinions at our show-and-tell at noon EST.

Even these?  https://www.lmfdb.org/EllipticCurve/?bf_deg=6&search_type=List

Checking this out now...

JohnCremona:
EC/NF: looks really good!  Just one suggestion: in the mod-ell images column, when ell is maximal, perhaps put something into the column instead of leaving a blank? e.g. "all maximal" or similar.

EC/Q: I spotted that "class size" is not in math mode,, perhaps it should be for consistency.

Over Q you have a Conductor column, showing the factored conductor, but there is no such column over NF.  There could/should be?  The database does contain the exponents as well as the bad primes (for which there is a column) so having a factored-conductor column should not be hard.

Discriminant column: perhaps factored as for conductor?  The exponents are available.

In general, for any column in either ECQ or ECNF which could be in the other, it should be (if not possible because of the database not having the appropriate column, say,  we can do something about it).

Thanks for doing this @AndrewVSutherland (and of course @roed314 -- this is a brilliant feature!).

AndrewVSutherland:
@JohnCremona Thanks for the feedback!  I put class size (and degree) in math mode for ECQ (as they are on ECNF), and the factored conductor norm is now shown on https://blue.lmfdb.xyz/EllipticCurve/?search_type=List&all=1.  Let me know what you think about the column header [O_K:n], which I used because it is narrower than "Conductor norm" and I didn't want to just write "Conductor", since that is a different column (although we could use Conductor as the column heading but show Conductor norm in the selector))

I agree with making the same columns available for ECQ, ECNF (and G2C!) whenever possible.  I'm working on adding Sato-Tate group for all three now.  I'll also add factored discriminant (norm) for all three.

AndrewVSutherland:
@JohnCremona You might also want to check out the maximal primes and mod-ell image search options I added to ECNF to address #294.

AndrewVSutherland:
@JohnCremona If you [search for curves of conductor 14](https://blue.lmfdb.xyz/EllipticCurve/Q/?hst=List&conductor=14&search_type=List) and show the "Class size" column you will see that 14.a3 has class size 4 while all the other members of its isogeny class have class size 6 (which is the number of curves in the isogeny class).  Is this a [data error](https://beta.lmfdb.org/api/ec_curvedata/?lmfdb_iso=14.a), or is there something special about 14.a3 (and if so, what does its class size mean)?

The same issue arises in several other isogeny classes.

JohnCremona:
> @JohnCremona If you [search for curves of conductor 14](https://blue.lmfdb.xyz/EllipticCurve/Q/?hst=List&conductor=14&search_type=List) and show the "Class size" column you will see that 14.a3 has class size 4 while all the other members of its isogeny class have class size 6 (which is the number of curves in the isogeny class). Is this a [data error](https://beta.lmfdb.org/api/ec_curvedata/?lmfdb_iso=14.a), or is there something special about 14.a3 (and if so, what does its class size mean)?
> 
> The same issue arises in several other isogeny classes.

It's a data error.  The class size is stored both in the table ec_classdata, but also in ec_curvedata so it can be searched on.  Obviouly all curves in the class should have the same class size there.  That should be easy to fix  in the database and to debug to see how it happened.  I'll do that tomorrow.

JohnCremona:
> @JohnCremona Thanks for the feedback! I put class size (and degree) in math mode for ECQ (as they are on ECNF), and the factored conductor norm is now shown on https://blue.lmfdb.xyz/EllipticCurve/?search_type=List&all=1. Let me know what you think about the column header [O_K:n], which I used because it is narrower than "Conductor norm" and I didn't want to just write "Conductor", since that is a different column (although we could use Conductor as the column heading but show Conductor norm in the selector))

I don't much like the notation [O_K:n] in the column heading, too much notation (though explained in the knowl certainly), for example the notation n for conductor is not anywhere wlse on the page.  Also, I thought that we could have the factored conductor (ideals to powers) rather than just the label. -- unless we can (not for his PR!) -- have an "ideal knowl" (like a number field knowl) which could be attached to all ideal labels everywhere to show factorization, norm, etc.

If we're allowed 2-line column headings then Conductor Norm would not be too wide?

> 
> I agree with making the same columns available for ECQ, ECNF (and G2C!) whenever possible. I'm working on adding Sato-Tate group for all three now. I'll also add factored discriminant (norm) for all three.

I'll come back to this tomorrow, I know that I have not answered all your questions yet.

AndrewVSutherland:
@JohnCremona I changed the column header to "Conductor norm", which will automatically display in two rows if your screen is narrow.  Let me know if you think "Cond norm" would be better.  I've also added factored discriminants to EC/ECNF/G2Q.

Once tests pass I think this is ready to merge if you are happy with it.  I decided to wait to add/update the Sato-Tate group columns for EC/ECNF/G2C until after I have implemented #4707 (so I can link them to Sato-Tate knowls).

I think we should also add "ideal knowls" as you suggest (which would just show the ideal label as the anchor but when opened would tell you the number field and show the factorization into prime ideals).  Presumably we should also do this for the discriminant? (which I think would require computing the ideal labels, which I don't see stored).  But I agree this should be a separate PR!

It would be good to merge this PR soonish, as it includes a fix to the BMF tests (currently all other PRs against master will fail tests because of this).  We can always make more tweaks later.

AndrewVSutherland:
@JohnCremona One further comment, in the feedback session we discussed omitting equations for ECNF when the base field has degree > 3 but decided not to on the grounds that putting the equation in the rightmost column ensures that it never hides something you want to see, even if it is ridiculously long, it will just scroll off to the right and is there for anyone who wants to see it -- I think this is better than the potential confusion people might have if they were running several searches or paging through search results and the equations sometimes disappeared, and people generally felt that it was really nice to have the equations there in the low degree cases (which is what people will see most of the time).

JohnCremona:
OK, fine

JohnCremona:
> > @JohnCremona If you [search for curves of conductor 14](https://blue.lmfdb.xyz/EllipticCurve/Q/?hst=List&conductor=14&search_type=List) and show the "Class size" column you will see that 14.a3 has class size 4 while all the other members of its isogeny class have class size 6 (which is the number of curves in the isogeny class). Is this a [data error](https://beta.lmfdb.org/api/ec_curvedata/?lmfdb_iso=14.a), or is there something special about 14.a3 (and if so, what does its class size mean)?
> > The same issue arises in several other isogeny classes.
> 
> It's a data error. The class size is stored both in the table ec_classdata, but also in ec_curvedata so it can be searched on. Obviouly all curves in the class should have the same class size there. That should be easy to fix in the database and to debug to see how it happened. I'll do that tomorrow.

I tracked down how this happened (when we implemented a search by class size and had to copy the 'class_size' column from table ec_classdata to ec_curvedata).  It should only effect isogeny classes of size>2, but I am in the process of updating all the class_size column entries in ec_curvedata.

AndrewVSutherland:
Great, let me know when it's down so I can copy `ec_curvedata` to proddb.  BTW, I added the columns `nonmax_primes` and `nonmax_rad` to `ec_nfcurves` and have already copied it over.  Once this PR is pushed to web I will drop the column `non-surjective_primes` which is superseded by `nonmax_primes` and no longer referenced in the code.

JohnCremona:
> Great, let me know when it's down so I can copy `ec_curvedata` to proddb. BTW, I added the columns `nonmax_primes` and `nonmax_rad` to `ec_nfcurves` and have already copied it over. Once this PR is pushed to web I will drop the column `non-surjective_primes` which is superseded by `nonmax_primes` and no longer referenced in the code.

Still in progress. An update_from_file times out while I had lunch so I am running it again. That is for conductors <500k, then I'll do the rest (but there will be nothing to do for the large primes conductors.

I'm glad you told me about the extra column in ec_nfcurves since I am preparing more data for upload and I need to know when the schema changes, or there will be problems.  Is 'non_max primes' just a renaming of 'non-surhective primes' which was always a misnomer for CM curves?

AndrewVSutherland:
@JohnCremona Yes, `nonmax_primes` is just a copy of `non-surjective_primes`.  I need to make a few changes to it for some of the CM curves in ec_nfcurves at primes dividing the CM discriminant (some that are marked as non-maximal are actually maximal -- the identified image is correct, but it is actually as large as the endomorphisms permit (even though a proper subgroup of the Borel), hence should be marked maximal -- this has already been done for elliptic curves over Q).  I'll update the script that computes the Galois images to handle this correctly so the issue won't arise again in the future.  I'll try to do this on Friday.

JohnCremona:
> @JohnCremona Yes, `nonmax_primes` is just a copy of `non-surjective_primes`. I need to make a few changes to it for some of the CM curves in ec_nfcurves at primes dividing the CM discriminant (some that are marked as non-maximal are actually maximal -- the identified image is correct, but it is actually as large as the endomorphisms permit (even though a proper subgroup of the Borel), hence should be marked maximal -- this has already been done for elliptic curves over Q). I'll update the script that computes the Galois images to handle this correctly so the issue won't arise again in the future. I'll try to do this on Friday.

Thanks.   Did I send you the version of your scripts which I slightly patched for ease of interface from Sage? See ~/galrep on legendre where the changed file is nfgalrep.m. The wrapper is in ~/ecnf-data/sage/galrep.py.

Timing is fine.  I have (or had) upload files ready to go for Q(sqrt(-d)) for d=59 and 89, d=47 is almost done and d=71, 79 have nearly finished finding BMFs for level norm up to 1000.  That will include 7 fields with odd class number 3, 5, 7 (countinf d=23, 31 which have been in for a while).  I have more work to do for even class number but I am confident they will follow.

JohnCremona:
OK @AndrewVSutherland ec_curvedata is ready to be copied to prod.  Thanks again for spotting this. Apart from it becoming blatant with the new multi-column search results display, the only situation where this glitch would have made any difference is when specifically searching for a particular class size + some confusion in the affected home pages (see https://www.lmfdb.org/EllipticCurve/Q/14/a/3 which wrongly says it's one of 4 curves in its class).

JohnCremona:
OK, so after you have done this there will be a discrepancy (perhaps one that you do not care about, but which may cause me grief at some point) between the database contents and my own data files in my ecdata repository.  When you have done what is necessary I'll run your revised Magma code to get the corrections in there too.

AndrewVSutherland:
ec_curvedata and ec_nfcurves have both been updated to remove extraneous nonmax_primes and glaois_images for CM curves that are actually maximal and copied to proddb.  So the data should all be correct now, but I still need to update the scripts, which I will try to do on Friday.

alexjbest:
Hi Drew (@AndrewVSutherland), in this PR you removed searching on "is a base change Q" (`include_base_change`) for ecnf it seems, but I can't see any discussion of that here, was this intentional? I just wanted to use this feature today and was sad to see it gone.

alexjbest:
Ok now I see that it is part of Q-curve, I find this a bit unintuitive as Q-curve has a different, somewhat related but distinct meaning, but the option exists so I'm happy, please ignore me

alexjbest:
The distinction is clear! But the location of a search option that used to be a separate option but is now combined with something else wasn't.
I wasn't confused by that either really, I know that base changes only come from Q in the lmfdb (for now)

I would have preferred an "advanced search" option dropdown if saving real estate was the only concern that made you want to combine these options, but ce la via.

In my head Q curve is a far more advanced concept than base change, so I believe there may well be many users who have no idea what a Q curve is but want to for instance find high rank (over some nf) curves that actually are defined over Q.

AndrewVSutherland:
The one other comment I will make is that I do think that in the current setup everything is "discoverable" in that someone looking for base changes from Q who doesn't know what a Q-curve is and is looking for the search option they want is likely to eventually click on "Q-curve", since just from the name, it could plausibly be shorthand for what they want, and once they read the knowl things will become clear.  I think the category of user you have in mind won't necessarily be familiar with the term "base change" either, and might not use a dropdown that was simply labelled "base change" even if it was present.

JohnCremona:
By the way, there is no reason why we should not compute and store information about whether curves are base-change from any subfield k, not just Q.  This is currently only relevant for the base fields of degree 4 and 6 of course (and I think most of those fields have no intermediate fields, I have not looked into that in detail).  The Sage method E.descend_to(k) returns a list (possibly empty) of curves over k, where k should be a subfield of K=E.base_field(), whose base-change from k to K is isomorphic to E.  The only tricky part would be to look up whether those curves are in the database too -- I assume that they will be, but that is only a guess based on why the HMFs were computed.  This would make a nice little project for someone!

I'm happy with the boxes labelled as they are for now.

PR #4965, opened 2022-01-22
Make unit group rank be in math mode in nf search pages

jwj61:
On the number field search pages, if you add unit rank as a display column, it was not in math mode even though it is just an integer.  This puts it in math mode like other integers in the search columns.

Seeing it involves clicking to show the extra columns, so a single url can't do it, but any number field search should do.


roed314:
FYI, you can include the displayed columns in a url, like this: https://blue.lmfdb.xyz/NumberField/?hst=List&search_type=List&showcol=unit_rank.  But the url only changes when you hit "Search again".

Issue #4662, opened 2021-07-03
Links to the corresponding API page


roed314:
Most (all?) object pages in the LMFDB have a main corresponding entry in some database table (though some of them also draw data from auxiliary tables).  It would be nice to have a link to the corresponding API page(s) in the Downloads section.  So, for example, the number field page [8.0.568002355840000.24](https://beta.lmfdb.org/NumberField/8.0.568002355840000.24) would have a link to the corresponding [API result](https://beta.lmfdb.org/api/nf_fields/?label=8.0.568002355840000.24).  When there are auxiliary tables, I think we would ideally have links there as well.  So, for example, the elliptic curve page [11.a3](https://beta.lmfdb.org/EllipticCurve/Q/11/a/3) would link not just to the [curvedata record](https://beta.lmfdb.org/api/ec_curvedata/?lmfdb_label=11.a3) but also to the [localdata record](https://beta.lmfdb.org/api/ec_localdata/?lmfdb_label=11.a3), the [classdata record](https://beta.lmfdb.org/api/ec_classdata/?lmfdb_iso=11.a), the [galrep records](https://beta.lmfdb.org/api/ec_galrep/?lmfdb_label=11.a3), the [mwbsd record](https://beta.lmfdb.org/api/ec_mwbsd/?lmfdb_label=11.a3), the [2-adic record](https://beta.lmfdb.org/api/ec_2adic/?lmfdb_label=11.a3), the [Iwasawa record](https://beta.lmfdb.org/api/ec_iwasawa/?lmfdb_label=11.a3), and the [torsion growth records](https://beta.lmfdb.org/api/ec_torsion_growth/?lmfdb_label=11.a3).

AndrewVSutherland:
I think this is a good idea, but in your second example I don't think we want to have a separate link for each record.  It would be better to have a single "data" link (e.g. in the downloads box) that takes you to a page where all of these records are shown. I realize it would take a bit more work to do this, but it wouldn't be that hard to implement a page that takes a variable length list of table/key pairs, and in this particular example it would actually be really handy to be able to see all the records on the same page.

roed314:
This is now live in production.

Issue #4812, opened 2021-10-17
Improvements to schema display


AndrewVSutherland:
For tables with a small number of columns the schema display is a bit weird, it creates a scrollable window that is much larger than necessary (and which cannot actually scroll), see

https://beta.lmfdb.org/api/mf_hecke_traces/

for example.  Also, the table  has a lot of unused space in each row, it could be nice to add another column to the table which is an example of the contents (e.g. taken from a random row in the table).

JohnCremona:
In #4971 I suggested adding a column to the schma display which the description (no a knowl under the column name), with appropriate display modification for when the description is long.

roed314:
If the column name is no longer a knowl, how do you edit the description easily?

JohnCremona:
> If the column name is no longer a knowl, how do you edit the description easily?

Would the same way as with  top knowl, e.g. (when logged in) https://beta.lmfdb.org/EllipticCurve/Q/37a1/  work?

AndrewVSutherland:
It would probably make sense to address ##4811 at the same time as this one.

PR #4971, opened 2022-01-27
Datapage

### LINKED ISSUE #4969: Server error on bad HGM label
The URL https://beta.lmfdb.org/Motive/Hypergeometric/Q/banana throws a server error.

roed314:
This PR adds "Underlying data links" to most homepages in the lmfdb, as described in #4662.  It is available at [purple](https://purple.lmfdb.xyz) for experimentation.

When the page depends only one table, we just give a link to the appropriate `api_query` search result.  When there are multiple underlying tables (e.g. elliptic curves, classical modular forms, abstract groups), there is a new page which shows the entries in these different that contribute to the page.

Some things I'd like feedback on:
* I ignored the links that were already present and just added "Underlying data" at the bottom.  In some cases, this may overlap with links like "Download all stored data to text".  I'm not sure whether to include both or to remove one or the other.
* There are some homepages where I didn't add "Underlying data" links, mainly isogeny classes of various kinds where there was no separate table (e.g. isogeny classes of elliptic curves).  I'm open to adding something here, either giving just one entry from the isogeny class or all entries.
* I chose not to include data from things that felt more like "friends" than part of the object.  So, for example, I don't include an entry from `nf_fields` for the coefficient field of a modular form.  I also may have missed tables that should have been included.  I'm happy to add more tables.

Since it was a simple fix, this PR also resolves #4969 and fixes a bug in the "Random subgroup" button [here](https://purple.lmfdb.xyz/Groups/Abstract/?search_type=Subgroups).

### DIFF
--- a/Postgres_FAQ.md
+++ b/Postgres_FAQ.md
@@ -658,7 +658,7 @@ Statistics
    refresh them manually using, for example, `db.g2c_curves.stats.refresh_statistics()`.
    Sometimes this isn't sufficient (if there are extra stats that shouldn't
    be present for example).  A more drastic option is to call
-   `db.g2c_curves._clear_stats_counts()`.  If you do this, make sure to
+   `db.g2c_curves.stats._clear_stats_counts()`.  If you do this, make sure to
    visit the relevant statistics page while logged in as editor,
    in order to regenerate the statistics used there.
 
--- a/lmfdb/abvar/fq/main.py
+++ b/lmfdb/abvar/fq/main.py
@@ -204,7 +204,7 @@ def download_search(info):
 
 @abvarfq_page.route("/data/<label>")
 def AV_data(label):
-    bread = get_bread((f"Data - {label}", " "))
+    bread = get_bread((label, url_for_label(label)), ("Data", " "))
     extension_labels = list(db.av_fq_endalg_factors.search({"base_label": label}, "extension_label", sort=["extension_degree"]))
     tables = ["av_fq_isog", "av_fq_endalg_factors"] + ["av_fq_endalg_data"] * len(extension_labels)
     labels = [label, label] + extension_labels
--- a/lmfdb/abvar/fq/main.py
+++ b/lmfdb/abvar/fq/main.py
@@ -18,6 +18,7 @@ from lmfdb.utils import (
     search_wrap, count_wrap, YesNoMaybeBox, CountBox, SubsetBox, SelectBox
 )
 from lmfdb.utils.interesting import interesting_knowls
+from lmfdb.api import datapage
 from . import abvarfq_page
 from .search_parsing import parse_newton_polygon, parse_nf_string, parse_galgrp
 from .isog_class import validate_label, AbvarFq_isoclass
@@ -130,6 +131,7 @@ def abelian_varieties_by_gqi(g, q, iso):
 
     if hasattr(cl, "curves") and cl.curves:
         downloads.append(('Curves to text', url_for('.download_curves', label=label)))
+    downloads.append(("Underlying data", url_for('.AV_data', label=label)))
 
     return render_template(
         "show-abvarfq.html",
@@ -198,6 +200,16 @@ def download_search(info):
     strIO.write(s.encode('utf-8'))
     strIO.seek(0)
     return send_file(strIO, attachment_filename=filename, as_attachment=True, add_etags=False)
+
+@abvarfq_page.route("/data/<label>")
+def AV_data(label):
+    bread = get_bread((f"Data - {label}", " "))
+    extension_labels = list(db.av_fq_endalg_factors.search({"base_label": label}, "extension_label", sort=["extension_degree"]))
+    tables = ["av_fq_isog", "av_fq_endalg_factors"] + ["av_fq_endalg_data"] * len(extension_labels)
+    labels = [label, label] + extension_labels
+    label_cols = ["label", "base_label"] + ["extension_label"] * len(extension_labels)
+    sorts = [[], ["extension_degree"]] + [[]] * len(extension_labels)
+    return datapage(labels, tables, title=f"Abelian variety isogeny class data - {label}", bread=bread, label_cols=label_cols, sorts=sorts)
 
 class AbvarSearchArray(SearchArray):
     jump_example = "2.16.am_cn"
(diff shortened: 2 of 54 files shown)


JohnCremona:
This looks very promising, and I have been playing with it, so I have some minor comments.

Having the schema available for all tables on the database data page is surely essential.  I saw it for a BMF but not for an EC/Q (e.g. https://purple.lmfdb.xyz/EllipticCurve/Q/data/238590.e4)  By the way, I like that URL system.  Is that because more than one table is in play?  I can see that if I click on the name of one of the tables there, e.g. ec_curvedata, then I get that subset of the data and now "show schema" appears.  Please can you put "show schema" next to each table on the main data page too?

Also concerning schemas: I see that the descriptions of what each column actually represents can be seen as a knowl linked to the column name.  I would like that to be visible for all columns in the "show schema" table, as a column, if that is possible.  (The knowl opening does not work too well for items near the bottom of the table, since the opened knowl is invisible).

I have only looked at BMF, ECQ and ECNF so far.   Could there be a data page for an entry in the bmf_forms table, e.g. for https://purple.lmfdb.xyz/ModularForm/GL2/ImaginaryQuadratic/2.0.23.1/72.11?

roed314:
Getting the schema to show for each table is currently technically challenging because of the way the javascript was written (it assumed that there was only one "show schema" link per page), but will think about whether there's a good way to make it work.

My worry about showing the contents of the knowl as a separate column is that it could be long.  Certainly many descriptions are very short at the moment, and there's a lot of unused horizontal space.  Maybe truncate it if it gets too long?

What would you like the data page for https://purple.lmfdb.xyz/ModularForm/GL2/ImaginaryQuadratic/2.0.23.1/72.11 to show?  There are data pages for bmf forms, just not bmf spaces since I didn't know what the appropriate tables to show there would be.  See https://purple.lmfdb.xyz/ModularForm/GL2/ImaginaryQuadratic/2.0.23.1/72.11/a/ for example.

JohnCremona:
Yes, some descriptions will be long.  Perhaps truncate after N characters (from some N) but if truncation was needed have some kind of "show more" indication?

Also, if this is not too circular, the ability to download a table's schema with the descriptions would be nice (if not already possible).

The page https://purple.lmfdb.xyz/ModularForm/GL2/ImaginaryQuadratic/2.0.23.1/72.11 shows data from the row of  table bmf_forms with label column '2.0.23.1-72.11', which contains
sage: db.bmf_dims.lucky({'label': '2.0.23.1-72.11'})                
{'gl2_dims': {'2': {'new_dim': 1, 'cuspidal_dim': 7}},
 'field_absdisc': 23,
 'label': '2.0.23.1-72.11',
 'field_label': '2.0.23.1',
 'level_norm': 72,
 'level_label': '72.11',
 'gl2_new_totaldim': 1,
 'gl2_cusp_totaldim': 7}

roed314:
I've added a link to https://purple.lmfdb.xyz/ModularForm/GL2/ImaginaryQuadratic/2.0.23.1/72.11.  Let me know if that looks right.  It seems to have two identical rows there....

roed314:
Where would you want the schema download to go?  As comments in the yaml file wouldn't be too hard, but json doesn't have comments.

roed314:
Ok, I think it won't be too hard to add schemas to the html (when multiple tables are shown).  I'll work on that tomorrow (later today for you).

If we added a column with the contents of the knowl (and possibly a "show more" link?), I think we still need the column name to be a knowl so that you can get the "edit" and "permalink" buttons.  That feels a little weird then, but I don't have a better suggestion at the moment.

roed314:
Also note that #4812 suggests adding an example of the contents as another column to the schema table, which would use up some of the extra horizontal space.

JohnCremona:
> I've added a link to https://purple.lmfdb.xyz/ModularForm/GL2/ImaginaryQuadratic/2.0.23.1/72.11. Let me know if that looks right. It seems to have two identical rows there....

Thanks!  The duplicate rows are a data glitch I only just notced and have made issue #4972 .  Otherwise looks perfect.

JohnCremona:
> Also note that #4812 suggests adding an example of the contents as another column to the schema table, which would use up some of the extra horizontal space.

OK, we can move the rest of this discussion to #4812

roed314:
Any thoughts on the bullet points in the original post?

JohnCremona:
> Any thoughts on the bullet points in the original post?

   1. I think we can remove the old data download links, assuming that we can download from the new pages. Some editorial input needed here.
   2. Fine by me to leave these out. We are not hiding any data.  There has always been a case for a table for isogeny class data for elliptic curves, but I ve never been convinced that anything would be gained. Not yet anyway.
   3. Omitting friends seems right to me.
Just my opinion!

edgarcosta:
Regarding (1), I would keep both links there, as usually some data already has been processed to make into a single coherent JSON, e.g., CMFs.

AndrewVSutherland:
I'm generally very happy with this, my only complaint/suggestion is that I think that the difference in how objects are handled when they have data in more than one table vs just one table is confusing, and (related) I would change the bread.  Compare the two links one arrives at if you go to an elliptic curve page and click the "Underlying data" download link:

- ECQ: https://purple.lmfdb.xyz/EllipticCurve/Q/data/11.a1 (bread is Elliptic curves -> Q -> data-11.a1)
- ECNF: https://purple.lmfdb.xyz/api/ec_nfcurves/?label=2.0.10691.1-1.0.1-a1 (bread is database -> ec_nfcurves)

In the second case the bread does not show the path that was taken to reach the page, and its confusing that the URLs are so different, I think it would be better to use the same type of URL in both cases.  In the first case I think the bread should include the object's home page, e.g. Elliptic curves  -> Q -> 11 -> a -> 1 -> data (and maybe the URL should also look like this?

AndrewVSutherland:
I think Elliptic curves -> Q -> 11.a1 -> data would also be fine.  Also, on https://purple.lmfdb.xyz/EllipticCurve/Q/data/11.a1 I think the list of rows in ec_torsion_growth should be a bullet list not a numbered list (being the 3rd entry in the list really doesn't mean anything).

roed314:
Alright, here's what I'm going to do:
* Add schema links for each table
* Change the ordered list to an unordered list
* Never use `API.api_query`; instead always make a new data route
* Change the bread so that there's a single link back to the object, e.g. `Elliptic Curves -> Q -> 11.a1 -> data`.

I'm going to work on group computations for a bit (since I want to get them started running), but will try to come back to this soon.

roed314:
Ok, I think I've implemented the changes requested by @AndrewVSutherland and @JohnCremona.

JohnCremona:
Thanks, @roed314 I think you have done everything!

edgarcosta:
This looks good to me.
We can always tweak things further later on.

Issue #3946, opened 2020-07-31
Acknowledgements/citation knowl


jvoight:
Create a new "Acknowledgements/Citations" knowl which deletes the "data computed by" at the bottom of pages.  This new knowl can overlap the "Source" knowls, but it would be intended to describe contributions made that aren't really about the "data source" (to include e.g. student contributions).  Perhaps each section of the LMFDB could say how to cite it, rather than citing the whole database?  The Magma handbook gives a possible example of how to be more detailed/specific in how we give acknowledgements.

AndrewVSutherland:
We added acknowledgments knowls, but I think we should also add an optional "how to cite" knowl for each section that would be displayed after the acknowledgments in the source pages and also be listed on https://www.lmfdb.org/citation (with the section identified).

Also, we should add a prominent "How to cite" link to the front page (in addition to the citation link that appears in the footer on every page but is easy for people to miss).

BarinderBanwait:
I'm working on this ... as discussed with Drew I'll add a `multi.html` template which takes a list of kids, and replace instances of `double.html` with `multi.html`; and create lots of placeholder knowls.

PR #4979, opened 2022-01-30
Lmfdb#3946 cite each section separately

BarinderBanwait:
This addresses #3946.

1. New 'How to cite' knowls for each section of the LMFDB which had existing "rcs.source" and "rcs.ack" knowls. These are currently empty, when content is added to them they will show up on the "Source and acknowledgements" pages as well as the "Citation" page
2. A new "Citation" link in the top banner of every page. This was discussed last week with Sam Schiavone and John Voight.

### DIFF
--- a/lmfdb/abvar/fq/main.py
+++ b/lmfdb/abvar/fq/main.py
@@ -786,9 +786,10 @@ def how_computed_page():
     t = "Source and acknowledgments for Weil polynomial data"
     bread = get_bread(("Source", " "))
     return render_template(
-        "double.html",
-        kid="rcs.source.av.fq",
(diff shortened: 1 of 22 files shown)


AndrewVSutherland:
@BarinderBanwait This looks like exactly what was proposed in #3946, but after reviewing it and discussing it with others I think we would prefer to not have all the citation knowls show up on the citing the LMFDB page (it's going to be too much), can I ask you to please remove the links to the rcs.cite.* knowls from citation.html?

BarinderBanwait:
@AndrewVSutherland Sure, latest commit removes those knowls.

PR #4991, opened 2022-02-03
Sort everything

roed314:
Addresses #3109.  Try it out on [purple](https://purple.lmfdb.xyz).


edgarcosta:
I have pushed to purple the latest updates.
This looks good to me. We can tweak it further, but I'm happy to merge it.

edgarcosta:
A tweak I would do is to reversethe order for ST group in genus 2 curves.

roed314:
I've reversed the order.  Note that it's still by label, so `D_{3,2}` and `D_{2,1}` come between `J(C_4)` and `J(C_2)` for example.

AndrewVSutherland:
I think it is confusing that in cases where we had search options before we provide (asc) and (dec) options but for the new ones we do not.  If we don't want to provide (asc) and (dec) options for everything, then I might remove the (asc) and only show (dec) for the exceptional cases where we decided to provide a descending search.  But I wonder if it would be simpler to just automatically provide (asc) and (dec) options for everything (and do this in the code rather than having to remember to add both).

JohnCremona:
+1 for asc and desc options for everything.  WOuld that double the number of entries in the drop-down list?

Pedantry: "inc" and "dec" would be short for "increasing" and "decreasing". 
"asc" and "desc" ... "ascending" and "descending".
But "asc" and "dec"?|

Are there up-arrow and down-arrow symbols which we could use instead?   &8593;  &8595; ?

AndrewVSutherland:
It would double the number of options, but maybe a better idea would be to add a sort direction selector? (so rather than choosing  "Sort order" you would choose "Sort column" and "Sort direction" (or something similar), but @roed314 should chime in on the feasibility of this.  +1 to using &8593; &8595;  rather than (asc) and (dec) if we stick with the current layout.

roed314:
@JohnCremona Edgar suggested adding a column to elliptic curves over number fields with the root analytic conductor, since that's a natural thing to sort on.  I've added the data for existing curves, and here's a formula to add to your upload scripts:
```
root_analytic_conductor = sqrt(0.00798504020212804*conductor_norm^(1/degree)*abs_disc)
```

AndrewVSutherland:
@roed314 I like the new sort selector, but here are some suggestions for improvement. 

- Only show the split button when the drop down list of columns is open, when closed show the sort column with a direction triangle appended to it (this would make it look less busy and let you see more of the column name).
- Make the drop down list of columns the width of the entire button (like the column selector).
- I think we should change the appearance of the search again button when the results don't reflect the current settings (if you change the sort order but don't click search again it's easy to get confused, especially if you switch to another tab and then back).  I would suggest making the default color of the search button the same color as the selectors (lighter blue) when it is in an inactive state (i.e. pressing it will just reload the search results currently shown) and then making the search button the color it is now (darker blue) whenever it is in an active state (pressing it will apply new search/sort criteria).

The third item could be done in a separate PR -- it would mean adding some javascript that serializes the current input state on page load and then checks for changes on each "onchange" event (excluding the column selector).  There is code to do this in psetpartners, see the [checkChanges](https://github.com/AndrewVSutherland/psetpartners/blob/c0a1f3cb5ae1586e61756261f3d7cd1a4f0df834/psetpartners/templates/student.html#L1058) function.

JohnCremona:
> @JohnCremona Edgar suggested adding a column to elliptic curves over number fields with the root analytic conductor, since that's a natural thing to sort on. I've added the data for existing curves, and here's a formula to add to your upload scripts:
> 
> ```
> root_analytic_conductor = sqrt(0.00798504020212804*conductor_norm^(1/degree)*abs_disc)
> ```

What is that constant and where does it come from? I assume that there or will be a knowl to explain this definition?

AndrewVSutherland:
@JohnCremona See https://www.lmfdb.org/knowledge/show/lfunction.root_analytic_conductor.  The leading constant comes from the gamma factor at infinity that appears in https://www.lmfdb.org/knowledge/show/lfunction.analytic_conductor.

AndrewVSutherland:
> I'm not sure how to implement this in javascript, and how you would actually edit the direction if the dropdown is only visible while you have another dropdown open. I totally agree that the current split solution is busy and ugly. The benefit for me was that it didn't require writing javascript. If we're going to add custom javascript to the solution, maybe a better option would be to echo the style of the column selecter dropdown and have an up/down arrow appear to the left when clicking on an option (clicking again would reverse it).

I like this idea, this is better than what I was suggesting.
 
> > * Make the drop down list of columns the width of the entire button (like the column selector).
> 
> I'm not sure what you mean by this....

It will be taken care of if you implement the idea above.

roed314:
Here's a list of text columns that we would need to split up into numerical columns (or possibly single columns consisting of a list of integers) in order to have proper sorting.
* `artin_reps`: `Container`
* `belyi_galmaps_fixed`: `base_field_label`
* `ec_curvedata`: `Ciso`
* `g2c_curves`: `st_label`
* `gps_groups`: `center_label`, `commutator_label`, `central_quotient`, `abelian_quotient`
* `gps_st`: `st0_label`
* `hgcwa_passports`: `group`
Similarly, in `ecnf` the `cm` column is strange to sort on because we use the sign to indicate actual vs potential CM.

There are also various points where we have numerical columns for most of the label and just use it as a tiebreaker when the other values are the same (e.g. for `ec_nfcurves`, we have `degree`, `signature` and `abs_disc` and just use `field_label` as a tiebreak).  I think that these are okay, since in such cases the missing part of the label usually just a tiebreaker anyway.

For which of these columns do we think it's worth the space and effort (which includes changing the upload scripts) to include them?

I'll also collect here which of @AndrewVSutherland's suggestions I haven't implemented (these are all scattered in comments on the issue #3109):
* In `CMF`, I wasn't able to add a sort on `CM/RM`.
* In `ECQ`, I couldn't find a column for cyclic isogeny degree.

Finally, we should make a decision on how to handle the sorting for mu(N) in Sato-Tate groups.

edgarcosta:
I have resized the boxes, so it the direction box takes less space:
Screen Shot 2022-02-05 at 16 04 13

JohnCremona:
> Here's a list of text columns that we would need to split up into numerical columns (or possibly single columns consisting of a list of integers) in order to have proper sorting.

> * `ec_curvedata`: `Ciso`

> 
> For which of these columns do we think it's worth the space and effort (which includes changing the upload scripts) to include them?

I would say no to ec_curvedata:Ciso.  When I added the numerified iso label column I decided that doing it for the LMFDB labels was enough. It could of course be added.

> 
> * In `CMF`, I wasn't able to add a sort on `CM/RM`.
> * In `ECQ`, I couldn't find a column for cyclic isogeny degree.

There is a column isogeny_degrees f type smallint[] which is a list of all cyclic isogeny degrees in the class (with some well-defined convention for CM curves which would otherwise have an infinite list), so this is not a sensible sort column.  You could have a column with the maximum such degree, which is also the LCM of the degrees, and sort by that, but it would need adding.

JohnCremona:
Two things I note on purple right now:
   - Sort order buttons: the second smaller button to the right is possibly too small, as I only see half or less of the current arrow, so until I click on that button it is not very clear which way it is pointing.  (chromium on ubuntu on a large screen).
   - If I choose to sort on a column not currently displayed (e.g. for EC/Q, Analytic Sha), it would be nice if that column was automatically selected to be displayed.

edgarcosta:
@JohnCremona, the second box being too small should be fixed now. I checked it on firefox and chrome.

roed314:
@JohnCremona I'll try to work on having the sort order affect the columns displayed today, as well as implementing the javascript suggestion I made to Drew so that we can compare with the double select.

roed314:
There aren't many cases where we want to sort in descending order by default, but there's at least one: first zero of an L-function.  With the current setup, it's easy to make this descending by default, but the arrow would be in the opposite direction.  Is this okay, or should we add infrastructure to support changing the arrow direction?

AndrewVSutherland:
I'm not sure I understand the question, can you point to an example?

roed314:
Take a look at https://beta.lmfdb.org/L/rational?hst=List&search_type=List&sort_order=zero.  It's sorted by first zero, but descending.  With the plan currently, if you choose to sort on first zero, it would sort ascending by default.  Is that okay?

edgarcosta:
that is okay

AndrewVSutherland:
I think that's fine, it's better to be consistent.

roed314:
Ok, I added some javascript so that the sort direction is now toggleable.  Please play around with it: I don't have full confidence in my javascript abilities.  One known issue is that it pushes the results table down as the column selecter used to.  @edgarcosta  solved that issue in the other case by using `position: absolute;`, but if you use absolute positioning on both select boxes they overlap with each other.  I'm sure there's a way to solve this (maybe using the labels to get the spacing right?), but I wanted to get feedback before putting too much time into figuring it out.

I have also added code for showing the column referred to by a given sort order, as suggested by @JohnCremona.  The mechanism is just whether the `SearchCol` entry has the same name as the `sorts` entry.

I propose postponing to another PR the data changes required for proper sorting on the columns the currently rely on text.  I can leave them in or comment them out, as desired.

roed314:
These sort orders reveal some problems with how some columns are handling null values.  For example, consider https://purple.lmfdb.xyz/ModularForm/GL2/Q/holomorphic/?hst=List&search_type=List&sort_order=hecke_ring_index_factorization&sort_dir=op

This should be showing the highest coefficient ring indexes, but they're all 1!  The problem is that they're actually null, and the factorization constructor is somehow producing "1" from that....

roed314:
It's also a bit frustrating to try to use reverse sort to find the "largest" example of some column, only to run into a bunch of cases where it hasn't been computed.  See https://purple.lmfdb.xyz/EllipticCurve/Q/?hst=List&search_type=List&sort_order=degree&sort_dir=op for example.  There may be a way around this with "NULLS LAST" in the sort order, but that will require some changes to the backend.

roed314:
Another data problem revealed, which I just fixed.  A minute ago, the following held:
```
sage: db.nf_fields.count({"label":None})                                                                                                          
1
```
There was no data in the row, so I deleted it.

roed314:
Number fields with [biggest discriminant](https://purple.lmfdb.xyz/NumberField/?hst=List&search_type=List&sort_order=disc&sort_dir=op) is kind of fun.

JohnCremona:
I was trying out sorting elliptic curves by CM discriminant, in both directions.  For ECNF the behaviour is unexpected, I think because we use  a hack whereby when the CM is actual (as opposed to potential) we stors the absolute value of the CM discriminant.  So the sort order puts all these together, the 0's (for no CM) in the middle and the potentials at the end, or ditto in reverse.

Perhaps this is an indication that it is time to add a column for potential/actual CM and leave all the doscriminants as negative?

roed314:
Yeah, this is the issue I was referring to a few days ago.  I think that if we want to sort by CM discriminant, adding a new boolean column for potential/actual CM is the way to go.

JohnCremona:
> Yeah, this is the issue I was referring to a few days ago. I think that if we want to sort by CM discriminant, adding a new boolean column for potential/actual CM is the way to go.

I'll make a new issue for that, since as well as the data change there will also be code changes.  We can start with just adding the new column.  (It is only an issue for imaginary quadratic fields of class number 1, since those are the only fields we have curves for which are not totally real, and you can only have actual CM when the field contains the Hilbert class field of an IQF.)

AndrewVSutherland:
When sorting on a column that may contain unknown values I'm wondering if it would be better to modify the query to exclude these rows from the search, with a warning to the user that only rows where the value is known are included (this would require us determining and noting which to which columns this applies, possibly this should be stored in the meta data for the table).

We could also display such a warning when the user specifies search criteria for the column.  I think this would be consistent with current behavior, for example, if I search for newforms with coefficient ring index >= 1, I get 207711 results, but of course all 281965 newforms in the database have coefficient ring index >= 1.

I also note that None is not always used to denote an unknown value, for booleans often -1 is used (the column is actually a short integer with values -1,0,1 for unknown, true, false), and these really should be handled the same way (or converted to booleans that can take values None,False,True).

Independent of this, the code that displays a factored integer should return the empty string on the input None, not "1".

AndrewVSutherland:
On a positive note, the fact that you are turning up all these data issues is a good thing!

AndrewVSutherland:
I really like the new sort order toggling interface!  Two minor UI tweaks to make (in addition to fixing the push-down problem you already noted):

- You should add a little more space (maybe one thinspace) to the sort columns that are unselected so the line up better with the selected option.
- Get rid of the useless scroll bar on the right (it should look like the column selector), I think this involves setting the size to the right value (possibly just one larger than you are using).

JohnCremona:
I also think that the new interface is great.  We'll sort out (ha ha) the data issues which have arisen soon enough, and will learn some lessons about table and schema design as we do so!

roed314:
@edgarcosta Thanks, that works for dealing with absolute positioning.  Do you know how to get rid of the useless scrollbar Drew noted?

edgarcosta:
```
select{
    scrollbar-width: none; /*For Firefox*/;
    -ms-overflow-style: none;  /*For Internet Explorer 10+*/;
}

select:-webkit-scrollbar { /*For WebKit Browsers*/
    width: 0;
}

**Comment by roed314** [comment]
It looks like
```
#sort-selecter {
  overflow-y: auto;
}
```
also works (I found this for `#column-selecter` in `style.css` just now).

edgarcosta:
On Firefox, I still see the sidebar with `overflow-y:auto`

roed314:
I fixed the scroll issue and fiddled with the space a bit (it's still not quite optimal on Firefox, but I don't want to spend any more time on it).

I've created followup issues for Sato-Tate searches, sorting on text columns, omitting nulls when sorting on that column, not displaying 1 when a factorization is NULL, and changing the color of the search button if results need to be updated.

Does anyone have more issues they'd like to see addressed before merging this PR?

AndrewVSutherland:
Sato-Tate group tests are going to fail until #5019 is merged and you merge upstream

roed314:
I merged #5019 into master and then master into this PR.

AndrewVSutherland:
@roed314 if you merge master into this PR tests should pass.

roed314:
Tests pass.  @AndrewVSutherland, do you want to merge this before or after the outstanding PRs to dev and prod?

AndrewVSutherland:
@roed314 Sorry I've been tied up with teaching.  I'll look at this and plan to merge it tonight.

roed314:
No worries.  I've been working on other stuff too.

AndrewVSutherland:
This looks really good.  I noticed a few cosmetic things, which I list below so I won't forget, but I'd be happy to merge this and address those in a follow up PR if you don't want to deal with them now.

-We currently are capitalizing column names in the column selector but not in the sort order selector; given that we generally don't capitalize options in selectors, I think we should probably switch the column names in the column selector to lower case (so nothing to do with this PR really, I'm just noting it).

- In groups I think "num. conj. classes" and "num. subgroup classes" should be "conjugacy classes" and "subgroup classes" (just as we write "rational points" not "num. rational points"), and similarly in Galois groups (for num. conj. classes).  I think its clear from context you won't actually be sorting on conjugacy classes (whatever that would mean).

- I would change "number ramified primes" to "ramified prime count" in Artin reps (to match the input caption); here I would not write "ramified primes" since you might actually want to sort on the list of ramified primes!

roed314:
I should be able to fix these later tonight, then merge once tests pass.

roed314:
Test failure was spurious.  I'm merging to master and will create another PR for merging to dev.

Issue #4982, opened 2022-01-31
Labeling of plus/minus Iwasawa invariants for supersingular elliptic curves 


JohnCremona:
From the bug report:

"For an elliptic curve with good supersingular reduction at a prime p, LMFDB lists both plus/minus Iwasawa invariants but it apparently does not specify which is which. For example, at the URL provided above, the curve 43a1 has lambda-invarinats 1 and 3 at p=7,  though it is not explicitly stated which one is the plus-invariant and which is the minus. (I believe these invariants are ordered "plus, minus", but it might be good to state this outright.)

I'll double check with Pollack who computed this data and if necessary make the web page clearer.

JohnCremona:
Pending that, the simplest thing to do would be to add to the knowl under "\lambda-invarinta" and "\mu-invariants" to say that the + inariant is listed first, but that departs from our conevntion that knowls are context-free.  The alternatice is to add to the displayed pair of numbers so for example at https://www.lmfdb.org/EllipticCurve/Q/43a1/ the \lambda-invariants at p=7 could be shown as +:1, -:3.   This is only for supersingular primes, and the signs could be omitted when the invariants are equal.

JohnCremona:
Rob Pollack has confirmed that in the case of supersingular primes the invariants are always given with the + one first.

JohnCremona:
Fixed in PR#4992

PR #4992, opened 2022-02-03
fix for #4982 (ss Iwasawa invariants)

JohnCremona:
This adds "+,-" under the table header when that is "ss" (for supersingular), to make it clear that in the pairs of numbers below, the first is the lambda^+ (or mu^+) and the second is the -.

For example: http://localhost:37777/EllipticCurve/Q/43a1/

There may be a better way of making the alignment work: the table headers are left-justified which does not look right here as that puts the + under ss, so I cheated by inserting a hard space before the ss.  I'm sure that it could have been done using a custom css style, and I am happy to do it that way if someone else tells me what to do.

### DIFF
--- a/lmfdb/elliptic_curves/templates/ec-curve.html
+++ b/lmfdb/elliptic_curves/templates/ec-curve.html
@@ -495,7 +495,7 @@ $p$-adic regulators are not yet computed for curves that are not $\Gamma_0$-opti
 <tr style="border-bottom: 1px solid #000;">
 <th>$p$</th>
 {% for pdata in data.iw.data %}
-<td>
+<td align=center>
 {{pdata[0]}}
 </td>
 {% endfor %}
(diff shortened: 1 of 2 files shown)


AndrewVSutherland:
Looks good, but I have three suggestions for improving the presentation of the table of reduction types:

- I would center the lambda and mu invariant columns
- I would abbreviate "ordinary" as "ord" to save space (or alternatively, use "supersingular" rather than "ss")
- In the knowl https://beta.lmfdb.org/knowledge/show/ec.reduction_type I would change "good ordinary" to "ordinary (ord)" and "good supersingular" to "supersingular (ss)" (if you use the abbreviations).  In particular I would drop the adjective "good" (the knowl already makes clear that "ordinary" and "supersingular" are subcases of good reduction)

JohnCremona:
Thanks, I have dealt with all three points.  I edited a couple of extra knowls too, but these are in a bit of mess:  ec.reduction_type makes ec.q.reduction_type redundant, and the latter could but does not link to the separate knowls for ordinary and supersingular reduction, though these are linked to from the Iwasawa invariants column.  Tidying these up would not require any more code changes, however we could make the knowls for the column headings "ss", "ord", "nonsplit" etc be the specific knowls rather than the one which defines all the types.

I think that last one is worth doing, so will add a commit to the PR in a minute.

AndrewVSutherland:
The test failure was caused by a G2C data change I just made (adding more 2-torsion fields) -- there is a data dependent test that needs to be changed, I'll take care of that in a separate PR

JohnCremona:
OK, ready for re-review: now the column headings "ord", "ss" etc are specific knowls while the column heading "reduction type" is as it was, to the general knowl.

PR #4995, opened 2022-02-04
fix mw test to account for data change

AndrewVSutherland:
Removed a test that would succeed only if the two torsion field is not in the database (which it now is).


AndrewVSutherland:
I'll merge this as soon as tests pass

Issue #4270, opened 2020-12-12
Hecke cutters display


edgarcosta:
There is something weird going on with the display of the Hecke cutters for https://beta.lmfdb.org/ModularForm/GL2/Q/holomorphic/8449/2/a/h/

AndrewVSutherland:
I think this will be addressed by #4394, but let's leave this issue open for them moment.

roed314:
This is resolved by Edgar's changes in #4394.  Assigning this issue to Edgar since that isn't yet merged.

PR #4999, opened 2022-02-05
Enabling raw/typeset view in classical modular forms

### LINKED ISSUE #4394: Badly formatted polynomial knowls
On many of the CMF pages (and possibly elsewhere) knowls for large polynomals garble the display.  For example, click on the knowl "Q[x]/(...)" for the cofficient field on the page

http://www.lmfdb.org/ModularForm/GL2/Q/holomorphic/1/118/a/a/

If katex can't handle these large polynomials we should just format them as raw text, and in fact they will probably look better as raw text in situations like this where the coefficients are wider than the knowl (just let the text wrap).

Here is a bold suggestion that wouldn't look as nice but would be much more useful: always display polynomial knowls as raw text, e.g. "x^2 + 3*x + 1", including the *'s so that you can cut and paste them -- there are a lot of situations where it is really annoying not to be able to copy a polynomial you want that you can see clearly right on the LMFDB page.

In fact, I could see an argument to making all polynomials knowls that open with raw text when you click on them, or even better, make it so you can right click on them and copy the raw text to the clip board.

edgarcosta:
This resolves #4394 and #4270 

In particular, I have applied the raw/typeset view to:
- defining polynomial (inside and outside the know) [1.118.a.a]
- the integral basis for eigenvalues,  [1.118.a.a]
- Hecke cutter [8449.2.a.h]
- Hecke characteristic polynomial [1.118.a.a]

[1.118.a.a]: https://tori.lmfdb.xyz/ModularForm/GL2/Q/holomorphic/1/118/a/a/
[8449.2.a.h]: https://tori.lmfdb.xyz/ModularForm/GL2/Q/holomorphic/8449/2/a/h/
This is available on https://tori.lmfdb.xyz

A way to generate potential interesting examples is via:
https://tori.lmfdb.xyz/ModularForm/GL2/Q/holomorphic/?dim=16-20&search_type=Random

Suggestions welcome on how to improve the CSS of the `x->$x$` and `$x$->x` buttons.

@jwj61, I rewrote your javascript and tried to better style the button, I would appreciate your feedback.


AndrewVSutherland:
See https://tori.lmfdb.xyz/ModularForm/GL2/Q/holomorphic/820/2/cc/d/ for another example where the latex on the exponents on factored Hecke polys has a problem (multi-digit exponents need {})

AndrewVSutherland:
I might suggest changing the wording in the header of the Hecke polynomial table so that the polynomial appears at the end (as it stands there is an annoying gap where the x->x button appears, see https://tori.lmfdb.xyz/ModularForm/GL2/Q/holomorphic/3966/2/a/p/, for example.

I would change it to say something like "This newform subspace can be constructed as the kernel of the action on blah of the linear operator defined by blah" (or something similar).

Other than that and the latex issue noted above, I think these look really good!

AndrewVSutherland:
The test failure is in test_hecke_charpolys at lmfdb/classical_modular_forms/test_cmf2.py:394.

edgarcosta:
Everything that you pointed has been fixed.

edgarcosta:
I missed some corner cases
https://tori.lmfdb.xyz/ModularForm/GL2/Q/holomorphic/820/2/cc/d/
https://tori.lmfdb.xyz/ModularForm/GL2/Q/holomorphic/10/3/c/a/

I will reopen this when I solve those

edgarcosta:
The previous issues have been fixed

edgarcosta:
I fixed the error that we were hitting when testing random forms: https://gist.github.com/edgarcosta/1ae2fade5f743fb57723f8c8c359d1e5

But I couldn't find the form that we failed on, as it doesn't make sense to display basis in terms of zeta_20...

Issue #5012, opened 2022-02-08
Sato-Tate search order


roed314:
With #4991, it's now possible to change the sort order for Sato-Tate group search results.  As discussed on #3109, in some sort orders the mu(N) should no longer appear first, but they do because it's hard coded.  There are several possible solutions:
* Add all the mu/nu groups for n <= 10^6 (we can still show the Sato-Tate groups on dirichlet character pages of modulus above that, since we can keep the code for displaying the mu(N) homepages even if they don't show up in search results). It would be less than 1GB data.  Switch from using special purpose code for Sato-Tate searches and use the standard `search_wrap` instead.
* Just add mu(N) for N that are small enough that they can be mixed in with some sort order (the component group order comes to mind, but there are possibly others). When irrational ST-groups are omitted, there will be no special case handling required; when they are, either the mus all come first (in which case we display the ones from the database and then endlessly include the others), or the extra mus all come at the end.  We won't be able to use `search_wrap` with this solution.
* For each of the different search orders, hard code how the mu(N) interleave with the database search results.  This doesn't seem like a good plan....

AndrewVSutherland:
I think we should go with the first option.  If we have learned anything over the years it is that in the long run it is better to just put everything in the database and treat it all consistently.

AndrewVSutherland:
@roed314 Does it make sense to make this change before merging the sort order PR?

roed314:
It depends on how much https://purple.lmfdb.xyz/SatoTateGroup/?hst=List&search_type=List&sort_order=trace_zero_density&sort_dir=op bothers you. 😄

AndrewVSutherland:
Okay, I'll do this now (but you might have to deal with some minor merge conflicts).

AndrewVSutherland:
This has been done (and is now in production).

PR #5019, opened 2022-02-09
ST groups in database for search results

AndrewVSutherland:
This PR is the first step in addressing #5012.  At the moment only the two rational weight 0 groups are in the database -- I don't want to add all the others yet because it will break beta even worse than it already is (right now 0.1.1 and 0.1.2 show up twice).  Once this PR is merged I will add the first 100,000 mu groups (and when char_orbits is extended to 1,000,000 I will extend the mu groups as well).

It will still happily render pages and knowls for groups that are not in the database as it did before, but with this PR the only groups that will appear in the search results are those in the database.

I also added a label_components column that can be used for sorting if we wish to.

### DIFF
--- a/lmfdb/sato_tate_groups/main.py
+++ b/lmfdb/sato_tate_groups/main.py
@@ -1,19 +1,16 @@
 # -*- coding: utf-8 -*-
-import itertools
 import re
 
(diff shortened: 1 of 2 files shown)


roed314:
Looks good to me!

Issue #5004, opened 2022-02-08
Change the way potential/actual CM are stored for ECNF


JohnCremona:
Currently th ECNF table (ec_nfcurves) has one column 'cm' which holds 0 for non-cm curves, a negative integer -D for curves with potential CM by the order of discriminant -D, and a positive integer +D for curves with actual CM by the order with discriminant -D.  For the last possibility to occur, the base field must contain the class field associated to discriminant -D and in particular contain sqrt{-D}, so we only see it over imaginary quadratic fields of class number 1 and one of the 13 possible D's (since all the other fields over which we have elliptic curves are totally real, or 3.1.23.1).

This makes sorting search results by CM discrimimnant problematic (see #4991).  We should change the convention:
   1. add a new column with values 0,+1,-1 for no CM, actual CM, potential CM
   2.  change the 'cm' column back to always being 0 or negative
   3. adapt the code to deal with this properly.

I think that it will be possible to do these in the order 1,3,2, and starting with 1 is certainly easy as the new column is just the sign of the 'cm' column.

JohnCremona:
Item 1 has been done, on beta there is now a new column 'cm_type' with values 0,1,-1 equial to the current sign of the 'cm' column.

AndrewVSutherland:
Let me know when the data is ready to be copied to prod.

JohnCremona:
> Let me know when the data is ready to be copied to prod.

OK -- it will not be until after a PR.  I hope I can get the code working both with the data as it is now and and as it will be (with no positive entries in the 'cm' column).

I'm also ready to add more rows to this table, as soon as I can get the correct Galois images for them :) (I also need to add the 'cm_type' column and root analytic conductor columns).

AndrewVSutherland:
Sorry for being slow on the Galois images (we had our Simons collaboration meeting Friday so I didn't have much time to work on it).  I will get it done this week (today if I can).

JohnCremona:
After #5005 is reviewed and merged and on beta (step 3), I can implement step 2. Then a second small PR to remove two redundant lines of code before pushing to prod.  It would be OK to copy ec_nfcurves to prod now, but we might as well wait until all these steps are completed.

AndrewVSutherland:
ec_nfcurves has been copied and #5005 has been merged.

JohnCremona:
Good!  I'll now do the second stage of data adjustment, putting back the - signs in the 'cm' column.

[...]

PR #5020, opened 2022-02-09
final cm code fixes for #5004

JohnCremona:
Two tiny changes, now that (on the beta database) the 'cm' column of ec_nfcurves now always contains 0 or the actual negative discriminant, as the 'cm_status' column holds the sign of the old 'cm' column (+1 for actual CM and -1 for potential).  Specifically, 
   1. in processing the search for specific discriminants we no longer need to append the negatives of the ones requested;
   2. in preparing the web page data we no longer have to replace positive 'cm' value by  their negatives
It will be safe to push this to prod *after* ec_nfcurves has been copied over, and at that opint#5004 can be closed.

### DIFF
--- a/lmfdb/ecnf/WebEllipticCurve.py
+++ b/lmfdb/ecnf/WebEllipticCurve.py
@@ -368,7 +368,6 @@ class ECNF(object):
         self.End = r"\(\Z\)"
         self.rational_cm = self.cm_type>0
         if self.cm:
-            self.cm = -abs(self.cm) # this line can be deleted when we no longer store abs(-D) for rational CM by -D
             self.cm_sqf = integer_squarefree_part(ZZ(self.cm))
             self.cm_bool = r"yes (\(%s\))" % self.cm
             if self.cm % 4 == 0:
--- a/lmfdb/ecnf/main.py
+++ b/lmfdb/ecnf/main.py
@@ -473,9 +473,6 @@ def make_cm_query(cm_disc_str):
     for d in cm_list:
         if not ((d < 0) and (d % 4 in [0,1])):
             raise ValueError("CM discriminants are negative and congruent to 0 or 1 mod 4.")
-    # the next line is because actual CM with disc -D is stored as +D in the table;
-    # it can be removed once we only store -D itself.
-    cm_list += [-el for el in cm_list]
     return cm_list
 
 @search_parser


roed314:
This looks like it may be a real test failure: `/EllipticCurve/?cm_disc=-11&include_cm=PCMnoCM` should include `14641.1-a1`....

JohnCremona:
> This looks like it may be a real test failure: `/EllipticCurve/?cm_disc=-11&include_cm=PCMnoCM` should include `14641.1-a1`....

Thanks for spotting this, I will look into it.

JohnCremona:
> /EllipticCurve/?cm_disc=-11&include_cm=PCMnoCM

14641.1-a1 over which field?  I do see curves with that label over 2.0.19.1, but the sort order hides them.  If you add conductor norm = 14641 to the search criteria you see that there are 12 matches.

AndrewVSutherland:
I wouldn't worry about errors on devmirror (I haven't copied ec_nfcurves over yet).  But I do think there is a problem with https://beta.lmfdb.xyz/EllipticCurve/?start=50&hst=List&cm_disc=-11&include_cm=PCMnoCM which is including curves with CM (not just PCM), and this is not fixed by this PR, I see the same thing on http://127.0.0.1:37777/EllipticCurve/?cm_disc=-11&include_cm=PCMnoCM

If I look at https://beta.lmfdb.xyz/EllipticCurve/data/2.0.11.1-24025.7-CMa1 (which has CM not just PCM) I see that `cm_type` is set to 1 which I think is correct based on what you wrote in https://github.com/LMFDB/lmfdb/issues/5004#issue-1127010400 (I just updated the schema doc to reflect this).  I don't see a column named `cm_status`, did you mean `cm_type` in what you wrote above?

So I think the search is not filtering on `cm_type` as it should, and indeed if you look at https://github.com/LMFDB/lmfdb/blob/903b104b3b33877b194c29d1231e42d376e0bcc1/lmfdb/ecnf/main.py#L616 you will see that the logic there needs to change.

AndrewVSutherland:
@JohnCremona I'm going to go ahead and merge this and fix the problem I noted above in a separate PR, I really want to get back to a state where tests are passing (I'm busy breaking sato_groups and its dangerous to have things in such a confused state).

JohnCremona:
Fair enough -- I'll work out what is wrong (and yes I did mean 'cm_type').

AndrewVSutherland:
I'm not convinced anything is actually wrong, but let's see if the PR I just submitted now passes the test that was failing (possibly the failure was simply that by including extra results it was pushing the result that the test was looking for onto the next page.

JohnCremona:
I have fixed it and will make a new PR.  You are right that I had not revised the 'include_cm' part of the logic, which is now a lot simpler than it was (thought there were several unnecesarily repeated lines already, in fact).   I'll make the PR as soon as I have checked locally that the relevant test pass.  I also updated the schema description for 'cm' which will need review.

JohnCremona:
@AndrewVSutherland See my comment on your PR!  I did not see that you had already done this, with almost the identical changes.

PR #5022, opened 2022-02-09
update include_cm search to use cm_type

AndrewVSutherland:


JohnCremona:
Thanks @AndrewVSutherland I was about to make a PR with almost the same changes, and now I don't have to. I agree with the new code here, so this can be merged when tests pass.

PR #5024, opened 2022-02-09
reenable irrational tests and fix bug in rational/irrational filtering

AndrewVSutherland:


AndrewVSutherland:
I've now added the irrational ST groups 0.1.n for 3 <= n <= 10000.  I think 10000 is sufficient for searching (I don't see any incremental benefit to going to 100000 or even 1000000, the links from the Dirichlet pages will all work regardless, up to 10^20, and I'd rather not bother changing the type of the components column from smallint to integer).  I have now reenabled the tests that involve searches that include irrational ST groups.

This should fix the test failures that were showing up when I tried to push to dev (there was a bug in the rational/irrational filtering).  I will merge this as soon as tests pass.

PR #5025, opened 2022-02-09
master --> dev

AndrewVSutherland:


AndrewVSutherland:
The ST test failure on proddb is expected (I haven't copied the updated table over yet and won't until this PR is pushed to web), the test is complaining about not finding irrational groups in search results that are not in the database, which is fine.

PR #5032, opened 2022-02-12
Uniformizing raw/typeset 

### LINKED ISSUE #5011: Uniformize raw/typeset view for large numbers and polynomials
We now have two different ways of allowing the user to cut and paste numbers and polynomials on LMFDB object pages, compare

[CMF page](https://beta.lmfdb.org/ModularForm/GL2/Q/holomorphic/1/118/a/a/)
[NF page](https://www.lmfdb.org/NumberField/17.17.13202363705223218603863184487755024630582383374680032800435939704158233027841788543229812635151072074455438626669432759450899735549243294921243306854586939663055181335880903211540683502950935087551803286773305471032026037990460071805978402769130092019752634101527352693814943791414905342572852579428180829030727661698750961477408713596845181881409326113271279701186934711912924535979375368089471422837282452421683506180048924876364820225210846226099372878635447796099920054239935070472348193791925328690838569369798938742618818040823435236370214291386195573062935332270993039421387687573397220106241394348907347636983408333263308765625.1)

We should uniformize these.  I think it makes sense to use the new control used on the CMF pages everywhere, both because it improves the page layout, and because it allows you to copy the data to the clipboard with one click.  But I know @jwj61 expressed some reservations about earlier versions of the control used on the CMF pages, so we should discuss this before making any changes.

There are also many pages where the numbers or polynomials are not unreasonably large but it would still be useful to have the typeset/raw button available just for the ability to cut-and-paste numbers and polynomials. It's unobtrusive enough that I could see an argument for putting it on every object's home page (but this will require updating a bunch of templates).

edgarcosta:
This partially resolves #5011, but honestly, now that we have agreed on what we want, we should create a new list of things that need to be converted.

This PR is available on https://tori.lmfdb.xyz

In short, now we:
- use `textarea` all the time
- display a copy button
- double click switches from tset to raw mode and selects it
- the buttons `$x$->x` and `x -> $x$` are 👁️ and `\not` 👁️.

I've also fixed a bug in CMFs, where we used to display `\beta_0 = 1`, even though we never made use of `\beta_0`: [old](https://beta.lmfdb.org/ModularForm/GL2/Q/holomorphic/9660/2/a/r/) vs [new](https://tori.lmfdb.xyz/ModularForm/GL2/Q/holomorphic/9660/2/a/r/)

I tried to take care of all easy in-place replacements of `raw_typeset` with `raw_typese_poly`. For example, the defining polynomial of a number field, or its knowl: [old](https://beta.lmfdb.org/NumberField/17.17.13202363705223218603863184487755024630582383374680032800435939704158233027841788543229812635151072074455438626669432759450899735549243294921243306854586939663055181335880903211540683502950935087551803286773305471032026037990460071805978402769130092019752634101527352693814943791414905342572852579428180829030727661698750961477408713596845181881409326113271279701186934711912924535979375368089471422837282452421683506180048924876364820225210846226099372878635447796099920054239935070472348193791925328690838569369798938742618818040823435236370214291386195573062935332270993039421387687573397220106241394348907347636983408333263308765625.1) vs [new](https://tori.lmfdb.xyz/NumberField/17.17.13202363705223218603863184487755024630582383374680032800435939704158233027841788543229812635151072074455438626669432759450899735549243294921243306854586939663055181335880903211540683502950935087551803286773305471032026037990460071805978402769130092019752634101527352693814943791414905342572852579428180829030727661698750961477408713596845181881409326113271279701186934711912924535979375368089471422837282452421683506180048924876364820225210846226099372878635447796099920054239935070472348193791925328690838569369798938742618818040823435236370214291386195573062935332270993039421387687573397220106241394348907347636983408333263308765625.1)

Obviously, there are plenty of things left to do, for example taking care of the integral basis, for that it should be easy to do where we just put `\cdots` when it gets too long...
I should also point out that @jwj61 also has a function called `unlatex` that converts typeset strings into raw strings.
Thus, it should be easy to use `raw_typeset` widely. However, going this way also makes it harder to compress the typeset strings, as in the case of the integral basis.


jwj61:
I should point out that unlatex expects nothing more complicated that exponents and fractions, so one should check that it can handle a situation before using it.

AndrewVSutherland:
If you change line 9 of `raw_typeset.js` to `$ta.height(rect.height - (2 + 3 + 2));` (subtract 2 more pixels) it will only show one line rather than one and a bit.

edgarcosta:
In what example do you see a line and a bit?
The current formulas keeps the outside span exactly with the same height and width:
Screen Shot 2022-02-11 at 22 28 03
Screen Shot 2022-02-11 at 22 28 17

I don't see "one and a bit", but for that perhaps the best thing is to tweak the `font-size` which is set to 95%

jwj61:
I looked at a typical number field and notice that the integral basis is not all visible in the textarea on https://tori.lmfdb.xyz/NumberField/9.1.35028793.1

edgarcosta:
@jwj61, by design, the initial size of the textarea is exactly the same as the span of the typeset display. The user may, later on, resize the text area as they wish.

jwj61:
Then I misunderstood @AndrewVSutherland 's comment earlier which I thought said that when something was shown in its entirety, the text area would show the raw form in its entirety (without scrolling or resizing).

edgarcosta:
@jwj61 that is the case in the current behavior, but then we also don't use a textarea.

edgarcosta:
I should clarify that today we decided that we discussed our options and @jvoight, @AndrewVSutherland  and I decided that we would use textarea's everywhere. From your previous comments, I assumed that you and @JohnCremona were in support of that.

@AndrewVSutherland also expressed the desire that the page doesn't move around when one switches between modes. Which I agree with, given that now are using textarea's everywhere.
I should also point, the obvious that raw mode will always take more space than the typeset mode, for things like small degree polynomials in a single roman letter variable the difference is small, but away from those cases, the sizes are quite different, e.g., the basis of [5.20.a.b](https://beta.lmfdb.org/ModularForm/GL2/Q/holomorphic/5/20/a/b/)

AndrewVSutherland:
I stand by what I said to @jwj61, we should avoid hiding information when switching to raw mode.  When the latex content is small and easily fits on one one (e.g. "x^2-17") it should be visible in its entirety in raw mode, even if that means the text area and the buttons next to it take a bit more horizontal space.  For content that is large enough to be just barely rendered on a single line in latex form, I think we should be compressing it (so you would still see more in raw mode).  To sum up, at least on a standard size screen (1920x1080) I think the following should always hold

- nothing moves when you switch to raw mode
- nothing is hidden when you switch to raw mode

I think both are easily achievable by setting the limit at which we start compressing expressions appropriately.  Another option is to shrink the raw text a bit more than the 95 percent we are currently using to gains some more horizontal space for the copy/show buttons.  I also think those buttons could be a bit smaller (which will help).

AndrewVSutherland:
One further thought, I think we should be using single character variable names in raw mode -- e.g. "b" rather than beta, and "n" or "v" rather than "nu".

AndrewVSutherland:
I would also not use underscores `_` in raw mode, I would turn `beta_2`  into `b2` (I think anyone writing magma/sage/pari code involving a variable they are thinking of as `beta_2` would likely do the same, including underscores just makes variable names harder to type -- if I'm copying raw text into magma/sage/pari I'd rather not have to type underscores when defining the polynomial ring).

jwj61:
I don't understand why it is a priority to not have other things move on the page with the click.  When the user clicks on a knowl, things move to make room for the information.  Since they click there, not only have they essentially asked for it, but their focus is on the spot of the click and not something down the page.  This seems to be an analogous situation.

AndrewVSutherland:
@jwj61 I'm thinking of the situation where the user clicks on the show button at the top, which I think is a likely use case, since the other show buttons only appear when you hover.  Its also a way to quickly find out what you can access in raw mode.

But I can see an argument for expanding completely (like a knowl) and wouldn't necessarily object to that now that you can copy without having to click the show button at all.  I also think there is an argument for keeping everything in place (as I have suggested), especially when clicking the top show button.  What I don't like is randomly sizing the text area somewhere in between those two options.

AndrewVSutherland:
@edgarcosta Hmm, possibly this is a mac o/s vs ubuntu issue?  With the change I suggested, does the first line get cut off or is it all still visible?

jwj61:
Even for the eye at the top, I don't see a problem.  The user gets what they asked for, and can quickly flip back.  I think people will pretty quickly find the ones next to individual entries as well.

Finally, the eye at the top is right next to the show commands links which force things on the page to shift down more dramatically than the eye would.

AndrewVSutherland:
@edgarcosta Do you agree with @jwj61 ?  If so, then we can forget about fiddling with extra pixels, just make the text area big enough to show everything (and I guess fix the width, as with knowls).

It might be worth trying this just to see what it looks like on the two pages I linked to on #5011.

AndrewVSutherland:
If we wanted to get fancy we could make the eye at the cycle through three states: closed, narrowly open (corresponding to the current PR), wide open (in shock at the size of the data).  I'm not asking you to implement this :)

jwj61:
Note, I am only suggesting that the text area show everything in cases where the typeset was showing everything, not when it is so massive.

edgarcosta:
@jwj61 thanks for the clarification, otherwise I was starting to wonder why use `textarea` at all

AndrewVSutherland:
Ah, then what I was proposing is not inconsistent with what @jwj61 is saying.  If the data is small enough to show everything typeset I think we should also be able to see everything in raw mode in a textarea of the same height, which would mean that things don't move.  It's just an issue of managing the width (using smaller variable names and no underscores will help).

jwj61:
I guess we will have to see how small the font will have to be.  I am sure that there are cases with integral bases or unit group generators for number fields where the typeset form is not crazy to show in full (which to me is the right threshold), but runs for multiple lines.  When converted to raw mode, the length difference might either require more lines or much smaller fonts.  But, we can cross that bridge when we get to it.

edgarcosta:
I have fixed the double-click bug.

I have also changed the behavior so that:
- when the input is not compressed, we set `cols` and `max-width`  to ensure that everything is displayed, and we increase height based on the scroll position, ie, to show everything.
- when the input is compressed, we keep the previous behavior. It tries to keep the same size as the typeset display. For formulas that wrap over more than one line, changing them into raw format will most likely induce a different wrapping.

Here is a page where you can see the different behaviors.

If this is not what you both want, I think you should try to sort out what you want and write out the specifics together.

edgarcosta:
I didn't address
> in cases where the compressed text is really compressed, making the text area big enough so the user can see that there is really a lot there

One partial way to address that is to increase `min-width` and add `word-break: break-all;`

AndrewVSutherland:
@edgarcosta Thanks for the fixes, I see the double-click is working now.  I think increasing `min-width` and adding `word-break: break-all;` both make sense, I would go ahead and do that, and also add a `min-height` parameter that would be used when the raw text is not completely visible).

I had a  question about https://tori.lmfdb.xyz/ModularForm/GL2/Q/holomorphic/1775/1/p/a/, why is `zeta` being changed to `b` rather than `z`?

I also have some questions about this [NF page](https://tori.lmfdb.xyz//NumberField/17.17.13202363705223218603863184487755024630582383374680032800435939704158233027841788543229812635151072074455438626669432759450899735549243294921243306854586939663055181335880903211540683502950935087551803286773305471032026037990460071805978402769130092019752634101527352693814943791414905342572852579428180829030727661698750961477408713596845181881409326113271279701186934711912924535979375368089471422837282452421683506180048924876364820225210846226099372878635447796099920054239935070472348193791925328690838569369798938742618818040823435236370214291386195573062935332270993039421387687573397220106241394348907347636983408333263308765625.1):

- the defining polynomial is compressed but the basis is not, is there a reason for that?  I think we should compress the basis (or if we think bases should not be compressed, then we shouldn't compress them on [CMF pages](https://tori.lmfdb.xyz/ModularForm/GL2/Q/holomorphic/1/118/a/a/) either).  Also, I notice that on the NF page you make the entire basis one text area, but not on the CMF pages.  Wouldn't it make sense to make them both the same?  I realize that might mean throwing a list of equations (including the equal sign) on the CMF page into the raw text, but I think that would be okay (or we could make it a vector equation so the RHS is just a list of polynomials).
- the discriminant is still using a number field knowl, but the point of issue #5011 is to uniformize things; I think it should be a text area (or if we think knowls are better, then we should use knowls everywhere, but I would still like a copy button).
- there is no way to easily copy the largest prime factor of the discriminant, I think it should also be using our copy/show mechanism, and I also think it should be compressed everywhere it appears, and in particular in the table of local algebras at the bottom of the screen; as it stands you cannot see most of the information in that table without a lot of horizontal scrolling.

@jwj61, @JohnCremona, @jvoight and I are planning to meet on Tuesday, so we can try to hammer out the specifics then (this is probably easier to do over Zoom rather than in a GitHub thread).  But for that discussion I think it would be useful to have a version of the [NF page](https://tori.lmfdb.xyz//NumberField/17.17.13202363705223218603863184487755024630582383374680032800435939704158233027841788543229812635151072074455438626669432759450899735549243294921243306854586939663055181335880903211540683502950935087551803286773305471032026037990460071805978402769130092019752634101527352693814943791414905342572852579428180829030727661698750961477408713596845181881409326113271279701186934711912924535979375368089471422837282452421683506180048924876364820225210846226099372878635447796099920054239935070472348193791925328690838569369798938742618818040823435236370214291386195573062935332270993039421387687573397220106241394348907347636983408333263308765625.1) that uses the new mechanism consistently  so that we can see what the impact is and debate what we want the default behavior to be.

edgarcosta:
- I have fixed the bug of the `b` instead of the `z`.

- I have added `min-width: 250px` and `word-break: break-all` for raw displays of compressed typsets.

As I said in the initial comment, I'm only addressing the uniformization of the interface, and making sure that it behaves as desired when the typeset is compressed or not, and I would like us to focus on that.

I agree that many things could be improved, in particular, your points for CMF basis or the display of several things on the NF page, but that should not be done on this PR, as many of those things require nontrivial changes, and should be done independently and at another time.

AndrewVSutherland:
@edgarcosta Thanks, the changes you made are definitely an improvement.  I had a question about the q-expansion on https://tori.lmfdb.xyz/ModularForm/GL2/Q/holomorphic/7/45/b/a/, why is there extra white space in the raw text:

image

I would change all sequences of repeated spaces to a single space.

edgarcosta:
I fixed the underlying bug and also added a `re.sub(r'\s+', ' ', str(raw).strip())` to clean up extra white spaces.

jwj61:
I looked at some examples in number fields and think it looks good.

In fact, there are cases where we don't show fundamental units which have been precomputed.  In the future we can show a bit of it and allow them to appear in the copy box (but I am not suggesting that should be done as part of this pull request).

PR #5033, opened 2022-02-14
update for congruent number data

JohnCremona:
Our data on congruent number curves includes 8 N under a million for which the curve had rank 2 but ony one generator was known.  Rathbun has recently provided the missing generators.

This PR  updates the congruent number curve web page to (1) change the list of 8 missing cases to a list of 3; (2) adds an acknowledgement to the two people who found these last generators (as communicated by Rathbun).

In addition, there are 3 data files which need updating from my home directory on legendre to wherever they are stored for beta and prod.  My directory is /scratch/home/jcremona/data/congruent-number-curves (with hyphens) and the database directory is data/congruent_number_curves (with uderscores), and the 3 files to be updated are:

 - CN.MWgroup
 - CN.height
 - CN.Tate-Sha

Of these, the last two are only provided for users to download, while the first is used in displaying on congruent number curve pages (e.g. https://beta.lmfdb.org/EllipticCurve/Q/CongruentNumber/378953).  In each case just 3 lines have changed, namely for N= 378953, 462073, 548953, 764593 and 822722.  And the last 3 cases are N=799657, 937121, 998617.

I suggest that the data is copied first.

### DIFF
--- a/lmfdb/elliptic_curves/templates/congruent_number_data.html
+++ b/lmfdb/elliptic_curves/templates/congruent_number_data.html
@@ -236,8 +236,7 @@ Each file consists of 1,000,000 lines, indexed by the positive integer $n$ and t
 </p>
 
 <p>
-  In eight cases ($n = 378953, 462073, 548953, 764593, 799657, 822722,
-  937121$, and $998617$), all of rank $2$, only one generator is
+  In three cases ($n = 799657, 937121$, and $998617$), all of rank $2$, only one generator is
   known, and hence neither the regulator nor &#1064; are known.
 </p>
 
@@ -261,7 +260,7 @@ Each file consists of 1,000,000 lines, indexed by the positive integer $n$ and t
 <p>
 The data files were created by Randall L. Rathbun on February 14, 2013
 with updates on May 18, 2014.  Some data was provided by Allan J
-MacLeod, the rest computed by Rathbun using <tt>PARI/GP</tt>
+MacLeod, Ralph Buchholz and Brett Wittby, the rest computed by Rathbun using <tt>PARI/GP</tt>
 and <tt>eclib</tt> (in particular <tt>mwrank</tt>).
 </p>
 


AndrewVSutherland:
The data has been copied on beta, but when running the new code I get a server error on  http://127.0.0.1:37777/EllipticCurve/Q/CongruentNumber/378953

JohnCremona:
When you run the website locally it doesn't find the data files since they are a relative directory specified in the python file.  When I tested it I had to change that path temporarily to point to my local copy.  So if it works ok on beta without the cosmetic change in this PR then that's fine. If you look at the PR diffs you'll see that they oy affect the template info, nothing else.

AndrewVSutherland:
@JohnCremona This is now live on beta and you can see the (very impressive) second generator on https://beta.lmfdb.org/EllipticCurve/Q/CongruentNumber/378953.  I'm copying the data to proddb now.

AndrewVSutherland:
(actually copying to prodweb1 and prodweb2, this data is stored under  the data directory).

edgarcosta:
@AndrewVSutherland the folder `/home/lmfdb/data` is a readonly file system, one needs:
1- to unmount it on one of the servers
2- disconnect it on GCE, and change it to read/write on the other server
3- remount the filesystem as read/write 
4- upload data
5- undo the first 3 steps

AndrewVSutherland:
@edgarcosta thanks, I forgot about this.  I don't have time to do this right now, but the three files are in prodweb1:/home/lmfdb if you want to take care of it (otherwise I will do it tonight).

JohnCremona:
Thanks both

edgarcosta:
I will take care of it tonight.

edgarcosta:
done

AndrewVSutherland:
Thank you!

JohnCremona:
Thanks Edgar.  Next time we push code to prod the page http://www.lmfdb.org/EllipticCurve/Q/CongruentNumbers will be updated and only say there are 3 missing cases, not 8.

PR #5036, opened 2022-02-15
Search by signature in ECNF searches

### LINKED ISSUE #5035: Search elliptic curves by signature of base field
It would be useful to be able to search for elliptic curves over totally real fields, so to specify by signature.  Can we replace the field degree with degree/signature?  (It's in the database.)

AndrewVSutherland:
This PR resolves #5035 by adding options to search for real/imaginary quadratic fields and real/mixed cubic fields to the base field degree selector (which is now the base field degree/signature selector).  Also adds signature as a column you can include in the search results table.


roed314:
Looks good.

Issue #5034, opened 2022-02-15
Two things on a big number field page


jvoight:
https://www.lmfdb.org/NumberField/17.17.13202363705223218603863184487755024630582383374680032800435939704158233027841788543229812635151072074455438626669432759450899735549243294921243306854586939663055181335880903211540683502950935087551803286773305471032026037990460071805978402769130092019752634101527352693814943791414905342572852579428180829030727661698750961477408713596845181881409326113271279701186934711912924535979375368089471422837282452421683506180048924876364820225210846226099372878635447796099920054239935070472348193791925328690838569369798938742618818040823435236370214291386195573062935332270993039421387687573397220106241394348907347636983408333263308765625.1

* The root discriminant shouldn't have commas.  We propose writing in scientific notation whenever it's so big that commas would be needed.
* The Galois group is listed as 177843714048000.0, but the knowl doesn't work--presumably we haven't computed all groups of that order, and we never will?  It's probably better not to try to give a label in this case, and just describe the group.

roed314:
There will soon be abstract group homepages for groups like 17T9, though that shouldn't prevent us from improving the situation in the short term.

jwj61:
I am going to work on this today

PR #5043, opened 2022-02-18
Taming the A_17 field

jwj61:
This addresses issue #5034 

The page which reflects many of the changes is

http://127.0.0.1:37777/NumberField/17.17.13202363705223218603863184487755024630582383374680032800435939704158233027841788543229812635151072074455438626669432759450899735549243294921243306854586939663055181335880903211540683502950935087551803286773305471032026037990460071805978402769130092019752634101527352693814943791414905342572852579428180829030727661698750961477408713596845181881409326113271279701186934711912924535979375368089471422837282452421683506180048924876364820225210846226099372878635447796099920054239935070472348193791925328690838569369798938742618818040823435236370214291386195573062935332270993039421387687573397220106241394348907347636983408333263308765625.1

The small group knowl is now just a typeset name of the Galois group if we do not have a gap ID for the Galois group.  If we do have the gap ID, then it is still a small group knowl, as in

  http://127.0.0.1:37777/NumberField/4.0.125.1

 - The discriminant is abbreviated if it is too long
 - Ramified primes are abbreviated if they are too long, both in the list of ramified primes and in the factored form of the discriminant.
 - Note, in this case the factored form of the discriminant has a shortened prime.  When you go to raw mode, only the discriminant is put in a textarea.  The prime is available in raw mode in the list of ramifying primes (but not in the factorization).
 - In the local algebras (bottom of page), the giant prime is abbreviated both in the left column of the table and where it appears in a Q_p (one of the factors for that prime)
 - The integral basis is sort of shown in full, but every numerator or denominator deemed to be too big gets the ... treatment.

The page above does not have fund. units.  If we like the treatment of the basis, we can do the analogous thing with the units for pages which have them.

### DIFF
--- a/lmfdb/galois_groups/transitive_group.py
+++ b/lmfdb/galois_groups/transitive_group.py
@@ -219,7 +219,7 @@ def group_pretty_and_nTj(n, t, useknowls=False, skip_nTj=False, cache={}):
     pretty = group_obj.display_short(True) if group else ''
     if pretty != '':
         # modify if we use knowls and have the gap id
-        if useknowls:
+        if useknowls and group['gapid']:
             gp_label = f"{group['order']}.{group['gapid']}"
             pretty = abstract_group_display_knowl(gp_label, cache=cache)
         if skip_nTj:
--- a/lmfdb/number_fields/number_field.py
+++ b/lmfdb/number_fields/number_field.py
@@ -504,7 +504,6 @@ def render_field_webpage(args):
         ram_primes = r'\textrm{None}'
     data['phrase'] = group_phrase(n, t)
     zkraw = nf.zk()
-    Ra = PolynomialRing(QQ, 'a')
     zk = [compress_poly_Q(x, 'a') for x in zkraw]
     zk = ['$%s$' % x for x in zk]
     zk = ', '.join(zk)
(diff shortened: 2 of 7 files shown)


jwj61:
This now also deals with unit groups.  Before, some fund. units were not shown because they were too long.  Now they are.  For example,

  http://beta.lmfdb.org/NumberField/10.0.14628283371.1
  http://127.0.0.1:37777/NumberField/10.0.14628283371.1

In the process, this fixes a bug in unlatex and makes the raw version of the units nicer.

jwj61:
Also adjusted the display of root discriminants.

Not it uses thin space instead of commas for middle sized numbers:

http://127.0.0.1:37777/NumberField/3.1.1003857398828083800.3

and scientific notation on both the main page and the properties box for really large numbers:

http://127.0.0.1:37777/NumberField/3.1.1086061775432017340256300.6

AndrewVSutherland:
This looks really good!  The only question I had was whether you considered adding raw/copy options for the polynomials in the local algebra table (it would be handy, although I can see an argument against this since it will make the column wider).  I'm happy to merge this either way, I just thought I would ask in case this is something you want to add.

jwj61:
I would be inclined to leave it as it is.  The user can click to get to the home page of the local fields to get the polynomial if they want.

PR #5046, opened 2022-02-23
various cosmetic pycodestyle details in the code

fchapoton:
about useless final ; 
about membership testing
all more all less trivial changes

### DIFF
--- a/lmfdb/classical_modular_forms/web_newform.py
+++ b/lmfdb/classical_modular_forms/web_newform.py
@@ -965,7 +965,7 @@ function switch_basis(btype) {
         """
 
         hecke_polys_orbits = defaultdict(list)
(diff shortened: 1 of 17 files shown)


edgarcosta:
@fchapoton could you point me to why one would prefer `not a in x` instead of the one more close to colloquial English `a not in x`?

fchapoton:
It's the contrary to what you say : one should indeed prefer "a not in x"..

This is the style that is recommended by pep8. 

https://www.python.org/dev/peps/pep-0008/#programming-recommendations

It is implemented in pycodestyle as check E713 :

https://pycodestyle.pycqa.org/en/latest/intro.html#error-codes

edgarcosta:
Upsidaisy, my brain was still booting up :)

Issue #5044, opened 2022-02-21
Mongenicity and index divisors for number fields


AndrewVSutherland:
It would be nice if the LMFDB could tell you for a given number field whether or not it is monogenic (meaning its ring of integers admits an integral power basis), and in the nonmonogenic cases, the index divisor i(K) = GCD({[O_K:Z[a]]: a in O_K}) (which may be 1 even when K is not monogenic).  In particular, it would be nice to be able to search  on these quantities.

It almost certainly won't be feasible to do this for all  number fields in the LMFDB (or even most of them), but I think we could do enough to provide a rich set of examples that would be really useful for people who are thinking about these questions (and there seem to be quite a few of them these days -- its a problem that has been around since Dedekind but never really goes away, perhaps because it is one of the simplest invariants of a number field that we don't really know how to compute, at least not comprehensively).

We could start by marking as monogenic all the fields for which our standard generator (root of the poldredabs polynomial)  already gives an integral power basis, then compute the monogenicity and index divisors for low degree fields (I think this is completely doable for degree 3 and 4, and there are now quite a few results for degrees 5 and 6), and we could apply some known results for producing families of non-mongenic fields and mark those that show up in the LMFDB.  Computing index divisors (or at least their prime divisors)  is probably easier than testing monogenicity (primes dividing the index divisor are bounded by the degree and there a number of [standard tests](https://www.ams.org/journals/proc/1952-003-05/S0002-9939-1952-0050627-4/S0002-9939-1952-0050627-4.pdf) one can apply, as well as some [new ones](https://arxiv.org/abs/2202.09342)), and that will already yield a lot of non-monogenic fields.

I'll note that developing an efficient algorithm for computing these quantities, even just for a subset of the fields in the LMFDB could make for an interesting [LuCANT](https://beta.lmfdb.org/LuCaNT) paper.  But even without inventing anything new there are quite a few existing results in the literature that could be applied.

I should note that I am not an expert on this topic.  I invite those who are to chime in with pointers to the literature and ideas on how to go about computing these quantities.  I'd be very happy for someone to tell me that this is actually easy to do!  But if it isn't that's all the more reason to try to do as much as we can easily do, since it then leaves a target for others to aim at (e.g. the number field of least absolute discriminant whose monogenicity is unknown should be marked as interesting number field).

jwj61:
This looks like a good general reference too:  https://www.ams.org/journals/tran/1985-289-01/S0002-9947-1985-0779058-2/

[...]

PR #5047, opened 2022-02-24
Monogenic fields etc.

jwj61:
Addressing issue #5044, this adds the display and search for number fields for whether or not a field is monogenic, its index, and the list of inessential primes.

Searching is visible on the main number field page

and the display can be seen on any particular field page, e.g., 

where we have data
http://127.0.0.1:37777/NumberField/2.0.567764.1

and where we (currently) do not have data

http://127.0.0.1:37777/NumberField/3.1.244.1

### DIFF
--- a/lmfdb/number_fields/number_field.py
+++ b/lmfdb/number_fields/number_field.py
@@ -452,7 +452,7 @@ def render_field_webpage(args):
     else:  # fallback in case we haven't computed them in a case
         data['frob_data'], data['seeram'] = frobs(nf)
     # This could put commas in the rd, we don't want to trigger spaces
-    data['rd'] = '\(%s\)' % fixed_prec(nf.rd(),2)
+    data['rd'] = r'\(%s\)' % fixed_prec(nf.rd(),2)
     # Bad prime information
     npr = len(ram_primes)
     ramified_algebras_data = nf.ramified_algebras_data()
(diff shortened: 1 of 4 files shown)


jwj61:
I notice that parse_posints accepts 0 as a positive integer.  I can fix this, but would that break the expected behavior elsewhere?

AndrewVSutherland:
@jwj61 Well, it shouldn't, but of course it might.  My vote would be to try changing it, and if tests pass go ahead and fix it, what do you think @roed314 and @edgarcosta ?

JohnCremona:
I would be more worried if we had more French developers :)

AndrewVSutherland:
FYI, we are experimenting with a change to the search results interface that will allow us to show a warning to the user when the search results are known to exclude rows with unknown values that might match the search criteria if the values were known.  I want to hold off merging this PR until that one is ready.

jwj61:
That's fine.  I am mainly working on data for this anyway.

roed314:
The test failures are probably due to adding some null statistics in support of #5052, and should be fixed by the changes there.

roed314:
I thought I had merged in fixes for the underlying data change in #5054; I'm not sure why tests are still failing.  I can take a look soon, but maybe not until Monday.

PR #5048, opened 2022-02-25
Fix parsing of positive integers

jwj61:
The regex for parsing positive integers was changed to not allow 0.  It then uses this function for searching the degree and class number of a number field.

This also changes the regex for bracketed lists of positive integers (e.g., for entering a class group structure).  Before, [02] would be rejected, but now it is accepted as equivalent to [2], but [00] or [0,2] are rejected.


AndrewVSutherland:
in number fields, test_split_ors is failing.

jwj61:
Right.  That should be fixed now.

On further inspection, it seems that parse_posints was never being used, so unlikely to cause any other issues.

There is a case where it will still allow 0 in the mix, but it doesn't seem worth sorting that out.

PR #5051, opened 2022-02-25
Very minor UI tweak to raw/typeset textarea

AndrewVSutherland:
With this change you should not see vertical changes on chrome/firefox/opera/brave/safari when switching between raw/typeset if all the text areas are one line (the most common case), and you should not see partial lines.  Some pages where you can try this

https://blue.lmfdb.xyz/ModularForm/GL2/Q/holomorphic/1/118/a/a/
https://blue.lmfdb.xyz/NumberField/3.1.2688652.1


roed314:
I thought I had merged in fixes for the underlying data change in #5054; I'm not sure why tests are still failing.  I can take a look soon, but maybe not until Monday.

PR #5052, opened 2022-02-25
Add warning when searching on columns with null values

roed314:
This PR adds a warning when a user searches on a column with null values.  To see it in action, visit http://localhost:37777/Groups/Abstract/?aut_group=4.2.


AndrewVSutherland:
Hmm, this is going to cause problems in cases where the nulls don't mean the value is unknown, they mean the value in n/a, for example, the projective_image column in mf_newforms is non-null in every case where it applies (newforms of weight 1), but null in cases where it down not apply (newforms of weight > 1).  We need a mechanism for handling this.

AndrewVSutherland:
See http://localhost:37777/ModularForm/GL2/Q/holomorphic/?projective_image=D15&search_type=List for example.  We don't want to flash message for this query.  We could fix this by setting the projective_image column to some non-null value that means not applicable (rather than not computed), but this would require a code change and the there are probably other columns that have the same issue.  It might be better to explicitly identify columns where null means uncomputed (ideally in the metadata for the table rather than in the code) and only add the warning when those columns are queried.

AndrewVSutherland:
I think there are a quite a few in both categories.  I don't think it would take long for us to go through this list and mark which is which.  I suggest we add a boolean column to the schema metadata that indicates "null means not computed" and set them all to True.  We can then mark the ones where we know it should be False -- if we miss any it's better to err on the side of showing a warning.  I can take care of all the mf_*, g2c_*, char_dir*, gps_st* columns.  That leaves av_fq_*, bely_*, bmf_*, ec_*, gps_*, hcwga_*, lf_*, lfunc_*, nf_* (we can ignore hecke_* and smf_*).

jwj61:
I can do nf_, especially since some of it looks like data we could compute.

edgarcosta:
We only need to worry about `lfunc_search`. There the 22019422 for the first columns are all expected to be NULL as those arise from irrational L-functions.
For the 5370 remainders are a bigger issue that even breaks the search function. I will create a separate issue for those

JohnCremona:
All the ec_* and bmf_* entries in the list are "not computed" rather than "not applicable".

jwj61:
I missed at first that there was a lf_ entry.  Missing entries are "not computed".

AndrewVSutherland:
- char_dir_orbits columns are "not computed".
- gps_st and gps_st0 columns are "not applicable".
- g2c_ columns are "not applicable".
- mf_boxes columns are "not applicable".
- mf_gamma1.a4_dim, mf_gamma1.a5_dim, mf_gamma1.s4_dim, mf_gamma1.dihedral_dim are "not applicable", other mf_gamma1 columns are "not computed".
- mf_hecke_cc and mf_hecke_nf columns are "not applicable".
- mf_newforms.analytic_rank and mf_newforms.analytic_rank_proved are "not computed"
- mf_newforms.is_polredabs, mf_newforms.projective_ are "not applicable", other mf_newforms columns are "not computed"
- mf_newspaces.a4_dim, mf_newspaces.a5_dim, mf_newspaces.s4_dim, mf_newspaces.dihedral_dim are "not applicable" other mf_newspaces columns are "not computed"

PR #5069, opened 2022-03-11
Fix search on residue field degree

jwj61:
This fixes a bug reported in the feedback.  When setting "f" in a search of p-adic fields, before it was ignored:

https://beta.lmfdb.org/padicField/?f=3&search_type=List
http://127.0.0.1:37777/padicField/?f=3&search_type=List


roed314:
I'm going to merge this, but it might also be worth adding a test.

PR #5083, opened 2022-03-16
fix wrong usage of builtin keywords as variables

fchapoton:
This is a partial fix only, some remains to be fixed.

This was found using:

pylint -d all -e W0622 lmfdb/


edgarcosta:
Thank you.

PR #5090, opened 2022-03-23
all trailing whitespaces removed

fchapoton:
this was scripted using

    pep -q --select=W293 lmfdb | xargs autopep8  --select W293 --in-place

and the same for W291.

One could activate these checks, or not, at your choice.


edgarcosta:
@fchapoton Thank you. We will not add those checks, as such a check could be nuisance to a novice contributor.

PR #5101, opened 2022-04-11
Showing monogenic in search results

jwj61:
This adds a class for a search column for yes/no/maybe.  It just uses "yes", "no", "not computed".  I can see people wanting something else, but I didn't see a clearly better option.  Then it is used in displaying number field search results.

Showing yes and not computed:

http://127.0.0.1:37777/NumberField/?hst=List&degree=11&search_type=List&showcol=monogenic

Showing no and not computed:

http://127.0.0.1:37777/NumberField/?hst=List&degree=11&monogenic=not_yes&search_type=List&showcol=monogenic


roed314:
I think that's a good change.  I'm curious whether we could use a question mark to better match the style of the checkmarks (and to save horizontal space).  Maybe there's a unicode character that would be styled similarly?  I haven't tried it, but perhaps https://unicode-table.com/en/2753/?

jwj61:
I thought of using a question mark, but liked "not computed" better myself since it is very direct as to what is going on, seems classier than displaying a question mark, and it is already used in the class group column.  I also thought of the issue of width, but that seems less important since it is an optional column, and the polynomial column can get to be so much wider.

But, whatever gets merged here, that particular choice is something which can be refined later if we can determine a consensus on what it should be.

AndrewVSutherland:
I  can see an argument for yes/not/not computed version checkmark/blank/not computed in cases where not computed is an option (but I don't feel strongly about it and happy to go with the consensus).

But I don't think we should use question marks in the column headings, I would just have the column heading be "Monogenic" (and similarly "Is Galois ?"), as we do for boolean columns elsewhere (e.g. "Solvable", "Semistable", "Q-simple", "Primitive", etc...).

PR #5105, opened 2022-04-16
various details, as suggested by pylint

fchapoton:
about using set and dict comprehension

and about better boolean tests


roed314:
Thanks!

PR #5109, opened 2022-04-20
remove inheritance from object (not needed in python3)

fchapoton:
this was scripted using

    grep -rl "(object):" lmfdb/ | xargs sed -i 's/(object)/()/g'


roed314:
Is the parenthesis necessary at all?  I thought `class AbvarFq_isoclass:` is also okay in Python 3, rather than `class AbvarFq_isoclass():`, which is longer.

fchapoton:
no, the parentheses are not needed, true. I like them for aesthetic reasons. Do you prefer a pull request without parentheses ?

roed314:
No, that's fine.

PR #5110, opened 2022-04-24
pycodestyle cleanup for belyi.main

fchapoton:
a lot of little details, mainly about code formatting

also changing a few things in code:

 including fixing a bad default value `[]`

and fixing one misformatted Error

### DIFF
--- a/lmfdb/belyi/main.py
+++ b/lmfdb/belyi/main.py
@@ -1,13 +1,12 @@
 # -*- coding: utf-8 -*-
-
-
 import re
 
 from flask import render_template, url_for, request, redirect, abort
+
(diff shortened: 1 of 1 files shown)


edgarcosta:
what is wrong with having '[]' has a default value?

roed314:
If you modify the default value, that modification persists:
```
sage: def bad(x=[]): 
....:     x.append(1) 
....:     print(x) 
....:                                                                                                                             
sage: bad()                                                                                                                       
[1]
sage: bad()                                                                                                                       
[1, 1]
sage: bad()                                                                                                                       
[1, 1, 1]
```

Issue #5107, opened 2022-04-19
Home page for the L-function of the Riemann zeta function could be improved


AndrewVSutherland:
https://www.lmfdb.org/L/1/1/1.1/r0/0/0 should have "Riemann zeta function" in the title and a link to the zeta zeros page: https://www.lmfdb.org/zeros/zeta/

PR #5122, opened 2022-05-10
Modify title of Riemann zeta page and add zeros to Learn more

jwj61:
Fixes issue #5107 by adding "Riemann zeta function to the title and a link to the zeta zeros page in the learnmore.  The style of the title matches that of special number fields and the learnmore is the same as on the general L-function page.  To test:

http://127.0.0.1:37777/L/1/1/1.1/r0/0/0
http://beta.lmfdb.org/L/1/1/1.1/r0/0/0


edgarcosta:
I'm hoping that #5124 fixes the tests, regardless, the changes LGTM

Issue #5078, opened 2022-03-12
Server error on large abelian group


AndrewVSutherland:
https://beta.lmfdb.org/Groups/Abstract/ab/1000000007 currently throws a server error

PR #5131, opened 2022-05-20
Big abelian groups

jwj61:
This fixes issue #5078 and should now handle pretty large abelian groups on the fly.  There will still be some limitations, e.g., if you ask for a cyclic group where the order cannot be readily factored by sage.

It changes the behavior a little in that it will rewrite the url and label for a big abelian group to use invariant factors (which meshes with how we show smaller abelian groups).

Some examples to see:

From the issue:
http://127.0.0.1:37777/Groups/Abstract/ab/1000000007 

even bigger:
http://127.0.0.1:37777/Groups/Abstract/ab/10000123451234500007

not cyclic:
http://127.0.0.1:37777/Groups/Abstract/ab/2.2.4.10.100.1000.10000

multiple factors which need to be canonified:
http://127.0.0.1:37777/Groups/Abstract/ab/10.6.14.15.35.21

### DIFF
--- a/lmfdb/groups/abstract/main.py
+++ b/lmfdb/groups/abstract/main.py
@@ -663,7 +663,12 @@ def by_abelian_label(label):
             {"abelian": True, "primary_abelian_invariants": primary}, "label"
         )
     if dblabel is None:
-        return render_abstract_group("ab/" + label, data=primary)
+        snf = primary_to_smith(primary)
+        canonical_label = '.'.join([str(z) for z in snf])
+        if canonical_label != label:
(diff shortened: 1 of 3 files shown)


roed314:
Some lint failures:
```
lmfdb/groups/abstract/main.py:49:1 '.web_groups.LiveAbelianGroup' imported but unused
lmfdb/groups/abstract/web_groups.py:8:1 'sage.all.gap' imported but unused
```

jwj61:
lint fixed

PR #5133, opened 2022-05-25
p-adic field update

jwj61:
This adds "visible" slopes to the home page of each p-adic field (it is the last entry in the top area of a p-adic field home page).  One can also optionally display them in a search results
There are some other changes under the hood so that defining polynomials can be changed.  In particular, subfields are now stored in the database by label, and the unramified subfield of a field does not need to be stored since it is determined by "f".  The structure of the subfield information was also changed to facilitate a future feature where one can search for a field with a given subfield.

Before this goes to production, lf_fields should be copied to the cloud.

Because we have both old and new columns in the database table, some entries will be deleted once this code makes its way to production.

### DIFF
--- a/lmfdb/local_fields/main.py
+++ b/lmfdb/local_fields/main.py
@@ -153,6 +153,12 @@ def show_slopes(sl):
         return "None"
     return(sl)
 
+def show_slopes2(sl):
+    # uses empty brackets with a space instead of None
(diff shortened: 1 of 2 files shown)


AndrewVSutherland:
lf_fields has been copied to the production database.

PR #5136, opened 2022-06-02
fix pycodestyle E306

fchapoton:
by adding the missing empty lines before function defs

also changes a few very minor details on the way


edgarcosta:
LGTM.
Okay to merge once tests pass

Issue #5130, opened 2022-05-20
Display inconsistency for elliptic curves


jwj61:
If you search for elliptic curves (over Q or over number fields), the torsion subgroup (if it is not cyclic) is displayed in the search result list using \oplus.  If you click to see a curve, the Mordell-Weil group is displayed with \times to indicate product.

I think \times is used more in the lmfdb

AndrewVSutherland:
They are of course both correct, but I would vote for \oplus as the "more correct" choice (they are Z-modules, not just abstract groups).

jwj61:
So the same would hold for any group which is automatically abelian?  Class groups come to mind as an example.

AndrewVSutherland:
Every abelian group can of course be given the structure of a Z-module, but the mordell-Weil group of an elliptic curve is naturally equipped with multiplication-by-n maps for every integer n in Z; this is in some sense more fundamental than, and independent of, the fact that it is an abelian group that happens to be finitely generated for elliptic curves over number fields.  Over the complex number the torsion subgroup of an elliptic curve is a Z-module, but it does not make sense to view it as as a product of finite cyclic groups; the direct sum and direct product are no longer the same in this setting, and it is the direct sum that corresponds to the torsion subgroup.

In the case of the class group of a number field there isn't a "multiplication-by-n" map that is as natural (and the group operation is usually written multiplicatively), and here I think it makes more sense to view them as products of cyclic groups.

JohnCremona:
I am not entirely convinced. Why is multiplying a point by n more "natural" than raising an ideal class to power n?  Is it that there's an endomorphism *ring* (of the elliptic curve) around, which happens to be Z?

We have some curves with actual CM e.g. with endomorphism ring Z[i] over K=Q(i).  One might want to show the structure of E(K) or the torsion subgroup as an End(E)-module in such cases, perhaps.

AndrewVSutherland:
I guess I am making two distinct arguments.  The first is that torsion subgroups really are direct sums, not direct products, as can be seen in the situation where they have infinite cardinality (as occurs for elliptic curves over C, for example), and the second is the point that you made, which is that they are naturally End(E)-modules (I agree that it would also be interesting to show the End(E)-structure of the torsion subgroups of CM elliptic curves E).

[...]

PR #5137, opened 2022-06-02
change torsion options to use ℤ/nℤ

fchapoton:
to fix part of #5130


fchapoton:
not sure if unicode ℤ is appropriate (hopefully yes) and whether to use "times" or "oplus"..

Issue #5139, opened 2022-06-06
Units error for number fields


roed314:
I'm not sure if this is a data error or a code error, but https://www.lmfdb.org/NumberField/8.0.97571192547662768845333987328.43 causes a server error.  The underlying problem is that the `units` column for this field is
```
['143/185002993*a^6 + 1885777/185002993*a^4 + 2558986/97937*a^2 + 1795063489/97937', '6105/83066343857*a^6 + 106107397/83066343857*a^4 + 270799866/43973713*a^2 + 330142479593/43973713', '3433705/83066343857*a^6 + 45267308080/83066343857*a^4 + 61499784194/43973713*a^2 + 43182980632147/43973713']
```
which seems fine, but we call `unlatex` on each entry, which transforms it into
```
['143/185002993**a^6 + 1885777/185002993**a^4 + 2558986/97937**a^2 + 1795063489/97937', '6105/83066343857**a^6 + 106107397/83066343857**a^4 + 270799866/43973713**a^2 + 330142479593/43973713', '3433705/83066343857**a^6 + 45267308080/83066343857**a^4 + 61499784194/43973713**a^2 + 43182980632147/43973713']
```
which Sage then can't parse.

jwj61:
I will take a look.  It looks like a code error in unlatex which is pretty simplistic.

jwj61:
The asterisks are harmless, so I will just close this.

PR #5141, opened 2022-06-06
Fix unlatex to handle number * a

jwj61:
Fixes issue #5139 .  See

http://127.0.0.1:37777/NumberField/8.0.97571192547662768845333987328.43


roed314:
Works for me.

Issue #1522, opened 2016-06-07
Simplified Weierstrass model?


jvoight:
In many circumstances, it would be nice to have not just the minimal model for an elliptic curve or a hyperelliptic curve, but you'd prefer the equation
  y^2 = f(x)
which amounts to completing the square.  Maybe we could list this as well?  

[UPDATE] This feature is now available for genus 2 curves.

JohnCremona:
At the moment I am concentrating on getting models for elliptic curves) which are minimal (or almost) and reduced, and do not care much about completing the square, which would of course be trivial to show.  So I won't stand in anyone's way but don't myself give it a high priority.

AndrewVSutherland:
I actually thought about doing this at one point but decided against it for the following reasons:

- Screen real estate -- displaying two equations would inevitably push something that you can currently see without scrolling further down the screen where you have to scroll to see it.  One could instead have two modes (e.g. as with G2/Igusa/Igusa-Clebsch invariants), but this would make my next concern worse.
- Model dependent data -- some information (e.g. coordinates of rational points) depends on the model, which could be confusing, especially if there are two modes. In the modes sceneario you would really want to switch everything to match the model on the screen, but recomputing this data on the fly might get expensive (and in some cases it might not even apply).
- Not necessary -- it is easy to get this whenever you need it (in Magma use SimplifiedModel for genus 2 curves and WeierstrassModel for elliptic curves, and there are similar commands in Sage).

Having said all that, I could be convinced to throw this into to the code snippets if you think that would be useful.

jvoight:
This is definitely not high priority, but it has shown up for me in at least two ways: first, in wanting to understand the Galois representation on the 2-torsion, and second, in talking to students/postdocs who don't understand why we don't show the simpler equation (to their mind).  

We can ignore me, I am just in the habit of passing on all ideas above a certain threshold.

Drew: in terms of "not necessary", I don't think that's the mode we want to be in.  There are many things you can dump into Magma or Sage and compute, but the point is the browser shouldn't have to do this.  In terms of "screen real estate", I don't know if we're really so successful with that right now anyway (we seem to show a ton of things requiring scrolling down).

Here's a modified suggestion: put the 2-torsion polynomial = minimal polynomial for Weierstrass points on the page and put it in a different variable.  We can make a comment that this gives a minimal Weierstrass model, and I think it meets all of the above constraints.

edgarcosta:
This could also be done with a hide/show button.

[...]

PR #5147, opened 2022-06-23
Addresses #1522 for EC/Q by adding option to show short Weierstrass equation (a la G2C)

AndrewVSutherland:
After being asked by the umpteenth person why we don't show simplified Weierstrass equations on elliptic curve pages in the LMFDB (we have a feature request for this that has been open for 8 years) I went ahead and implemented on option to cycle through minimal/homogenized/simplified models as we do on the genus 2 curve pages.  I will do the same for ECNF once this is merged.


lmfdb/elliptic_curves/web_ec.py
```diff
@@ -168,6 +168,32 @@ def latex_equation(ainvs):
...
+                    r'\)'])
+
+def short_latex_equation(ainvs):
+    a1,a2,a3,a4,a6 = [ZZ(a) for a in ainvs]
+    c4 = -27*a1^4 - 216*a1^2*a2 + 648*a1*a3 - 432*a2^2 + 1296*a4
+    c6 = 54*a1^6 + 648*a1^4*a2 - 1944*a1^3*a3 + 2592*a1^2*a2^2 - 3888*a1^2*a4 - 7776*a1*a2*a3 + 3456*a2^3 - 15552*a2*a4 + 11664*a3^2 + 46656*a6
+    for p in c4.gcd(c6).prime_divisors():
+        while c4.valuation(p) >= 4 and c6.valuation(p) >= 6:
+            c4 = c4.divide_knowing_divisible_by(p**4)
+            c6 = c6.divide_knowing_divisible_by(p**6)
```
roed314, line 189: c4 //= p**4 is surely fast enough, but this is fine too.

AndrewVSutherland: I generally try to avoid using / or // in python (who knows, the semantics could change in python4).

JohnCremona:
Thanks for doing this -- I am about to take a look.

JohnCremona:
One triviality: when you click on "simplify", it is the second option (after "homogenize") but then to go back, the "minimize" option comes first.  Better for it to be the second option.

More serious:  http://localhost:37777/EllipticCurve/Q/36/a/3 has minimal equation y^2=x^3-27 which is surely already simplfied as it is a minimal short Weierstrass equation, but "simplify" turns it into y^2=x^3+6x-1259707 !

I have not yet identified the bug in the code but (c4,c6)=(0,-1259712) has been displayed as if it were (6,-1259707), so somewhow 6 has been added to both.  

[It's also slightly bad to call the new coefficients c4 and c6.  Something like A and B might be better -- they are not c4 and c6, even if they start out as -27*c4 and -54*c6.]

Finally, given that the long equation is always minimal, the only primes p which can occur in the loop are 2 and 3 and the valuations cannot be too big, so it might be simpler to just have two while loops, one for 2 and one for 3.

AndrewVSutherland:
@JohnCremona I think this should be fixed now. I swapped the order of the options as you suggested (and did the same in genus 2 curves for consistency).

JohnCremona:
OK -- thanks! -- thate example looks good now.  With the change from c4,c6 to A,B (thanks again) I could not see what else had changed, but it is probably not very interesting!

AndrewVSutherland:
@JohnCremona Yes, in python (but not in Sage) "^" is bitwise xor, but perhaps this will change in python4 :)

PR #5163, opened 2022-07-15
Resolves #5159

### LINKED ISSUE #5159: "raw_typeset undefined" error on sibling fields that have a polynomial but no label
The number field knowls in the sibling data do not work properly in cases where a defining polynomial for the sibling field is known but the sibling field itself is not in the LMFDB,

Click on the "Deg 12" link in the list of sibling fields on https://www.lmfdb.org/NumberField/5.5.106588382947225.1 to reproduce: you will see a knowl with the error message

`ERROR in the template: 'raw_typeset' is undefined. Please edit it to resolve the problem.`

I'm sure this is easy to fix, I'm just not sure exactly where the fix needs to be made.

AndrewVSutherland:
As @edgarcosta noted, the raw_typeset errors in #5159 are the caused by not exposing raw_typeset to jinja (via `app.context_processor`).  I'm not sure why no one has noticed/encountered it before, but this fixes the problem (at least for number fields).

### DIFF
--- a/lmfdb/number_fields/number_field.py
+++ b/lmfdb/number_fields/number_field.py
@@ -93,6 +93,10 @@ def fixed_prec(r, digs=3):
     print(head)
     return str(head) + '.' + n[-digs:]
 
+@app.context_processor
+def ctx_raw_typeset():
+    return {'raw_typeset': raw_typeset}
+
 
 @app.context_processor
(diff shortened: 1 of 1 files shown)


edgarcosta:
Looks good to me. Feel free to merge

Issue #5166, opened 2022-07-18
Ramified prime count not showing on number field searches


davidlowryduda:
This came up during Drew's PCMI talk. To replicate, I go to https://www.lmfdb.org/NumberField/?hst=List&discriminant=0-5000&search_type=List&sort_order=num_ram and for whatever reason, the `Ram. prime count` column isn't displaying.

PR #5167, opened 2022-07-18
Fix issue #5166

jwj61:
Display number of ramified primes in search results

 https://www.lmfdb.org/NumberField/?hst=List&discriminant=0-5000&search_type=List&sort_order=num_ram 
 http://127.0.0.1:37777/NumberField/?hst=List&discriminant=0-5000&search_type=List&sort_order=num_ram


davidlowryduda:
I'm very impressed with how quickly you acted!

I confirm that this resolves #5166 and works. I didn't investigate why the abstract group test fails, but that seems unrelated.

edgarcosta:
I will merge, and what happens when the tests run again

PR #5169, opened 2022-08-07
some pep8 details and empty sets

fchapoton:
* pep cleanup for one file
* a few details about empty sets


edgarcosta:
There is a keys removal that should be keys().

fchapoton:
indeed, thanks ; now fixed

PR #5181, opened 2022-08-16
update conda env

edgarcosta:


edgarcosta:
the tests will take a while as we need to generate a new sage installation

roed314:
Looks good.

PR #5184, opened 2022-08-20
Fix list of knowls for source/ack

jwj61:
On the source and acknowlegments page for number fields, the extra acknowledgments appears twice and the "how to cite" not at all.  This fixes that.

http://127.0.0.1:37777/NumberField/Source
http://beta.lmfdb.org/NumberField/Source


roed314:
This is certainly an improvement, but there's currently nothing in [rcs.cite.nf](https://beta.lmfdb.org/knowledge/show/rcs.cite.nf).  Should there be?

jwj61:
I think most of the cite knowls are empty.

roed314:
Ok.  Improving those can be a separate issue.

PR #5188, opened 2022-08-23
various enhancements in web_family

fchapoton:
some little changes in the doc, for efficiency purposes

plus full pep8 codestyle

### DIFF
--- a/lmfdb/hypergm/web_family.py
+++ b/lmfdb/hypergm/web_family.py
@@ -1,5 +1,3 @@
-from __future__ import annotations
-
 import re
 from flask import url_for
 from collections import defaultdict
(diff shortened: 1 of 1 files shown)


roed314:
I'm happy with most of these changes, but I'm not sure about starting to add annotations like `def cyc_to_QZ(A) -> list`.  I think having this in just one part of the LMFDB code base is confusing; if we want to add them it should be a conscious choice.

That being said, if you are interested in adding annotations more broadly, I'm happy to have a discussion about that.

fchapoton:
closed ? :disappointed:

fchapoton:
I have just undone the annotations..

JohnCremona:
I agree with @roed314 here.  We should make a decision and make such changes globally if we decide to implement them.

More generally, I certainly approve of the stylistic changes which @fchapoton has been making systematically, but I do also fear that unless we make coding conventions explicit and check pull requests for adherence the job will never end.  We do want good quality code of course, but we do not want to put barriers in the way if new (or old) contributors.

JohnCremona:
Sorry I didn't mean to close it. I'm on a phone and gave fat fingers

fchapoton:
No harm, John. I guesses it was a wrong manoeuver.

The usual advice about annotations is to start somewhere, and not do that everywhere at the same time. I agree that this should be discussed first.

roed314:
This looks fine.  The error is from the `/random` route failing and doesn't have anything to do with the changes here.  I'll plan on making a new issue about annotations.

PR #5192, opened 2022-08-28
refresh two files in siegel_modular_forms (pep8, etc)

fchapoton:
namely full pycodestyle cleanup of `__init__.py` and `dimensions.py`

also some minor other details, in inparticular about using dict comprehensions


roed314:
Thanks!

PR #5193, opened 2022-08-28
fix imports of ComplexField

fchapoton:
so that they work with recent Sage


roed314:
Thanks; that deprecation warning has been annoying me.

PR #5197, opened 2022-09-03
Change wording

jwj61:
On the page for an individual group, the label for the line listing semidirect products says semidirect product.  This is technically not right since direct products are semidirect products.  This pull request just adds the word nontrivial, which is now explained in the knowl.

http://127.0.0.1:37777/Groups/Abstract/32.37
http://beta.lmfdb.org/Groups/Abstract/32.37

### DIFF
--- a/lmfdb/groups/abstract/templates/abstract-show-group.html
+++ b/lmfdb/groups/abstract/templates/abstract-show-group.html
@@ -202,7 +202,7 @@
   {% endif %}
   {% if gp.semidirect_products %}
     <tr>
-      <td >{{KNOWL('group.semidirect_product', 'Semidirect product')}}:</td>
+      <td >{{KNOWL('group.semidirect_product', 'Nontrivial semidirect product')}}:</td>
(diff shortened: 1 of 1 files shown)


jwj61:
That brings up another issue with this part of the display.  If there are no non-trivial direct product decompositions, that line is not included.  As it stands, on a group like 12.1, one only sees the semidirect (and non-split) lines.  Regardless of what happens with this pull request, I think that should change so that it says either "there are no nontrivial direct product decompositions" or "data not computed" when we don't have any decompositions for the direct product line (conditioned on whether we know about all normal subgroups).

AndrewVSutherland:
I guess I have a slight preference for not including the word "nontrivial" but I agree with the suggestion that when there are no nontrivial semidirect product constructions we should display "no nontrivial semidirect products" (and similarly for direct products).  This would mean that the user always sees either semidirect products that are evidently nontrivial (using an rtimes symbol that I think everyone would assume means it is not a direct product) or text saying there are no nontrivial semidirect products (and the trivial semidirect products will appear above in the direct products list).

PR #5199, opened 2022-09-09
Fix rewrite so that it no longer deletes rows not matching query

roed314:
We now use `update_from_file` under the hood rather than `reload`.


JohnCremona:
Thanks!

PR #5206, opened 2022-09-13
some care for E251 (no space around = in arguments of functions)

fchapoton:
only partially fixed, remains some in other sub-folders


edgarcosta:
LGTM

PR #5208, opened 2022-09-16
Empty or clause evaluate to False rather than True

roed314:


roed314:
Despite the claim that all tests have passed, it looks like there are actually test failures.  See localhost:37777/L/4/1/1.1/r0e4/p2.27m6.04m13.14p16.90/0, which now gives an error "Unable to render L-function page due to the following problem(s): given string 'NaN' is not a complex number."

I'm not going to have a chance to track this down this afternoon.

edgarcosta:
I will try to look into this next week.

edgarcosta:
This issue that you noticed has been fixed in #5217

PR #5210, opened 2022-09-21
fix code snippets for real period of EC/Q

### LINKED ISSUE #5209: EC/Q magma code for real period uses different convention
From the bug tracker:  "The real period listed in the database and the real period computed by Magma using the command RealPeriod are sometimes off by a normalisation factor of two. The first instance of this is the rational curve 14.a3 but there are many other examples. Perhaps the Magma command for real periods shown in each page should be modified accordingly?"

This curve has recatngular period lattce [x,yi] with x=0.99 (approx), and this is the value returned by magma's RealPeriod(E), but our convention is that we multiply the least real period by the number of components (=2 here).  So the magma code we give should have the number of compnents as a factor too.

JohnCremona:
This fixes #5209.

### DIFF
--- a/lmfdb/elliptic_curves/code.yaml
+++ b/lmfdb/elliptic_curves/code.yaml
@@ -63,7 +63,7 @@ reg:
 
 real_period:
   sage:  E.period_lattice().omega()
-  pari:  if(e.disc>0,2,1)*e.omega[1]
+  pari:  if(E.disc>0,2,1)*E.omega[1]
   magma: (Discriminant(E) gt 0 select 2 else 1) * RealPeriod(E);
 
 cp:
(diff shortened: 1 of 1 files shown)


JohnCremona:
I don't understand the test failure, can someone explain (even if it is irrelevant to this PR, which it surely is)?

AndrewVSutherland:
It's definitely irrelevant and will likely succeed when it runs again, I just restarted the test (with our current CI setup there is a small probability that one or more tests will not start correctly, restarting the failed job usually fixes it).

JohnCremona:
> It's definitely irrelevant and will likely succeed when it runs again, I just restarted the test (with our current CI setup there is a small probability that one or more tests will not start correctly, restarting the failed job usually fixes it).

Thanks

AndrewVSutherland:
Tests passed, merging.

PR #5223, opened 2022-09-25
fix all remaining E251 in lmfdb/

fchapoton:
fix all remaining E251 in lmfdb (no whitespace around = in function calls)

this could be activated in the linter, if one wishes. I have not done that here.


AndrewVSutherland:
lgtm

PR #5227, opened 2022-10-05
partial pep8 cleanup of web_newform

fchapoton:
in particular, get rid of all E227 errors in lmfdb/  (all inside the same line of this file)


fchapoton:
test have finished and lights are all green

PR #5236, opened 2022-10-20
Automatic style fixing after PR's are merged

### LINKED ISSUE #5224: Extra linting options when merging into dev/web?
When merging from master->{dev, web} we should perhaps take the opportunity to check for extra linting options that we want our code to satisfy but don't want to bother contributors with it.
In particular, several of @fchapoton contributions that fixed:
- E251 : unexpected spaces around keyword / parameter equals
- E241: multiple spaces after ‘,’
- E203: whitespace before ‘,’, ‘;’, or ‘:’
- E304: blank lines found after function decorator
- E702: multiple statements on one line (semicolon)

edgarcosta:
Note, we will not be enforcing any new style, just keeping the styles that we already satisfy.
If some style got broken, it creates a PR to fix it.
To see the kind of PR that this creates see https://github.com/edgarcosta/lmfdb/pull/12

In particular, when we merge this PR, we will get exactly the same PR submitted here.

This resolves #5224

### DIFF
--- a/.github/workflows/autopep8.yml
+++ b/.github/workflows/autopep8.yml
@@ -1,6 +1,7 @@
 name: Format python code
-on: push
-  branches: [ 'master' ]
+on:
+  push:
+    branches: [ 'master' ]
 jobs:
   autopep8:
     runs-on: ubuntu-latest
--- a/.github/workflows/autopep8.yml
+++ b/.github/workflows/autopep8.yml
@@ -41,7 +41,7 @@ jobs:
           # E714 - Use 'is not' test for object identity.
           # E721 - Use "isinstance()" instead of comparing types directly.
           # E722 - Fix bare except.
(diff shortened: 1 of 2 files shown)


edgarcosta:
@roed314 I think this solves the problem that we were talking about last time of who would have to fix the style...

edgarcosta:
@roed314, checkout the new automatic PR https://github.com/edgarcosta/lmfdb/pull/10, where I disabled `E301`

roed314:
I'm in favor of this overall approach: I think it strikes a good balance between annoying to new contributors, helping them learn, and not taking too much of our time.  I do think we need to get feedback more broadly from other LMFDB editors before activating this.

Maybe also remove E401 (personally, I prefer putting common libraries like os, sys, collections on a single line).  We shouldn't use both W503 and W504 since they conflict with each other.

edgarcosta:
After discussing with @roed314 we decided to test this out in practice, and we will get our first auto-generated PR.

PR #5240, opened 2022-10-23
fix pep8 E305 (done using autopep8)

fchapoton:
I am no longer sure what to do after you put the recent mechanism in place.

Have you decided once and for all which pep8 codes you want to enforce ? Or may I continue to fix some of the less controversial ones that are not yet done ?

### DIFF
--- a/lmfdb/abvar/fq/__init__.py
+++ b/lmfdb/abvar/fq/__init__.py
@@ -12,6 +12,7 @@ abvarfq_logger = make_logger(abvarfq_page)
 def body_class():
     return {"body_class": "abvarfq"}
 
+
 from . import main
 assert main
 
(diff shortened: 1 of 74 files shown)


roed314:
@edgarcosta and I talked about which pep8 codes to enforce and selected a set of them (and omitted others).  Personally, I find E305 annoying: I don't want two blank lines between top level functions and classes, and while I usually want a blank line between methods there are times when I don't (multiple one line methods that are related to each other).  So we chose not to enforce it, and my vote would be not to merge this PR.

That being said, we're not dictators and other people (including you) can argue against our opinion.

fchapoton:
ok, I understand. Let me close this one then.

edgarcosta:
Also, we decided to use autopep8 as we don't want to force contributors to
follow the guidelines, but we also don't want to be correcting the style
for every contribution manually.

PR #5242, opened 2022-10-27
Check if wild inertia group has been computed

jwj61:
If not, just give a message saying so.  The current code produces a bad knowl.

http://127.0.0.1:37777/padicField/2.18.35.488
https://beta.lmfdb.org/padicField/2.18.35.488


AndrewVSutherland:
LGTM, I'm just rerunning the failed test (the failure looks spurious) and will then merge.

PR #5246, opened 2022-10-28
Reflex fields

rbommel:
DIsplaying reflex fields for CM fields.


jwj61:
Since this is specific to CM fields, maybe it would make sense to have the information immediately after the statement that it is a CM field, and as a regular entry in the list of invariants as opposed to being its own subsection.

Also, on a field like

  http://127.0.0.1:37777/NumberField/16.0.1048576000000000000.1

it should either say "none" or "data not computed" depending on which is the right information.

Are there cases where only some reflex fields are computed for a given field?  If so, can you give an example so we can see how that is indicated?

AndrewVSutherland:
@rbommel can you respond to @jwj61 comments when you get a chance?  I'd like to merge this PR soon.

rbommel:
@AndrewVSutherland, currently my priority for the LMFDB work has been with the modular curves project. I am happy to change the way reflex fields are displayed when the problems with the modular curves are resolved.

@jwj61, there are indeed a lot of examples where some but not all reflex fields have been computed. I don't know a label by heart, but if you look at degree 12 CM fields, there should be plenty of examples. I believe there is an explanation in the knowl that not all such fields have been computed.

AndrewVSutherland:
@rbommel I'm happy to wait on this until modular curves are in good shape.

jwj61:
On the display of incomplete information, I think it needs to be more apparent.  The usual role of the knowl would be just to define relfex field (which it does).  I don't think the user should be expected to look in the knowl to learn that they are getting partial information.

The number of reflex fields of each degree should depend only on the Galois group.  So, that information can be stored in the transitive group database.  Then when displaying them, the code can compare how many reflex fields we have to how many exist and either display no extra message (meaning this is all of them), "some data not computed" (if there are some, but not all), or "data not computed" if none of them have been computed.

rbommel:
Thanks for all the suggestions. I have changed the display of the data accordingly. While it is true that the degrees only depend on the Galois group, I did not go as far to implement this last suggestion, so right now the unavailable reflex fields are not counted by degree but just in total.

jwj61:
On the whole this is looking good.  I am confused by a page like

  http://127.0.0.1:37777/NumberField/4.0.144.1

This is a V field and it says that 2 of the reflex fields are not computed.  Aren't the two fields which are listed are the entire list?  It could be I am missing something important.

rbommel:
Okay, yes, I see there was a factor 2 missing for all the counts as I was computing the reflex fields for each complimentary pair before.

jwj61:
I am not an expert in this, so this is an honest question.  Since a set and its complement always produce the same field, should it be counted as a multiplicity?

rbommel:
I don't know and I also wouldn't call myself an expert. Right now, it seems to me that this is the easiest way to explain it; just having for each reflex field the number of CM types.

jwj61:
The first reference I can find to help resolve this is on page 13 of Milne's notes

https://www.jmilne.org/math/CourseNotes/CM.pdf

He talks about the 32 CM types for Q(zeta_13), which means that he is counting complementary CM structures as the same.

One more minor thing (which I can do in a subsequent pull request if you want), is to put a colon after Reflex fields (to match the structure of other lines that have a descriptor and then data, as opposed to the ones which are sentences).

JohnCremona:
The simplest case us that of an imaginary quadratic field. Do we consider it to have one CM type or two?

rbommel:
Okay, I currently took the definition that is in Marco Streng's notes https://pub.math.leidenuniv.nl/~strengtc/cm/cmtypes.pdf, but if this other definition is standard, we can also change it, I don't mind that much.

JohnCremona:
Is it true that complementary CM types have the same reflex field?  I could not find that stated in Milne but I think it may be an easy consequence of the definition of reflex field.  In that case the 2^g CM-types (as defined literally by Milne in his Def.1.3, and by everyone) form 2^{g-1} complementary pairs which have 2^{g-1} reflex fields.

I am no expert on this.

rbommel:
Yes, that is true and that is also how I computed it to save myself half the work. If an element of the (absolute) Galois group fixes a CM type, then it will automatically fix its complement too.

jwj61:
I think we should just adjust the explanation in the knowl to explain that for each CM type, its complement is a CM type with the same reflex field, so multiplicities are of complementary pairs of CM types (and divide all multiplicities by 2).

AndrewVSutherland:
Just to clarify, we aren't changing the definition of CM type, of which there will still be $2^g$, but when counting reflex fields we account for the fact that there is a baked-in 2-1 map that sends a CM type and its complement to the same reflex field.  The multiplicities of the reflex fields should always sum to $2^{g-1}$, and it is possible for the reflex fields to all occur with multiplicity one.

JohnCremona:
> Just to clarify, we aren't changing the definition of CM type, of which there will still be $2^g$, but when counting reflex fields we account for the fact that there is a baked-in 2-1 map that sends a CM type and its complement to the same reflex field. The multiplicities of the reflex fields should always sum to $2^{g−1}$, and it is possible for the reflex fields to all occur with multiplicity one.

Agreed.  Regarding Milnor's example where g=6, he only says that there are 32 CM types *containing the given embedding of E into C*  which is one way of cutting out half of the total number of 64, I think.

jwj61:
That is my suggestion.

rbommel:
This is changed.

PR #5259, opened 2022-11-07
display adelic gens

BarinderBanwait:
We can now see the generators for the adelic image, live on pink, eg [here](https://pink.lmfdb.xyz/EllipticCurve/Q/175791/h/1)

### DIFF
--- a/lmfdb/elliptic_curves/templates/ec-curve.html
+++ b/lmfdb/elliptic_curves/templates/ec-curve.html
@@ -467,7 +467,7 @@ The {{KNOWL('ec.galois_rep', title=' $\ell$-adic Galois representation')}} has {
 {% endif %}
 
 {% if data.adelic_level %}
-<p> The image of the {{KNOWL('ec.galois_rep', title='adelic Galois representation')}} has {{KNOWL('ec.galois_rep_adelic_image', title='level')}} ${{data.adelic_level}}$, {{KNOWL('ec.galois_rep_adelic_image', title='index')}} ${{data.adelic_index}}$, {{KNOWL('ec.galois_rep_adelic_image', title='genus')}} ${{data.adelic_genus}}$, and generators ${{data.data.adelic_data.adelic_gens}$</p>
+<p> The image of the {{KNOWL('ec.galois_rep', title='adelic Galois representation')}} has {{KNOWL('ec.galois_rep_adelic_image', title='level')}} ${{data.adelic_level}}$, {{KNOWL('ec.galois_rep_adelic_image', title='index')}} ${{data.adelic_index}}$, {{KNOWL('ec.galois_rep_adelic_image', title='genus')}} ${{data.adelic_genus}}$, and generators ${{data.data.adelic_data.adelic_gens}}$</p>
 {% endif %}
 
 <h2>$p$-adic regulators</h2>
--- a/lmfdb/elliptic_curves/templates/ec-curve.html
+++ b/lmfdb/elliptic_curves/templates/ec-curve.html
@@ -467,8 +467,11 @@ The {{KNOWL('ec.galois_rep', title=' $\ell$-adic Galois representation')}} has {
 {% endif %}
 
 {% if data.adelic_level %}
+
+  {{ place_code('adelicimage') }}
+
 <p> The image of the {{KNOWL('ec.galois_rep', title='adelic Galois representation')}} has {{KNOWL('ec.galois_rep_adelic_image', title='level')}} ${{data.adelic_level}}$, {{KNOWL('ec.galois_rep_adelic_image', title='index')}} ${{data.adelic_index}}$, {{KNOWL('ec.galois_rep_adelic_image', title='genus')}} ${{data.adelic_genus}}$, and generators</p>
-<p>${{data.data.adelic_data.adelic_gens}}$</p>
+<p>${{data.data.adelic_gens_latex}}$</p>
 {% endif %}
 
 <h2>$p$-adic regulators</h2>
--- a/lmfdb/elliptic_curves/templates/ec-curve.html
+++ b/lmfdb/elliptic_curves/templates/ec-curve.html
@@ -467,7 +467,7 @@ The {{KNOWL('ec.galois_rep', title=' $\ell$-adic Galois representation')}} has {
 {% endif %}
 
 {% if data.adelic_level %}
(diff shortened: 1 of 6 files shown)


lmfdb/elliptic_curves/templates/ec-curve.html
```diff
@@ -467,7 +467,8 @@ <h2> {{KNOWL('ec.galois_rep', title='Galois representations')}} </h2>
 {% endif %}
 
 {% if data.adelic_level %}
-<p> The image of the {{KNOWL('ec.galois_rep', title='adelic Galois representation')}} has {{KNOWL('ec.galois_rep_adelic_image', title='level')}} ${{data.adelic_level}}$, {{KNOWL('ec.galois_rep_adelic_image', title='index')}} ${{data.adelic_index}}$, and {{KNOWL('ec.galois_rep_adelic_image', title='genus')}} ${{data.adelic_genus}}$.</p>
+<p> The image of the {{KNOWL('ec.galois_rep', title='adelic Galois representation')}} has {{KNOWL('ec.galois_rep_adelic_image', title='level')}} ${{data.adelic_level}}$, {{KNOWL('ec.galois_rep_adelic_image', title='index')}} ${{data.adelic_index}}$, {{KNOWL('ec.galois_rep_adelic_image', title='genus')}} ${{data.adelic_genus}}$, and generators</p>
+<p>${{data.data.adelic_data.adelic_gens}}$</p>
```
roed314, line 471: It looks like there's an extra data here?  If you're only using the gens, why not just change the code so that you can write this is `data.adelic_gens`?

lmfdb/elliptic_curves/web_ec.py
```diff
@@ -313,8 +314,17 @@ def make_curve(self):
...
-        data['galois_data'] = [r for r in list(db.ec_galrep.search({'lmfdb_label': lmfdb_label})) if r["prime"] > 0]
-
+        galois_data = list(db.ec_galrep.search({'lmfdb_label': lmfdb_label}))
+        data['galois_data'] = [r for r in galois_data if r["prime"] > 0]
+        adelic_data = [r for r in galois_data if r["prime"] == 0]
+        if adelic_data:
+            assert len(adelic_data) == 1
+            my_adelic_data = adelic_data[0]
+            my_adelic_data['adelic_gens'] = ",".join([str(latex(dispZmat_from_list(z,2))) for z in my_adelic_data['adelic_gens']])
+            data['adelic_data'] = my_adelic_data
```
roed314, line 324: I would probably just set `data['adelic_gens'] = ",".joing(...)`

lmfdb/lfunctions/main.py
```diff
@@ -349,14 +349,21 @@ def common_parse(info, query):
                  lambda origins: " ".join('<a href="%s">%s</a>' % (url, name) for name, url in origins),
                  default=True)],
     db_cols=['algebraic', 'analytic_conductor', 'bad_primes', 'central_character', 'conductor', 'degree', 'instance_urls', 'label', 'motivic_weight', 'mu_real', 'mu_imag', 'nu_real_doubled', 'nu_imag', 'order_of_vanishing', 'primitive', 'rational', 'root_analytic_conductor', 'root_angle', 'self_dual', 'z1'])
-lfunc_columns.dummy_download = True
+# lfunc_columns.dummy_download = True
```
roed314, line 352: This looks like an unrelated change that accidentally got included in this PR (certainly you should change the name of the class from `G2C_download` to `Lfunc_download` eventually).

lmfdb/utils/web_display.py
```diff
@@ -783,3 +784,14 @@ def list_to_latex_matrix(li):
...
     return mm
+
+
+def dispZmat_from_list(a_list, my_dim):
+    r"""Display a matrix with integer entries from a list
+    """
+    num_entries = len(a_list)
+    assert num_entries == my_dim ** 2
+    output = []
+    for i in range(my_dim):
```
roed314, line 795: This won't work.  You need `range(0,my_dim**2,my_dim)`.  I might also just call it `dim` rather than `my_dim` for simplicity.

AndrewVSutherland:
We should add code snippets to create the adelic image as a subgroup of GL(2,Z/NZ) using the generators (at least for Sage and Magma).  This could be done in a later PR but would be easy to add to this PR.

lmfdb/elliptic_curves/templates/ec-curve.html
BarinderBanwait: This extra `data` is used in many other places in this file, eg:

https://github.com/LMFDB/lmfdb/blob/bbadf384c6306d82de37fd7b13b3ee2094cd0106/lmfdb/elliptic_curves/templates/ec-curve.html#L445

It seems to arise when a database lookup is done from another table. In the above line, `data.adelic_level` works because `adelic_level` is a column in the `ec_curvedata` table, but when we add other things in the data blob in `web_ec.py`, it becomes `data.data`. I wouldn't know how to get around this ...

AndrewVSutherland: This is a known issue with how web_ec is implemented, changing it would require a lot of changes that are outside the scope of this PR.  I'm not sure it is worth doing, but if it is, I think it should be done in a separate PR.

lmfdb/elliptic_curves/web_ec.py
BarinderBanwait: done (now called `adelic_gens_latex` to make clear it's a string)

lmfdb/lfunctions/main.py
BarinderBanwait: yes indeed, this was from another thing ... removed now

lmfdb/utils/web_display.py
BarinderBanwait: oops ... yes indeed, sorted now

tox.ini
```diff
@@ -30,4 +30,4 @@ commands =
     # E111 indentation is not a multiple of four
     # E702 multiple statements on one line (semicolon)
     # W391 blank line at end of file
-    pycodestyle --select=E722,E711,E702,W391 lmfdb/
```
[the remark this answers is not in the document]
BarinderBanwait: that was something @edgarcosta suggested I do (but that might have been intended only for the modcurve branch ...)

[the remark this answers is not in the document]
edgarcosta: Because, autopep8 will fix it, and it is not that relevant to annoy PR authors

BarinderBanwait:
<working on adding code snippets ...>

BarinderBanwait:
Code snippets have been added to pink, but somehow the latest changes (even from last week) are not showing on [pink.lmfdb.xyz](pink.lmfdb.xyz)? @roed314 they show when i run pink locally

edgarcosta:
Please use olive or purple.

PR #5264, opened 2022-11-15
Workshop page update

jwj61:
Update workshop page


lmfdb/templates/workshops.html
```diff
@@ -6,6 +6,35 @@
 
 <ul>
 
+<li>
+Modular curves workshop 2, <br>
```
roed314, line 10: Might as well include a link to the website: https://math.mit.edu/~edgarc/MCW2.html

PR #5265, opened 2022-11-15
Add label for group descriptions

jwj61:
The first piece of information for an abstract group or subgroup is a descriptive name for the group (which does not always uniquely identify the group).  It was pointed out that this information did not have a label as to what it is, which is unusual for the site.  This adds labels for it.

### DIFF
--- a/lmfdb/groups/abstract/templates/abstract-show-subgroup.html
+++ b/lmfdb/groups/abstract/templates/abstract-show-subgroup.html
@@ -6,7 +6,7 @@
 <h2> Subgroup ($H$) information</h2>
 <p>
   <table>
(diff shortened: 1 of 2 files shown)


lmfdb/groups/abstract/templates/abstract-show-subgroup.html
```diff
@@ -6,7 +6,7 @@
 <h2> Subgroup ($H$) information</h2>
 <p>
   <table>
-    <tr><td><a href="{{url_for('.by_label', label=seq.subgroup)}}">${{seq.subgroup_tex}}$</a></td><td>  </td></tr>
+    <tr><td>{{KNOWL('group.name', title='Description:')}}</td><td><a href="{{url_for('.by_label', label=seq.subgroup)}}">${{seq.subgroup_tex}}$</a></td></tr>
```
roed314, line 9: We should also add the knowl for the other places names show up (the ambient group and quotient group on the subgroup page, for example).

PR #5276, opened 2022-12-05
slow bad bots

edgarcosta:
You can test this by using developer mode on your browser and changing your user agent.


JohnCremona:
> Test failure is from separate BMF data update.

Feel free to fix by merging my PR :)

PR #5280, opened 2022-12-15
Nonsurjective primes for g2c

BarinderBanwait:
Currently, the homepage of each genus 2 curve lists information about the mod 2 and mod 3 Galois images.

This PR augments this information with the list of primes p for which the mod-p image is not surjective, whenever this information is available. Information about whether this smaller image is explained by the existence of torsion is also shown. See [603.a.603.1](https://olive.lmfdb.xyz/Genus2Curve/Q/603/a/603/1) for an example.

Since labels for mod-p images have not been computed for p > 3, we say only that the image is 'not computed' for these larger primes. All available information is also presented in a **single** table, as discussed during last friday's LMFDB Friday with @AndrewVSutherland and @shiva-chid .

This nonsurjective information is available whenever the geometric endomorphism ring is Z; so for any other larger endomorphism ring, we explicitly say that such information is not available for primes > 3, eg [294.a.8232.1](https://olive.lmfdb.xyz/Genus2Curve/Q/294/a/8232/1)

This is all live on olive (see above links). How does this look? If it's fine, then I'll add some tests.


AndrewVSutherland:
Three comments:
- The header "Is torsion prime?" in the galrep tabel needs a knowl.
- I think `nonmax_primes` should be a column in g2c_curves so that we can search on it, and I think `torsion_primes` should be computed on the fly as `torsion_order.prime_divisors()`.  The table `g2c_nonmaximal_test` is then superfluous and can be removed (which would fix the other thing I would complain about, which is that you don't see the data in this table when you click the "underlying data" link).
- Once `nonmax_primes` is a column in `g2c_curves` it should be added as a display column in search results.  We will eventually want to add an option to search on it, as well as to search on a particular mod-2 or mod-3 Galois image, as we can for elliptic curves, but I think that can/should wait for a separate PR.

BarinderBanwait:
Latest commit addresses your comments @AndrewVSutherland ; in particular:

- The header "Is torsion prime?" now shows the `torsion order` knowl, which I modified to define `torsion primes`;
- On the backend, I added a new column to `g2c_curves` called `non_maximal_primes`, and removed the table `g2c_nonmaximal_test'
- The nonmaximal primes can be viewed in search results, by selecting it in the optional "columns to display" drop-down

I have raised a separate issue #5291 for the searching on nonmaximal primes

AndrewVSutherland:
This looks good except for one tiny issue: the column non-maximal primes should be listed before the invariants and equation columns in the search results show columns drop down (we want these wide columns at the end so they can extend off the screen to the right).  I might put it after Sato-Tate group (which is roughly where it is on the EC pages).

BarinderBanwait:
@AndrewVSutherland done

PR #5285, opened 2022-12-16
adelic level/index and mod-m image search

AndrewVSutherland:
Adds search options for adelic level and index, and the ability to enter a modular curve label into the Galois image search box (note: the search results will include all elliptic curves that have that image, which is not the same as all elliptic curves that correspond to points on the modular curve X_H, since these could correspond to elliptic curves whose mod-m image is strictly contained in H).

There are search display columns for adelic level, index, and mod-m images.  The mod-m images column is now the rightmost column (after equation) because it can be quite long (and it will get longer, at the moment only images with level <= 23 are shown, but that will change soon).

The mod-m image levels don't currently have knowls attached to them the way the mod-ell and ell-adic images do, I will add these later in a separate PR.


AndrewVSutherland:
We should hold off merging this until I update the labels in the `modm_images` column, which will happen tomorrow.

PR #5287, opened 2022-12-17
Base change2

JohnCremona:
This displays the newly uploaded data about base change of elliptic curves from base fields other than Q (as well as base change from Q as before).

The properties box now only says "yes" or "no" for base change.  The Base change section at the bottom of a curve's home page now display base change information as a table, with links when the downstairs curve is in the database and a suitable message otherwise (so even then those curves are over Q there is more informatin than before).

Examples:

http://localhost:37777/EllipticCurve/6.6.453789.1/189.1/b/1 : cyclic sextic base field, curve is b.c. from Q so also of curves over the quadratic and cubic subfields

http://localhost:37777/EllipticCurve/6.6.820125.1/289.1/g/1 : sextic base field, curve is b.c. from a cubic subfield (but not Q)

http://localhost:37777/EllipticCurve/6.6.820125.1/64.1/b/1 : sextic base field, b.c. from quadratic subfield

http://localhost:37777/EllipticCurve/4.4.13824.1/24.1/b/6 : quartic base field with one quadratic subfield, base change from that and from Q

http://localhost:37777/EllipticCurve/4.4.1600.1/400.1/a/5 : biquadratic base field, b.c. from Q(x4) and from three quadratic fields (x2 each)

Use the random faility for more examples.  Note that with the current data without this patch, you might ask for base-change curves only but see curves which say that they are not base change.  The knowl for base change suggests that "base change" is short for "base change from Q" because that matched the old data;  I think it should be changed (after this) to being short for "base change from a proper subfield".

I also think that the search box labelled "Q-curves" should be changed to something like "Q-curves / base change", as I suggested recently (but have not put into this PR).  I still find myself looking for a search box called "base change" for several seconds before I remember that it is hidden behind "Q-curves".

### DIFF
--- a/lmfdb/ecnf/WebEllipticCurve.py
+++ b/lmfdb/ecnf/WebEllipticCurve.py
@@ -11,7 +11,7 @@ from lmfdb.number_fields.web_number_field import WebNumberField, field_pretty
 from lmfdb.lfunctions.LfunctionDatabase import (get_lfunction_by_url,
                                         get_instances_by_Lhash_and_trace_hash)
 from lmfdb.sato_tate_groups.main import st_display_knowl
-from lmfdb.elliptic_curves.web_ec import conductor_from_label
+from lmfdb.elliptic_curves.web_ec import conductor_from_label, cremona_label_to_lmfdb_label
 
 # The conductor label seems to only have three parts for the trivial ideal (1.0.1)
 # field 3.1.23.1 uses upper case letters for isogeny class
@@ -691,6 +691,7 @@ class ECNF():
 
         if not self.base_change:
             self.base_change = []  # in case it was False or None instead of []
+        self.nbc = len(self.base_change)
 
         # add base_change yes/no to Properties box
         if self.base_change:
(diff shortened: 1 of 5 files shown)


AndrewVSutherland:
Looks good, I agree the base-change knowl should change once this is merged.  Regarding the search options, I think we should address this in combination with switching to and advanced search options display, as we (and @alexjbest) discussed elsewhere.  You can see an example of what this might look like for EC/Q on the modular curves branch at https://red.lmfdb.xyz/EllipticCurve/Q/ (we really wanted the adelic image search options to be there but this pushed the complexity of the the search/browse page over the edge, and the same thing will happen to ECNF (and perhaps has already).

On a completely separate note, the page https://www.lmfdb.org/EllipticCurve/6.6.820125.1/289.1/g/1 makes me wonder about the wisdom of showing the conductor in the properties box for on the ECNF pages.  What if we just showed the conductor norm?  The conductor ideal is of course available on the page and I don't see much benefit to having it in the properties box (and there is the obvious non-benefit of making the page formatting more unwieldly).

JohnCremona:
I agree about having only conductor norm in the properties box. That's an easy change, I'll do it

PR #5288, opened 2022-12-21
Belyi plane equations

SamSchiavone:
Added new plane model equations for Belyi maps. The changes can be viewed at https://purple.lmfdb.xyz/Belyi/ .

I seem to be missing a few maps from the old table, notably 7T7-6.1_5.2_4.2.1-a, which is causing a test to fail, so I'll look into that. In the meantime, I'm happy to get any feedback on the new equations and layout. There is a button which toggles between the new plane equations and the old equations.


AndrewVSutherland:
@SamSchiavone Looks like tests are still failing.

edgarcosta:
the test that fails is:
```
     def test_orbit_size_search(self):
>       self.check_args("/Belyi/?orbit_size=20-", "7T7-6.1_5.2_4.2.1-a")
```

SamSchiavone:
Yes, I haven't had a chance to fix the data (and probably won't until after JMM). Yes, that's one of the maps that's missing---that test failing alerted me to this fact.

AndrewVSutherland:
Any update on this?  If it is going to be a while longer I would suggest closing the PR and resubmitting when it is ready.

SamSchiavone:
Hi Drew, I'm planning to work on this Friday. I figured out the bug in my upload script, so I think I should be able to fix it then.

PR #5303, opened 2023-01-02
hide conductors from property box (ECNF), keeping conductor norm

JohnCremona:
As suggested in #5287.  For example, compare https://www.lmfdb.org/EllipticCurve/6.6.820125.1/289.1/g/1 with http://localhost:37777/EllipticCurve/6.6.820125.1/289.1/g/1

The compplete conductor, as an ideal, is still displayed on the page.


AndrewVSutherland:
I think this looks much better, thanks!

PR #5307, opened 2023-01-05
fix for Issue 5306 (BMF banana)

JohnCremona:
Test with http://localhost:37777/ModularForm/GL2/ImaginaryQuadratic/banana/banana

### DIFF
--- a/lmfdb/bianchi_modular_forms/bianchi_modular_form.py
+++ b/lmfdb/bianchi_modular_forms/bianchi_modular_form.py
@@ -310,16 +310,17 @@ def bmf_field_dim_table(**args):
 def render_bmf_space_webpage(field_label, level_label):
     info = {}
     t = "Bianchi modular forms of level %s over %s" % (level_label, field_label)
-    bread = get_bread([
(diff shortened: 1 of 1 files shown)


lmfdb/bianchi_modular_forms/bianchi_modular_form.py
```diff
@@ -310,16 +310,17 @@ def bmf_field_dim_table(**args):
...
-    bread = get_bread([
-        (field_pretty(field_label), url_for(".render_bmf_field_dim_table_gl2", field_label=field_label)),
-        (level_label, '')])
+    bread = get_bread()
     friends = []
     properties = []
     downloads = []
 
     if not field_label_regex.match(field_label):
-        flash_error("%s is not a valid label for an imaginary quadratic field", field_label)
```
edgarcosta, line 321: I don't understand the benefit of removing the flash_error.
This prevents cross script  attacks.

JohnCremona:
OK, I will put it back.  You have to set info['err'] or it tries to render the page and that fails.

JohnCremona:
This is better now -- has the flash error and rediectes to the BMF index page as with other similar errors.

Thanks @edgarcosta

Issue #5320, opened 2023-01-27
Test failure in local fields


AndrewVSutherland:
The top slope test is currently failing (presumably due to data changes):

```
    def test_search_top_slope(self):
        L = self.tc.get('/padicField/?p=2&topslope=3.5')
>       assert '81' in L.get_data(as_text=True) # number of matches
```

The number of matches is now 1443.  But I'm not sure checking for the number of matches is a good idea, since this can so easily change, and isn't always included in the initial response when there are more than 1000.  Perhaps a better test would be to check that `2.4.9.1` (the first result with top slope 3.5) shows up.

PR #5321, opened 2023-01-27
Fix top slope test for local fields

jwj61:
Fix issue #5320 as suggested


jwj61:
Tests passed, and this is a change of tests

PR #5323, opened 2023-01-30
Fix multiplicity in aut diagram

jwj61:
In a group like

`http://beta.lmfdb.org/Groups/Abstract/144.111
http://127.0.0.1:37777/Groups/Abstract/144.111

there are 5 conjugacy classes of subgroups of order 2 each containing 3 subgroups, which are all in one class up to automorphism.  In the diagram of subgroups up to automorphism, the count should be 15, and this fixes it.

In the group D_4,

http://beta.lmfdb.org/Groups/Abstract/8.3
http://127.0.0.1:37777/Groups/Abstract/8.3

subgroups of order 2 fall into 2 classes up to automorphism, one class has 2 conjugacy classes with 2 subgroups each, and the other has one conjugacy class with one subgroup.

There should be no changes to the conjugacy class diagram, or diagrams when we only store subgroups up to automorphsim.  An example of the latter is

http://127.0.0.1:37777/Groups/Abstract/32.51
http://beta.lmfdb.org/Groups/Abstract/32.51

### DIFF
--- a/lmfdb/groups/abstract/web_groups.py
+++ b/lmfdb/groups/abstract/web_groups.py
@@ -1119,6 +1119,23 @@ class WebAbstractGroup(WebObj):
             edges = [list(edge) for edge in edges]
         return [nodes, edges]
 
+    # Figuring out the subgroup count for an autjugacy class might not be stored
+    # directly.  We do them all at once.  If we only computed up to aut
+    # return {} since we don't need this
+    # output is a dictionary of aut_labels and counts
+    @lazy_attribute
+    def aut_class_counts(self):
+        counts = {}
+        if self.outer_equivalence:
+            return counts
+        subs = self.subgroups
+        for s in subs.values():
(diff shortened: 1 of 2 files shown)


lmfdb/groups/abstract/web_groups.py
```diff
@@ -1119,6 +1119,23 @@ def subgroup_lattice_aut(self):
...
+    def aut_class_counts(self):
+        counts = {}
+        if self.outer_equivalence:
+            return counts
+        subs = self.subgroups
+        for s in subs.values():
+            if s.aut_label in counts:
+                counts[s.aut_label] += s.count
+            else:
+                counts[s.aut_label] = s.count
```
roed314, line 1136: Works for me, though I'll note that `from collections import Counter` makes this code cleaner.

jwj61:
I will roll the code simplification into my next set of changes related to groups.

PR #5330, opened 2023-02-09
Lots of minor changes

jwj61:
This contains a bug fix: the number of conjugacy classes for

http://beta.lmfdb.org/Groups/Abstract/27.1
http://127.0.0.1:37777/Groups/Abstract/27.1

had some numbers in the wrong position.

It also starts to uniformize pages.  It lists composition factors for all groups:

http://127.0.0.1:37777/Groups/Abstract/ab/9.9.9.9.9.9  (on-the-fly abelian group)
http://127.0.0.1:37777/Groups/Abstract/1920.240418 (on-the-fly non-solvable group)
http://127.0.0.1:37777/Groups/Abstract/512.402873  (on-the-fly solvable nonabelian group)

 lists the correct number of irreducible complex representations for on-the-fly abelian groups (see the 9.9.9.9.9.9 example above), says data not computed for on-the-fly groups for numbers of divisions and autjugacy classes, and lists the transitive permutation degree for on-the-fly abelian groups (now says data not computed for on-the-fly nonabelian groups), always lists permutation degree in the properties box, always includes rank in the properties box (says not computed when that's the case).

There will be more uniformization changes, but I don't want bundle too many together.

### DIFF
--- a/lmfdb/groups/abstract/templates/abstract-show-group.html
+++ b/lmfdb/groups/abstract/templates/abstract-show-group.html
@@ -143,8 +143,15 @@
       {% else %}
       <tr>
         <td>{{KNOWL('group.division', 'Divisions')}}</td>
-          <td  colspan="{{gp.order_stats|length}}">data not computed</td>
-          <td class="border-left"></td>
+          {% if gp.division_stats %}
+            {% for c in gp.division_stats %}
+              <td>{{ c[1] }}</td>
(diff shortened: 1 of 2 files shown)


edgarcosta:
LGTM 
@roed314, any opinions?

roed314 (approved):
Looks good after a couple fixes.

Issue #5345, opened 2023-02-27
Number field ramified primes


jvoight:
In the properties box, does it always just say "see page"?  Even
  http://www.lmfdb.org/NumberField/2.0.4.1
says this.  If so, why do we bother saying anything there at all?

jwj61:
I had no idea we did that at all!  I am sure it was put in to save space when there are really large primes

jvoight:
Sorry for picking on such a small thing, I just thought it was funny.  :)

jwj61:
It should be changed.  It is kind of ridiculous now.

PR #5347, opened 2023-02-27
Fix issue #5345

jwj61:
It was always saying 'see page' for ramified primes in the property box.  This is because the string for ramified primes had javascript code for toggling raw mode.  We just reset the variable.

Fixed page:

http://127.0.0.1:37777/NumberField/2.0.4.1
http://beta.lmfdb.org/NumberField/2.0.4.1

Still says see page:

http://127.0.0.1:37777/NumberField/43.1.11615818670249741926786783635700464005809046848845495205636184619664502520703084303881154330624.1


edgarcosta:
LGTM, will merge once tests pass

PR #5353, opened 2023-02-28
Gp code ec

pascalmolin:
Added GP code for elliptic curves

and changed a little bit the code rendering mechanism so that more info is put in the yaml file


edgarcosta (approved):

edgarcosta:
LGTM

Issue #5348, opened 2023-02-27
Uneditable knowls


AndrewVSutherland:
If you go to https://beta.lmfdb.org/ModularForm/GL2/ImaginaryQuadratic/2.0.7.1/10000.1/b/ when you are logged in and try to edit the top knowl you will receive the error message
```
Label 'bianchi.2.0.7.1-10000.1-b' contains characters not allowed by knowl database; updated allowed_id function or change label scheme
```
If you go directly to https://beta.lmfdb.org/knowledge/show/mf.bianchi.2.0.7.1-10000.1-b.top, and try to edit the knowl, the same error appears.

roed314:
This was fixed by #5355.

PR #5355, opened 2023-02-28
Fix annotation re

roed314:


roed314:
Fixes #5348

lmfdb/knowledge/main.py
```diff
@@ -41,7 +41,7 @@
 
 # know IDs are restricted by this regex
 allowed_knowl_id = re.compile("^[a-z0-9._-]+$")
-allowed_annotation_id = re.compile("^[a-zA-Z0-9._-~]+$")  # all unreserved URL characters
+allowed_annotation_id = re.compile(r"^[a-zA-Z0-9._\-~]+$")  # all unreserved URL characters
```
edgarcosta, line 44: While `\` is not reserved, it is weirdly treated.

roed314: The problem is that `-` has a special meaning inside brackets, so this previously was including all the characters between `_` and `~`.  Instead, we want the three characters `_`, `-` and `~`, which is achieved by escaping `-`.

PR #5360, opened 2023-03-01
Make elliptic curve pics transparent

davidlowryduda:


AndrewVSutherland:
I like this, but I think others should comment, it does reduce the contrast and may make some of the ECNF pictures slightly harder to read (I think that's okay, the pictures are not meant to convey precise mathematical content, and we may eventually want to add the ability to provide a larger picture that could be shown on a separate page with a white background).

Also, we should do the same thing on the genus 2 and Sato-Tate group pictures.

roed314:
Would it be possible to change the colors used in the totally real case to improve the contrast?  I agree that the yellow isn't great (also note that the actual pictures are a bit smaller and thus less readable):

(with PR)
Screen Shot 2023-03-01 at 11 58 41 AM

(previously)
Screen Shot 2023-03-01 at 12 02 11 PM

edgarcosta:
thicker lines might help with the contrast

alexjbest:
I like this change on the code side, even if for display we still set the background colour in css, as it makes it easier for users to copy past pictures for reuse

AndrewVSutherland:
I think my vote would be to merge this (and also make corresponding changes to the genus 2 and Sato-Tate portraits).  We can then separately think about whether to make changes to individual portraits (e.g. ECNF) to improve readability.  But if anyone disagrees with this, please say so.

I don't think we want to set the background color globally in css, because this will impact portraits like CMFs and Maass forms that already have transparent backgrounds and we really don't want to add a background to them.

alexjbest:
Well you can set the colour only for one type of portraits in css. My point was really that I think this should be merged even if people like white backgrounds on the screen

roed314:
Merging it is fine with me.

PR #5376, opened 2023-03-02
Change eyeball icon for raw/typeset toggle

AndrewVSutherland:
This PR changes the icon that is used for the button that toggles between raw/typeset mode on pages like

https://teal.lmfdb.xyz/EllipticCurve/Q/26400/d/1
https://teal.lmfdb.xyz/NumberField/10.2.6802444800000000000000000.3

compare to

https://www.lmfdb.org/EllipticCurve/Q/26400/d/1
https://www.lmfdb.org/NumberField/10.2.6802444800000000000000000.3


edgarcosta:
I prefer the eyeball, to be honest. The ~RAW~ is quite hard to read

AndrewVSutherland:
I'd welcome feedback from others, I'm used to the the eyeball and probably prefer it as well, I created this PR in response to feedback we got at COUNT (a few people complained that they didn't know what the eyeball meant and that our use of it was nonstandard).  Possibly just keeping the tool-tip that I added (hover over the raw icon to see it) would be enough.

edgarcosta:
If there is a Show Commands, we could perhaps add "/Raw"

JohnCremona:
I was one who did not much like the eyeball, but I don't like the new one either.

Would there be room for a line in the top left under "Show commands" to say "Formula display style: raw/typeset" with some kind of toggle button?

If not, it is not too important.

jwj61:
I don't have a strong opinion, but prefer the eyeball to Raw, especially with the slash through it.

AndrewVSutherland:
I like the idea of adding "/ Raw" (and maybe then change "Show commands" to just "Show").   But we still need an icon to show when the users hovers over typeset text that can be shown in raw mode (alongside the copy/paste icon).  But maybe it would be okay to keep using the eyeball when you hover (I think the eyeball icon is jarring when it is always present at the top of the screen, and without any context it might not be clear what it does, but having it pop up next to the copy/paste icon seems less ambiguous and less intrusive).

One other suggestion was is to use the icon: [</>](https://fontawesome.com/icons/code?s=solid&f=classic), which is sometimes used to show html source (but its not clear what icon one would change this to).

AndrewVSutherland:
I'm reverting this back to the eyeball so the only change in this PR is the addition of hte tool tip, which might at least help prevent confusion.  I encourage anyone who is bothered by the aesthetics to submit an alternative PR (keeping in mind that the eyeball is used in two places).  I'll merge this one as soon as tests pass.

PR #5381, opened 2023-03-02
Add magma code snippets for most of Galois group pages

jwj61:
Go to any Galois group page to see them


jwj61:
Don't understand why tests stopped running

PR #5385, opened 2023-03-03
Save input names in data-name on submit

### LINKED ISSUE #5346: Random number field sometime ignores CM
If you go to https://www.lmfdb.org/NumberField set CM to yes and click the random button, you will often get a field that is not a CM field.  Note that this can be tricky to reproduce, you may need to first to a random search without CM set, then set it and click random again.

Reproduced in firefox on windows and in chrome on linux

roed314:
Fixes #5346.

To test, on www.lmfdb.org go to any page, enter a search query, click the "back" button to get back to the browse page, and then note that inputs in other search boxes are ignored (whether you're doing a normal search or looking for a random page).

The behavior should be fixed with this PR merged.

### DIFF
--- a/lmfdb/static/lmfdb.js
+++ b/lmfdb/static/lmfdb.js
@@ -524,10 +524,18 @@ function cleanSubmit(id)
   var allInputs = myForm.getElementsByTagName('input');
   var allSelects = myForm.getElementsByTagName('select');
   var item, i, n = 0;
+  // We need to reload saved names from their data-name attributes, since pressing the back button would otherwise return us to a broken state.
+  for(i = 0; item = allInputs[i]; i++) {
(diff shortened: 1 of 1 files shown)


alexjbest:
When I looked at this I thought the cleanest (no pun intended) way to go is actually to use html formdata rather than custom javascript to clean the form.
I'll try an alternative pr tonight and we can compare the approaches then.
Feel free to merge this now if you want of course, but just wanted to mention a different proposal

AndrewVSutherland:
I'm happy to wait for your PR @alexjbest (thank you for looking at this!), but please let us know if it's going to be more than a day or two, I'd like to get a fix for this into production soon.

roed314:
Closing in favor of #5386, though we can reopen if @alexjbest needs some more time for that one.

PR #5386, opened 2023-03-04
remove cleansubmit in favour of a formdata handler

### LINKED ISSUE #5346: Random number field sometime ignores CM

alexjbest:
Instead of using js to change the html of a webpage to force some search options not to appear in the url, we use the more canonical method of setting the formdata via a callback, this prevents weird bugs when using the back button

We enable this method on all re-search and search forms.

alternative to #5385 fixes #5346

I'm not sure what the setting all=1 thing was meant to accomplish, if someone can tell me what to do to test that that would be great.

https://developer.mozilla.org/en-US/docs/Web/API/FormData

@roed314 @AndrewVSutherland


lmfdb/static/lmfdb.js
```diff
@@ -518,40 +518,6 @@ function resetStart()
...
-        n++;
-      }
-    }
-  }
-  if (!n) {
-    var all = document.createElement('input');
-    all.type='hidden';
-    all.name='all';
-    all.value='1';
-    myForm.appendChild(all);
```
alexjbest, line 551: which pages is this relevant for now?

AndrewVSutherland: I think that was just there in order to make things work properly when the user didn't enter any selections (it is only set in that case).  I think you should be able to remove it now (the thing to test would simply be that searches with no selections work).

lmfdb/static/lmfdb.js
```diff
@@ -722,3 +688,31 @@ function hide_schema(tbl) {
   $("#"+tbl+"-schema-shower").show();
   return false;
 }
+
+
+/* add handler for search forms to clean their own
+   form data and remove keys for empty (default) values */
+$(document).ready(function () {
+  let forms = [document.getElementById('search'),
+               document.getElementById('re-search')];
```
roed314, line 697: There are several examples where you deleted `cleanSubmit` but the form was named something else (e.g. `hgmsearch` or `knowl-search`).  You should either rename the forms or add other options here.

roed314:
I agree that this overall solution is nicer than mine.  Thanks!

lmfdb/static/lmfdb.js
alexjbest: Thanks, i think renaming the search forms makes more sense, or we could add a class `clean` that the js uses to find which forms to clean? Downside of that is people have to remember to add it, if all search forms want this functionality anyways naming them all search seems simpler

lmfdb/static/lmfdb.js
alexjbest: Yeah i tried a couple and it seemed that the search type = List was always set so that seems ok. I'll try a search on every subdirectory of the lmfdb later

lmfdb/static/lmfdb.js
roed314: I think just renaming the forms to `search` or `re-search` in each case makes sense.  If people want to use this functionality on another kind of form, they can add the id to the list here.

edgarcosta:
I will fix the `id`s, indeed I will convert them to classes

alexjbest:
lgtm now, thanks edgar

PR #5401, opened 2023-03-10
Downloads for Belyi plane equations

SamSchiavone:
Add new plane equations to downloads. Add test for plane equations.


edgarcosta:
You will need to merge with upstream master for the tests to pass

PR #5405, opened 2023-03-11
Disable test

jwj61:
Test is currently broken because of a change in the data.  The data may change again.  When finalized, the test will be reinstated.


AndrewVSutherland:
Added a small change so that the PR to make tests pass passes tests (lint was complaining about an unused variable).

PR #5427, opened 2023-03-25
Use 'minimal sibling of splitting field' on mod-l galrep pages

AndrewVSutherland:
Minor tweak to mod-l galrep pages to clarify that the number field used to represent the splitting field is the minimal sibling.


JohnCremona (approved):
LGTM

PR #5429, opened 2023-03-26
Add Galois closure to dynamic number field knowls

jwj61:
A good place to look for this is on degree six number field pages because the twin sextic algebra and sibling data hit the various options:

 - the field is Galois, as in one of the cubic factors of the twin on http://127.0.0.1:37777/NumberField/6.0.9747.1
 - the field is not Galois, but its Galois closure is in the database (the other cubic in the twin algebra on the same page)
 - the Galois closure has degree <48 so it might be in the database, but we haven't computed it (the degree 9 sibling on the same page)
 - the Galois closure has degree which is too big for our database, as in the degree 10 sibling on http://127.0.0.1:37777/NumberField/6.0.14731.1

If the Galois closure has been computed, there are two options (in the database or not), but this is handled by the number field knowls.

### DIFF
--- a/lmfdb/number_fields/web_number_field.py
+++ b/lmfdb/number_fields/web_number_field.py
@@ -340,11 +340,12 @@ def nf_knowl_guts(label):
     if wnf.is_galois():
         galstring = r'this field is Galois over $\Q$'
     else:
+        galord = db.gps_transitive.lookup('%dT%d'%(wnf.degree(), wnf.galois_t()), 'order')
         res = wnf.resolvents()
         if 'gal' in res:
             galstring = formatfield(string2list(res['gal'][0]))
         else:
-            galstring = 'data not computed'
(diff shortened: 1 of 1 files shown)


AndrewVSutherland:
Looks good, my only suggestion (if its easy to do) would be to include the degree of the Galois closure in all non-computed cases (whether it is within the range of the LMFDB or not).  I think we could probably also show the (factored) discriminant of the Galois closure (but that might be more work).

jwj61:
Not including the degree is easy, and I will do that.

Adding the discriminant is significantly harder.  We don't store that data so it would have to be computed first, and even then, my programs won't do some cases.

AndrewVSutherland:
So actually I was suggesting that we should always show the degree in the uncomputed cases, e.g. it would say something like "degree 18 extension of Q not in the database" on http://127.0.0.1:37777/knowledge/show/nf.field.data?label=9.3.1270238787.1

PR #5435, opened 2023-03-28
Ramification polygons

roed314:
This update adds ramification polygons to p-adic field homepages (together with some other invariants, like residual polynomials, associated inertia and indices of inseparability).  It also allows for searching based on slopes, visible slopes, indices of inseparability and associated inertia.

It's running on [olive](https://olive.lmfdb.xyz/padicField/) for testing.


edgarcosta:
@roed314, did this imply some change on the database? 
code against proddb seems to fail https://github.com/LMFDB/lmfdb/actions/runs/4558491468/jobs/8042053534#step:11:273

roed314:
Yes, lf_fields needs to be pushed to prod, though we should probably check with @jwj61 since he was changing defining polynomials.  John, is it okay to push lf_fields to production?

edgarcosta:
and then the next question is, do we need to push the code first?

jwj61:
Yes it is ok to push the database.  There will be future changes to the data, but I believe that the current data is correct.

I believe data first, then code.

edgarcosta:
data has been pushed

PR #5441, opened 2023-04-07
Add discriminant root field for number fields

jwj61:
There was a request for adding some classical resolvents (as opposed to all resolvents) for number fields.  This adds the discriminant root field since it can be done without adding to the database.  It is something which is already on pages for p-adic fields.  It adds it as a number field knowl near the top of the page, and as a friend if it is in the database.  The main cases are:

It is a quadratic field in the database:
http://127.0.0.1:37777/NumberField/6.0.10571.1
It is a quadratic field not in the database:
http://127.0.0.1:37777/NumberField/47.1.31280045577397422652719953713309913701940250885770507510815871257152976174188094627629760512.1
It is Q:
http://127.0.0.1:37777/NumberField/4.0.23104.1

While adding the friend, I updated the friend links for L-function and Galois group to use url_for.

### DIFF
--- a/lmfdb/number_fields/number_field.py
+++ b/lmfdb/number_fields/number_field.py
@@ -604,7 +604,6 @@ def render_field_webpage(args):
         # note that the first degree 4 number field missed the zero of the zeta function
         if abs(D**n) < 50000000:
             info['friends'].append(('L-function', url_for('l_functions.l_function_nf_page', label=label)))
-            #info['friends'].append(('L-function', "/L/NumberField/%s" % label))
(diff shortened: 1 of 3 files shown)


jwj61:
I will add, the other classical resolvent is the resolvent cubic algebra of a quartic. People are free to comment here as to whether they feel this is worth it or not.

JohnCremona:
+1 from me .  There's a simple formula for the cubic resolvent of a quartic but finding in the database would involve polredabsing that, so not trivial.

Thanks for the quadratic resolvent s!

PR #5449, opened 2023-04-16
Make labels more general for mod ell Galois representations

jwj61:
This is accomodate new data which, for now, uses a variation on the labels for mod-ell Galois representations (namely ones of dimension 1).


jwj61:
Tests passed, and this only affects a secret section, so self-merging.

Issue #5451, opened 2023-04-21
Galois Group jump box handling leads to server error


edgarcosta:
```
Exception on /GaloisGroup/ [GET]
Traceback (most recent call last):
  File "/home/sage/sage-9.7/local/var/lib/sage/venv-python3.8/lib/python3.8/site-packages/flask/app.py", line 2525, in wsgi_app
    response = self.full_dispatch_request()
  File "/home/sage/sage-9.7/local/var/lib/sage/venv-python3.8/lib/python3.8/site-packages/flask/app.py", line 1822, in full_dispatch_request
    rv = self.handle_user_exception(e)
  File "/home/sage/sage-9.7/local/var/lib/sage/venv-python3.8/lib/python3.8/site-packages/flask/app.py", line 1820, in full_dispatch_request
    rv = self.dispatch_request()
  File "/home/sage/sage-9.7/local/var/lib/sage/venv-python3.8/lib/python3.8/site-packages/flask/app.py", line 1796, in dispatch_request
    return self.ensure_sync(self.view_functions[rule.endpoint])(**view_args)
  File "/home/lmfdb/lmfdb-git-web/lmfdb/galois_groups/main.py", line 93, in index
    return galois_group_search(info)
  File "/home/lmfdb/lmfdb-git-web/lmfdb/utils/search_wrapper.py", line 171, in __call__
    data = self.make_query(info, random)
  File "/home/lmfdb/lmfdb-git-web/lmfdb/utils/search_wrapper.py", line 64, in make_query
    errpage = self.f(info, query)
  File "/home/lmfdb/lmfdb-git-web/lmfdb/galois_groups/main.py", line 197, in galois_group_search
    parse_galgrp(info, query, qfield=['label','n'],
  File "/home/lmfdb/lmfdb-git-web/lmfdb/utils/search_parsing.py", line 122, in __call__
    rval = self.f(inp, tmp, qfield, *args, **kwds)
  File "/home/lmfdb/lmfdb-git-web/lmfdb/utils/search_parsing.py", line 1063, in parse_galgrp
    gcs = complete_group_code(inp.upper())
  File "/home/lmfdb/lmfdb-git-web/lmfdb/galois_groups/transitive_group.py", line 615, in complete_group_code
    code = "[%s,%s]"%tuple(code.split("."))
TypeError: not all arguments converted during string formatting
[...]

PR #5452, opened 2023-04-21
Fix regex in parsing Galois group codes

jwj61:
Fixes issue #5451 

To test, put something like 5.1.442.1 in the jump box for Galois groups.  It should be rejected, but not give a server error.


edgarcosta:
LGTM

Issue #4924, opened 2022-01-20
Extend Dirichlet character orbit data


AndrewVSutherland:
Currently searches for Dirichlet characters of modulus > 10000 yield no results.  You can access individual pages by label, e.g. https://teal.lmfdb.xyz/Character/Dirichlet/10001/17, but you get less information than on pages like https://teal.lmfdb.xyz/Character/Dirichlet/10000/17 where the modulus is <= 10000.

Dirichlet character orbit data for all characters up to modulus 100000 has been computed and is available in the table [char_orbits](https://beta.lmfdb.org/api/char_orbits/).  The Dirichlet pages need to be updated to use this and use pari/gp to compute character specific data on the fly (which can now be done very efficiently via the interface in [TinyConrey.py](https://github.com/LMFDB/lmfdb/blob/master/lmfdb/characters/TinyConrey.py).  This would allow us to eliminate the tables [char_dir_orbits](https://beta.lmfdb.org/api/char_dir_orbits/) and [char_dir_values](https://beta.lmfdb.org/api/char_dir_values/).

This is a necessary precursor to extending the CMF pages beyond level 10000 (which is planned at least in weight 2).

BarinderBanwait:
@AndrewVSutherland just a clarification: your first reference to "char_dir_orbits" above should have been to [char_orbits](https://beta.lmfdb.org/api/char_orbits/) which you showed us yesterday?

Also, could I just check the requirements of this issue: the work here is essentially "just" a refactor of the code in the characters folder, and mostly in `web_character.py` and `main.py`? The user will still be able to compute character pages with less information for huge modulus like https://lmfdb.org/Character/Dirichlet/100000000000001/17 ? In particular, there should be no user-facing changes beyond the following?

1. the 10,000 cut-off above being increased to 100,000
2. possible performance increase

AndrewVSutherland:
@BarinderBanwait Yes, I should have written "char_orbits" above (now fixed), and yes, for characters of modulus > 100,000 there would be no change in the UI.  Also, I confirmed yesterday that it is feasible to go to 1,000,000 and plan to do this (but 100,000 is enough for now).

[...]

PR #5454, opened 2023-04-23
Remove char_dir_orbits and char_dir_values across codebase

BarinderBanwait:
This PR addresses #4924 

Specifically, this PR removes the tables `char_dir_orbits`and `char_dir_values` from the codebase, and replaces it with the new `char_orbits` table. Other quantities not in this smaller table are computed on the fly, which has been achieved by extending the PARI interface in `TinyConrey.py`.

At the moment, these two tables are still present in various files in the `lmfdb/verify` folder. This is a part of the codebase I'm totally unfamiliar with, and I'm unsure what to do with this. For example, in `lmfdb/verify/mf_subspace.py` one finds the `check_conrey_indexes` function, which checks that the conrey indexes in the mf_subspaces table agree with the galois_orbit column of the char_dir_orbits table. But if we're removing this latter table, and the new char_orbits table doesn't have galois_orbit data (these will be computed on the fly after this PR is merged), then how does one carry out the verification @AndrewVSutherland  ? Should this verification compute ALL galois orbits on the fly? Or (more likely) the mf_subspaces table will be updated (in a future PR) to remove conrey indexes, and these will be computed on the fly in the cmf folder.

There remains the task of refactoring and cleaning up the code in `web_character.py`, for which a separate ticket will be created once this PR is merged.


BarinderBanwait:
This is ready for review now. The 10,000 cutoff in the code has been replaced with 100,000, so now we have Galois orbit data for characters like `/Character/Dirichlet/99999/17`.

One thing to note is that rendering pages takes a bit longer, at least on my localhost running from legendre. This is particularly noticeable when rendering pages for Dirichlet Groups, e.g compare `/Character/Dirichlet/30/` between this branch and the main site.

This seems to be because querying `char_orbits` on modulus and first_index (in order to get the orbit of a Dirichlet character as suggested by @AndrewVSutherland  in a comment in #4924 ) seems to take about 2 seconds per query. Maybe this is because we are querying the table on fields that are not the index? In any case these 2 seconds per query add up fast, meaning that even computing the page with modulus 30 is painfully slow. I don't know how to improve this ...

AndrewVSutherland:
Possibly we just need to add an index, I'll take a look.

edgarcosta:
If the issue is indeed the slow queries, this should show up on the lot
when you run it locally.
Perhaps you need to lower the cutoff in config.ini to see it.

BarinderBanwait:
@edgarcosta 

```
2023-04-27 16:43:19,895 - SELECT "label" FROM "char_orbits" WHERE "modulus" = 30 AND "first_label" = '30.29' LIMIT 1 ran in  2.0534677505493164s
```

It's a similar time for each character in the orbit, 8 characters, so about 16 seconds to compute

AndrewVSutherland:
There were no indexes and sort order was not set (this code is using an entirely different table), I just fixed this and it should be fast now.

AndrewVSutherland:
In addition to test failure, the random button isn't working.  But it should be easier to debug now that it is fas.

AndrewVSutherland:
I'm about to get on a plane, but I can take a look again tomorrow.  If you spot anything else that looks slow let me know, everything should be fast (faster than it was even when we only went to modulus 10,000) and if it isn't there should be an easy fix.

BarinderBanwait:
Thanks @AndrewVSutherland it is indeed much faster now! The last commit fixed the jump and random character issues, they are now working

edgarcosta:
If the pages are still slow, there is a profiler option when you run lmfdb
locally, that will tell where most time is being spend.

AndrewVSutherland:
Loading pages for characters of large conductor and order is very slow, and I don't think this is a database issue, it probably has to do with calculations that are being done on the fly, can you take a look at http://127.0.0.1:37777/Character/Dirichlet/99997/hj for example?

BarinderBanwait:
Using the profiler mentioned by @edgarcosta (via `--enable-profiler`) the time is taken in computing the kernel field polynomial, which (despite what that page says) is actually computed. The page says it is "not computed" because another function has hardcoded a limit degree of 100, where if the kerpoly has degree larger than this, then it's not even displayed.

So I propose the following change: only compute the kerpoly if the degree is <= 100. This should fix this timing issue, and not have any user facing change. The next commit will do this.

BarinderBanwait:
Regarding the use of `char_dir_orbits` and `char_dir_values` in the `verify` code: I spoke to @roed314 on Friday, after which I'm going to take the following strategy:

1. Remove the files `char_dir_orbits.py` and `char_dir_values.py`
2. Remove the specific `check_*` functions from `mf_subspaces.py` that reference one of these tables
3. Open an issue listing which specific `check_*` functions have been removed from which verification files, and leave it to future work to figure out how to verify these in the new schema, using `char_orbits`, and however the mf tables end up looking. 

This will mean that, once this PR is merged, there will be no mention of  `char_dir_orbits` and `char_dir_values` in the codebase.

AndrewVSutherland:
So now the pages for characters of larger order are fast, but when the order is below 100 they are very slow, much slower than the current pages, e.g. compare http://127.0.0.1:37777/Character/Dirichlet/199/k to https://www.lmfdb.org/Character/Dirichlet/199/k

edgarcosta:
this is now running on olive.lmfdb.xyz

lmfdb/number_fields/web_number_field.py
```diff
@@ -271,12 +271,17 @@ def formatfield(coef, show_poly=False, missing_text=None, data=None):
     if isinstance(coef, str):
         coef = string2list(coef)
     if data is None:
-        thefield = WebNumberField.from_coeffs(coef)
+        from lmfdb.number_fields.number_field import poly_to_field_label
+        pol = coeff_to_poly(coef)
+        label = poly_to_field_label(pol)
+        thefield = WebNumberField(label)
```
BarinderBanwait, line 277: OK, the delay is happening because of this change I made to `formatfield` in `web_number_field.py`

I thought this change was necessary for the following reason: previously, the kerpoly was stored in `char_dir_orbits`, it was looked up in the number field database, it was found, and all was well. 

In the new schema, with kerpoly being computed on the fly, the coefficients defining the number field might not be exactly the same as those found in the number field database, and I was finding that, with `formatfield` as previously implemented, it was not finding the number field from a possibly different polynomial (that actually defines the same number field).

So I decided to add these four lines, which essentially get the field label from the kerpoly. And getting this field label is precisely the step that takes time (about 30s in the `199.k` example).

So the question is: how to do this without taking this long? I wouldn't have thought finding the label of a number field from a polynomial would take that long ...

jwj61:
To look up a field in the database, you need to run polredabs first.  You definitely don't want to do this in formatfield since then you are forcing polredabs to be called everytime that function is called.

If you have low degree polynomials, then calling polredabs should not be a problem.  If the degree is large, it will be harder to avoid.  I have not been following this thread closely, but I assume we are talking about abelian fields, in which case you might be able to do something with Frobenius matching.

lmfdb/number_fields/web_number_field.py
AndrewVSutherland: I started to write exactly what @jwj61 wrote below: we cannot afford to call polredabs here unless the degree is quite small.

Note that this situation is different from when a user types a polynomial into the number field jump box, there the user presumably is willing to wait a bit to find the number field they care about and if it takes a few seconds (or even times out) that's okay, because the user asked for this.  But whatever we do here needs to be fast (e.g. 100 milliseconds or less).  Most of the time a user goes to a Dirichlet character page they are not going there to find the LMFDB label of the kernel field and we can't impose a performance penalty for computing it on the fly.

If we think it's important to have the LMFDB labels of the kernel fields, then perhaps we should precompute them up to some modulus and degree bound (where the modulus bound might be much less than the range of the Dirichlet character database) and store them in a separate table.  I also think it would be okay to polredabs the kernel field when the degree is quite small (maybe 12 or less?).  It might be worth asking @alexjbest what he thinks, I believe he is the one who added the kernel fields to the Dirichlet character pages.

As a side note, once this PR is merged I plan to push the modulus bound up to 1,000,000, so this is the scale we should be thinking about.

I'm going to resolve this conversation but would encourage others with opinions/suggestions to post them below.

BarinderBanwait:
Thanks John; so one idea is to call polredabs in the function that defines the kerpoly; this takes less time for the `199.k` example, but still about 10 seconds.

However, I note that the largest degree of a number field in the database is 47, so one idea is as follows: only call `polredabs` on the kerpoly if its degree is at most 47?

This still takes about 10 seconds for an order 47 orbit, e.g. `283.e`; however, it *does* find the fixed field in the number field database now, whereas on the live site, it reports that the number field is not in the database. (So actually this is an existing issue with prod, saying the number field is not in the database when in fact it is.)

lmfdb/number_fields/web_number_field.py
BarinderBanwait: Just saw this reply ... only calling polredabs if the degree is <= 12 seems to be just as fast as prod on a few examples i checked. I still think the language needs to be changed to make clear that, when we're not calling polredabs on the kerpoly, then the number field *might* be in the database, and the user can check this via the number field jump.

For now I'm going to go ahead and commit these changes to this PR: revert `formatfield` back to its original, and call polredabs on the kerpoly if the degree <= 12.

jwj61:
Here's another suggestion.  I haven't looked closely at what you are doing, but I assume you have a polynomial, know its Galois group (or at least that it is abelian), and know the list of ramifying primes (or at least a superset of them).  I will describe this in terms of gp, sage should have wrappers for these functions.

You can search the database with the information you know.  First, loop through the results looking for exact matches.

If you don't have a match, use nfinit([polynomial, [list containing ramified primes]) to get a p-maximal order at all possible ramifying primes.  Take the valuation of the discriminant for each potentially ramifying prime to get the discriminant on the nose.  Use that to prune the list you got from the database.

All of this should be fast since there should be relatively few hits in the database and because we don't ask gp to fully factor a polynomial discriminant.

Then use nfisisom to directly test for isomorphism with candidates from the database.  This is the only part where I don't have a feel for speed when the degree gets up near 47, but I don't think it will be as slow as polredabs.

edgarcosta:
I agree with @jwj61, I think this should be much faster than polredabs as long as the number of matches is small.

BarinderBanwait:
Latest commit implements a link to the knowl for the kernel field; clicking this will then subsequently search for the number field, an idea suggested by @roed314. I'll raise a separate issue to implement @jwj61 's idea at a later date (db search for abelian number fields).

AndrewVSutherland:
This is now wonderfully fast!  My only further suggestion for improvement would be to not offer to lookup the kernel polynomial in the number field database when the degree exceeds 47, since this invites the user to run a time-consuming computation that is guaranteed to fail.

BarinderBanwait:
latest commit should have limited the lookup to deg <= 47

PR #5461, opened 2023-05-03
Moving search boxes around

jwj61:
Moved ell and dimension to the top since they are probably the first criteria one would set.


jwj61:
Also added abstract group of the image to the representation display page

PR #5489, opened 2023-06-05
ECQ: better information about twists

JohnCremona:
This PR gives home pages of elliptic curves over Q a new small section "Twists", showing the minimal quadratic twists and (where relevant) the minimal quartic/sextic twist.  The Related Object section of the page also has the latter information.

There are two new knowls: https://beta.lmfdb.org/knowledge/show/ec.twists and https://beta.lmfdb.org/knowledge/show/ec.q.minimal_twist and a review of these along with this PR would be appreciated.  The new improved definition for minimal twists was discussed on Zulip (https://lmfdb.zulipchat.com/#narrow/stream/250930-ECQ/topic/minimal.20quadratic.20twists) and the revised data is already on beta and prod.

### DIFF
--- a/lmfdb/elliptic_curves/templates/ec-curve.html
+++ b/lmfdb/elliptic_curves/templates/ec-curve.html
@@ -465,34 +465,114 @@ The {{KNOWL('ec.galois_rep', title=' $\ell$-adic Galois representation')}} has {
 <p>The torsion field $K:=\Q(E[{{data.adelic_level}}])$ is a degree-${{data.data.adelic_image_size}}$ Galois extension of $\Q$ with $\Gal(K/\Q)$ isomorphic to the projection of $H$ to $\GL_2(\Z/{{data.adelic_level}}\Z)$.
 {% endif %}
 
-<h2>$p$-adic regulators</h2>
 
+<h2>{{ KNOWL('ec.isogeny', title='Isogenies') }} </h2>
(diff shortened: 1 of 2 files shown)


AndrewVSutherland:
On http://127.0.0.1:37777/EllipticCurve/Q/223587/bn/1 (which has j-invariant 0) the section on twists uses the phrase "minimal quartic twist" where I think "minimal sextic twist" is meant.

JohnCremona:
Oops, looks like a cut and paste glitch. I'll fix it

AndrewVSutherland:
Looks good, but I have one more question before I merege.  The knowl:

http://127.0.0.1:37777/knowledge/show/ec.q.minimal_quadratic_twist

is a stub (which would need to be filled in before we push to beta).  However, the knowl

http://127.0.0.1:37777/knowledge/show/ec.q.minimal_twist

that is linked to for quartic and sextic twists appears to define minimal quadratic twists.

Would it make sense to just use the ec.q_minimal_twist know everywhere and delete the knowl ec.q.minimal_quadratic_twist?  Alternatively, we could move the definition of minimal quadratic twist from ec.q_minimal_twist to ec.q.minimal_quadratic_twist and replace it with a knowl link (we should not repeat the definition in two different places).  I think I tend to favor the former.

AndrewVSutherland:
I think one knowl is good (anyone looking at one is likely to be interested in the others).  I'll merge once tests pass.

PR #5495, opened 2023-06-08
catch timestamp error

edgarcosta:


roed314:
The test failures suggest you also need to catch TypeErrors in addition to ValueErrors.

edgarcosta:
Thanks. I'm afk, but I will fix that tomorrow

edgarcosta:
I have also upgraded some of the cations

PR #5496, opened 2023-06-08
Added link to the modular curve page from elliptic curves page over Q

omisheel:
Also fixed a regular expression causing search bug.


roed314 (approved):
Looks good

PR #5497, opened 2023-06-09
Typo fix

havarddj:
Fixed a very minor typo I found in the elliptic curve BSD check script in sage. Let me know if I should submit minor things like this differently!


JohnCremona:
Thanks!

PR #5498, opened 2023-06-10
With subgroups show order

jwj61:
On the page of an abstract group, if you click on a subgroup in the diagram you get some information about the subgroup.  This adds the order of the subgroup to the list of information.  If the order is not one and not a prime, it also shows it in factored form.

http://127.0.0.1:37777/Groups/Abstract/12.1
http://beta.lmfdb.org/Groups/Abstract/12.1


roed314:
Looks good.

PR #5499, opened 2023-06-11
CMF changes to remove depency on conrey_indexes

AndrewVSutherland:
As a prelude to extending the range of the CMF database to higher levels, this PR removes all code dependency on the `conrey_indexes` column in the tables: `mf_newforms`, `mf_newspaces`, and `mf_subspaces`, as this data is no longer stored in the Dirichlet character database (only the min and max elements of the list of Conrey indexes in a character orbit are now stored -- this ensures that the space required grows quasi-linearly with the modulus rather than quasi-quadratically).

A new column `conrey_index` has been added to each of these tables that contains the first element of `conrey_indexes` (the minimal Conrey index in the character orbit).  In most situations only the first element of the list was being used, and in cases where the entire list is needed it is now  computed on the fly using the TinyConrey interface to Pari/GP implemented by @BarinderBanwait (this is only used when listing embedded newforms or L-functions).  This code is fast enough to easily compute all the character orbits that we currently use on the fly (the largest is https://alpha.lmfdb.org/Character/Dirichlet/3997/cz, which has 144 elements) and I have tested it with the modulus ranging up to a million (character orbits with hundreds of thousands of elements take about 5 seconds but I don't expect we will ever add any of these).

None of these changes should be visible to the user.

The updated tables with the new `conrey_index` column have already been copied to proddb.  Once this change has been merged and pushed to production I will drop the `conrey_indexes` column.

### DIFF
--- a/lmfdb/classical_modular_forms/download.py
+++ b/lmfdb/classical_modular_forms/download.py
@@ -357,7 +357,7 @@ class CMF_download(Downloader):
     def download_spaces(self, info):
         lang = info.get(self.lang_key,'text').strip()
         query = literal_eval(info.get('query', '{}'))
-        proj = ['label', 'analytic_conductor', 'conrey_indexes', 'char_order']
+        proj = ['label', 'analytic_conductor', 'conrey_index', 'char_order']
         spaces = list(db.mf_newspaces.search(query, projection=proj))
         s = ""
         c = self.comment_prefix[lang]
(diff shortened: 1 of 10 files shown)


AndrewVSutherland:
@roed314, @edgarcosta, @BarinderBanwait if one or more of you has time to look at this, another pair of eyes would be much appreciated!

BarinderBanwait:
I've looked over the code changes and poked around on a local instance with these changes and it seems fine to me. It looks like `conrey_indexes` are still referenced in `lmfdb.js` in the static folder, but I'm not sure where exactly that gets used?

(looks like it's used in a commented out line in `web_newform.py`? so maybe just remove this `properties_lfun` function in `lmfdb.js`? )

AndrewVSutherland:
Thanks for the quick code review.  The `properties_lfun` function is actually used in the java script snippet that is constructed in what is now line 381 of `web_newform.py` (immediately below the line that is commented out that you mentioned above).  But I changed that line to pass in the attribute self.conrey_orbit which is computed on the fly (the `conrey_indexes` in `properties_lfun` is just a variable name, it is not a reference to the `conrey_indexes` column of the `mf_newforms` table so I didn't bother to change it).

roed314:
I just read through it and didn't notice any problems.

PR #5501, opened 2023-06-13
Quotient group knowls

jwj61:
On the page of a group, various quotient groups appear.  This makes them small group knowls if the group is in the database.  For live groups, they are just links to the group (unless the group is the one we are looking at).  This is done for special subgroups (center, socle, etc.) and in the information about a subgroup when clicked on from the diagram (if it is normal).

Standard page:
http://127.0.0.1:37777/Groups/Abstract/68.4

Live abelian group:
http://127.0.0.1:37777/Groups/Abstract/ab/3.3.9.9.9.9.9.36.72

Live gap group:
http://127.0.0.1:37777/Groups/Abstract/512.1

Notes:
  * there may be other quotient groups which come up in some circumstances.  They can always be handled at a later date.
  * in some cases, a quotient of a live group might be in the database in which case one could make a small group knowl for it.  However, it might not be worth the time to try to look them up.
  * The code also has a start at trying to set a variable inside the web abstract group object to say what type it is.  In the long run, this is would be clearer so we wouldn't have to infer the information by data types of "_data".  However, it wasn't fully working, so isn't actually used yet.

### DIFF
--- a/lmfdb/groups/abstract/main.py
+++ b/lmfdb/groups/abstract/main.py
@@ -1155,7 +1155,6 @@ def render_abstract_group(label, data=None):
             )
             friends += [("As the component group of a Sato-Tate group", st_url)]
 
-    info['display_group_knowl'] = abstract_group_display_knowl
     bread = get_bread([(label, "")])
     learnmore_gp_picture = ('Picture description', url_for(".picture_page"))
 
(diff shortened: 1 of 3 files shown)


roed314:
For http://localhost:37777/Groups/Abstract/ab/3.3.9.9.9.9.9.36.72, only some of the subgroups and quotients have links (it looks like this PR just adds links for the trivial group in those cases).  It's fine to put off for another day, but it would be nice to have all of those be links (using `ab/` urls).

I agree with the notes you've made: it's too expensive to do isomorphism testing for live groups when outside the range where `GroupId` works, and I fully approve of the `source` variable.

I'm going to merge this PR, with the note that we should try to further improve links to abelian groups in the live case.

jwj61:
There are links for S and G/S.  It does not do the link if it would be to the current page, which happens a lot with abelian groups.

PR #5509, opened 2023-06-24
improved modm reduction, added code blocks

omisheel:


roed314:
For now, this should be into LMFDB:alpha.  I'll make a new PR with that target.

PR #5515, opened 2023-07-03
fix bread for ecnf

alexjbest:
Should fix the bread on https://www.lmfdb.org/EllipticCurve/stats and interesting, but I didn't actually try to build lmfdb with this change


lmfdb/ecnf/main.py
```diff
@@ -43,6 +43,8 @@
 def get_bread(*breads):
     bc = [("Elliptic curves", url_for(".index"))]
     for x in breads:
+        if not isinstance(x, tuple):
+            x = [(x, " ")]
```
alexjbest, line 47:
```suggestion
            x = (x, " ")
```

PR #5518, opened 2023-07-05
New front page layout, resolves #4963

### LINKED ISSUE #4963: Announcements/homepage cleanup?
The current homepage could use an update.  

Three thoughts:

* The "hall of fame" has been subsumed by "interesting" objects, which are now throughout.  On pages these are commingled with the browse functionality, so I propose we include this in the browse/search bubble in some way.  (I wouldn't be sad to see the phrase "hall of fame" go away.)  To create space there, I wonder if we can compress or omit the static text "Search for objects with specific properties, or browse categories."  Is that adding something essential?  How is this distinct from the "Database"?  
* The title "code and open software" is a bit confusing, because Magma is not "open", and anyway we meant "open source".  Do we really want to display this at the same visibility as the content of our website?  I would relegate this to the footnote and our acknowledgements/citation knowls.

If we do at least one of these bubbles, that would create space for:
* An "announcements" bubble.  We could make this a different color, or bold it, to draw the user's attention.  Here, we could add our LuCaNT announcement and put "Citing the LMFDB" in a visible place, as well as an announcement/release notes for 1.3.  (So we should do this last, after the rest of 1.3 is done.)

AndrewVSutherland:


roed314:
* Groups should probably stay beta in sidebar.yaml.
* There are some knowls that still need to be written (most notably the [rcs summary knowl](https://beta.lmfdb.org/knowledge/show/rcs), but it might also be helpful to add some UI tips to the [knowl knowl](https://beta.lmfdb.org/knowledge/show/doc.knowl) like how to open and close knowls, the dotted underline that indicates them....)
* Once the design stabilizes, we should probably add tests back in since all the old homepage tests were deleted.

AndrewVSutherland:
Good catch on groups, thanks.  I just commented out the home page tests and plan to revisit them post-LuCaNT (unless someone beats me to it).  I'll try to work on the knowls tomorrow night (again, unless someone beats me to it).

PR #5521, opened 2023-07-08
Modular curves

roed314:
Adds modular curves section.  Changes are visible on alpha.lmfdb.org.


roed314:
@edgarcosta or @AndrewVSutherland, we should push several tables to prod now that this is going into master. `gps_gl2zhat_fine`, `gps_gl2zhat_coarse` are ready to go now.  My plan was to wait until this was into beta before renaming `modcurve_points_test` to `modcurve_points` (etc).

AndrewVSutherland:
I'm copying the tables over now and will rerun tests when that is done.  But there is at least one test failure that won't fix
```
lmfdb/modular_curves/main.py:333:60: W0108: Lambda may not be necessary (unnecessary-lambda)
```

AndrewVSutherland:
@roed314 I was unable to copy `gps_gl2zhat_fine` to proddb due to the following error:
```
UndefinedObject                           Traceback (most recent call last)
~/github/lmfdb/lmfdb/backend/base.py in _execute(self, query, values, silent, values_list, template, commit, slow_note, reissued, buffered)
    312                 try:
--> 313                     cur.execute(query, values)
    314                 except (OperationalError, ProgrammingError, NotSupportedError, DataError, SyntaxError) as e:

UndefinedObject: data type text has no default operator class for access method "gin"
HINT:  You must specify an operator class for the index or define a default operator class for the data type.
```

AndrewVSutherland:
The mod-m image data has been updated in `ec_curvedata` and copied to proddb.

roed314:
I fixed the unnecessary lambda and deleted the offending index on `["label"]` (we had both a btree index and a GIN index; I don't know where the GIN index came from).

AndrewVSutherland:
Still getting errors trying to copy to prod:
```
UndefinedObject: data type text has no default operator class for access method "gin"
HINT:  You must specify an operator class for the index or define a default operator class for the data type.
 happens while executing b'CREATE INDEX "gps_gl2zhat_fine_label_gin_tmp" ON "gps_gl2zhat_fine_tmp" USING gin ("label")'
```

roed314:
Try again: I didn't delete the index in the right way.

AndrewVSutherland:
@roed314 Thanks, that fixed the problem, tables are now on prod and I'm rerunning the failed tests.

roed314:
@AndrewVSutherland There are still tests failing because `modcurve_points_test` is missing.  We should decide how we want to deal with these test tables, since we're intending to rename them.  In this case, I don't think anything is using `modcurve_points`, so we could just rename it now.  For the upcoming groups changes, I'm not sure we want to copy the 500GB or so of tables to production and then rename them afterward.  I'm currently building indexes on `gps_subgroups_test`, and once those are done and I can double check that everything's working correctly I'll be making a groups PR.  If we're alright with some downtime on beta.lmfdb.org/Groups/Abstract, we can wait until we're almost ready to merge the groups PR, rename the tables and push to production, run tests, then merge.

I'm open to other suggestions as well.

PR #5538, opened 2023-07-10
Fullpage diagrams, change default group search columns

roed314:


roed314:
To review the first change, go to some small group and click on the "see full page diagram" links for each of the four diagram types.  To review the second, look at the search results.

Both changes should be visible on groups.lmfdb.xyz shortly.

PR #5542, opened 2023-07-11
Special subgroup knowls

jwj61:
For special subgroups, the subgroups themselves are now subgroup knowls when possible.

Also, commutator length is 0 when the group is abelian.

https://beta.lmfdb.org/Groups/Abstract/132.2
http://127.0.0.1:37777/Groups/Abstract/132.2

Also fixed highlighting for Sylow subgroups.

### DIFF
--- a/lmfdb/groups/abstract/main.py
+++ b/lmfdb/groups/abstract/main.py
@@ -1367,13 +1367,16 @@ def shortsubinfo(ambient, short_label):
 
     def subinfo_getsub(title, knowlid, lab):
         full_lab = "%s.%s" % (ambient, lab)
-        h = WebAbstractSubgroup(full_lab)
+        h = WebAbstractSubgroup(full_lab) if lab else None
         prop = display_knowl(knowlid, title)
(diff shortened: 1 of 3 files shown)


jwj61:
There are a couple of highlight things going on.  When hovering, all examples of that subgroup get a background highlight.  In the list, characteristic subgroups have a different font color.  The latter one confused me for a while and others.  The question is whether to explain it, or undo it.

AndrewVSutherland:
Okay, I understand that the characteristic font color issue is separate from this PR.

But it seems the hover highlight is part of this PR (e.g. if I hover on C_11 only one gets highlighted on beta, but two copies of C_11 get highlighted in this PR) and I also have a question about that.  Should the hover highlight also apply to quotients (e.g. C_2 appears twice as G/Soc and G/Fit, Soc and Fit both are both C_66 and get highlighted together when you hover, but the corresponding quotients don't)?

roed314:
I'm fine with highlighting quotients to show their equivalence.  It could also highlight them in the normal subgroup profiles....

AndrewVSutherland:
I'm happy to merge this as is, just wanted to make sure it was intentional.

PR #5544, opened 2023-07-11
Group fixes

jenpaulhus:
1. Fixes font for permutation degree.
2. Gets rid of “None” in Normal Subgroup profile if quotient was unknown. Replaces it with “unknown group of order blah”

Compare https://beta.lmfdb.org/Groups/Abstract/10485760.mp
with the local version. Click on “normal subgroup” in a subgroup profile to see difference.

3. Adds \POmega to base.html


jenpaulhus:
Been a while since I did a PR to master branch and forgot to run tests locally.  I'm fixing this.

PR #5546, opened 2023-07-11
Fix bug in search by point count on modular curves

roed314:
Currently the point count search for modular curves is broken and always searches for non-cuspal points, regardless of the value of the dropdown.  This PR fixes the issue.

### DIFF
--- a/lmfdb/modular_curves/main.py
+++ b/lmfdb/modular_curves/main.py
@@ -656,11 +656,11 @@ def modcurve_search(info, query):
     parse_ints(info, query, "rational_cusps")
     parse_ints(info, query, "nu2")
     parse_ints(info, query, "nu3")
-    if not info.get("points_quantifier"): # default, which is non-cuspidal
+    if not info.get("points_type"): # default, which is non-cuspidal
         parse_ints(info, query, "points", qfield="num_known_degree1_noncusp_points")
-    elif info["points_quantifier"] == "noncm":
+    elif info["points_type"] == "noncm":
         parse_ints(info, query, "points", qfield="num_known_degree1_noncm_points")
-    elif info["points_quantifier"] == "all":
+    elif info["points_type"] == "all":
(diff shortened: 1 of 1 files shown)


AndrewVSutherland:
I don't understand the results of this search:

http://127.0.0.1:37777/ModularCurve/Q/?hst=List&level=6&index=12&points_type=all&contains_negative_one=yes&points=1&search_type=List&showcol=num_known_degree1_points

The unique search result 6.12.1.b.1 claims to have 2 rational cusps and no known rational points.

Also (and probably a separate issue), the Weierstrass model for this curve is 144a1, but this is not compatible with the decomposition of the Jacobian as 36.2.a.a

AndrewVSutherland:
Should "non-CM" mean "non-CM and non-cuspidal" (probably it should, and maybe it does, but this isn't clear in the drop down).  I'm also confused by 6.12.0.b1, which based on the point count searches appears to satisfy all=17, non-CM=15, non-cusp=16, which is consistent with

http://127.0.0.1:37777/ModularCurve/Q/low_degree_points?curve=6.12.0.b.1&degree=1

if one supposes there is just one rational cusp, but there are two rational cusps.

Is there perhaps a problem with how cuspidal points are being counted?  I think this would explain both the inconsistencies that are bothering me.

AndrewVSutherland:
I like j-points better (the cusps all correspond to a single j-point at  infinity).

I think there is still an issue with the non-CM points, if you look at

http://127.0.0.1:37777/ModularCurve/Q/?hst=List&level=6&index=18&points_type=noncm&contains_negative_one=yes&points=1&search_type=List&showcol=num_known_degree1_points

The counts don't make sense.

roed314:
I'll take a look tomorrow; going to sleep early tonight.

roed314:
The non-CM counts were including cusps.  I've updated the data, so it should be okay now.

PR #5547, opened 2023-07-11
Removing some math mode

jwj61:
When looking at the normal subgroup profile, this changes plain text from being text-inside-latex-math-mode to just text (for a quotient group).

http://127.0.0.1:37777/Groups/Abstract/400000000.bk
http://beta.lmfdb.org/Groups/Abstract/400000000.bk

Note, I am aware of another issue with this the display of quotients in the normal subgroup profile on this page, but am putting that off to another time.

### DIFF
--- a/lmfdb/groups/abstract/main.py
+++ b/lmfdb/groups/abstract/main.py
@@ -2383,12 +2383,12 @@ def group_data(label, ambient=None, aut=False, profiledata=None):
         profiledata[0] = None
         tex_name = profiledata[2]
         if tex_name is None:
-            ans = "Unknown group<br />"
+            ans = "Unidentified group<br />"
         else:
             ans = f"Group ${tex_name}$<br />"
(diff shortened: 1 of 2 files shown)


jwj61:
The actual changes are to say when things are not identified and adding statements explaining the font colors of subgroups

jenpaulhus:
When I look at the page above or another page like: 12856620672.u this PR seems to add "Group None" in a bunch of places in the profile.  Is that intended?  It looks strange to me.

jwj61:
I didn't realize I did that ... looking into it.

jenpaulhus:
It actually might be something from earlier (maybe my earlier PR?)

jwj61:
It looks like it is already in master.

jenpaulhus:
It looks like the issue is in web_groups.py line 99+.  I'm going to merge this and then try to fix that issue after it is merged.

PR #5555, opened 2023-07-12
Compress polynomials in the search results display

jwj61:
In number field search results, interesting information is sometimes way off the screen because polynomials are large.  This limits the space used for those polynomials.

http://127.0.0.1:37777/NumberField/?hst=List&degree=37&search_type=List
http://beta.lmfdb.org/NumberField/?hst=List&degree=37&search_type=List


jwj61:
Test error is from abstract groups

jenpaulhus:
Looks like the following search is timing out even on beta:
beta.lmfdb.org/Groups/Abstract/?perfect=yes&nontrivproper=yes&search_type=Subgroups

AndrewVSutherland:
Looks good.  One thing I noticed while testing this is that you cannot use unbounded search ranges on discriminant the way you can on degree (I would expect 10- to give me all number fields with discriminant > 10), which is slightly annoying.

PR #5574, opened 2023-07-13
add a button that flips vertically group diagrams

pascalmolin:
When doing Galois Theory we usually want this flip.


pascalmolin:
example: /Groups/Abstract/diagram/240.89

Issue #5575, opened 2023-07-14
Slow Dirichlet character pages


AndrewVSutherland:
Pages like https://www.lmfdb.org/Character/Dirichlet/79337/20310 (and others of large conductor/order) takes 10s of seconds to load, leading to occasional timeouts.

AndrewVSutherland:
I think this is a duplicate of #5488

AndrewVSutherland:
Actually not really a duplicate.  The underlying issue is the same (we are spending a lot of time computing Galois orbits), but the issue here is for a single character page, not the entire group, and the solutions may need to be different, here we cannot really avoid computing the orbit here, and we need to figure out why it is taking so long (doing this directly in gp is much faster)

AndrewVSutherland:
I think the issue is that there is a loop in python that is making 25932 separate calls to pari, we need to move this loop into a single pari call.

pascalmolin:
I'm taking care of it

PR #5579, opened 2023-07-14
fix slow galois orbit

pascalmolin:
solves #5575


pascalmolin:
ops, clearly not ready yet

Issue #5572, opened 2023-07-13
Poor formatting of class number formula


davidfarmer:
The class number formula on most number field pages would look better if it had better line breaks.

For example, see here:

https://www.lmfdb.org/NumberField/6.0.26512640.1

davidfarmer:
The same changes should be done for the BSD formula, as in

https://beta.lmfdb.org/EllipticCurve/Q/75933/i/1

However, it turns out that this was not a "good first issue", even though someone is
addressing it as their first issue.  Also, the BSD case needs more subtle treatment,
because of how it tries to illustrate the equality of the two ways of computing the
central value.

The typesetting of "assuming GRH", as on this page

https://beta.lmfdb.org/NumberField/17.1.1532892906308884008951324459158712209.1

needs some attention, because we do not have a way of putting a knowl in
display math.

davidfarmer:
The original part of this issue has been addressed in a pull request.  I will take care of the remainder.

AndrewVSutherland:
This is definitely an improvement, but I have a suggestion for a further improvement: use two lines rather than three by moving the RHS of the third line to the (currently vacant) LHS of the second line (which would both save vertical space and make it clearer that the former is the approximate value of the residue at s=1 rather than a simplification of the latter).

davidfarmer:
This happened before:  I reply with a comment in email, and it does not appear on GitHub.

I will copy from my sent email.

davidfarmer:
I am not sure I understand the suggestion.  The current format is

```
A = B
  approx C
  approx D
```
I interpret that as:
```
A equals B
and
B (viewed in isolation) is approximately C
and
C (viewed in isolation) is approximately D .
```
The rearrangement you suggest would have a different interpretation.

Another possible rearrangement, which also saves a line and does not
change the interpretation, is

A = B
  approx C approx D

AndrewVSutherland:
Sorry, I was incorrectly assuming that we were actually showing the (independently computed) special value as we are on the ECQ pages, and intended the LHS of the second line to be an approximation of the LHS of the first line, and the RHS of the second line to be an approximation of the RHS of the first line, which would be true on ECQ pages and I think is the most natural interpretation (we are checking that both sides of a known/conjectured formula are approximately equal when we instantiate them with the values associated to this particular object).

[...]

PR #5581, opened 2023-07-14
Add Fix for Poor Formatting of Class Number Formula

kammitama5:
- fixed formatting so displays class number formula on each line.
- Right aligned GRH for rational number fields with a span right align.
- Fixes issue #5572


davidfarmer:
Looks much better.

@jwj61 pointed out that the same issue exists for the BSD formula.  I have updated #5581 to mention this, but I suggest merging this and addressing the BSD case separately (which I can do).

PR #5605, opened 2023-07-24
add nonmaximal prime search option to genus 2 curves

AndrewVSutherland:
This PR adds the ability to search on nonmaximal primes of genus 2 curves in the same way one can for elliptic curves.

Non-maximal primes are currently available only for generic curves (geometric endomorphism ring Z), and the search results will warn the user about uncomputed quantities (the knowl for nonmaximal primes also notes this).

Available to test on https://purple.lmfdb.xyz.

We will switch to using advanced search options on genus 2 pages soon (and this will be one of the advanced options), but that change will come in a separate PR.


BarinderBanwait (approved):
LGTM: if this passes the tests, Ship It!

PR #5621, opened 2023-07-31
Fix a bug in min_conrey_conj when there is no small orbit rep

roed314:
Compare https://beta.lmfdb.org/Character/Dirichlet/9675/6296 and http://localhost:37777/Character/Dirichlet/9675/6296.


AndrewVSutherland:
LGTM, happy to merge when tests pass, thanks for the fix!

Issue #5632, opened 2023-08-08
Possible improvements for Belyi maps


jwj61:
The knowl for geometry of a Belyi map says that it corresponds to a triangle group Delta(a,b,c).
  - the values of a, b, and c for the page we are on are not given
  - "triangle group" should probably be a knowl

The curve and map should be raw/typeset toggles

The monodromy group is the geometric Galois group of the map.  It would be nice to include the generic Galois group this gives over the base field

jvoight:
Agree with all of these suggestions!

jwj61:
Triangle group is now a knowl (it already existed!)

PR #5645, opened 2023-08-17
Using raw/latex in Belyi pages

jwj61:
Addresses most of #5632, using raw/latex toggles for the equation curves and maps.  The one bit not done here is the factored constant when the map is c*t because it is stored as latex and it is too complicated for the current version of unlatex.

Example page:

https://beta.lmfdb.org/Belyi/8T23/8/2.2.2.1.1/6.2/a/
http://127.0.0.1:37777/Belyi/8T23/8/2.2.2.1.1/6.2/a/


roed314:
There's something weird happening where the polynomial is displayed differently in raw and typeset mode.  Consider http://127.0.0.1:37777/Belyi/7T4/6.1/3.3.1/2.2.2.1/a/?abc=6.1_3.3.1_2.2.2.1
The raw mode is
```
(5*nu - 8)*t*x^6 + (-32*nu + 12)*t*x^5 + (24*nu + 40)*t*x^4 + (44*nu - 76)*t*x^3 + (-56*nu + 28)*t*x^2 + (16*nu + 8)*t*x - 4*t + 4*x^7 + (nu - 3)*x^6=0
```
but typeset looks like
Screenshot 2023-08-26 at 11 05 27

Comparing to https://beta.lmfdb.org/Belyi/7T4/6.1/3.3.1/2.2.2.1/a/?abc=6.1_3.3.1_2.2.2.1, the typeset version looks correct.  I don't see anything obvious with the code.

@SamSchiavone, can you take a look?

jwj61:
FWIW, those two particular expressions are equal (using the defining relation for nu).

jwj61:
I can look at refactoring the code.  It looks like it makes changes to the raw mode and then puts that into latex, so I would have to separate those steps into two functions instead of one.

jwj61:
I took another look, and this instance seems to be the difference between the columns plane_model and plane_model_latex.  So, some rearranging is done on the equation outside of the lmfdb code.

SamSchiavone:
Hi David, yes as John says, there's a separate `plane_model_latex` column where a factored version of the equation is stored for display purposes. Similarly, there is a `plane_map_constant_factored` for displaying the coefficient of the Belyi map for the plane model.

SamSchiavone:
We have a script in the Gm-Reduce repo that takes a polynomial and outputs a string to be LaTeX-ed (https://github.com/SamSchiavone/Gm-Reduce/blob/main/Code/utils.m#L257). We store this in the `plane_model_latex` column, but the unfactored version which can be eval-ed in Magma is stored in `plane_model`.

jwj61:
Can the output of that function still be eval'ed properly in magma?

SamSchiavone:
Not without some parsing and work. For instance, the entry of `plane_model_latex` for the first example `8T23-8_2.2.2.1.1_6.2-a` you gave is
```
'\\left(84 \\nu - 123\\right) t^{2} + 2 \\left(-4 \\nu x^{6} + \\left(6 \\nu + 93\\right) x^{4} + \\left(276 \\nu + 60\\right) x^{2} + 172 \\nu - 355\\right) t + x^{8}'
```
Meanwhile, the entry for `plane_model` is
```
'(84*nu - 123)*t^2 - 8*nu*t*x^6 + (12*nu + 186)*t*x^4 + (552*nu + 120)*t*x^2 + (344*nu - 710)*t + x^8'
```

jwj61:
So, it is after the latex prep.  In the code is a variable test_str.  That looks like it will be magma readable, and easily converted to latex (there are only a few lines after that which do string replacements).  Right?

SamSchiavone:
Yes, I think so.

jwj61:
I extracted what I needed from the file @SamSchiavone linked to and ran it on the Belyii database.  So, I now have raw strings for plane_model which match the current typeset versions (plane_model_latex).  The magma code checks that the new string evaluates to the same thing as the old one in magma.

I don't want to step on any toes here and see 3 ways forward:

1. replace the contents of plane_model
2. make a new column plane_model_old, copy plane_model into that, then replace plane_model with the new versions
3. make a new column plane_model_good and put the new versions there (and update this pull request to match that).

Any preferences on which way to proceed?

SamSchiavone:
Hi John, thanks for checking in. I'm sort of in the middle of uploading new plane models for the Belyi maps, which are currently in `db.belyi_galmaps_test`. I had pushed this to olive so that people could take a look and give me feedback, but it looks like now there's some weird proxy error when I go to https://olive.lmfdb.xyz/ .

I had also noticed that there was a bug in my script for taking the contents of `plane_model` and populating `plane_model_latex`; -1s were getting dropped in some examples. I was planning to look into this on Friday, but it sounds like your script might already do the trick here, and eliminate the need for this column altogether. Is that right?

jwj61:
Possibly (or probably).  A quick way to find out is for me to follow plan #1 and have my pull request display sage's latex on that, and see what it looks like.  I don't know if the magma code was doing something in constructing the latex to work around something sage would have done.

jwj61:
I did the replacement on plane_model (in belyi_galmaps) and updated this code.  The current version is just for testing: it shows the plane model twice, once as it was before using plane_model_latex and once using the new plane_model string, and converting it to latex in python (which has the raw/typeset toggle).  The conversion is just a python translation of what you had in magma (except I added handling of multi-digit exponents).

I am not sure where the extra -1's were showing up, but you can look to see if they are still there on the new version.

I did notice on http://127.0.0.1:37777/Belyi/9T30/9/6.2.1/4.2.1.1.1/a/ that the two latex versions are slightly different, but it comes down to distributing a minus sign (so equivalent).

SamSchiavone:
Hi John,

Looks good! I clicked on random a bunch of times and only came across 2 issues. (I'm not worried about the distribution of the minus signs that you pointed out.)

1) Something strange seems to be going on with this one for me, at least:
http://127.0.0.1:37777/Belyi/8T50/5.2.1/3.2.2.1/4.3.1/a/

2) When the curve on which the Belyi map is in the LMFDB, we include a link to its homepage after the equation, e.g,

http://127.0.0.1:37777/Belyi/8T2/4.4/4.4/2.2.2.2/a/
http://127.0.0.1:37777/Belyi/8T49/5.3/5.3/2.2.2.2/a/
http://127.0.0.1:37777/Belyi/9T27/7.1.1/3.3.3/3.3.3/a/

I think this may have gotten dropped when you modified the template.

What's the script you used on the `plane_model` column? I would like to try running this on `db.belyi_galmaps_test`, too. It looks like it works well, and then we could drop the `plane_model_latex` column (and probably `plane_map_constant_factored`, too).

jwj61:
I leave for the airport in less than an hour, so this is just a quick response.

The item 2 is mostly an artifact of the two version being displayed.  I just pushed a modification removing the old display.  On the downside, there is now a space before the "isomorphic to" for the icons.  This can be fixed so that the icons come after the text, but that has to be done in python, so I won't get to it in the next few minutes.

I will have have to look at the first item more closely.  It might be triggered by the "double latexing" of the P^1.

On lovelace, you can find the script at ~jj/belyi/makeraw.m.  It is recreating your function, but exiting before converting the string to latex, so the future fix is to modify PolynomialToFactoredString to return test_str at line 346.

jwj61:
I think this is ready now.

On http://127.0.0.1:37777/Belyi/8T50/5.2.1/3.2.2.1/4.3.1/a/ we still don't have latex, but we don't have it on beta either.  I suspect that there is a limitation in katex on the length of the string it will process.  A solution which might be worthwhile in any case is to compress large integers in the display, e.g., if an integer is more than 20 digits long it becomes ab...cd as happens elsewhere in the lmfdb.

SamSchiavone:
Hm, I see LaTeX on beta and olive...

https://beta.lmfdb.org/Belyi/8T50/5.2.1/3.2.2.1/4.3.1/a/
https://olive.lmfdb.xyz/Belyi/8T50/5.2.1/3.2.2.1/4.3.1/a/

I certainly agree with your suggestion about compressing large numbers, but I think that will have to be done in the future.

jwj61:
Weird.  I see typeset on olive and latex code on beta

AndrewVSutherland:
On Firefox on Ubuntu I get typeset on both, with Chrome (and Brave/Opera/Vivaldi) on Ubuntu I get typeset on olive and LaTex on beta.  On Safari on IOS I get LaTeX on both, which Firefox and Chrome on IOS I get typset on olive and LaTeX on beta.

So my guess is Sam is using Safari on MacOS and will see typeset on olive if he switches to Firefox or Chrome (but I can't be sure, MacOS and IOS versions can differ).  It would be nice to find a solution that renders correctly on all of Chrome/Firefox/Safari/Edge (I don't have a Windows machine handy so I cannot test Edge right now, but it would be good for someone to try it).

AndrewVSutherland:
On Edge/Windows, I get typeset on olive and LaTex on beta.

jwj61:
I thought compressing large integers would fix this, but it didn't.  I made a second pull request so others can look at that too (it includes these changes).  Unfortunately, I still see latex code for brave/linux on the example above.  That pull request links to an example which runs through the same part of the template but the latex is typeset.

jwj61:
Closing this in deference to pull request #5676 which should be good on all browsers.

Issue #5660, opened 2023-09-05
server error in ModularCurves


AndrewVSutherland:
https://beta.lmfdb.org/ModularCurve/Q/30.60.3.b.1/ is throwing the following error:
```
Exception on /ModularCurve/Q/30.60.3.b.1/ [GET]
Traceback (most recent call last):
  File "/home/sage/sage-9.7/local/var/lib/sage/venv-python3.10.5/lib/python3.10/site-packages/flask/app.py", line 2525, in wsgi_app
    response = self.full_dispatch_request()
  File "/home/sage/sage-9.7/local/var/lib/sage/venv-python3.10.5/lib/python3.10/site-packages/flask/app.py", line 1822, in full_dispatch_request
    rv = self.handle_user_exception(e)
  File "/home/sage/sage-9.7/local/var/lib/sage/venv-python3.10.5/lib/python3.10/site-packages/flask/app.py", line 1820, in full_dispatch_request
    rv = self.dispatch_request()
  File "/home/sage/sage-9.7/local/var/lib/sage/venv-python3.10.5/lib/python3.10/site-packages/flask/app.py", line 1796, in dispatch_request
    return self.ensure_sync(self.view_functions[rule.endpoint])(**view_args)
  File "/home/lmfdb/lmfdb-git-dev/lmfdb/modular_curves/main.py", line 140, in by_label
    return render_template(
  File "/home/sage/sage-9.7/local/var/lib/sage/venv-python3.10.5/lib/python3.10/site-packages/flask/templating.py", line 147, in render_template
    return _render(app, template, context)
  File "/home/sage/sage-9.7/local/var/lib/sage/venv-python3.10.5/lib/python3.10/site-packages/flask/templating.py", line 130, in _render
    rv = template.render(context)
  File "/home/sage/sage-9.7/local/var/lib/sage/venv-python3.10.5/lib/python3.10/site-packages/jinja2/environment.py", line 1301, in render
    self.environment.handle_exception()
  File "/home/sage/sage-9.7/local/var/lib/sage/venv-python3.10.5/lib/python3.10/site-packages/jinja2/environment.py", line 936, in handle_exception
    raise rewrite_traceback_stack(source=source)
  File "/home/lmfdb/lmfdb-git-dev/lmfdb/modular_curves/templates/modcurve.html", line 1, in top-level template code
    {% extends 'homepage.html' %}
[...]

PR #5661, opened 2023-09-05
Deal with Weierstrass models where the coefficient of w^2 is not 1

roed314:
Addresses #5660.


edgarcosta:
LGTM. The test failure doesn't seem to be related to this PR, but please confirm

roed314:
Yeah, the test failure is something I need to take care of with groups, and is unrelated to this.

PR #5679, opened 2023-10-09
use ruff to remove superfluous parentheses

fchapoton:
ruff is a recent and powerful linter

here I used
`
ruff check --fix --select UP034 .
`
to remove superfluous parentheses


roed314:
Thanks!

PR #5681, opened 2023-10-11
fix some details found by ruff

fchapoton:
ruff is a fast linter, with many available useful checks, and the capacity to fix some of its issued warnings.

The changes here were done using

`ruff check --fix --select UP008,UP028,UP004 lmfdb/`


roed314:
Thanks!

PR #5682, opened 2023-10-11
master -> dev

AndrewVSutherland:


AndrewVSutherland:
The proddb belyi map tests are failing, is there a table that needs to be copied over? (we could disable belyi tests on prod, but given that I am hoping we will be able to push belyi to prod soon, I'd rather copy the data and get tests to pass if that is all that needs to be done).

jwj61:
Yes, there were data changes in belyi_galmaps.  That should be uploaded to prod.

AndrewVSutherland:
belyi_galmaps has been updated on proddb.  Hopefully tests will now pass.

PR #5684, opened 2023-10-11
ruff fixes for list, dict, set comprehension

fchapoton:
Enhancing or simplifying various kinds of list, set or dict comprehensions.

this was scripted using

`ruff check --fix --select=C4 lmfdb/`


roed314:
LGTM

PR #5699, opened 2023-10-18
better threshold for Artin L-functions

edgarcosta:
Lowering the threshold on the conductor from
```
> N := 729000000000000; [RealField(8) | N^(1/d) : d in [2..4]];
[ 27000000., 90000.000, 5196.1524 ]
```
to
```
# degree -> bound on conductor
artin_Lfunction_threshold = {
    2: 10000,
    3: 10000,
    4: 4000,
}
```

I also have tested some of the L-functions at the edge of the threshold, and they seem to be computed in about 5s


edgarcosta:
Once this is merged, one should update: https://beta.lmfdb.org/knowledge/show/rcs.cande.lfunction

PR #5702, opened 2023-10-20
Download columns

roed314:
This PR changes the downloads for search results across the LMFDB.  Now the set of columns downloaded is the same as the columns displayed, and the sort order is the same as well.  In addition, column knowls are included at the bottom of the file for documentation (this can be overridden on a per column basis).

In addition, we've changed some underlying mechanisms so that downloads are no longer subject to the 30 second limit imposed by gunicorn, so we're able to remove all limits on number of records.  Download buttons are now available on every LMFDB section, regardless of the number of results.

It also adds a new backend capability, `join_search`.  We haven't switched over to using this yet, but it may be necessary to do so for some sections that currently do separate lookups for data from other table.

There are a lot of changes in here, so let me know if you notice any problems.  It's currently available to experiment with on [olive](https://olive.lmfdb.xyz).

@jenpaulhus The only section of the LMFDB that I didn't change was higher genus curves, since you'd written some code for doing something more clever there and I didn't want to clobber it.  I'm happy to add it in if you'd like me to.


roed314:
I've updated the PR to use generators, which I had hoped would improve the problems with gunicorn's 30 second timeout, but it doesn't seem to have done so.  Hopefully @edgarcosta and I can investigate next week.

There are some test failures, which I will work on fixing after the weekend.

jenpaulhus:
It has been so long since I wrote that higher genus code, that I don't even remember what it does!  I'll look at those pages next week and give you thoughts on how to proceed. Thanks for checking.

roed314:
Alright, I think this is ready for review, though there's obviously more to do.  After some help from @edgarcosta today I've removed all download limits: this should no longer time out after 30 seconds, and the download speed is okay after some changes made to buffering (maybe more improvements could be made).

edgarcosta:
We might need to do something in the backend to support async workers:
https://www.psycopg.org/docs/advanced.html#green-support

if you try to download multiple instances, only one download survives

roed314:
Good point.  Threads always make my head hurt.  Maybe we should talk about this in person on Friday.

edgarcosta:
Something like this should suffice:
https://github.com/psycopg/psycogreen#module-psycogreengevent

roed314:
Ok, I just tested that I can download two large files from different sections of the LMFDB simultaneously.

roed314:
I added a few columns to support the things that had been downloaded previously.  @SamSchiavone, can you take a look at [Belyi](https://olive.lmfdb.xyz/Belyi/?hst=List&search_type=List&showcol=triples) and tell me if the new triples column looks okay?  I also added [generators](https://olive.lmfdb.xyz/ModularCurve/Q/?hst=List&search_type=List&showcol=generators) for modular curves.

SamSchiavone:
Hi David, I'm a bit swamped at the moment, but I should be able to take a look later next week.

roed314:
I took the opportunity to switch the default for `default` in the `SearchCol` constructor (previously if you forgot to include it, your column wouldn't show up).

Issue #5696, opened 2023-10-15
Server error on many dynamically computed Artin L-functions


AndrewVSutherland:
We are getting regular server errors on pages like

https://www.lmfdb.org/L/ArtinRepresentation/1.145860.2t1.a.a/

that time out trying to compute the L-function (which is accessible from the L-function link on https://beta.lmfdb.org/ArtinRepresentation/1.145860.2t1.a).  Currently the conductor limit for dynamically computed L-functions of 2-dim Artin reps is set to 27000000, according to https://beta.lmfdb.org/L/Completeness, but perhaps it needs to be lower.  In the longer term, we should precompute all of these L-functions (currently only a few thousand are), as we have been planning to do for many years now.

My personal preference would be to simply turn off all dynamic computation of L-functions.  This produces pages that are both incomplete and unreliable, often leads to error messages in the flasklog that require investigation, and places an unnecessary computational burden on the server.  Doing so would put more pressure on us to actually precompute the L-functions we keep saying we plan to precompute.

For those who don't appreciate why the timeout/server-errors are a problem, the issue is that every time a bot crawls the website (which happens much more frequently than it used to, these days everyone is busy crawling the web looking for data to feed their latest LLM) it traverses each and every L-function link.  This makes having exposed links that place a computational burden on the server is a bad idea, even if an individual user might be willing to put up with waiting 30 seconds, maybe getting timed out, and maybe getting incorrect results.  It also leads those of us who get email warnings every time the flasklog is empty to start ignoring these warnings, and that has recently led to a number of problems going unnoticed for longer than they should have.

edgarcosta:
I have lowered the thresholds significantly in #5699

edgarcosta:
I take it back, maybe we should disable them all. Something like this will never be computed in any reasonable amount of time
https://www.lmfdb.org/ArtinRepresentation/2.3659.29t2.a.b/

AndrewVSutherland:
So do you want to amend or close your PR?  Maybe we should talk about this tomorrow at LMFDB Friday.

edgarcosta:
Let's talk friday

PR #5709, opened 2023-10-30
No L-functions on the fly on beta/www

edgarcosta:
I'm sure that there will be a couple of tests failing, but I will fix those later


edgarcosta:
The commit d8b01de66fccba87c405b07e97b94ffc143b520e gives a good sample of what URLs will no longer work.
If we agree with all those routes giving errors, then this resolves #5696

PR #5710, opened 2023-11-01
a few details as suggested by the linter ruff

fchapoton:
This is just simplifying a few details, following some suggestion of `ruff`.

### DIFF
--- a/lmfdb/hilbert_modular_forms/q_expansion.py
+++ b/lmfdb/hilbert_modular_forms/q_expansion.py
@@ -41,7 +41,7 @@ def qexpansion(field_label=None):
             v = next(S)
             continue
 
-        if field_label is None or not field_label != v["field_label"]:
+        if field_label is None or field_label == v['field_label']:
             field_label = v["field_label"]
             print("...new field " + field_label)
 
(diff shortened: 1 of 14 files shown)


lmfdb/hilbert_modular_forms/q_expansion.py
```diff
@@ -41,7 +41,7 @@ def qexpansion(field_label=None):
             v = next(S)
             continue
 
-        if field_label is None or not field_label != v["field_label"]:
+        if field_label is None or field_label == v['field_label']:
```
fchapoton, line 44: this looks dubious

roed314: I agree.  I'll take a look.  Thanks for the PR.

roed314: The file has a comment at the top that "As of August 2017 only one newform had this field, so apart from testing this has not been run."  But reading the code it's clearly a bug, so I'll fix it in the PR.

roed314 (approved):
Looks good.  I made a few more manual fixes.

roed314:
I'll merge once tests pass.

roed314:
Thanks!

PR #5725, opened 2023-11-07
Escape percent signs when printing error

roed314:
With bad input to the api, it tries to catch the error and print it, but if there are percent signs then printing the error fails.

See https://beta.lmfdb.org/api/gps_groups/?permutation_degree=12&abelian=no%&_format=json for an example.


edgarcosta:
lgtm, merge once tests pass

PR #5729, opened 2023-11-12
Fix buggy subgroup page 

jwj61:
On beta

http://127.0.0.1:37777/Groups/Abstract/sub/167772160.bng.40960.CP
https://beta.lmfdb.org/Groups/Abstract/sub/167772160.bng.40960.CP

it says that the subgroup is nilpotent, but that the nilpotency class is -1 which is inconsistent.  The subgroup is in the database, but only minimal information was computed about it.  The code was substituting information about the ambient group for it and the subgroup.  Fixing that led to various places missing data caused an error.

Also, fixing that uncovered that if it is not computed whether or not the group is simple, that fact that it was not computed was being reported twice (internally, there are flags for abelian-non-solvable and nonabelian-non-solvable).

Also fixed a place where plain text was put in textrm inside math mode.

### DIFF
--- a/lmfdb/groups/abstract/web_groups.py
+++ b/lmfdb/groups/abstract/web_groups.py
@@ -1892,7 +1892,7 @@ class WebAbstractGroup(WebObj):
         try:
             if self.aut_group is None:
                 if self.aut_order is None:
-                    return r"$\textrm{not computed}$"
+                    return r"not computed"
                 else:
                     return f"Group of order {pos_int_and_factor(self.aut_order)}"
             else:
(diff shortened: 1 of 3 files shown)


lmfdb/groups/abstract/web_groups.py
```diff
@@ -2386,6 +2389,11 @@ def _full(self):
     def sub(self):
         S = self._lookup(self.subgroup, self._full, WebAbstractGroup)
         # We set various properties from S for create_boolean_subgroup_string
+        if not S:
+            order = self.subgroup_order
+            self.order = order
+            self.pgroup = len(ZZ(order).abs().factor())==1
+            return self
```
roed314, line 2396: I think it might make more sense to create a new `WebAbstractGroup` object, so that the return type doesn't depend on whether it's in the database or not.

jwj61: I imagine there is an elegant way to do that, but I tried several variations and keep getting errors.

roed314:
I started working on a change to make the return value a `WebAbstractGroup`, but it got very involved.  I'm not going to finish it today, but will try to by Monday at the latest.

jwj61:
I am not sure about the sequence of things here, but I just pushed a change which does that.

AndrewVSutherland:
Is this ready to merge @roed314 ?

roed314:
Yep, looks good to me.

PR #5732, opened 2023-11-17
Smf week

manamiroy:
Displaying the dimensions of spaces in the dimension search page even if the forms are not computed but the space is computed.


manamiroy:
Need to do the pull request to a different branch

PR #5750, opened 2023-11-29
Add minimal siblings

jwj61:
On number field pages, this displays the minimal sibling of the field.  The possibilities are

it is a different field: http://127.0.0.1:37777/NumberField/4.2.4107.1

it is the same field: http://127.0.0.1:37777/NumberField/4.0.117.1

we computed it, but it is not in the database (yet): http://127.0.0.1:37777/NumberField/6.0.932501808.1

not computed yet: http://127.0.0.1:37777/NumberField/20.0.287532152115567788032.1

I put it with the section on the Galois group since the idea is that there is only one per Galois closure, but an argument could be made to put it in the sibling section.

### DIFF
--- a/lmfdb/number_fields/templates/nf-show-field.html
+++ b/lmfdb/number_fields/templates/nf-show-field.html
@@ -131,6 +131,8 @@ table.ntdata a {
             {{ info.cclasses | safe }}</tr>
             <tr><td>
             {{ info.character_table | safe }}</tr>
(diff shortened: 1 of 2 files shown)


AndrewVSutherland:
I think it's probably better to put it in the sibling fields section (which is where I looked first).

AndrewVSutherland:
Otherwise looks good!

PR #5755, opened 2023-12-01
multiple cosmetic changes to abstract groups

jenpaulhus:
This PR is for a bunch of cosmetic changes to the abstract groups pages.

I. Fixes a bug where we have a link for normal subgroups but the search returned all subgroups, not just normal subgroups. See group 150493593600000000.b and scroll down to "There are 6 normal subgroups" in the Subgroups section to see the difference between beta and local. 

II. Made rank, inequivalent generating pairs, and commutator length NOT LaTeX to match the fonts for other integer values on the pages. Compare for group 8.4 in beta vs local.

III. Added the header and footer in each LMFDB page to complex character table pages, rational character table pages, and pages for generators of the automorphism groups.  This allows for the breadcrumb to the ambient group.  We will need this feature once we create a search page on characters. (The link for the search will be to the character table and a user may want to then go on to the group itself.)
/Groups/Abstract/auto_gens/1728.10278
/Groups/Abstract/char_table/1728.10278
/Groups/Abstract/Qchar_table/1728.10278

IV. Added Schur index as a column in rational character tables if there are any nontrivial Schur indices. I will edit the schur index knowl tomorrow to give an actual definition.
Check out the rational character table for group 63.1 to see an example.

V. After adding Schur index, the rational character table had more information than the complex character table so we also switched which one is displayed when they are the same.  See group 8.4 again.

### DIFF
--- a/lmfdb/groups/abstract/main.py
+++ b/lmfdb/groups/abstract/main.py
@@ -880,6 +880,13 @@ def show_factor(n):
         return "$0$"
     return f"${latex(ZZ(n).factor())}$"
 
+#for irrQ_degree and irrC_degree gives negative value as "-"
+def remove_negatives(n):
+    if n < 1:
+        return "-"
+    return n
+
+
 def get_url(label):
     return url_for(".by_label", label=label)
 
@@ -952,8 +959,8 @@ group_columns = SearchColumns([
     ProcessedCol("outer_order", "group.outer_aut", r"$\card{\mathrm{Out}(G)}$", show_factor, align="center", short_title="outer automorphisms", default=False),
     MathCol("transitive_degree", "group.transitive_degree", "Tr. deg", short_title="transitive degree", default=False),
     MathCol("permutation_degree", "group.permutation_degree", "Perm. deg", short_title="permutation degree", default=False),
-    MathCol("irrC_degree", "group.min_complex_irrep_deg", r"$\C$-irrep deg", short_title=r"$\C$-irrep degree", default=False),
-    MathCol("irrQ_degree", "group.min_rational_irrep_deg", r"$\Q$-irrep deg", short_title=r"$\Q$-irrep degree", default=False),
+    ProcessedCol("irrC_degree", "group.min_complex_irrep_deg", r"$\C$-irrep deg", remove_negatives, short_title=r"$\C$-irrep degree", default=False, align="center"),
+    ProcessedCol("irrQ_degree", "group.min_rational_irrep_deg", r"$\Q$-irrep deg", remove_negatives, short_title=r"$\Q$-irrep degree", default=False, align="center"),
     MultiProcessedCol("type", "group.type", "Type - length",
                       ["abelian", "nilpotent", "solvable", "smith_abelian_invariants", "nilpotency_class", "derived_length", "composition_length"],
                       show_type,
--- a/lmfdb/groups/abstract/main.py
+++ b/lmfdb/groups/abstract/main.py
@@ -733,10 +733,9 @@ def auto_gens(label):
     return render_template(
         "auto_gens_page.html",
         gp=gp,
-        title="Generators of automorphism group for %s" % label,
-        bread=get_bread([("Automorphism group generators", " ")]),
-        learnmore=learnmore_list(),
-    )
+        title="Generators of automorphism group for $%s$" % gp.tex_name,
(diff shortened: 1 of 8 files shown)


jwj61:
The LMFDB style is to display integers in math mode, so these should be in math mode and others should be changed to match.

jenpaulhus:
I'll deal with that tomorrow then.

jwj61:
The style guide says "In general, all mathematical quantities that are offset in some way (e.g. listed in a table or following a caption) should be displayed in math mode, including integers.  This does not apply to integers that appear in a sentence or as part of a phrase (e.g. "Genus 2 curves" not "Genus $2$ curves")"

Until I just read this just now, I hadn't realized we had an exception to what I think is considered the best practice in math writing.

There may be many places where this is not carried out, but it should be.  The abelian variety page is a clear example.

I think the numbers in the search result table for Sato Tate groups are already in math mode (except labels, but then the number is not on its own).

I don't think some lines saying "not computed" in plain text and other rows having numbers in math mode are a problem at all.

Places where numbers appear in a phrase are a judgement call.  In the case of degree that you mention, I would lean toward using math mode since it is followed by numbers in math mode on the same line.  

By the style guide, the numbers appearing in the boolean statement can be in plain text, but I am not sure that is an improvement.  I somewhat randomly looked at 68.1 and the boolean line there has a 2 in plain text, but later in the sentence it has $p=2$.  So the digit 2 appears twice as a mathematical quantity and in two different fonts.

roed314:
Is the style guide publicly available somewhere?  I'm not sure I remember seeing it....

One difference between our context and something like latex is that katex doesn't fit as well with the surrounding plain text (I think texed numbers have a slightly different height, and if there are too many of them you get a flash of dollar signs as the rendering happens).  So I think it's good to have the rule about numbers appearing in the flow of text.

I'm happy to help change the tabulated numbers in other parts of the LMFDB though.

jwj61:
The style guide is in the github repository at the top level as StyleSheet.md

In the case of the numbers appearing in the boolean sentence, it is violating what I quoted above in that there are mathematical quantities embedded in the sentence rather than being set off.  Here we could fix that by having those quantities appear on their own lines when relevant (e.g., just say solvable, and then have a line for solvable length; or if not solvable don't have the line for the length).  For the properties box, it could be just yes/no on solvable.

jwj61:
Good point @davidfarmer

Do you think the entries in the $p$-adic browse table should be in math mode?

davidfarmer:
I think there is an argument for large tables of integers not being in
math mode.  That helps the speed of loading the page, and it is less
likely to encounter the problem of the same number appearing twice, in
different fonts.

jwj61:
One can see the p-adic browse table in math mode in pull request #5756

davidfarmer:
I think I also prefer to old way (so I stick with my suggestion that when the "data"
is only numbers in a large table, don't put in math mode).

AndrewVSutherland:
@jwj61 

I don't have a strong opinion, but my suggestion would be to separate them.  Just to be completely explicit, I would remove the parenthetical (of class blah) and (of length blah) both from the list of properties (so it would be just a list of boolean properties) and from the property box (so it will just be yes/no), and for groups that are nilpotent/solvable, add a nilpotency class and solvable length item to the body of the page (I would omit them for groups that are not nilpotent/solvable)

You could save some vertical space by using two columns in the first table in the body of the page (e.g. as on the modular curves pages).  It wouldn't bother me to see order and exponent on the same line, automorphism group and outer automorpshisms on the same line, and then solvable length and nilpotency class on the same line.  But you have been staring at these pages much longer than I have and there may be reasons why that is a bad idea.

jenpaulhus:
I'm not clear on the timing of comments and merging of PR last night,  but I thought the changes for the boolean phrasing were being made by David or John on the other PR which is now merged? I'm not planning to make those changes on this PR.

jenpaulhus:
One addition.  Changed irrC_degree and irrQ_degree  to show "-" instead of "-1" if the group has not faithful representatives.

You can see the change if you search for order 12 groups:
beta.lmfdb.org/Groups/Abstract/?order=12 
And then select "C-irrep degree" or "Q-irrep degree" to display.

jwj61:
Isn't there a problem with the automorphism generators of the page noted above.  The column headers are all powers of d, so not generators of the group.

jenpaulhus:
I agree something suspicious is going on with the automorphism group generators. I will investigate the issue, but whatever is wrong  was not introduced with this current PR.

PR #5764, opened 2023-12-05
Move nilpotency class and derived length out of sentence

jwj61:
Changes made to both abstract group and subgroup pages, e.g.

http://beta.lmfdb.org/Groups/Abstract/144.41
http://127.0.0.1:37777/Groups/Abstract/144.41

http://beta.lmfdb.org/Groups/Abstract/sub/144.41.2.c1.a1
http://127.0.0.1:37777/Groups/Abstract/sub/144.41.2.c1.a1


jenpaulhus:
I like this change.  It looks good to me.

PR #5774, opened 2023-12-09
Latex values shown on group page

jwj61:
On

http://127.0.0.1:37777/Groups/Abstract/8.4
http://beta.lmfdb.org/Groups/Abstract/8.4

this puts values under minimal presentations in math mode


jwj61:
Also added latex to the Jacobian count on

http://127.0.0.1:37777/Variety/Abelian/Fq/1/223/v

PR #5778, opened 2023-12-12
Style sheet update: no math mode for tables of integers.

AndrewVSutherland:
This PR updates the style sheet to exclude tables of integers from the provision that mathematical data (including integers) should generally be displayed in math mode.  This mostly agrees wtih current practice (e.g. [CMF dimension tables](https://www.lmfdb.org/ModularForm/GL2/Q/holomorphic/?search_type=SpaceDimensions) and [p-adic field counts](https://www.lmfdb.org/padicField/) do not use math mode) and is desirable because it reduces page render times, especially when the tables are large.

As noted in the discussion on #5755, the style guide is not always consistently followed  (especially on newer pages).  It would be good to correct instances of this as and when they are noticed, and to review new PRs with the style guide in mind.  If there are cases where conforming to the style guide seems undesirable, please open an issue where it can be discussed (which may lead to changes to the style guide, as in this PR).  The style guide is meant to be used as a tool to help us create a consistent interface but it is not meant to bind us to any particular choice (there are always trade offs).

### DIFF
--- a/StyleSheet.md
+++ b/StyleSheet.md
@@ -34,7 +34,7 @@ contains other conventions and information for the development process.
 - Search results table should use row striping (ntdata table), headings should all have knowls (or pseudo-knowls).
 - Refine search page headings should have the form "Blah search results", where blah is singular.
 - Refine search pages should have captions above input boxes, no example to right, gray example inside input box.
-- Mathematical values listed in search results, including integers, should be in math mode.
+- Mathematical values listed in search results, including integers, should be in math mode, with the exception of tables whose entries are integers.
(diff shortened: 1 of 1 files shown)


AndrewVSutherland:
Updated to also mention percentages, we don't want [stats tables](https://www.lmfdb.org/ModularForm/GL2/Q/holomorphic/stats) in math mode.

roed314:
I think the rational character table of a group should be in mathmode, even if it's entirely integers.

jenpaulhus:
I agree with David, but also this table is not in standard table style for the LMFDB anyway. So I see this as a decision about this unique little snowflake, and not an exception to the style guide.

AndrewVSutherland:
I also agree (and would in any situation where the table entries may be integers but often are not).

PR #5779, opened 2023-12-12
Rename style sheet file

jwj61:
StyleSheet may make people think of css.  Maybe this makes people think more about reading this file.

Also, mention this file in the README file so it is more visible to new developers.


roed314:
Looks good.

PR #5780, opened 2023-12-12
Small issue with rational char tables

jenpaulhus:
A recent PR introduced a small bug.  The rational character tables with only trivial schur indices had their  columns shifted by one (the first column of #s didn't match the corresponding header column)

See
http://localhost:37777/Groups/Abstract/465.2
vs
https://beta.lmfdb.org/Groups/Abstract/465.2


roed314:
Looks good.

PR #5785, opened 2023-12-19
Fix google groups url on contact.html

koffie:


roed314:
This looks good.  The test failures in abstract groups result from some unrelated things I've been doing.

PR #5788, opened 2023-12-21
evidently none isn't always none

jenpaulhus:
Some group searches are failing on beta because I forgot to add an option for the empty string instead of None.

For example, search groups in the range below and you get  a server error because irrQ_degree or irrC_degree is None but being passed to a function as "".:

https://beta.lmfdb.org/Groups/Abstract/?order=190000-200000


AndrewVSutherland:
Test failure is a timeout that has nothing to do with this PR

PR #5794, opened 2023-12-21
More updates to abstract groups

jenpaulhus:
This PR updates and improves a number of things on the abstract groups pages.

I Fixed an issue with those subgroups of order 6561 which are not in the abstract group database (These groups are in Magma's db but not GAPs so we have labels for them but can't show them on live pages).  
If you compare beta to localhost for the following group id and scroll down to "Subgroup diagram and profile" and click on the knowl for the group of order 6561, you will see an error in beta and a better message on localhost.

http://beta.lmfdb.org/Groups/Abstract/150493593600000000.b 
http://localhost:37771/Groups/Abstract/150493593600000000.b 

Note that those groups of order 6561 which are also in our db still have knowls that give basic info about the group.   Go to this group and scroll down to subgroups.
http://beta.lmfdb.org/Groups/Abstract/13122.bo
http://localhost:37771/Groups/Abstract/13122.bo

II. Updates to the Construction section. We added the group as transitive group, if relevant.  We also made the presentation and the permutation group copiable as raw text and truncated how much of those lines are actually displayed.  And updated the html in this section to align entries better.
A few examples
http://localhost:37771/Groups/Abstract/200000.bp
http://localhost:37771/Groups/Abstract/800000000.cqs
http://localhost:37771/Groups/Abstract/3200000000.jdy

III. Got rid of message where we claimed we didn't know information about subgroups of 6.1 (and other subgroups too). This was an issue with non-normal subgroups where we didn't store the fact that they were not (semi) direct product factors.  
Compare beta where we have extra message about (semi)direct factors and localhost where no longer claim to not know if they are (semi)direct factors.
http://beta.lmfdb.org/Groups/Abstract/sub/6.1.3.a1.a1
http://localhost:37771/Groups/Abstract/sub/6.1.3.a1.a1

IV. Got rid of "?" in subgroup searches for quotient information if subgroup is non-normal. Also change quotient column in searches label as "Order" to be labeled as"Size", and made a knowl for size of quotient.  
Go here:
https://beta.lmfdb.org/Groups/Abstract/?search_type=Subgroups
http://localhost:37771/Groups/Abstract/?search_type=Subgroups
and scroll to see non-normal subgroups.

V.  Changed a hyphen to mathmode for search results when irrepQ and irrep C are not defined because there are no faithful irreducible reps.  
Enter this search:
https://beta.lmfdb.org/Groups/Abstract/?order=12
http://localhost:37771/Groups/Abstract/?order=12
and choose to show Cirrep deg or Qirrep deg


jenpaulhus:
(I know we're into the holidays so not expecting an immediate look at this PR.)

jwj61:
LGTM

The test failure may be a fluke (the database was busy at the moment).  Can you retrigger tests?

PR #5798, opened 2024-01-02
Delay several database calls

roed314:
Helps with startup time when on a slow connection.  Should not change any displayed pages; pages to check include CMF stats (http://localhost:37777/ModularForm/GL2/Q/holomorphic/stats), abelian variety count pages (http://localhost:37777/Variety/Abelian/Fq/?search_type=Counts) and Galois group alias searches (https://beta.lmfdb.org/padicField/?inertia_gap=D13)


lmfdb/abvar/fq/stats.py
```diff
@@ -172,17 +173,17 @@ def dynamic_parse(info, query):
     dynamic_parent_page = "abvarfq-refine-search.html"
     dynamic_cols = ["q", "g", "p_rank", "angle_rank", "size", "geometric_extension_degree", "jacobian_count", "hyp_count", "twist_count", "max_twist_degree", "is_simple", "is_geometrically_simple", "is_primitive", "has_jacobian", "has_principal_polarization", "jacobian_count", "hyp_count"]
 
-    @lazy_attribute
+    @cached_method
```
edgarcosta, line 176: Just for curiosity, what is the benefit of using `cached_method` here?

roed314: The problem is that it's passed as a keyword argument to `count_wrap` [here](https://github.com/roed314/lmfdb/blob/startup_time/lmfdb/abvar/fq/main.py#L660).  A lazy attribute can't be passed unevaluated, while a cached method can.

PR #5799, opened 2024-01-05
Fix errors with groups of Lie type

### LINKED ISSUE #5562: Projective Lie groups
Currently in the groups database, elements of Lie groups of a projective family are stored with code as if they are permutations instead of matrices.

jenpaulhus:
This PR fixes some server errors caused by elements of particular groups of Lie type which were not decoding correctly.

Specifically groups of the form *(n,q) where * is CSU, CU, SU, GU, and PSU consist of matrices with entries in the field with q^2 elements, not q elements.  And if * is AGL, ASL, or Spin, the matrices are of dimension n+1, not n.   

When * is Spin+ and Spin-, we have matrices of dimension 2^(n/2), and for Spin- we also have q^2 instead of q for the size of the field the elements are in.

Here are a few group IDs to try (click on "generators" of the automorphism group or if there is a complex character table, click on the knowl for the conjugacy classes to see a representative). These all cause errors on beta (other examples don't necessarily cause errors on beta but are not correctly decoded matrices).

98784.a         AGL(2,7)
151632.a        ASL(3,3)
240.89           CSU(2,5)
20462400.a  CU(2,29)
291456.a        GU(2,23)

I also took this opportunity to show representatives of projective groups in many cases (basically if the order of the group is >2000 then group elements stored in the DB are correct, except for the automorphism group generators which are given as permutations not matrices).  This partially fixes #5562 in many cases. I added "[ ]" around each matrix to indicate that the element we are showing is really the equivalence class of the given matrix.  Suggestions for whether this is appropriate notation, and if we should make this choice of notation more clear, are welcome.

Here are some group IDs, (again, you can see by following the knowl for conjugacy classes in the complex character tables where you will see a representative which doesn't exist in beta). 
6072.a           PSL(2,23)
4680000.a   PSp(4,5)
378000.b      PGU(3,5)
1030200.b    PGL(2,101)

We also now show generators of subgroups for most projective groups
Here are PSL(2,23) subgroups and if you click on any of them you now see generators of the subgroup on the page for each subgroup
http://localhost:37777/Groups/Abstract/?ambient=6072.a&search_type=Subgroups

I also changed the "Construction" section to always show Lie type even if the name of the group matches the particular family.  This fixes an issue that is on the master branch but not yet on production. If you compare the master branch to this and look at 168.42 you'll see that PSL(2,7) is not listed for this group on master but is listed on this branch.

### DIFF
--- a/lmfdb/groups/abstract/main.py
+++ b/lmfdb/groups/abstract/main.py
@@ -2255,7 +2255,7 @@ def cc_data(gp, label, typ="complex"):
         else:
             gp_value = WebAbstractGroup(gp)
             if gp_value.representations.get("Lie"):
-                if gp_value.representations["Lie"][0]["family"][0] == "P":  #Problem with projective lie groups
+                if gp_value.representations["Lie"][0]["family"][0] == "P" and  gp_value.order < 2000:  #Problem with projective lie groups of order <2000
                     pass
                 else:
                     repn = gp_value.decode(wacc.representative, as_str=True)
--- a/lmfdb/groups/abstract/web_groups.py
+++ b/lmfdb/groups/abstract/web_groups.py
@@ -1717,10 +1717,23 @@ class WebAbstractGroup(WebObj):
 
     def _matrix_coefficient_data(self, rep_type, as_str=False):
         rep_data = self.representations[rep_type]
+        sq_flag = False # used later for certain groups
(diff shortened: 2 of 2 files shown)


roed314:
I made some small fixes, happy to merge once tests pass.

Issue #5800, opened 2024-01-08
Timeouts in Dirichlet characters


roed314:
We're getting timeouts from this query on production:
```
SELECT SUM("degree") FROM "char_orbits"
```
This is referenced from the [main Dirichlet character page](https://www.lmfdb.org/Character/Dirichlet/) via [this line](https://github.com/LMFDB/lmfdb/blob/master/lmfdb/characters/main.py#L737) of in the init method for the statistics object.  We're currently recomputing this each time; it would be good to cache it as a statistic.  While we currently have a [_record_extreme](https://github.com/LMFDB/lmfdb/blob/master/lmfdb/backend/statstable.py#L671) method, there isn't an analogous _record_sum method.  It would be good to add this infrastructure so that we can eliminate these timeouts, which have been occurring fairly frequently in the last few months.

roed314:
Fixed by #5889.

PR #5801, opened 2024-01-08
make nchars static

edgarcosta:
a temporary fix for #5800

### DIFF
--- a/lmfdb/characters/main.py
+++ b/lmfdb/characters/main.py
@@ -734,7 +734,7 @@ class DirichStats(StatsDisplay):
                   "is_real": yesno}
 
     def __init__(self):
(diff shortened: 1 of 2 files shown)


AndrewVSutherland:
We may eventually add more char_orbits, so this really should only be temporary...

edgarcosta:
Once that happens, the test below will fail, and inform us that we need to update the number:
```python
class DirichletSearchTest(LmfdbTest):
    def test_nchars(self):
        from lmfdb import db
        nchars = db.char_orbits.sum_column('degree')
        assert nchars == 3039650754 # if this fails, one also needs to update DirichStats.__init__
        W = self.tc.get('/Character/Dirichlet/')
        assert comma(nchars) in W.get_data(as_text=True)
```

PR #5802, opened 2024-01-08
Add minimal sibling to search

jwj61:
Now that we have minimal sibling information for more than 99% of the number field database, this adds it as a search option.  You can compare the counts for

D_4 quartic which is not a minimal sibling:
http://127.0.0.1:37777/NumberField/?degree=4&galois_group=d4&is_minimal_sibling=no
D_4 quartic which is a minimal sibling:
http://127.0.0.1:37777/NumberField/?degree=4&galois_group=d4&is_minimal_sibling=yes
D_4 quartic, both cases:
http://127.0.0.1:37777/NumberField/?degree=4&galois_group=d4

### DIFF
--- a/lmfdb/number_fields/number_field.py
+++ b/lmfdb/number_fields/number_field.py
@@ -1178,7 +1178,7 @@ class NFSearchArray(SearchArray):
             example_span="[8,3], 8.3, C5 or 7T2")
         is_galois = YesNoBox(
             name="is_galois",
-            label="Is Galois",
+            label="Galois",
             knowl="nf.galois_group")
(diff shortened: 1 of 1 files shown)


AndrewVSutherland:
I'm wondering if it would be clearer if we changed the caption "Minimal sibling" to "Is minimal sibling" (or if not, should we change "Is Galois" to "Galois"?)

jwj61:
I switch "Is Galois" to Galois since we have on the page "CM field", and on both abstract groups and Galois group index pages we don't say "Is" for boolean properties.

PR #5805, opened 2024-01-09
More updates to groups pages

jenpaulhus:
This PR should not be merged before my previous PR to avoid any possible conflicts.  

I. Added scroll bar to Group Statistics table.  A good example is a large Sn like 87178291200.a

In messing around with the scrolling, I managed to "fix" the small space issue in the scrolling newform tables by adding two \&nbsp; in front of "Label". There is probably a more elegant solution involving spacing, but this worked. You can check out scrolling on these two pages in beta and local:
http://localhost:37777/ModularForm/GL2/Q/holomorphic/9394/2/a/u/
http://localhost:37777/ModularForm/GL2/Q/holomorphic/4641/2/a/m/

II. Added a line to the Construction section to let users know which construction we use to store elements of the group (permutation, PC group, matrix group, Lie type, etc.).  Also included a line on the auto_gens pages and information about what the "[]" around matrices means for projective groups.  You can just hit random group a few times and see the message for permutation or PC groups.  For matrix groups, here are a few examples. 
GLFp   192.486
GLFq, 58806.d
GLZ, 4608.bt   
GLZN, 384.2000
GLZq 8192.mz

III. Prevent the  character table pages and the automorphism group generators pages from showing up if this data isn't stored in the db.  Beta loads a blank page, while local returns an error message:  localhost:37771/Groups/Abstract/char_table/335544320.bda

IV. Added a banner at top of the page to inform observers when the group is being computed live.  Many groups of order 512 fit this category, eg 512.4251

### DIFF
--- a/lmfdb/classical_modular_forms/templates/cmf_newform.html
+++ b/lmfdb/classical_modular_forms/templates/cmf_newform.html
@@ -34,7 +34,7 @@ function get_all_embeddings(num_embeddings) {
 <table class="ntdata complex-cols">
   <thead>
     <tr>
-      <th class="sticky-head dark" style="min-width: 40px;">&nbsp;&nbsp;{{ KNOWL('cmf.embedding_label', 'Label')}}</th>
+      <th class="sticky-head dark" style="min-width: 40px;">{{ KNOWL('cmf.embedding_label', 'Label')}}&nbsp;&nbsp;</th>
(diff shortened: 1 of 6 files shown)


jenpaulhus:
I think I've successfully fixed the live group error from last round of tests without messing anything else up. But worth another pair of eyes confirming.

Issue #5812, opened 2024-01-11
Be able to limit number of objects downloaded


jwj61:
A suggestion relayed to me verbally:

On a search results page, one can download the results.  Add the option to specify how many objects to download.  For example, someone searches for all D_4 quartic number fields but really just wants the first 1000 of them.  They want to specify this directly instead to tinkering with discriminant bounds or getting a huge download and having to then pare it down.

roed314:
@jwj61: This was implemented in #5820.  If you're happy with the interface that's live, feel free to close this issue.

PR #5820, opened 2024-01-16
Add ability to limit number of downloads

roed314:
Addresses #5812.  Here's a picture of the changed section of the download page by default:

Screenshot 2024-01-16 at 17 16 26

and after clicking on "all rows":

Screenshot 2024-01-16 at 17 16 52

### DIFF
--- a/lmfdb/static/lmfdb.js
+++ b/lmfdb/static/lmfdb.js
@@ -576,16 +576,22 @@ function control_sort(S) {
 
 function update_download_url(link) {
   // console.log("before modification", link.href);
+  var newval;
   var url = new URL(link.href);
   var params = url.searchParams;
   var keys = ["showcol", "hidecol", "sort_order", "sort_dir"];
   for (var i = 0; i < keys.length; ++i) {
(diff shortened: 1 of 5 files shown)


edgarcosta:
Should we put "all" as default? or something similar? Now it gives me the impression that I need to put a number.

(Sorry, I cant test the PR at the moment)

roed314:
If you don't do anything, it will download all rows.  The mechanism to select something other than all the rows is to click on the "all rows" link, at which point it will give you a text input.  If you enter an integer there, it will limit to that number of rows.  Anything other than an integer will give you all the rows.

I'm certainly open to other UI suggestions....

roed314:
I guess your suggestion is to have the input box pre-filled with "all"?  I'd be fine with that...

edgarcosta:
I understand now and now I'm more confused.
I don't think it is intuitive to click "all rows" if a user wants to limit the number of rows.

roed314:
I just saw Edgar's comment.  Maybe we can talk about other UI possibilities sometime this week, and make a revision to this PR.

Issue #5818, opened 2024-01-16
Wrong Sylow subgroups


a234:
There are some Sylow group errors on abstract group pages:
The Sylow 3-subgroup of [PSU(3, 2)](https://beta.lmfdb.org/Groups/Abstract/72.41) should be C3^2, not C3.

edgarcosta:
This is an interesting bug.
Is it easy to add @roed314 an internal test checking the cardinality of Sylow sugroups?

roed314:
Thanks for the report; we will look into it.

roed314:
The problem is widespread, and we haven't been able to figure out where in the code for computing the data that it came from.  But since it's so easy to recompute (just a gcd between the order of the subgroup and the order of the quotient and a power-power check), we'll just redo the computation for all subgroups.  There are some other things that we're currently working on recomputing, and reloading the subgroup table is difficult (since it's so big).  I'll report here when the issue is fixed on the backend.  In the meantime, I'll add some workarounds so that there isn't incorrect data being displayed.

jenpaulhus:
Temp fix #5822 should be undone once we permanently fix the Sylow data.

PR #5822, opened 2024-01-18
Temp fix for Sylow problems

roed314:
See #5818.

Compare the Sylow subgroups on https://beta.lmfdb.org/Groups/Abstract/12.3 and http://localhost:37777/Groups/Abstract/12.3.  This also removes the search boxes for Sylow and Hall subgroups on the subgroup search page.

Once the underlying data is fixed, this change can be reverted.


jenpaulhus:
Looks good. I will merge.

PR #5828, opened 2024-01-22
Add information on minimal linear degrees

roed314:
Advanced search now includes linear degrees over C and Q (but not R); the "Minimal degrees of faithful linear representations" contains info on both irrep and rep degrees over C, R, and Q.

Up on [groups.lmfdb.xyz/](https://groups.lmfdb.xyz/Groups/Abstract/?showcol=irrC_degree.irrR_degree.irrQ_dim.linC_degree.linR_degree.linQ_dim) for testing (the new columns are not shown by default).


jenpaulhus:
Looks good to me.

PR #5830, opened 2024-01-22
Change knowl sort order on the knowl index page

roed314:
Compare the ? sections of http://localhost:37777/knowledge/ and http://beta.lmfdb.org/knowledge/

### DIFF
--- a/lmfdb/knowledge/main.py
+++ b/lmfdb/knowledge/main.py
@@ -901,8 +901,8 @@ def index():
             knowls.append(k)
 
     def first_char(k):
-        t = k['title']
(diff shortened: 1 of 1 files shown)


BarinderBanwait (commented):
This fixes the knowls that start with a math symbol, but there are still some `?` knowls there? These seem to start with an apostrophe character `, can anything be done for these? Or is it that they e.g. don't have a title?

AndrewVSutherland:
@BarinderBanwait the issue with these is they don't have a title, I'm going through them now, but this PR looks good.

PR #5832, opened 2024-01-24
Add search for complex characters of groups

roed314:
The new search page is at http://localhost:37777/Groups/Abstract/?search_type=ComplexCharacters.  This PR also uses mathml to render character tables, allowing for much larger tables to be successfully opened (Firefox seems to do better than Chrome when the tables are larger than about 250x250, but this is past what beta is currently capable of).


jwj61:
LGTM  Maybe retrigger the tests to run

PR #5833, opened 2024-01-25
Allow some group searches to use order of group

jwj61:
For Center, Commutator subgroup, Central quotient, and Abelianization, one can enter a group label or the order of the desired group.  In fact, one can enter a comma separated list which is a mix of these things:

Central quotient of order 8:
http://127.0.0.1:37777/Groups/Abstract/?central_quotient=8&showcol=central_quotient

Center which is either 4.2, or orders 5 or 7
http://127.0.0.1:37777/Groups/Abstract/?center_label=4.2%2C+5%2C+7


roed314:
Looks good to me.  I added an order to the examples; feel free to merge once tests pass.

Issue #5836, opened 2024-01-25
Feature request: add infinite-order Mordell-Weil generators as columns in Elliptic Curve search results


alonamit1729:
When searching for elliptic curves it is currently not possible to include the Mordell-Weil generators as columns in the search results. It may be cumbersome to do so for curves of high rank, but for rank 1 (and possibly 2, 3) it would be extremely valuable to include them.

edgarcosta:
This is doable. However, it is not straightforward, as MW generators live in a different table from the search table.

roed314:
The main problem is with downloading, if we're getting the data from another table.

edgarcosta:
I agree, another not straightforward point :)

alonamit1729:
I should have mentioned that downloading is my main motivation :) I hope a `join` is possible!

roed314:
Ah.  Yeah, I didn't get downloading to work.  If you set up the LMFDB locally, you can do the following:
```
from lmfdb import db
D = {rec["lmfdb_label"]:rec["gens"] for rec in db.ec_mwbsd.search({}, ["lmfdb_label", "gens"])}
```
This gives you a dictionary with all the generators for all the curves in the LMFDB, and takes about 3.5 minutes to run.

JohnCremona:
In case it is helpful, the repositories https://github.com/JohnCremona/ecdata and https://github.com/JohnCremona/ecnf-data have the raw data as well as Sage code for reading the files.  Also, the web page https://johncremona.github.io/ecdata/ (which long predates the LMFDB) allows you to download all the generators in text files called allgens.* split into conductor ranges of 10000.

Of course it really should be true that anything you can do with these files you can also do with the LMFDB.

alonamit1729:
Thank you!!

roed314:
This was fixed in #6259, though it will be slow if there are a lot of search results (I haven't tested, so I would be curious what the actual limit is before you hit a 30 second timeout).

PR #5837, opened 2024-01-25
Add MW-gens search column to EC

roed314:
Unfortunately, it won't be present in downloads, since the data is in another table.  I don't see a clean way to fix that without join searches.  Addresses #5836.


BarinderBanwait (approved):
I've run the branch locally and can see the `mw-generators' option in the columns to display box, and tested it on curves of varying rank, so looks good to me

edgarcosta:
Thank you @BarinderBanwait

PR #5842, opened 2024-01-29
Update formatting for download options

roed314:
First page of results:
Screenshot 2024-01-29 at 14 08 50

Second page of results:
Screenshot 2024-01-29 at 14 09 04

Text download is now the default, and has been updated to make it easier to parse.  Besides the data lines, every non-empty line starts with a # symbol, and the data lines are now just tab-delimited for easier copy-pasting into a spreadsheet.

I also added an option to download search results as a CSV file for loading into a spreadsheet, since I think that's how a lot of people work (the comments are all deleted, but the first row is the column labels, with hyperlinks to the corresponding knowls).

### DIFF
--- a/lmfdb/lattice/test_lattice.py
+++ b/lmfdb/lattice/test_lattice.py
@@ -82,7 +82,7 @@ class HomePageTest(LmfdbTest):
 
     def test_downloadstring_search(self):
         L = self.tc.get("/Lattice/?class_number=8").get_data(as_text=True)
-        assert 'Download displayed columns' in L
+        assert 'displayed columns' in L
 
     def test_download_shortest(self):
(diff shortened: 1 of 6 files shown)


edgarcosta:
The the text in the "Next" button looks a bit high, but overall these changes are quite good

roed314:
Yeah, I was looking in the inspector and couldn't figure out why "Next" was as high as it is.

I'm making some other changes (and fixing the test failure).  I'll report back here when this is ready.

roed314:
I ended up adding support for CSV downloads, since I realized that I was trying to get text downloads to paste into spreadsheets and thought that we should just use the correct format.  I think this is ready for review (assuming tests pass).

PR #5850, opened 2024-02-06
some fixes ; auto-done by ruff linter

fchapoton:
this was scripted using the very fast linter ruff

`ruff --fix --select=UP034,UP036,UP024,UP003,UP025 lmfdb/`

### DIFF
--- a/lmfdb/backend/base.py
+++ b/lmfdb/backend/base.py
@@ -784,7 +784,7 @@ class PostgresBase():
                     print("Created file %s" % filename)
 
     def _check_header_lines(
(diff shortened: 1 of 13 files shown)


roed314:
Why change `IOError` to `OSError`?  Also, this now fails our linter (now `sys` is imported but not used).

fchapoton:
IOError is now an alias, so better use the new error name directly

fchapoton:
maybe I should rather undo the change to this parsing file ?

roed314:
Yeah, maybe it's best to leave it in since some people like to run the LMFDB on old versions of Sage.

PR #5865, opened 2024-02-13
master -> web

edgarcosta:


roed314:
Fine with me, but shouldn't we do master->dev first?

edgarcosta:
yes, got disconnected while creating the 2nd PR (:

PR #5881, opened 2024-02-23
Passport search

SamSchiavone:
Add a search page for passports, imitating the subgroup search page for abstract groups, as well as tests.

### DIFF
--- a/lmfdb/belyi/main.py
+++ b/lmfdb/belyi/main.py
@@ -677,17 +677,7 @@ def query_convert_abc_list(query):
         query.pop("abc_list")
     return query
 
-@search_wrap(
-    table=db.belyi_galmaps,
-    title="Belyi map search results",
-    err_title="Belyi map search input error",
-    columns=belyi_columns,
-    shortcuts={"jump": belyi_jump, "download": Belyi_download()},
-    url_for_label=url_for_label,
-    bread=lambda: get_bread("Search results"),
-    learnmore=learnmore_list,
-)
-def belyi_search(info, query):
+def common_parse(info, query):
     parse_bracketed_posints(info, query, "abc_list", "a, b, c", maxlength=3)
     query = query_convert_abc_list(query)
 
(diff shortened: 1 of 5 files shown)


lmfdb/belyi/main.py
```diff
@@ -710,6 +717,60 @@ def belyi_search(info, query):
...
+    parse_bracketed_posints(info, query, "abc_list", "a, b, c", maxlength=3)
+    query = query_convert_abc_list(query)
+    # a naive hack
+    if info.get("abc"):
+        for elt in ["a_s", "b_s", "c_s"]:
+            info_hack = {}
+            info_hack[elt] = info["abc"]
+            parse_ints(info_hack, query, elt)
+
+    parse_ints(info, query, "g", "g")
```
roed314, line 755: It would be better to factor out the common functionality between this and `belyi_search` into a common function (like in [classical modular forms](https://github.com/LMFDB/lmfdb/blob/main/lmfdb/classical_modular_forms/main.py#L728))

lmfdb/belyi/main.py
```diff
@@ -908,3 +969,77 @@ def __init__(self):
         self.browse_array = [[deg], [group], [abc], [abc_list], [g], [orbit_size], [pass_size], [field], [geomtype], [is_primitive], [primitivization], [count]]
 
         self.refine_array = [[deg, group, abc, abc_list], [g, orbit_size, pass_size, field], [geomtype, is_primitive, primitivization]]
+
+class PassportSearchArray(SearchArray):
```
roed314, line 973: This is also almost identical to the search array for Belyi maps.  It should be possible to use inheritance to factor out a bunch of common code.

roed314:
Looks good to me aside from benefiting from refactoring to reduce code duplication.

lmfdb/belyi/main.py
SamSchiavone: Sounds good. I'll revise this PR this week.

SamSchiavone:
@roed314 Hi David, I factored out the repeated code. Let me know if you have any more comments or suggestions.

lmfdb/belyi/main.py
```diff
@@ -682,6 +675,11 @@ def belyi_search(info, query):
             a = query["abc_list"][0]
             query["$or"] = [{"a_s": a}, {"b_s": a}, {"c_s": a}]
         query.pop("abc_list")
+    return query
+
+def common_parse(info, query):
+    parse_bracketed_posints(info, query, "abc_list", "a, b, c", maxlength=3)
+    query = query_convert_abc_list(query)
```
roed314, line 682: The query is modified in place, so you don't need to return it and set it agian.

scripts/belyi/find_friends.py
```diff
@@ -16,7 +16,10 @@ def base_field_label(rec):
 
 def genus2_lookup_equation(rec):
     f,h = curve_string_parser(rec)
-    return genus2_lookup_equation_polys(str([f,h]))
+    lab, _ = genus2_lookup_equation_polys(str([f,h]))
+    if lab:
+        return lab
+    return None
```
roed314, line 22: Technically `return None` is unnecessary here since that's the default behavior when there's no return statement.  But it's fine.

PR #5882, opened 2024-02-24
Add grd to number fields

jwj61:
This adds Galois root discriminants to number fields, both displaying and searching.  The number field database would have to be copied to the production server before this pull request gets there.  There is currently data for over 21 million of the number fields (half a million are currently not computed).


AndrewVSutherland:
The search grid spacing is off (minimal sibling should be one column to the right)
Screenshot from 2024-02-24 09-53-30

PR #5889, opened 2024-03-01
Dirchar timeout issue

BarinderBanwait:
addresses #5800 

Previously, going to the Dirchars homepage would recompute the number of dirichlet characters every time via a time-consuiming SQL query. This led to some timeouts.

Now, this is done as for min and max: this number is computed once and cached with the statstable of `char_orbits`; it is then looked up whenever the page is loaded. This should speed up loading that page.

### DIFF
--- a/lmfdb/backend/searchtable.py
+++ b/lmfdb/backend/searchtable.py
@@ -1493,5 +1493,13 @@ class PostgresSearchTable(PostgresTable):
         """
         return self.stats.count_distinct(col, query, record=record)
 
-    def sum(self, col):
-        return self.stats.sum(col)
+    def sum(self, col, constraint={}):
+        """
+        The sum of a given column.
+
+        INPUT:
+
+        - ``col`` -- the name of the column
+        - ``constraint`` -- a query dictionary constraining which rows are considered
+        """
+        return self.stats.sum(col, constraint)
--- a/lmfdb/backend/searchtable.py
+++ b/lmfdb/backend/searchtable.py
@@ -1492,3 +1492,7 @@ class PostgresSearchTable(PostgresTable):
         - ``record`` -- (default True) whether to record the number of results in the stats table.
         """
         return self.stats.count_distinct(col, query, record=record)
+
+    def sum(self, col):
+        print("wakka wakka")
+        return self.stats.sum(col)
--- a/lmfdb/backend/searchtable.py
+++ b/lmfdb/backend/searchtable.py
@@ -1494,5 +1494,4 @@ class PostgresSearchTable(PostgresTable):
         return self.stats.count_distinct(col, query, record=record)
 
     def sum(self, col):
-        print("wakka wakka")
         return self.stats.sum(col)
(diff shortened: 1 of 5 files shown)


lmfdb/backend/statstable.py
```diff
@@ -753,6 +753,62 @@ def min(self, col, constraint={}, record=True):
...
+
+        INPUT:
+
+        - ``col`` -- the column
+        - ``m``   -- the value to store as sum
+        """
+        inserter = SQL(
+                "INSERT INTO {0} "
+                "(cols, value) "
+                "VALUES (%s, %s)"
```
roed314, line 782: You need to include the kind of statistic as well as constraint columns.  See `_record_extreme` for an example.

lmfdb/backend/statstable.py
```diff
@@ -753,6 +753,62 @@ def min(self, col, constraint={}, record=True):
...
+            [Json([col]), m],
+        )
+
+    def _quick_sum(self, col):
+        """
+        This just looks up the sum; returns None if it hasn't been computed
+        """
+        values = [Json([col])]
+        selecter = SQL(
+            "SELECT value FROM {0} WHERE cols = %s AND threshold IS NULL"
```
roed314, line 795: Again, you need to make sure that you're looking for statistics of `sum` type.  See `_quick_extreme`.

lmfdb/backend/searchtable.py
```diff
@@ -1492,3 +1492,7 @@ def count_distinct(self, col, query={}, record=True):
         - ``record`` -- (default True) whether to record the number of results in the stats table.
         """
         return self.stats.count_distinct(col, query, record=record)
+
+    def sum(self, col):
```
roed314, line 1496: This should take a constraint as an optional argument.  See `max` for an example.

lmfdb/backend/statstable.py
```diff
@@ -753,6 +753,62 @@ def min(self, col, constraint={}, record=True):
...
+        ).format(Identifier(self.stats))
+        cur = self._execute(selecter, values)
+        if cur.rowcount:
+            return cur.fetchone()[0]        
+
+    def sum(self, col, record=True):
```
roed314, line 801: As above, this should take `constraint` as an optional argument, like `max` does.

lmfdb/backend/searchtable.py
roed314: You should probably change `sum_column` rather than add a new `sum` function.

BarinderBanwait: optional `constraint` now added. I actually renamed `sum_column` (something i put in before) to `sum` to be more in line with `min` and `max` naming.

lmfdb/backend/statstable.py
BarinderBanwait: done; actually i just used the existing `_record_extreme` which i renamed to `_record_statistic` because it's not just for min and max anymore

lmfdb/backend/statstable.py
BarinderBanwait: piggybacking off the existing functions now (i.e. done)

lmfdb/backend/statstable.py
BarinderBanwait: done

BarinderBanwait:
@roed314 addressed your comments

PR #5890, opened 2024-03-02
Couple quick subgroup fixes

jenpaulhus:
Fixes a couple errors on subgroup pages:

 sub/327680.ijy.20480.a1.N 

sub/5598720000000.b.60.A

sub/6718464.km.256.A

/sub/1920.204530.1920.a1


roed314:
Thanks for fixing these!

PR #5892, opened 2024-03-03
Add ABC quality and Szpiro ratio to elliptic curves over Q

roed314:


koffie:
I left some review comments on the knowls.

AndrewVSutherland:
There is some inconsistency between the use of "abc-quality" and "ABC-quality" (e.g. in the sort order versus display columns dropdowns, and in the knowls, where lowercase abc are used in the formula but the defined term uses ABC).  I would suggest consistently using lower case abc everywhere; this is not an acronym, the formulas are more pleasant in lower case, and I note the Wikipedia uses lower case.

roed314:
I've updated the PR and the two knowls to use lower case mathmode rather than upper case.

PR #5893, opened 2024-03-04
Dynamically generate spaces with nontrivial characters

thealexcowan:


lmfdb/classical_modular_forms/web_space.py
```diff
@@ -406,7 +521,7 @@ def __init__(self, level, weight):
                     space['char_orbit_label'] = char['label'].split('.')[-1]
                     space['char_degree'] = char['degree']
                     space['dim'] = newspace_dims_by_label[char['label']]
-                    space['in_mf_newspaces_db'] = False # This is used in self.decomposition(), to avoid making a hyperlink to a non-existant newspace page.
+                    space['in_mf_newspaces_db'] = (space['char_orbit_label'] != 'a') # This is used in self.decomposition(), to avoid making a hyperlink to a non-existant newspace page.
```
roed314, line 524: I think this also needs to check that weight=2.  See http://localhost:37777/ModularForm/GL2/Q/holomorphic/5040/4/, where there are many broken links.

roed314:
Thanks.  The last thing I want to do before merging this is compare your implementation of dimension formulas with what we already have.  I'll try to do that today.

roed314:
Looks good.

PR #5901, opened 2024-03-08
Uploads

roed314:
This PR adds a new upload process where LMFDB users can add data to the database, with both automated verification and human review.

To see the interface in action, check out
* https://blue.lmfdb.xyz/ModularCurve/Q/upload
* https://blue.lmfdb.xyz/users/uploads (some test data is there, requires login).

I am continuing to work on this, but wanted to get feedback.


jenpaulhus:
I really like this idea.  Thanks for your work on it!

Couple questions perhaps more about the process. 
1. Will we require code used to create any added data to be publicly available? 
2. How will we handle updating  completeness knowls if the new data impacts them?
3. Who will do the "human verification"?

roed314:
1. Currently there's a stub [here](https://github.com/roed314/lmfdb/blob/uploads/lmfdb/templates/upload_data.html#L4) where I say we should include some policies and documentation on the process.  I'm in favor of asking for code used to generate data be publicly available, but I think that's a decision to be made by the managing editors. 😄 Regardless of what we decide to require, perhaps we should add a place for users to include a github repo link.

2. Good question.  Since we are requiring an account to upload data, both editor and submitter have the power to update the completeness knowl.  This process is not the best option for uploading very large amounts of data (the verification process often requires multiple database lookups for each row), so I don't anticipate the completeness knowl needing to be updated very often.  But we should settle on whose responsibility it is for doing so, and document it.

3. Currently, a user has the power to review uploaded data if and only if they have the power to review knowls, which is (almost) the same as being an LMFDB editor.  So I think responsibility should fall on the LMFDB editors interested in the particular section of the LMFDB (currently just modular curves).  In terms of implementation, I was thinking that we should set up a google alert so that these editors get an email when there is data that needs review (that's why the [needs_review](https://github.com/roed314/lmfdb/blob/uploads/lmfdb/utils/uploader.py#L491) [route](https://github.com/roed314/lmfdb/blob/uploads/lmfdb/modular_curves/main.py#L299) exists).

Issue #5903, opened 2024-03-08
Support transitive group label in abstract group jumpbox


SamSchiavone:
It would be nice to be able to enter a transitive group label (e.g., 24T138) in the jumpbox on the abstract group search page, and get the abstract group associated to this transitive permutation group. This can be accomplished by first going to the Galois groups page (e.g., https://beta.lmfdb.org/GaloisGroup/24T138) and scrolling down to Label in the Group invariants section (e.g., 96.67), but this is a bit cumbersome.

jenpaulhus:
This has been implemented now so I'm going to close this issue.

PR #5906, opened 2024-03-09
Add transitive group option to abstract group jump box

roed314:
@SamSchiavone had a good idea in #5903.  With this PR, http://localhost:37777/Groups/Abstract/?jump=8T34 works.  After it's merged, I'll update the `group.find_input` knowl.


jenpaulhus:
Great idea!  Tiny issue with PR. Currently it gives a server error if the transitive group id doesn't exist (instead of an error message to the user).  Eg: put 2T3 into the jump box.

roed314:
Thanks.  There were instances of raising errors further down in the function, so I thought they were caught, but that was not the case.  This PR now also makes groups pass the banana test.

PR #5910, opened 2024-03-11
add new test on searching for level

junbolau:
Added a test for level range, tested locally.


roed314:
The contents of this PR are in #5926

PR #5921, opened 2024-03-12
Isogeny classes of modular curves

assaferan:
Added pages for isogeny classes of modular curves, including data about the jacobian and other isogeny invariants, with a list of the coarse modular curves in the isogeny class, with some of their properties (cusp widths, cusp orbits, rational cusps, rational CM points, gonality)


roed314:
lint failure, but the pages look good.  I haven't read the code yet, but it would be cool to demo this this morning if we can get it in.

PR #5939, opened 2024-03-13
Update knowls and add tests for jump

junbolau:
Inconsistent knowls between labels before change.


junbolau:
Added tests for jump, interesting curves, image, random.

PR #5940, opened 2024-03-13
add check on conductor exponent from Brumer-Kramer-Martin bounds

BarinderBanwait:
sanity checking the conductor exponent for when -1 is in H, and the conductor is present. I have run this locally and it finishes in 575s.


BarinderBanwait:
the failing test seems unrelated to my changes

PR #5958, opened 2024-03-17
inexact gonality search, ui-tweaks

AndrewVSutherland:
This PR adds an "inexactly" search option to interval queries that will match only intervals that properly contain the specified value and uses it to provide an option to search for curves whose gonality bounds permit a specified gonality but for which the exact gonality is not known, for example

http://127.0.0.1:37777/ModularCurve/Q/?level=1-70&genus=3-&gonality_type=inexactly&q_gonality=2&contains_negative_one=yes

returns the 1293 curves of level up to 70 of genus >= 3 for which the gonality could be 2 but is not known to be 2 (as noted in the Zulip thread for the modular curves workshop, this includes cases where we really should know that the gonality has to be greater than 2).

The PR also cleans up a few tiny ui issues (number in property box not in math mode, option box, table alignment, etc..)

### DIFF
--- a/lmfdb/modular_curves/isog_class.py
+++ b/lmfdb/modular_curves/isog_class.py
@@ -106,7 +106,7 @@ class ModCurveIsog_class():
         if self.conductor is not None:
             self.properties.append(('Conductor', '$' + self.web_curve.factored_conductor + '$'))
         if self.rank is not None:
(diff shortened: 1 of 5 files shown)


roed314:
LGTM

PR #5972, opened 2024-03-27
Add info the group automorphism generators page

jwj61:
This adds the order of each listed automorphism, and if it is inner, and element which will induce the automorphism.  It also restores the sidebar for these pages so they look more like regular lmfdb pages.  Finally, these pages are turned off for all matrix groups since there seems to be issues with the data.

For a group given by a presentation:
http://beta.lmfdb.org/Groups/Abstract/auto_gens/1440.3334
http://127.0.0.1:37777/Groups/Abstract/auto_gens/1440.3334

Group given by permutations:
http://beta.lmfdb.org/Groups/Abstract/auto_gens/24.12
http://127.0.0.1:37777/Groups/Abstract/auto_gens/24.12

One where the permutations are large, so it doesn't fit on the screen:
http://beta.lmfdb.org/Groups/Abstract/auto_gens/204073344.mo
http://127.0.0.1:37777/Groups/Abstract/auto_gens/204073344.mo

Finallly, a group where we no longer give a like to automorphism generators (near the top is the line for the automorphism group)
http://beta.lmfdb.org/Groups/Abstract/192.486
http://127.0.0.1:37777/Groups/Abstract/192.486

### DIFF
--- a/lmfdb/groups/abstract/templates/auto_gens_page.html
+++ b/lmfdb/groups/abstract/templates/auto_gens_page.html
@@ -8,21 +8,35 @@
 <p>
  Each row represents an {{KNOWL('group.automorphism', 'automorphism')}} given by its image on a set of {{KNOWL('group.generators','generators')}} of the group (listed as the headers of the columns). <br>
 
+<p>
+The column "Inner by" gives an element which gives rise to the automorphism as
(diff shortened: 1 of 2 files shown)


roed314:
There are more things to be done in this direction, but this is certainly an improvement.

PR #5974, opened 2024-03-29
use pathlib.Path in platt_zeros

fchapoton:
as this make cleaner code than using the `os` module

plus pep8 cleanup of the modified file


roed314:
LGTM.  I hadn't know about pathlib; thanks for the suggestion.

PR #5983, opened 2024-04-12
Dirichlet characters to modulus 10^6

AndrewVSutherland:
The new table `char_dirichlet` contains Dirichlet character orbit data up to modulus 10^6, extending `char_orbits` which only covered modulus up to 10^5.  To keep the storage reasonable several columns were changed, necessitating a new table, but the UI should be unchanged (except for a bug fix, computing character values should now work consistently, where as it currently does not, e.g. try evaluating https://www.lmfdb.org/Character/Dirichlet/17/14 (no output is generated because there is a 404 on the request for the value due to a malformed URL).

This PR should not be merged until the table `char_dirichlet` has been copied to prod (in progress now).  Tests pass but I'd appreciate another set of eyes on the code to see if I have missed anything.  In the new schema, the integer `parity` column has been replaced by `is_even` (a boolean), the `first_label` and `last_label` text columns have been replaced by `first` and `last` integers that specify the Conrey index, the `primitive_orbit_label` text column has been replaced by a `primitive_orbit` integer specifying the orbit (as an ordinal starting from 1,  the modulus is the conductor), and the `orbit_index` integer column has been renamed `orbit` for consistency (was and still is an ordinal starting from 1).

### DIFF
--- a/lmfdb/backend/searchtable.py
+++ b/lmfdb/backend/searchtable.py
@@ -308,7 +308,7 @@ class PostgresSearchTable(PostgresTable):
             []
             sage: db.lfunc_lfunctions._parse_values({'bad_lfactors':[1,2]})[1][0]
             '[1, 2]'
-            sage: db.char_orbits._parse_values({'modulus':3})
+            sage: db.char_dirichlet._parse_values({'modulus':3})
             [3]
         """
 
--- a/lmfdb/characters/ListCharacters.py
+++ b/lmfdb/characters/ListCharacters.py
@@ -1,8 +1,6 @@
 # ListCharacters.py
 import re
 from sage.all import lcm, factor, Integers
-from sage.databases.cremona import cremona_letter_code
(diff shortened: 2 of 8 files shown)


roed314:
@AndrewVSutherland, some of the test errors look like they might be real.  Can you take a look?

AndrewVSutherland:
`char_dirichlet` has been copied to proddb. I'm looking into the test failures now.

AndrewVSutherland:
@roed314 Tests now pass.

roed314:
Great.  I'll take a look soon.

roed314:
All the changes look good to me.  One thing I noticed (not new to this PR):

On http://localhost:37777/Character/Dirichlet/123456, the structure is listed as C2 x C2 x C2 x C5136, but the generators have order 2, 16, 2, and 642.  There's no contradiction here, but it might be nice if the orders of the generators matched the description.

I'm merging this PR, since the generator orders is not a new phenomenon.

PR #5985, opened 2024-04-13
use pathlib.Path in backend/database

fchapoton:
another file converted from `os.path` to `pathlib.Path` for cleaner code

### DIFF
--- a/lmfdb/backend/database.py
+++ b/lmfdb/backend/database.py
@@ -1,12 +1,11 @@
 # -*- coding: utf-8 -*-
 import csv
 import logging
-import os
+from pathlib import Path
 import time
 import traceback
 import itertools
 from collections import defaultdict, Counter
-from glob import glob
 
 from psycopg2 import connect, DatabaseError
 from psycopg2.sql import SQL, Identifier, Placeholder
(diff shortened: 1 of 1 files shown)


roed314:
I would rather leave the files in `lmfdb/backend` alone for this kind of change, since they're awkwardly shared with some other projects ([psycodict](https://github.com/roed314/psycodict), [researchseminars](https://github.com/roed314/seminars) and [psetpartners](https://github.com/AndrewVSutherland/psetpartners).

fchapoton:
ok, too bad. I have carefully double-checked my changes...

AndrewVSutherland:
@roed314 So should we close this PR?

roed314:
I talked to @edgarcosta some more, and I don't think this is a big deal (syncing up different versions of this folder across different projects will have to be done anyway, and this won't really make it much harder).  So I'm merging this.

@fchapoton, I certainly wasn't objecting to this on principle (indeed, you've motivated me to start using pathlib in my own code).  It's just a local concern about my plans to straighten out the dependency situation with this folder (Edgar and I are going to meet on May 8 to try to make it its own pip package).

roed314:
Just a note about Pathlib, which I've been learning about.  Here's a quote from [this discussion](https://discuss.python.org/t/is-there-a-pathlib-equivalent-of-os-scandir/46626): "Pathlib is superb, but its primary benefit is code readability (and writeability). Not raw performance on unmanageably extreme numbers of files."  That makes it fine for the LMFDB, but others should be aware when using it elsewhere (I'm currently writing some scripts for finite groups dealing with folders with 500K files).

With that said, I think it works well in the LMFDB server contexts where @fchapoton is applying it.

PR #5994, opened 2024-04-17
Fix server error on links that are both old and broken

roed314:
Followup fix to #5982: now need to catch a `ValueError` on urls like https://www.lmfdb.org/ModularForm/GL2/Q/holomorphic/20/8/0/


edgarcosta:
LGTM

Issue #525, opened 2015-10-28
Improve character tables


jwj61:
Character tables are poorly displayed when they are very large, e.g. for 20T1030.  They are also not connected with other things like Artin representations, and the columns may not match the ordering of the conjugacy classes.

jwj61:
Also, they should be more clear as to what special symbols are, and what the data above the table means.

This will likely be handled through the finite groups project.

jenpaulhus:
Ok to remove the finite groups label now that we have a new beta with improved character tables and this is more about the other pages? (I'm trying to clean up old issues since we'll be adding new issues with the revamped beta pages.)

jwj61:
No

PR #5998, opened 2024-04-20
Better character tables for Galois groups

jwj61:
Addresses issue #525 

This replaces the old GAP character tables with ones computed for abstract groups.

The most typical case is where we have stored an explicit isomorphism to an abstract group in the database, such as

http://127.0.0.1:37777/GaloisGroup/5T4
http://beta.lmfdb.org/GaloisGroup/5T4

This is for almost all groups in degree <=15.  The conjugacy class table now has labels, and the character table is typeset with the conjugacy class labels being knowls with data on the class including a representative.  We still show power class data.  The old tables had the order of the centralizer for a class in a slightly cryptic manner; now the centralizer itself appears in the knowl.

There are some cases where we don't have an explicit isomorphism, or it is suppressed because of data issues with finite groups, such as

http://127.0.0.1:37777/GaloisGroup/5T5
http://beta.lmfdb.org/GaloisGroup/5T5

It now looks like the information above, but the conjugacy classes don't have labels in the main conjugacy class list, and the knowls for the classes in the character table do not list a representative.

Another case is where there are many conjugacy classes.  Before we just didn't give anything on conjugacy classes or the character table.  Now they are in knowls, so they are available if the user really wants to see them.

http://127.0.0.1:37777/GaloisGroup/32T12264
http://beta.lmfdb.org/GaloisGroup/32T12264

In some cases, the abstract group is not in the database or we don't have a character table for it.  

http://127.0.0.1:37777/GaloisGroup/46T12
http://beta.lmfdb.org/GaloisGroup/46T12

Then we give conjugacy class data but no character table.

Finally, the code shows some contortions for the trivial group which for some reason is always a problem here (gap does not have 1T1).

### DIFF
--- a/lmfdb/galois_groups/main.py
+++ b/lmfdb/galois_groups/main.py
@@ -14,7 +14,7 @@ from lmfdb.utils import (
     clean_input, prep_ranges, parse_bool, parse_ints, parse_galgrp,
     SearchArray, TextBox, TextBoxNoEg, YesNoBox, ParityBox, CountBox,
     StatsDisplay, totaler, proportioners, prop_int_pretty, Downloader,
-    search_wrap, redirect_no_cache)
+    sparse_cyclotomic_to_mathml, search_wrap, redirect_no_cache)
 from lmfdb.utils.interesting import interesting_knowls
 from lmfdb.utils.search_columns import SearchColumns, LinkCol, MultiProcessedCol, MathCol, CheckCol, SearchCol
 from lmfdb.api import datapage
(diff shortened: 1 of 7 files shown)


AndrewVSutherland:
There are some unused imports lint is complaining about:
```
lmfdb/groups/abstract/web_groups.py:302:13: local variable 'gpc' is assigned to but never used
lmfdb/groups/abstract/web_groups.py:3013:9: local variable 'force_repr' is assigned to but never used
```

AndrewVSutherland:
And it looks like there are some failures in the tests for abstract groups:
```
FAILED lmfdb/groups/abstract/test_browse_page.py::AbGpsHomeTest::test_highlighted_character
FAILED lmfdb/groups/abstract/test_browse_page.py::AbGpsHomeTest::test_lookup
```

AndrewVSutherland:
Still some tests failing:
```
lmfdb/galois_groups/test_galoisgroup.py::GalGpTest::test_underlying_data FAILED
lmfdb/tests/test_dynamic_knowls.py::DynamicKnowlTest::test_character_table_knowl FAILED
```

jwj61:
The underlying data test is now a faulty test.  It looks that "10.1" does not appear on a certain page, but that is the current version of sage, so it appears on all pages.

The dynamic knowl test passes for me locally, so I am not sure what's happening there yet.

AndrewVSutherland:
The dynamic knowl test is failing for the same reason (no `character-table.html`).

jwj61:
Oops.  I will need to remember that as a potential cause of test failures.

PR #6011, opened 2024-04-23
Added Zigzag functions and plots

sarangop1728:


roed314:
I'm going to merge this now because it looks good, but there are some more followup improvements to work on:
* Add the Hodge vector to the right hand side of the zigzag picture
* Provide a mode switch option on the zigzag picture that looks like a step function instead.

PR #6012, opened 2024-04-23
Add close tags to html tables, remove "Difficulty" from hgm page

roed314:


roed314:
The only visible change should be the removal of "Difficulty" from motive invariants.  Everything else is just cleaning up html.

PR #6028, opened 2024-05-06
Write cycle types with multiplicity

jwj61:
On Galois group pages, the conjugacy class section lists the classes with information on each including cycle types.  This changes something like 3,2,2,2,1,1 to 3, 2^3, 1^2.

Here is A_9 (which has a lot of cycle types):

http://beta.lmfdb.org/GaloisGroup/9T33
http://127.0.0.1:37777/GaloisGroup/9T33


roed314:
The test failure is from me deleting some `gps_subgroups` indexes on Friday in an effort to save space.  I'll add a few back in and rerun.

jwj61:
Thanks for the code improvements

PR #6030, opened 2024-05-07
Update backend code

roed314:
Fix refresh_null_counts bug, store table descriptions in knowls, other fixes in preparation for using psycodict directly


roed314:
@edgarcosta said he was happy with this in person.

PR #6037, opened 2024-05-10
Add rigorous Maass forms to the LMFDB

davidlowryduda:
Woohoo!

Also thanks to David Roe.


AndrewVSutherland:
I've put this up on https://blue.lmfdb.xyz/ModularForm/GL2/Q/Maass/ and would encourage people to take a look and give feedback (especially @davidfarmer).  Note that once this PR is merged and main is pushed to web, this will supersede the current Maass form section of the LMFDB at https://www.lmfdb.org/ModularForm/GL2/Q/Maass/.  We shouldn't merge this PR until we are ready to do that, and I think it would be good to get the portraits in place before that happens (I understand that @davidlowryduda is working on this).

The RCS knowls are currently all empty because you are using maass_forms rather than maass in the knowl identifiers, is that intentional?  In any case, I think all of these knowls need to be filled in before we merge.

Also, we need to copy the `maass_rigor` table to production (and the new portraits) before this is merged, but is that the table name we are planning to use indefinitely?

davidfarmer:
I see a lot of places where is says the Fricke eigenvalue is "not computed rigorously".
The examples I saw were for level 3 and 5, so it is just +1 or -1.

For composite levels, such as 15, it does not provide the Atkin-Lehner eigenvalues. 
I presume you know those, since (to avoid cusps) those were used in the computations.

Will L-functions be computed before this goes live?

For non-squarefree level, will those be empty, or will the old non-rigorous data be inserted?

On the browse page it offers 4, 8 and 9 as levels to click on, even though there is no data.

I can review the source/completeness/reliability knowls once available.

davidlowryduda:
I wrote RCS knowls under a different KNOWL hierarchy. I'll transfer them over.

> I see a lot of places where is says the Fricke eigenvalue is "not computed rigorously".
> The examples I saw were for level 3 and 5, so it is just +1 or -1.
> 
> For composite levels, such as 15, it does not provide the Atkin-Lehner eigenvalues.
> I presume you know those, since (to avoid cusps) those were used in the computations.

In these cases, we didn't use a rigorous Hejhal to compute the Maass forms. Instead they were computed using (Andrei's) trace formula implementation. And for some reason my rigorous Hejhal implementation found some problem doing something rigorously. In practice, this means there is some unlucky Bessel computation.

> Will L-functions be computed before this goes live?

I wasn't planning on it. Some of the coefficients are to poor (but rigorous) precision.

> For non-squarefree level, will those be empty, or will the old non-rigorous data be inserted?

I was thinking of only having rigorous data.

> On the browse page it offers 4, 8 and 9 as levels to click on, even though there is no data.

Good catch! I'll remove those.

> I can review the source/completeness/reliability knowls once available.

Thank you!

davidfarmer:
Maybe we have enough precision to compute the Fricke/Atkin-Lehner eigenvalues accurately
enough so that, invoking the fact that those exist and are +1 or -1, we have them rigorously?

The previous "interesting" Maass forms should be removed from the 'some interesting" list.

roed314:
> The previous "interesting" Maass forms should be removed from the 'some interesting" list.

These should stay until we're ready to merge the PR, since if we we delete them now they'll disappear from the live site.  But yes, once we're ready to merge we'll delete the old ones.

davidfarmer:
About the Fricke sign:

How about the Fricke sign (and Atkin-Lehner, if possible) is shown in all
cases, but (not computed rigorously) appears after that number, if appropriate?

davidlowryduda:
> Maybe we have enough precision to compute the Fricke/Atkin-Lehner eigenvalues accurately enough so that, invoking the fact that those exist and are +1 or -1, we have them rigorously?

This is almost always true. But I want to make sure that I implement that correctly and handle the error bounds correctly. It's a bit of a different way of working with error than I had been. I'll modify the data later after I've worked that out. And when I can't guarantee it rigorously, I'll add the heuristic in.

> The previous "interesting" Maass forms should be removed from the 'some interesting" list.

I'll do that as soon as this PR is approved.

davidlowryduda:
@davidfarmer @AndrewVSutherland I've also added in RCS knowls and a portrait knowl.

I think this PR is ready.

AndrewVSutherland:
I've copied the data to proddb.  Does anyone (including @arbooker @davidfarmer @jenpaulhus @jvoight @jwj61) have any remaining concerns or comments on this PR?  I'd like to merge it tomorrow if we think it is ready (but happy to wait if there are people who want to review it that need more time).

AndrewVSutherland:
Question for @davidlowryduda from @arbooker: "The Hecke eigenvalue data seems to be limited to about 8 decimal places of precision, even for level 1 (where we have much more accurate data). Is that by choice?"

Andy also noted that we need to delete the top/bottom knowls for the old Maass form data so that we don't get duplicates on the list of interesting Maass forms (I can take care of that).

AndrewVSutherland:
No, we should remove non-squarefree levels from the browse list

jenpaulhus:
I know very little about Maass forms so I mainly just clicked around to see if I could break anything or see anything weird. 

Neither of the two things below  necessarily need to happen before this PR is merged, but mentioning them now.

(1) It might be good to have the spectral parameters cut off at a certain length on the individual pages so that if your browser screen isn't very wide, you can still fit them in and not force a lot of white space at the top. Instead of showing the whole spectral parameter, you can click a button to see it all or to copy it. I believe Edgar originally coded this for somewhere else, but it's on the groups pages (eg: Permutation Group presentation part way down the page here https://beta.lmfdb.org/Groups/Abstract/1679616.oq)    Here's an example of a long spectral parameter:
https://blue.lmfdb.xyz/ModularForm/GL2/Q/Maass/1.0.1.46.1

(2) There was also something strange  when I looked at the pages on my phone.  A few entries in the tables on the individual pages were not LaTeXing.  I don't know what that's about.  I've attached a screenshot. It happened for many pages, not just this one.

IMG_0253

davidlowryduda:
Thanks!

> (1) It might be good to have the spectral parameters cut off at a certain length on the individual pages so that if your browser screen isn't very wide, you can still fit them in and not force a lot of white space at the top. Instead of showing the whole spectral parameter, you can click a button to see it all or to copy it. I believe Edgar originally coded this for somewhere else, but it's on the groups pages (eg: Permutation Group presentation part way down the page here https://beta.lmfdb.org/Groups/Abstract/1679616.oq) Here's an example of a long spectral parameter: https://blue.lmfdb.xyz/ModularForm/GL2/Q/Maass/1.0.1.46.1

Good idea. I've now pushed a change where I now use the same functionality that you pointed to. When the spectral parameter is very long, I truncate it and add a raw-toggle where the full value can be seen and copied.

> (2) There was also something strange when I looked at the pages on my phone. A few entries in the tables on the individual pages were not LaTeXing. I don't know what that's about. I've attached a screenshot. It happened for many pages, not just this one.

I have no clue!

AndrewVSutherland:
@jenpaulhus I think the latex display problems are likely inside  katex and something we have no control over.

I see the same issue on my iPhone when I use Safari, but if I switch to Chrome the coefficients are displayed correctly.

AndrewVSutherland:
The `maass_rigor` tables have been copied to proddb, I think this PR is ready to merge.  If anyone has any remaining concerns, please raise them now!  Unless there are any objections I will merge this tomorrow.

AndrewVSutherland:
@davidlowryduda @roed314 rigorous Maass forms are up on beta, please let me know if anyone spots anything amiss!

roed314:
I deleted the old top/bottom knowls.  I'll keep looking to see if I spot anything else wrong.

PR #6043, opened 2024-05-14
Add subgroups labels as option for jump box

jenpaulhus:
Title says it all.  Go to /Groups/Abstract/ and enter a subgroup label in the jump box and it will bring you to subgroup page.

Once this is merged, I will update the knowl group.find_input that describes what the label can be to include subgroup labels.


jenpaulhus:
These failures are all about an error in a backend file.

roed314:
I merged in #6044 to make tests pass, so someone else should look at that PR.

PR #6044, opened 2024-05-15
Fix cmf test

roed314:


edgarcosta:
why the change?

roed314:
Because the old one didn't work, presumably because @AndrewVSutherland has been adding data.

roed314:
See [here](https://beta.lmfdb.org/ModularForm/GL2/Q/holomorphic/?Submit=sage&download=1&query=%7B%27dim%27%3A+%7B%27%24gte%27%3A+30000%7D%2C+%27num_forms%27%3A+%7B%27%24exists%27%3A+True%7D%7D&search_type=SpaceTraces) for the current test, which is broken.

PR #6045, opened 2024-05-15
Remove restriction on lie groups

jwj61:
When showing conjugacy classes and character tables on Galois group pages, we added labels to conjugacy classes.  Initially, this was blocked if the abstract group was internally of Lie type because of data issues.  This is no longer needed, so this removes blocking on the level of code.

https://beta.lmfdb.org/GaloisGroup/16T28
http://127.0.0.1:37777/GaloisGroup/16T28


roed314:
I'm merging this, but it would be good to add knowls to the conjugacy class table headers.

PR #6048, opened 2024-05-16
Groups updates

jenpaulhus:
Fixes the link from group pictures.  Was showing xml code and now shows picture again.  (Click on your favorite group and then click on the picture which will take you to a page with just the picture.)

Also fixes an regex issue for Underlying Data link for subgroups.  Now a subgroup like [12800000000.com.32.BW](https://beta.lmfdb.org/Groups/Abstract/sub/12800000000.com.32.BW) will show underlying data when you click on link.


roed314:
Looks good.  I'm still not happy with the solution we came up with the for the image, but I'll make an issue since this is definitely an improvement.

PR #6054, opened 2024-05-18
add a few more cmf search options

AndrewVSutherland:
Adds the ability to search for twist minimal newforms, and newforms/newspaces of prime squared or powerful level, as well as adding twist minimal and minimal twist as optional search columns.

### DIFF
--- a/lmfdb/classical_modular_forms/main.py
+++ b/lmfdb/classical_modular_forms/main.py
@@ -1672,8 +1672,6 @@ class CMFSearchArray(SearchArray):
 
         results = CountBox()
 
-        wt1only = BasicSpacer("(only for weight 1)")
-
         trace_coldisplay = TextBox(
             name='n',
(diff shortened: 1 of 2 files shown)


roed314:
It looks like you added `prime_squared` and `powerful` to the dropdown for level, but I don't think the parser supports them so they'll be ignored.

AndrewVSutherland:
> It looks like you added `prime_squared` and `powerful` to the dropdown for level, but I don't think the parser supports them so they'll be ignored.

These are handled in the cmf code, all the Boolean level options just add "level_is" in front of the qualifier so no new parsing is required you just need to add the columns to the relevant tables, which I did.

roed314:
Got it.  I was confused because I had sage running from before you added the columns in beta, so when I looked at `db.mf_newforms.search_columns` on my computer I didn't see them.

PR #6055, opened 2024-05-18
main -> dev

AndrewVSutherland:


roed314:
@AndrewVSutherland I don't think the prime_square and powerful options will work (see my comment on the last PR).  I can take a look in a few minutes (trying to finish up some things on olive.lmfdb.xyz).

roed314:
Yep, that's right; currently powerful and prime_square will have no effect.  I'll fix it.

AndrewVSutherland:
Already done and it does work

roed314:
Ok, cool.  I guess I needed to restart my sage, since the list of search columns is cached.

AndrewVSutherland:
Ah, I think you mean on prod, yes they won't work there until the tables are copied over.  I can take of that tonight.

roed314:
`level_is_prime_square` seems to mean an even power of a prime, which isn't what I would expect from the name.

AndrewVSutherland:
It should mean square; I probably made a mistake when adding the columnSent from my iPhoneOn May 18, 2024, at 2:36 PM, roed314 ***@***.***> wrote:﻿
level_is_prime_square seems to mean an even power of a prime, which isn't what I would expect from the name.

AndrewVSutherland:
Fixed.

AndrewVSutherland:
The updated tables have been copied to prod.

PR #6061, opened 2024-05-20
fix sorts, resolves #5979

### LINKED ISSUE #5979: Sorting by label
In 
  http://www.lmfdb.org/padicField/?n=8&p=2&c=18&hidecol=p.c.e.f.gal.slopes
the fields are helpfully sorted by label, but instead of 
  2.8.18.1, 2.8.18.2, ... 
it does
  2.8.18.1, 2.8.18.10, ...
(sorting the strings).  Note that sorting by label is not an option in the dropdown list--perhaps we could add that, since we using it to break ties.  But wouldn't it also make more sense to sort them in the first way?  (It would make my life easier, when I'm grabbing number 18 from the list, it's easier when that's in the 18th place.)  

I have a workaround, so this is low priority.

AndrewVSutherland:


lmfdb/modular_curves/main.py
```diff
@@ -962,11 +962,12 @@ def __init__(self):
             [CPlabel],
         ]
 
+    _default = ["level", "index", "genus", "coarse_class_num", "coarse_level", "coarse_num", "fine_num"]
     sorts = [
-        ("", "level", ["level", "index", "genus", "label"]),
-        ("index", "index", ["index", "level", "genus", "label"]),
-        ("genus", "genus", ["genus", "level", "index", "label"]),
-        ("rank", "rank", ["rank", "genus", "level", "index", "label"]),
+        ("level", "level", _default),
```
edgarcosta, line 967: I think we want `"", "level", _default` for the first entry

AndrewVSutherland: Done (although I don't think it actually matters).

roed314: The way it matters is just in the url: if you have a blank default sort option it doesn't clutter urls.

roed314:
For the rank sort, do we want genus to be next (as was done before), or just `["rank"] + _default`?

edgarcosta:
lgtm

edgarcosta:
> For the rank sort, do we want genus to be next (as was done before), or just `["rank"] + _default`?

genus next

PR #6062, opened 2024-05-20
Remove unittest2, return to unittest

davidlowryduda:
We are currently using a custom-modified unittest2, which itself was a modified version of the standard library package unittest.

Originally, unittest2 allowed people to use later versions of unittest in early 2.* versions of python. It also backported certain utilities from python's 3.4 version of unittest back to python2.

We don't use python2 anymore. We don't need unittest2 anymore. This commit removes all unittest2 and removes the custom unittest2 dependency.


davidlowryduda:
As additional background, I'll note that I was experiencing a testing heisenbug that sometimes affected me (and the CI on my fork). I talked about this with David Roe and Edgar, but it felt really mysterious. This *doesn't* solve that heisenbug.

roed314:
Fine with me.

edgarcosta:
should we add `unittest` to `requirements.txt`?

also, given that you touched, I would consider removing `install-deps.sh`. @roed314 do you agree?

roed314:
Yes, I think removing `install-deps.sh` and adding `unittest` to `requirements.txt` is a good plan.

davidlowryduda:
We don't need to add `unittest` to `requirements.txt` because it's in the python standard library.

AndrewVSutherland:
This is likely going to fail the CMF tests as well, but if that is all that fails I will merge it.

PR #6063, opened 2024-05-20
Postgres_FAQ updates

davidlowryduda:
I make a few changes to the postgres FAQ, as this was my primary resource for uploading data into the LMFDB. Overall it was good, but I noticed a few things and correct them now.

Since #6032, it's a bit more out of date. I went ahead and tried to change appropriate references to `psycodict` instead of `backend`.

In practice, I think it's necessary to open `psycodict` and read/grep through various parts of its source in order to know how to upload data into the LMFDB.

### DIFF
--- a/Postgres_FAQ.md
+++ b/Postgres_FAQ.md
@@ -13,8 +13,10 @@ Changes
 
 1. What's an overview of the changes?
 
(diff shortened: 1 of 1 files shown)


roed314:
The best option, of course, is to improve the documentation of psycodict, but this is good in the meantime!  Thanks.

PR #6066, opened 2024-05-23
remove dependencies on mf_hecke_newspace_traces

AndrewVSutherland:
The table `mf_hecke_newspace_traces` takes up a lot of space (about 70GB) and is currently out of date.  Extending it to include traces for the new data that has been added (and will be added soon) would increase its size substantially.

This table is used to support trace table searches on newspaces (not newforms), but this search is essentially never used --all the examples we could find in the logs were either from us or bots), and I'm not aware of any use case where you would want to identify an entire newspace using traces (without knowing what the newspace is.  This PR removes the option for creating trace table search for newspaces from the UI and removes all code dependencies on this table.  Once it is merged and pushed to production we can delete `mf_hecke_newspace_traces`.

Note that traceforms for each newspace are still shown on newspace home pages and available for download, it is only the ability to search for, e.g. all newpaces with a3=17, that is being removed (supporting this requires a separate row for each individual trace, as well as associated index entries, and this is much more costly than storing an array of traces with each newspace, which we will continue to do).

None of this impacts the ability to do trace table searches on newforms, which we will continue to support.


AndrewVSutherland:
I also removed references to `mf_subspaces` which is no longer used (Alex Cowan removed this dependency a while ago, but there were a few lingering references).

We should delete both `mf_subpaces` and `mf_hecke_newspace_traces` after this PR is merged and live.

PR #6070, opened 2024-05-26
remove superflous utf8 declarations

fchapoton:
as this is the default encoding


AndrewVSutherland:
The test failure is caused by the recent data change and this PR not including the CMF changes I just merged (the random test hits a server error).  merging.

PR #6071, opened 2024-05-26
CMF tweaks to accommodate recently added data

AndrewVSutherland:
This PR contains a bunch of minor tweaks to handle the recent addition of about 700,000 newforms to the LMFDB (including all weight 2 newforms for Gamm0(N) for N up to 50000).  For most of these we do not have algebraic eigenvalues, just the trace form, and for dim <= 20, the Hecke field.  This PR prevents server errors in cases where the old code was expecting certain columns to always be defined (e.g. inner twists, self twists, hecke_ring info for dim <= 20, etc...).

It also updates the browse ranges and stats to better reflect the new data set (which has more than twice as many newforms as it did before).  This data has not been copied to proddb yet, but none of these code changes depend on any of the data changes, they just make the code more flexible.

### DIFF
--- a/lmfdb/classical_modular_forms/main.py
+++ b/lmfdb/classical_modular_forms/main.py
@@ -840,7 +840,7 @@ newform_columns = SearchColumns([
                  align="center", short_title="projective image"),
     MultiProcessedCol("cm", "cmf.self_twist", "CM",
                       ["is_cm", "cm_discs"],
-                      lambda is_cm, cm_discs: ", ".join(map(quad_field_knowl, cm_discs)) if is_cm else "None",
+                      lambda is_cm, cm_discs: ", ".join(map(quad_field_knowl, cm_discs)) if is_cm else ("None" if is_cm == False else "not computed"),
                       short_title="CM",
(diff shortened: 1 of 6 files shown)


AndrewVSutherland:
There are also a few UI tweaks.  I noticed the AL signs were centered in some tables but not others, now they are always centered, and I changed the summary line to account for the fact that we don't have embedded newforms for every Galois orbit (which is good, because the total number would be well over 3 billion and take many terabytes).

AndrewVSutherland:
I should note that the new data was computed by Alex Cowan (for prime levels 10^4 to 10^6) and Krit Boonsiriseth (for composite levels 10000 to 50000).  I will updated the source and completeness knowls once all the new data has been uploaded and pushed to prod (there is still more to come, but I want to merge this PR before adding it).

AndrewVSutherland:
I'm going to merge this now and push to dev (otherwise we are going to keep getting server errors whenever someone accesses a newform of non-prime level > 10000, which happens more than half the time if you click the random button).

PR #6079, opened 2024-05-28
Add conjugacy class searches to abstract groups pages

jenpaulhus:
This PR is to add conjugacy class searches and to update the groups pages with the newer gps_conj_classes table.

Could we please get this onto one of the color servers so people can mess around with it to make sure everything works, or to suggest any added features?

This PR will require the new table gps_conj_classes.


jenpaulhus:
Locally I had a test failure for underlying data on the groups pages which I suspect will also fail now. I think that is something that needs to be fixed elsewhere?

roed314:
Demo is up at https://purple.lmfdb.xyz/Groups/Abstract/?search_type=ConjugacyClasses.  I'll take a look a the test failure in a bit.

jenpaulhus:
Looks like the test errors might be something else (not the data issue I was getting locally). I can look into them tomorrow. 

And good for folks to mess around on the general groups pages too to see if there are any issues:  [https://purple.lmfdb.xyz/Groups/Abstract](https://purple.lmfdb.xyz/Groups/Abstract/)  as I had to make some changes to those pages with the new table.

Going to most group pages and clicking on the links in the Statistics table will also get you to the conjugacy class searches.

roed314:
Looks like using `magma.IsInSmallGroupDatabase` isn't good, since magma isn't available while running CI.

edgarcosta:
Can we avoid the magma call? Also, how large are the numbers that we are planning to have here? The factoring step might get us in trouble.

Here is the Magma code:
```pascal
intrinsic IsInSmallGroupDatabase(o::RngIntElt) -> BoolElt
{Return true iff the small groups database contains the groups of order o}
 requirege o, 1;
 if (o eq 1024) then
    return false;
 end if;
 if (o le 2000  or  o eq 5^5  or o eq 7^4 or o eq 3^8) then
    return true;
 end if;
 f := Factorisation(o);
 if forall{i : i in [1..#f] | f[i,2] eq 1 and f[i,1] lt 2^30} then
    return true;  // square-free order
 end if;
 if (#f eq 2  and  exists(i){i : i in [1,2] | f[i][2] eq 1}) then
    p := f[3-i]; // the other prime power...
    if ( p[2] le 2
	 or p[1] eq 2  and  p[2] le 8
         or p[1] eq 3  and  p[2] le 6
         or p[1] eq 5  and  p[2] le 5
         or p[1] eq 7  and  p[2] le 4 ) then
       return true;  // q^n*p for a  q^n | 2^8, 3^6, 5^5 or 7^4  and  p ne q
    end if;
 end if;
 if (#f le 3 and &+[p[2]:p in f] eq 4) then return true; end if;
 if (#f eq 1 and f[1,2] le 7) then return true; end if;
 return false;
end intrinsic;
```

roed314:
Yeah, I think we should just reimplement this in Sage.  In practice, the numbers are all quite smooth, so we can do factoring with a very small factor base (the largest prime dividing any order is probably 1999).  If the group order isn't in the database, it doesn't really matter if we get it right or not.

jenpaulhus:
We already have the factorizations in the groups database, right?  Is it faster to just query gps_groups on order and read off that factorization?  

I possibly can also get rid of a couple of the calls to cc_data_to_gp_label (the group label is floating around as data other places eg. line 2678 of main.py, the group has already been computed so the label exists and can be called there instead).

roed314:
99% of the time, the order will be fast enough to factor.  In the case of a large order (larger than 40 digits maybe), we could do a search for groups of that order, and look at the first label we find.

jenpaulhus:
Replaced the magma command with sage code that does the same. For the moment I'm just calling factor for all values.  Figured we can discuss cutoffs and efficiencies at mtg this afternoon.

jenpaulhus:
See Zulip for one outstanding issue.

PR #6092, opened 2024-06-16
main -> dev

roed314:


roed314:
@edgarcosta, @AndrewVSutherland, can one of you push `gps_conj_classes` to prod so tests against prod succeed?

AndrewVSutherland:
The copy is in progress.  This will take a while, but I'll let you know when it's done.

AndrewVSutherland:
@roed314 The reload on proddb is failing because the id column of gps_conj_classes is an integer, not a bigint:
```
sage: db.gps_conj_classes.col_type
{'counter': 'integer',
 'group_counter': 'integer',
 'id': 'integer',
 'order': 'integer',
 'powers': 'integer[]',
 'group_order': 'numeric',
 'representative': 'numeric',
 'size': 'numeric',
 'aut_label': 'text',
 'centralizer': 'text',
 'label': 'text'}
```
How on earth did that happen?  In any case, I think you need to fix this.

roed314:
That was on purpose; since we've been having problems with disc size I was experimenting with using integer instead of bigint.  I just need to update `psycodict` on prod.

AndrewVSutherland:
Ah, that's a good idea for files like this with many (but less than 2^31) small rows.

AndrewVSutherland:
`gps_conjugacy_classes` is now on prod, but there is still one test failure (occuring twice):
```
lmfdb/tests/test_dynamic_knowls.py::DynamicKnowlTest::test_conjugacy_classes_knowl FAILED [ 50%]
```

PR #6120, opened 2024-07-18
fixes by ruff

fchapoton:
just a few minor changes, as done automatically by

`ruff --fix --ignore=E501,E402,E741 lmfdb`

Maybe this too could be scripted inside the "autopep8" patching system ?


edgarcosta:
In theory, it should be possible, but then one needs to combine it with https://github.com/ChartBoost/ruff-action to not get a double PR for each PR...

PR #6121, opened 2024-07-23
Update affiliation + 2 tiny fixes

jenpaulhus:
Updated my affiliation and webpage on these two pages:
https://www.lmfdb.org/management
https://www.lmfdb.org/acknowledgment

And took this opportunity to fix 2 tiny things on groups pages.
(1) Made "presentation" lower case in "Minimal presentations". Compare
http://localhost:37777/Groups/Abstract/12.4
http://lmfdb.org/Groups/Abstract/12.4

(2) Made the header for the "Group statistics" section disappear if we don't compute order statistics.  Compare beta and local for groups like: 53687091200.di or 167384925748592640.a

~~EDIT: Don't merge this yet.  Realizing we should make (2) say "not computed" not disappear. Will fix right now.~~

### DIFF
--- a/CONTRIBUTORS.yaml
+++ b/CONTRIBUTORS.yaml
@@ -280,8 +280,8 @@ email: aurel.page@normalesup.org
 affil: University of Bordeaux
 ---
 name: Jennifer Paulhus
-affil: Grinnell College
-url: https://www.math.grinnell.edu/~paulhusj/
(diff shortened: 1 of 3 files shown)


jenpaulhus:
Ok! This can be reviewed and merged now.  I made it so that if order_stats is not computed, then we have a message saying so.

PR #6129, opened 2024-08-10
Enable reCaptcha for www.lmfdb.org

edgarcosta:
First of all, this is already running on www.lmfdb.org, as I needed to:
- test if it worked
- stop the bots from overwhelming our servers

This is the solution I have managed to devise so far that doesn't spoil the user interaction and keeps the bots at bay.
If someone has any other suggestions, I'm all ears, but over the last month, we have been observing severe persistence from bots that often lead to www.lmfdb.org going offline.
In particular, we start to notice serious deprecation of service whenever the servers are handling more than 4 requests a second (the blue/green line is the number of accepted/blocked requests):
Screen Shot 2024-08-10 at 15 29 46
As you can see, until yesterday (when reCaptcha was enabled), I was not succeeding at keeping the bots under control despite several hours spent tuning firewall rules to try to block bots but not humans.

PS: I still would like to restrict the loading of the captcha javascript to only when we are serving www.lmfdb.org; I will figure that out later.

### DIFF
--- a/lmfdb/app.py
+++ b/lmfdb/app.py
@@ -308,31 +308,6 @@ def netloc_redirect():
         return redirect(urlunparse(replaced), code=302)
 
 
-@cached_function
-def bad_bots_list():
-    return [
-        elt.lower()
-        for elt in [
-            "The Knowledge AI",
-            "Wolfram",
-            "petalbot",
-            "Bytespider",
-            "Sogou",
-            "MJ12bot",
-            "Amazonbot",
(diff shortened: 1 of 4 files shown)


AndrewVSutherland:
Looks good to me, feel free to merge and push to dev+web whenever you are ready (I think it would be good to do this only when you are in a place where you can easily monitor/intervene if something goes wrong, so maybe wait until tomorrow).

PR #6130, opened 2024-08-12
Safe index order

roed314:
See https://beta.lmfdb.org/Groups/Abstract/sub/1336653538935178044929642368586691379200000000.a.3450136419953716089934630092800000000.A, where bigint knowls are not properly displayed.


jenpaulhus:
LGTM!

PR #6140, opened 2024-08-12
removing jquery.dataTables

edgarcosta:
and the stylesheet code associated with it

@davidfarmer, do you remember what this was used for?


jwj61:
One part (maybe Dirichlet characters) used it for search results.  We tried it for number fields, but it was too slow (it loads all the search results before compressing the display).  I think it was abandoned for not being suitable for consistency.

AndrewVSutherland:
We removed data table that were being used in a couple of places (including Dirichlet characters) many years ago

PR #6147, opened 2024-08-19
Corrected autopep8 fixes

roed314:
Fixes #6146.


edgarcosta:
Thanks, if one doesnt correct these, then the autopep8 will keep suggesting the bad ones!

PR #6158, opened 2024-08-21
gl2_subgroup knowl tweaks

AndrewVSutherland:
Adds abstract group sub knowl


AndrewVSutherland:
To test go to an elliptic curve page or search results page that shows Galois images (e.g. http://127.0.0.1:37777/EllipticCurve/Q/?nonmax_primes=7) and click on one.  If the group is small enough to have a GAP id that can be looked up, it should have an "Abstract group" line in the gl2_subgroup knowl.

PR #6160, opened 2024-08-21
Deal with empty search

jwj61:
If there are no search arguments in mod-ell Galois reps and the user hits search, return everything.


jwj61:
Added better formatting on search result page for 1 dimensional reps

PR #6181, opened 2024-09-02
Add relative class number

jwj61:
Relative class numbers for CM fields has been computed in cases where we had both class numbers already.  They are shown on the page of a CM number field (it is in the class number/class group section).

http://127.0.0.1:37777/NumberField/6.0.399424.1
http://beta.lmfdb.org/NumberField/6.0.399424.1

Here is a case where it has been computed, but the computation assumed GRH

http://127.0.0.1:37777/NumberField/42.0.2281836760183646137444154412268560109828024514076489472840222217265158917203.1

and a case where it has not been computed

http://127.0.0.1:37777/NumberField/42.0.2654972529667057457433107680535256579009544347022866599647016024950419363310098668524052547.1

If the field is not CM, we don't do anything new since it is not relevant.

It is also now a search criterion

A possible future improvement is if relative_class_number is part of the search, automatically turn is_cm to true (and give a flash error if the user has set it to false).

The database nf_fields should be copied to the cloud before this goes to production.


jwj61 (commented):
Doesn't this cause a problem if it is a message that the value has not been computed?

roed314:
You're right; I'll fix it.

roed314:
This looks good to me, though we perhaps want to wait to merge it until we push `nf_fields` to prod.

PR #6188, opened 2024-09-12
missing self arguments

fchapoton:
add some missing self arguments in keywords, as suggested by `mypy`

### DIFF
--- a/lmfdb/classical_modular_forms/test_cmf.py
+++ b/lmfdb/classical_modular_forms/test_cmf.py
@@ -7,7 +7,7 @@ cmf_logger.setLevel(100)
 
 
 class CmfTest(LmfdbTest):
-    def runTest():
+    def runTest(self):
         pass
 
     def test_expression_divides(self):
--- a/lmfdb/classical_modular_forms/test_cmf2.py
+++ b/lmfdb/classical_modular_forms/test_cmf2.py
@@ -6,7 +6,7 @@ cmf_logger.setLevel(100)
 
 
 class CmfTest(LmfdbTest):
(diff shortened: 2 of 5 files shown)


roed314:
I agree that there's something to be fixed here.  I [think](https://docs.python.org/3/library/unittest.html#unittest.TestCase) having `def runTest(self): pass` disables scanning for tests.  I'm not sure why we want to do that; should we instead delete all these functions and use the default `runTest` from `unittest.TestCase`?

fchapoton:
I have no idea. There only a few definitions of `runTest` :
```

lmfdb/classical_modular_forms/cmf_test_pages.py:    def runTest(self):
lmfdb/classical_modular_forms/test_cmf.py:    def runTest(self):
lmfdb/classical_modular_forms/test_cmf2.py:    def runTest(self):
lmfdb/ecnf/ecnf_test_pages.py:    def runTest(self):
lmfdb/genus2_curves/g2c_test_pages.py:    def runTest(self):
lmfdb/groups/abstract/groups_test_pages.py:    def runTest(self):
lmfdb/siegel_modular_forms/smf_test_pages.py:    def runTest(self):

```

fchapoton:
should I change the PR to remove all these methods ?

roed314:
That sounds good.  We can see what happens with the tests.

edgarcosta:
I should have said that these tests are not run by default, as these are the tests that test **all** webpages of some kind

fchapoton:
so should I step back to just adding the missing ``self`` ?

fchapoton:
Voilà

PR #6194, opened 2024-09-28
fix pep E225 and E228

fchapoton:
this was purely scripted using autopep8
(first fixing E228 that  creates some E225 failures, then fixing E225)

should this be added to the linter / autopep8 patcher ?


roed314:
Looks good to me, and I'd be fine adding 225 and 228 to the autopep8 patcher.

Issue #3310, opened 2019-09-05
Browsing Features in Isogeny Classes of AbVarFq


tdupu:
People requested more browsing features (as in Elliptic Curves over QQ):
    *by p-rank
    *by simple 
    *some with interesting p-rank and delta rank

tdupu:
more search buttons:
*ordinary 
*supersingular
*p-corank

roed314:
Fixed by #6196.  I didn't add additional dropdowns for ordinary or supersingular (since the UI is already very full), but ordinary is available either via `p_corank=0` or `newton_elevation=0`, and supersingular is available either by `angle_rank=0` or `newton_elevation=1,2,4,6,9,12` (if `g` is fixed).

PR #6196, opened 2024-10-02
Update abvar page

roed314:
to show info when `hyp_count` is known but not `jacobian_count`.  To test, compare

* http://localhost:37777/Variety/Abelian/Fq/4/3/ae_d_i_aw with https://beta.lmfdb.og/Variety/Abelian/Fq/4/3/ae_d_i_aw (no hyperelliptic Jacobian, unknown if other Jacobian)
* http://localhost:37777/Variety/Abelian/Fq/4/4/a_a_ab_au with https://beta.lmfdb.org/Variety/Abelian/Fq/4/4/a_a_ab_au (1 hyperelliptic Jacobian)
* http://localhost:37777/Variety/Abelian/Fq/3/23/a_v_a with https://beta.lmfdb.org/Variety/Abelian/Fq/3/23/a_v_a (contains 1902 hyperelliptic Jacobians).


AndrewVSutherland:
I'm wondering if it is potentially confusing to have the "hence it is principally polarizable" after "it contains X hyperelliptic Jacobians", since the "hence" is independent of "hyperelliptic".  It might be clearer to make the the statement about containing (any sort of) Jacobian, hence PPAV first, then say how many are hyperelliptic.

roed314:
I tried rewording it; let me know if you think the new version is an improvement (it's a bit wordier).

AndrewVSutherland:
Hmm, the statements still seem to have the form "This isogeny class contains the Jacobian of X hyperelliptic curves, hence is principally polarizable, ...", which doesn't address the issue I was concerned about.  How about "This isogeny class is principally polarizable and contains the Jacobian of X hyperelliptic curves, ...".

I guess we are happy to call an isogeny class principally polarizable if it contains a principally polarizable AV (this is already implicit in the current wording).

roed314:
@AndrewVSutherland, I've made the change you suggested.  I also added more features, so this now fixes #3310 (by adding new entries to the top of the browse page), adds a bunch of hidden search columns and several new sort orders.

roed314:
To test, look at the browse page and play around with search columns and sort orders in abelian variety search results.

Before this is merged, we should push `av_fq_isog` to production.

AndrewVSutherland:
Looks good, and I see that @edgarcosta copied av_fq_isog to proddb.

Issue #6170, opened 2024-08-23
Notation for the order of the quotient G/N 


edgarcosta:
at the moment, we use |G/N|, as the table header in "Low degree resolvents",  see: https://beta.lmfdb.org/GaloisGroup/24T2668

We should switch to #G/N, or [G:N].

It also needs to be clarified what is N from that table.

I believe the first item is a good first issue.

jwj61:
It should use the latex macro \card (which we already have).

Issue #6177, opened 2024-08-30
Page for Galois group fails 34T82


edgarcosta:
https://beta.lmfdb.org/GaloisGroup/34T82

and the www version fail:
```
Exception on /GaloisGroup/34T82 [GET]
Traceback (most recent call last):
  File "/home/sage/sage-9.7/local/var/lib/sage/venv-python3.10.5/lib/python3.10/site-packages/flask/app.py", line 2525, in wsgi_app
    response = self.full_dispatch_request()
  File "/home/sage/sage-9.7/local/var/lib/sage/venv-python3.10.5/lib/python3.10/site-packages/flask/app.py", line 1822, in full_dispatch_request
    rv = self.handle_user_exception(e)
  File "/home/sage/sage-9.7/local/var/lib/sage/venv-python3.10.5/lib/python3.10/site-packages/flask/app.py", line 1820, in full_dispatch_request
    rv = self.dispatch_request()
  File "/home/sage/sage-9.7/local/var/lib/sage/venv-python3.10.5/lib/python3.10/site-packages/flask/app.py", line 1796, in dispatch_request
    return self.ensure_sync(self.view_functions[rule.endpoint])(**view_args)
  File "/home/lmfdb/lmfdb-git-dev/lmfdb/galois_groups/main.py", line 82, in by_label
    return render_group_webpage({'label': label})
  File "/home/lmfdb/lmfdb-git-dev/lmfdb/galois_groups/main.py", line 335, in render_group_webpage
    data['malle_a'] = wgg.malle_a
  File "sage/misc/lazy_attribute.pyx", line 127, in sage.misc.lazy_attribute._lazy_attribute.__get__
    result = self.f(a)
  File "/home/lmfdb/lmfdb-git-dev/lmfdb/galois_groups/transitive_group.py", line 224, in malle_a
    ccs = self.conjclasses
  File "sage/misc/lazy_attribute.pyx", line 127, in sage.misc.lazy_attribute._lazy_attribute.__get__
    result = self.f(a)
  File "/home/lmfdb/lmfdb-git-dev/lmfdb/galois_groups/transitive_group.py", line 204, in conjclasses
    cc = [isom.Image(z) for z in cc1]
  File "/home/lmfdb/lmfdb-git-dev/lmfdb/galois_groups/transitive_group.py", line 204, in <listcomp>
    cc = [isom.Image(z) for z in cc1]
[...]

PR #6199, opened 2024-10-09
Lessmalle

jwj61:
Addresses issue #6177 but cutting off the computation of conjugacy classes and the character table when there are more than 160 classes.  These computations are done on-the-fly and gap can have trouble doing it.  These pages mentioned in the issue now load:

http://127.0.0.1:37777/GaloisGroup/34T82
http://127.0.0.1:37777/GaloisGroup/42T8771

This also fixes issue #6170 on the notation for the order of G/N as a header for Low Degree Resolvents

http://127.0.0.1:37777/GaloisGroup/4T5


roed314:
LGTM

PR #6214, opened 2024-10-19
just a few typos

fchapoton:
also fixing some rst code.

NOTE : There is invalid rst code in `siegel_modular_forms/dimensions.py`, which is just html inside rst blocks...


roed314:
Thanks for pointing out the problem with `dimensions.py`,  We don't currently do any kind of documentation building, so docstrings are just read by developers editing the file.  That being said, it would be better if these were changed to rst instead of html; feel free to make that change if you're interested and I'll review it.

PR #6218, opened 2024-10-27
Resolves #6215

### LINKED ISSUE #6215: CMF newspace page times out on load
I haven't looked into why, but the page https://www.lmfdb.org/ModularForm/GL2/Q/holomorphic/3312/1/c/ seems to time out.

AndrewVSutherland:
The timeout noted in #6215 is a result of trying to compute dimensions of subspaces on the fly when computing the old space decomposition (this is done to avoid listing summands of dimension zero); this is very quick in general, but for weight 1 spaces it can be very slow.

This PR changes the code to instead perform a database lookup when the weight is 1.  It also modifies the code to only list newspaces of lower level that have positive dimension in the oldspace decomposition (this was incorrectly implemented in the update that changed the code to compute dimensions on the fly, which was using the cusp dim rather than the new dim in the call to Pari/GP).

Also fixes a knowl name typo I noticed while fixing this.


roed314:
I moved the database call outside the for loop, and checked that the problematic page from #6215 still loads okay on my branch.

PR #6226, opened 2024-11-04
Add stark units

roed314:
See the bottom of the "Newform invariants" section of http://localhost:37777/ModularForm/GL2/Q/holomorphic/2763/1/c/b/ for example.  Before this is merged, the new table `mf_stark` should get pushed to prod.


AndrewVSutherland:
I've copied `mf_stark` to prod and the PR looks good, but the source and acknowledgments knowl should be updated (and maybe also the completeness knowl)

roed314:
I've updated both rcs knowls; @AndrewVSutherland, do you want to review them?

AndrewVSutherland:
Done!

Issue #5840, opened 2024-01-29
Knowl title should be required


AndrewVSutherland:
When creating or editing a knowl, the user should not be allowed to save it without putting a title in the description field.

AndrewVSutherland:
Addressed by #6232

Issue #4321, opened 2021-01-01
Knowl link suggestions should prioritize knowls in the same category


AndrewVSutherland:
When offering link suggestions during knowl editing it would make sense to list knowls that are in the same category first (and to sort the suggestions by category).  This would make it easier to find the correct knowl for terms like "degree", "dimension", "weight", "parity", etc... that may have multiple knowls that define their meaning in different contexts.

AndrewVSutherland:
Addressed by #6232

Issue #4041, opened 2020-09-18
Knowl link suggestions should check singular form of plurals


AndrewVSutherland:
The link suggestion feature of the knowl editor could easily be made more useful if it checked not only for exact terms that appear in the text, but also the the terms that appear when you remove the trailing s from words (I think it would be safe to just do this blindly, I'm not asking the code to identify plural nouns).

For example, if you write a knowl containing the phrase "elliptic curves" it would be great if it would give the same link suggestion that it currently does for "elliptic curve".

JohnCremona:
+1  I sometimes mistype on purpose to get the link put in automatically, then correct it later.  I hope.

alexjbest:
There are some pluralizer libraries on pip already like pluralize and inflect, so adding a lightweight dependency should get us things like isogeny -> isognies for free.

AndrewVSutherland:
@alexjbest even better (but not essential).

AndrewVSutherland:
Addressed by #6232

PR #6232, opened 2024-11-09
Some knowl improvements

roed314:
Fixes a bug (visible when editing [ec.q.abc_quality](https://beta.lmfdb.org/knowledge/show/ec.q.abc_quality) for example) that gave far too much context for a change.  Also addresses #5840, #4321 and #4041.


AndrewVSutherland:
This is truly wonderful, thank you!

PR #6233, opened 2024-11-09
Make is_supersingular a property so that it is compatible with new data column

roed314:


roed314:
I'm merging since all abelian variety pages are currently broken.

Issue #5731, opened 2023-11-16
We should avoid using color to mark a mathematical feature on the webpage


edgarcosta:
e.g., "The blue line marks the conjugacy class containing complex conjugation." in https://www.lmfdb.org/ArtinRepresentation/4.87616.8t40.b.a

![image](https://github.com/LMFDB/lmfdb/assets/322822/c5a4de87-e9d6-40b4-bec7-70841afeb688)

https://www.toptal.com/designers/colorfilter

PR #6243, opened 2024-11-13
Remove use of color to designate complex conjugation

jwj61:
Fix issue #5731 -- instead of marking the conjugacy class of complex conjugation on an Artin representation page, added a column to the table and use a check mark to signal which class it is.

Can check on any Artin representation page, such as

http://beta.lmfdb.org/ArtinRepresentation/6.106906880000.18t51.a.a
http://127.0.0.1:37777/ArtinRepresentation/6.106906880000.18t51.a.a

### DIFF
--- a/lmfdb/artin_representations/templates/artin-representation-show.html
+++ b/lmfdb/artin_representations/templates/artin-representation-show.html
@@ -68,7 +68,7 @@
         </table>
     </div>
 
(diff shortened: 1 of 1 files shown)


roed314:
I added some knowls.  Feel free to merge if you approve of the changes.

PR #6246, opened 2024-11-14
Change color of Search again button when inputs change

### LINKED ISSUE #5003: Indicate when search results are out of date
A suggestion of @AndrewVSutherland in the discussion on #4991: 

"I think we should change the appearance of the search again button when the results don't reflect the current settings (if you change the sort order but don't click search again it's easy to get confused, especially if you switch to another tab and then back). I would suggest making the default color of the search button the same color as the selectors (lighter blue) when it is in an inactive state (i.e. pressing it will just reload the search results currently shown) and then making the search button the color it is now (darker blue) whenever it is in an active state (pressing it will apply new search/sort criteria)."

roed314:
Fixes #5003.

### DIFF
--- a/lmfdb/templates/refine_search_form.html
+++ b/lmfdb/templates/refine_search_form.html
@@ -30,6 +30,16 @@ function set_buckets(col_selecter, bucket_id) {
   <input type="hidden" name="sort_dir" value="{{info.sort_dir}}" />
 </form>
 
+<script>
 {% if info.search_array.has_advanced_inputs(info) %}
-<script> show_advancedQ(); </script>
+show_advancedQ();
 {% endif %}
+
+function set_fresh() {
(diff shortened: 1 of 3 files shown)


JohnCremona:
It certainly works, but the change is quite subtle.  Does it have to be another shade of blue?  Would something more eye-catching like flashing text be too much?!

roed314:
I don't know what you mean by flashing text; something like [this](https://stackoverflow.com/questions/16344354/how-to-make-blinking-flashing-text-with-css-3) seems unwise to me.

I tried using a grey background (using `color.button_background_inactive`) rather than the light blue, but I didn't think that was appropriate since the button is still functional.  Besides, I prefer having the difference be subtle, since I don't want it to be jarring for users.

JohnCremona:
Fair enough -- I wasn't being serious about blinking. Thanks for implementing this!

PR #6251, opened 2024-11-15
Add top-level rcs page

### LINKED ISSUE #5675: Learn more -> source, reliability, and completeness
At the moment, we point to https://www.lmfdb.org/knowledge/show/rcs

should we instead point to a page that includes the knowl?

That is what we do for:
`www.lmfdb.org/Database/{Source,Completeness,Reliability}`

roed314:
Fixes #5675.


JohnCremona:
I think this looks good, and should be merged now even if there is some tweaking possible.

PR #6254, opened 2024-11-16
Additional search options for number fields

### LINKED ISSUE #4004: Search for number fields on Galois / Abelian
I was surprised that I couldn't see a way to search only for Galois or only abelian number fields.
This seems like a useful feature that would probably be easiest to implement by adding a couple of columns to `nf_fields` or `nf_fields_extras` as currently it looks like these properties are computed on the fly given the Galois group.
Implementing this as a fancy query is probably also not too bad.

roed314:
Fixes #4004.


roed314:
I put this up on purple.lmfdb.xyz/NumberField (there are also some other changes to the sidebar on purple, but those would not be merged with this PR).

jwj61:
I like the idea.  If I select dihedral-non-Galois but not a degree I get a server error.

roed314:
Thanks. I've fixed the bug you found; purple should update in 3 minutes.

jwj61:
Thanks.  If you enter Galois-dihedral and 8,9,16-20 for degree, the output is "Error".  Also giving an error is a degree above 43 with either type of dihedral.

roed314:
Fixed.  I also added `multiquadratic` as an option.

PR #6266, opened 2024-11-18
add inducing character search option

### LINKED ISSUE #968: show a graph of characters above and below
From /roadmap: show a graph of characters above and below (parents and sons).

AndrewVSutherland:
Adds an option to search Dirichlet character orbits by specifying the label of the primitive character orbit that induces it.

We can close #968 once this is merged.

### DIFF
--- a/lmfdb/characters/main.py
+++ b/lmfdb/characters/main.py
@@ -4,6 +4,7 @@ from lmfdb.app import app
 import re
 from flask import render_template, url_for, request, redirect, abort
 from sage.all import euler_phi, PolynomialRing, QQ, gcd, ZZ
+from sage.databases.cremona import class_to_int
 from lmfdb.utils import (
     to_dict, flash_error, SearchArray, YesNoBox, display_knowl, ParityBox,
     TextBox, CountBox, parse_bool, parse_ints, search_wrap, raw_typeset_poly,
@@ -106,6 +107,12 @@ class DirichSearchArray(SearchArray):
             example="2",
             example_span="2 or 3-5"
         )
+        inducing = TextBox(
+            "inducing",
+            label="Induced by",
+            knowl="character.dirichlet.primitive",
+            example="3.b"
+        )
         parity = ParityBox(
             "parity",
             knowl="character.dirichlet.parity",
@@ -133,12 +140,13 @@ class DirichSearchArray(SearchArray):
         count = CountBox()
 
         self.refine_array = [
-            [modulus, conductor, order, is_real], [parity, is_primitive, is_minimal, count],
+            [modulus, conductor, order, inducing], [parity, is_primitive, is_minimal, is_real], [count],
         ]
         self.browse_array = [
             [modulus],
             [conductor],
             [order],
+            [inducing],
             [parity],
             [is_primitive],
             [is_real],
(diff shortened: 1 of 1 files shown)


lmfdb/characters/main.py
```diff
@@ -156,6 +164,24 @@ def common_parse(info, query):
...
+    if 'inducing' in info:
+        try:
+            validate_label(info['inducing'])
+            parts_of_label = info['inducing'].split(".")
+            if len(parts_of_label) != 2 or not str.isalpha(parts_of_label[1]):
+                raise ValueError("Invalid character orbit label format, expected N.a")
+            primitive_modulus = int(parts_of_label[0])
+            primitive_orbit = class_to_int(parts_of_label[1])+1
+            if db.char_dirichlet.count({'modulus':primitive_modulus,'is_primitive':True,'orbit':primitive_orbit}) == 0:
+                print("hi")
```
roed314, line 176: Should remove this debugging statement.

lmfdb/characters/main.py
```diff
@@ -156,6 +164,24 @@ def common_parse(info, query):
...
     parse_ints(info, query, "conductor", name="conductor")
     parse_ints(info, query, "order", name="order")
+    if 'inducing' in info:
+        try:
+            validate_label(info['inducing'])
+            parts_of_label = info['inducing'].split(".")
+            if len(parts_of_label) != 2 or not str.isalpha(parts_of_label[1]):
+                raise ValueError("Invalid character orbit label format, expected N.a")
+            primitive_modulus = int(parts_of_label[0])
+            primitive_orbit = class_to_int(parts_of_label[1])+1
```
roed314, line 174: Is it worth supporting a Conrey label for an actual character, instead of just an orbit?

lmfdb/characters/main.py
```diff
@@ -156,6 +164,24 @@ def common_parse(info, query):
...
+            if db.char_dirichlet.count({'modulus':primitive_modulus,'is_primitive':True,'orbit':primitive_orbit}) == 0:
+                print("hi")
+                raise ValueError("Primitive character orbit not found")
+            if 'conductor' in query:
+                flash_error("Conductor constraint %s ignored due to specification of primitive inducing character", info['conductor'])
```
roed314, line 179: I've tended to treat incompatible constraints like this as giving no results rather than just ignoring them.  It does make parsing things harder though.

roed314: See [this line](https://github.com/LMFDB/lmfdb/blob/5bd0fe0d6d5882a4e4fa81dd133d2a2416ca1ef1/lmfdb/number_fields/number_field.py#L925) for example, which I added a few days ago.

roed314: It would be nice to have a general policy about when approach like this (a flash error that we're ignoring input because another input is more specific) is okay.

AndrewVSutherland: > I've tended to treat incompatible constraints like this as giving no results rather than just ignoring them. It does make parsing things harder though.

I would normally do that too, but in this case you can't do that without inventing a constraint they did not enter (there is no column for primitive orbit label, and you can't specify two values for the conductor).

But thinking about it again I realize we could set `primitive_orbit` to zero, which would yield no results and not be visible to the user, I'll do that instead.

lmfdb/characters/main.py
```diff
@@ -156,6 +162,24 @@ def common_parse(info, query):
...
+        try:
+            validate_label(info['inducing'])
+            parts_of_label = info['inducing'].split(".")
+            if len(parts_of_label) != 2 or not str.isalpha(parts_of_label[1]):
+                raise ValueError("Invalid character orbit label format, expected N.a")
+            primitive_modulus = int(parts_of_label[0])
+            primitive_orbit = class_to_int(parts_of_label[1])+1
+            if db.char_dirichlet.count({'modulus':primitive_modulus,'is_primitive':True,'orbit':primitive_orbit}) == 0:
+                raise ValueError("Primitive character orbit not found")
+            if 'conductor' in query:
```
roed314, line 175: Shouldn't this be `'conductor' in query and query['conductor'] != primitive_modulus`?

AndrewVSutherland: yes, fixed

lmfdb/characters/main.py
AndrewVSutherland: Done, seems fast enough.

lmfdb/characters/main.py
AndrewVSutherland: I made a further change so that it never ignores constraints (if the induced label is invalid you get an error, if it is valid but incompatible with the conductor you get no results)

roed314:
There are still some issues; I'm mostly through fixing them and will try to update this PR this evening.

roed314:
See [this PR](https://github.com/AndrewVSutherland/lmfdb/pull/18) to your PR.

PR #6267, opened 2024-11-18
Add SW and BHKSSW elliptic curve datasets

roed314:
Also adds an index page for datasets, and links it from the new access options page.  See
* https://purple.lmfdb.xyz/api/options for the new links to the dataset index page
* https://purple.lmfdb.xyz/datasets/ for the dataset index page itself
* https://purple.lmfdb.xyz/EllipticCurve/Q/BHKSSW for the BHKSSW page
* https://purple.lmfdb.xyz/EllipticCurve/Q/stein-watkins for the Stein-Watkins dataset.


JohnCremona:
Looks good to me.  Two small points: you only list last names for BHKSSW but full names for SW, I suggest just last names for the latter.  Also, in the api page I think the wording of "you can install the LMFDB locally" is misleading!  People may think that this gives them all the data locally.  Perhaps insert "web frontend" after "LMFDB"? ALso possibly add a phrase making it 100% clear that the frontend you install will connect to the same SQL database as the main one.

roed314:
I've made the two changes John suggested (the second one through [this knowl](https://beta.lmfdb.org/knowledge/show/intro.direct_sql)).  I also added a statement adding that we ask for a CC-BY-SA license for contributions.  If the managing editors don't agree with that requirement, let me know!

It would be nice to have links in the "Source" column for zeta zeros and congruent numbers.  I wasn't able to find any links that were as good as the other three rows; I'd welcome suggestions from someone who knows more about the history of these datasets.  For Zeta zeros, [wikipedia](https://en.wikipedia.org/wiki/ZetaGrid) has a stub on ZetaGrid, which links to [a paper](http://numbers.computation.free.fr/Constants/Miscellaneous/zetazeros1e13-1e24.pdf) that might be reasonable but it looks like the author of that paper probably wasn't the person who computed our data.  For congruent numbers, I wasn't able to find any appropriate papers by Randall L. Rathbun.

jwj61:
Just a tiny quibble.  On the conditions at the end, all 3 are set up at continuations of the phrase "In order to be considered for inclusion, your dataset should", so they should not start with a capital letter, the second should not end in a period, I think commas are more appropriate than semicolons, and the third one is not parallel (currently: In order to be considered for inclusion, your dataset should Be willing to release your data under the ...).

roed314:
@jwj61 Fixed.

Issue #4943, opened 2022-01-21
Real numbers displayed with ellipses


bmatschke:
The regulator of https://beta.lmfdb.org/EllipticCurve/Q/37a1/ is given as "0.051111408239968840235886099757...". The ellipses "..." are misleading (to me), as this expansion is obtained by rounding, not by truncation. (A longer expansion is "0.05111140823996884023588609975694...".)
@JohnCremona

roed314:
I propose that we rephrase all the rounded quantities on the page as `X \approx x`, where `X` is an appropriately chosen symbol (often there's a standard symbol, but sometimes we may need to choose one), and `x` is a real approximation with no ellipses.

Here's a proposal for the symbols to use, but @JohnCremona please let me know if there are better or more standard choices!

* [Faltings height](https://beta.lmfdb.org/knowledge/show/ec.q.faltings_height): `h_{\mathrm{Faltings}}`
* [Faltings height](https://beta.lmfdb.org/knowledge/show/ec.q.faltings_height): `h_{\mathrm{stable}}`
* [abc quality](https://beta.lmfdb.org/knowledge/show/ec.q.abc_quality): `Q`
* [modified Szpiro ratio](https://beta.lmfdb.org/knowledge/show/ec.q.szpiro_ratio): `\beta`

We can then use all the standard notation for the BSD invariants, including `r` for the analytic rank (so that it's not the only quantity not of the form `X = x` in that section.

Finally, we should also fix the `(rounded)` knowl which is broken.  Since it just points to the same knowl as the row header I propose just removing it (that's the current state for anyone not logged in).

JohnCremona:
This is related to #5308 but this one is easier.  I approve of @roed314 in principle but I do not know a standard notation for either the abc quality or the Szpiro ratio.  If there is none yet, we can be the ones to set the standard, but it would be good to have some input, or look ni the literature, before committing ourselves.

roed314:
[...]

PR #6268, opened 2024-11-18
fixes for issue #4943

JohnCremona:
This addresses #4943  for ECQ.  For the two lists of invariants to look consistent I added a symbol/name for all the quantities there, not just the real ones which may be approximate.  All now do have names except the Sato-Tate group, and the same names appear in the appropriate knowls so a couple of those have had minor edits (Szpiro ratio, abc-quality).  We could show something like ST(E) for the ST group of that is suffiently standard -- @AndrewVSutherland ?

I fixed but left the "(exact)"/"(rounded)" for analytic Sha since the knowl (originally written just for this purpose) does say something concrete, which is not obvious, namely that for both rank 0 and rank 1 the quantity shown is exact, having been computed as a rational, while for ranks 2 and above it is a rounded floating point number (not known theoretically to even be rational, let alone integral).  I think this is markedly different from all the other approximate reals on the page.  But of course I am willing to be persuaded!

Once this is done (with further changes if desired) I will do the same thing for ECNF.

### DIFF
--- a/lmfdb/elliptic_curves/templates/ec-curve.html
+++ b/lmfdb/elliptic_curves/templates/ec-curve.html
@@ -196,7 +196,7 @@ white-space: pre in order to keep line breaks! #}
 
         <tr>
         <td>{{ KNOWL('st_group.definition', title='Sato-Tate group') }}:</td>
-        <td>{{ data.data.ST|safe }}</td>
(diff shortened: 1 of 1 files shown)


AndrewVSutherland:
I think ST(E) is fine.

roed314:
Should there also be symbols for conductor, discriminant and $$j$$-invariant?  I agree that "(exact)" / "(rounded)" are good as is.

roed314:
Looks great!

Issue #5040, opened 2022-02-17
Display is not as advertised


jwj61:
On https://beta.lmfdb.org/ModularForm/GL2/Q/holomorphic/1/118/a/a/ you get a link in the q-expansion to display 100 coefficients.  If you click on it, you don't get 100 coefficients.  Instead, you get the 99th coefficient instead of the 9th because it is compressed.

jwj61:
I don't think it should say it will display something if it isn't going to display it.  Maybe it should be a link to copy the expansion through q^99.

roed314:
I think the current behavior is fine.  I think saying "Display 99 coefficients" would look weird, as would having the precision be 101.  Besides, technically we're displaying 100 coefficients, it's just that the coefficient of q^0 is always 0 (note that we are including other 0 coefficients in the count even though the corresponding terms are not shown; for example the coefficient of q^4 is 0 [here](https://beta.lmfdb.org/ModularForm/GL2/Q/holomorphic/23/1/b/a/)).

jwj61:
This is not a question of off by one in the count, or of coefficients which are 0 being suppressed.  The user clicks to display 100 coefficients but it does not display all non-zero terms up to q^99, it only displays a couple of them.

roed314:
Oh, I see.  Yeah, I find that a bit jarring as well.

AndrewVSutherland:
Hmm, this is a bit tricky because we don't really know whether the q-expansion is going to be compressed or not.  Like David I don't think the "100 coefficients" part should change, but I agree that perhaps the word "Display" should change.  Maybe "Get" instead of "Display"?  If the user moves their mouse over the q-expansion at all (which I think someone unhappy with the dots would be likely to do, e.g. to try clicking on them), they will see the copy to clipboard and eyeball icons pop up, both of which will give them access to 100 coefficients of the q-expansion.

jwj61:
That would be a fix.  I may be the only one who thinks this, but I think is ok to have "Display 100 coefficients" and if the user clicks on that, they get what they asked for (without compression).

assaferan:
@jwj61, @AndrewVSutherland, @roed314 - have we come to a conclusion about the wanted behavior in this case? 
I'm fine with our current (compressed) behavior.

roed314:
I agree with @jwj61 that our current behavior is weird.  Here are some alternatives (some of which have been suggested already):
[...]

PR #6269, opened 2024-11-18
modifying display of q-expansion to always have 100 coefficients

assaferan:
Resolving issue #5040

### DIFF
--- a/lmfdb/classical_modular_forms/templates/cmf_full_gamma1_space.html
+++ b/lmfdb/classical_modular_forms/templates/cmf_full_gamma1_space.html
@@ -118,16 +118,9 @@ function show_qexp(qstyle) {
   <form>
     <table class="qexp-table">
       <tr>
-        <td class="qexp-output smalloutput">{{ space.trace_expansion(prec_max=10) | safe }}</td>
-        <td class="qexp-output mediumoutput nodisplay">{{ space.trace_expansion(prec_max=100) | safe }}</td>
+        <td class="qexp-output mediumoutput">{{ space.trace_expansion(prec_max=100) | safe }}</td>
       </tr>
     </table>
-    <p>
-      <div class="toggle">Display
-        <span class='smalloutput'><a onclick="show_qexp('mediumoutput'); return false;" href='#'>100 coefficients</a></span>
-        <span class='mediumoutput nodisplay'><a onclick="return show_qexp('smalloutput');" href='#qexp-anchor'>10 coefficients</a></span>
-      </div>
-    </p>
   </form>
 </div>
 {% endif %}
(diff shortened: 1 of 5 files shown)


roed314:
I think there are more places that need to be changed:
* The "Trace form" sections of `cmf_full_gamma1_space.html` and `cmf_space.html`,
* The `show_exp` javascript functions should be deleted in all three files.

assaferan:
Oops, thanks for catching this! Should be fixed now.

roed314:
I like this change, but I want to hear other opinions about whether the default $$q$$-expansion shown for something like [3.14.a.b](http://localhost:37780/ModularForm/GL2/Q/holomorphic/3/14/a/b/) is too much.  If we think there's too much shown by default there, then we might need to tweak the parameters to [raw_typeset_qexp](https://github.com/LMFDB/lmfdb/blob/main/lmfdb/utils/web_display.py#L589) which has `compress_threshold=100` as a default, and then that variable does not seem to be used....

roed314:
@AndrewVSutherland, what do you think?

AndrewVSutherland:
I'm slightly puzzled as to why it is not compressing the 3.14.a.b q-expansion at all, e.g. as on http://127.0.0.1:37777/ModularForm/GL2/Q/holomorphic/882/4/g/z/

My assumption was that large q-expansions were going to be compressed (indeed, the original objection was that we say "show 100 coefficients" but often don't show 100 coefficients due to compression.

assaferan:
It seems that until now the q-expansion would only be compressed if there was some compression in the coefficients. I modified it to instead use the compression threshold that was never used, as David suggested, and it seems better now, I think.

roed314:
Looks good to me.  There are still a few modular forms that aren't compressed (e.g. [3600.1.e.a (localhost)](http://127.0.0.1:37777/ModularForm/GL2/Q/holomorphic/3600/1/e/a/)) but most of them are now compressed, which was the intention of #5040.

I'm happy to merge this, but will give @AndrewVSutherland a chance to weigh in.

AndrewVSutherland:
Certainly 3600.1.e.a doesn't need any compression!

This looks pretty good, but I think we are being slightly too agresseive on the compression of the number of terms, for example compare

https://beta.lmfdb.org/ModularForm/GL2/Q/holomorphic/680/6/a/f/ (click to see 100 coefficients)
http://127.0.0.1:37777/ModularForm/GL2/Q/holomorphic/680/6/a/f/

I think we should be aiming to have compressed q-expansion extend roughly as far as the line above, and it's not the end of the world if it wraps to two lines.

assaferan:
Just to be clear, if one looks at https://beta.lmfdb.org/ModularForm/GL2/Q/holomorphic/680/6/a/f/, the compression does not work well (at least not as intended) - there is no compression for the trace, and the form itself is being compressed starting with the first compressed term (which happens to be $$q^{19}$$). 

If we are being too aggressive, and would like the threshold to have more characters (e.g. as above), it's easy to do, but line / two lines is display dependent. Do we want the string to depend on the display parameters?

AndrewVSutherland:
I thinks this looks good! (and no I don't think we should do anything display dependent)

PR #6271, opened 2024-11-20
Updated acknowledgements for congruent number data

JohnCremona:
After learning more about how the data was computed and by whom, I have updated the final section of the page https://beta.lmfdb.org/EllipticCurve/Q/CongruentNumbers

**However** looking at that page now, I see that its breadcrumbs place it under the "extra datasets" page -- good -- while the Learn More box still places it within ECQ.  That is because (I think) of the @ec_page.route("/CongruentNumbers")
 appearing within elliptic_curves.  Where is the datasets/ route handled?  Should these "home pages for datasets" such as this one be handled by routes there?  Then they could have their own Source, Completeness and Reliability data.

These remarks need nt stop *this* PR from being merged, it should either be a new Issue or be added to an open issue about datasets (but I could not find one).

### DIFF
--- a/lmfdb/elliptic_curves/templates/congruent_number_data.html
+++ b/lmfdb/elliptic_curves/templates/congruent_number_data.html
@@ -259,9 +259,12 @@ Each file consists of 1,000,000 lines, indexed by the positive integer $n$ and t
 
 <p>
 The data files were created by Randall L. Rathbun on February 14, 2013
-with updates on May 18, 2014.  Some data was provided by Allan J
-MacLeod, Ralph Buchholz and Brett Wittby, the rest computed by Rathbun using <tt>PARI/GP</tt>
-and <tt>eclib</tt> (in particular <tt>mwrank</tt>).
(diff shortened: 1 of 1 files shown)


roed314:
The learnmore is because of the `learnmore` list created explicitly [here](https://github.com/LMFDB/lmfdb/blob/main/lmfdb/elliptic_curves/elliptic_curve.py#L960), [here](https://github.com/LMFDB/lmfdb/blob/main/lmfdb/elliptic_curves/elliptic_curve.py#L996) and [here](https://github.com/LMFDB/lmfdb/blob/main/lmfdb/elliptic_curves/elliptic_curve.py#L934) from [this larger list](https://github.com/LMFDB/lmfdb/blob/main/lmfdb/elliptic_curves/elliptic_curve.py#L996).  I think it's helpful to have links to these datasets from the other elliptic curve pages.  I agree that the RCS knowls probably shouldn't be linked from the datasets; I'm not sure what should be in the list (do we want all auxiliary datasets to show up each one's learnmore list?).

I don't think there's an issue for this yet, so feel free to create one.

PR #6278, opened 2024-11-28
Ban outside domains like 375420.xxaza.cn

roed314:


roed314:
@edgarcosta came up with a different solution using our firewall.

Issue #6276, opened 2024-11-28
Listing of elliptic curve generators


JohnCremona:
As report by @AndrewVSutherland at #6273 :

> I notice that both ECQ/ECNF use a table with a column for each infinite order generator, which is already likely to force horizontal scrolling on pages like https://www.lmfdb.org/EllipticCurve/Q/19047851/a/1 for many users, and I worry things will get worse in the future if/when we add more higher rank curves (we have definitely been getting requests to include the new Elkies-Klagsbrun rank 29 curve, for example!) I wonder if it would make more sense to use a table with a row for each generator (which could also include torsion generators and a column for the order, giving you a complete Z-basis). This table will be very short for most curves, so I don't think screen real estate is a major concern.

JohnCremona:
I have done this, both for ECQ and ECNF, and dealing with #6275 at the same time.  It looks OK when we have all the data, but for ECNF in particular there are issues around not always having computed the rank exactly, so the number of generators we have to display may (for example) be 0 when the rank bounds are 0 and 2.  So I am check lots of representative examples before making a PR, and hope to give several links to those in the PR.

JohnCremona:
Question:  on an ECQ curve page the BSD invariants include "Special value" with knowl ec.q.special_value, while in an ECNF curve page we have instead "Leading coefficient" with knowl lfunction.leading_coeff.  There is no knowl ec.special_value, though there coud be.  Shall we make the ECQ page also show "Leadiing coefficient"?

roed314:
I like the ec.q.special_value knowl more than lfunction.leading_coeff for this purpose.  Maybe change the name of ec.q.special_value to ec.special_value and use it for both ECQ and ECNF?

JohnCremona:
> I like the ec.q.special_value knowl more than lfunction.leading_coeff for this purpose. Maybe change the name of ec.q.special_value to ec.special_value and use it for both ECQ and ECNF?

Can do.  The knowl will need editing of course.  I'll try to finish this tomorrow morning

roed314:
Looks good.

Issue #6275, opened 2024-11-28
ECQ and ECNF pages: make more consistent


JohnCremona:
As reported by @AndrewVSutherland at #6273: 

> I notice that the ECQ/ECNF Mordell-Weil Invariants sections use rather different layouts, are there reasons to format them differently? I like that everything is grouped under one heading in the ECNF layout, but I also like the way the structure of the MW group is shown on the ECQ pages (including both the free and torsion parts rather than giving the structure of the torsion subgroup on a separate line). I think it might be worth figuring out the optimal layout (perhaps combining elements of both, but also see (4) below), and then uniformizing them.

I set the milestone to 1.3 rather than 1.2.2 as 1 Dec is only a couple of days away.

JohnCremona:
See #6276:  I am doing both in one PR but will leave this open.

roed314:
@JohnCremona Now that #6279 has been merged and is on beta, can this and #6276 be closed?

JohnCremona:
Yes, I think so. I am still working on replacing the ECNF gens with smaller ones but that will take some time ( and will only affect stored data, not code).

roed314:
Great!

roed314:
For the record, here are some differences that still exist:
* ECNF is missing integral points
* ECNF is missing Faltings height, stable Faltings height, abc quality, and Szpiro ratio.
* For ECNF the period is labeled "Period" while for ECQ it's labeled "Real period."
* There is no displayed BSD formula for ECNF.
* In the local data for ECNF, the symbols for conductor and discriminant do not match those used above, and $$\mathrm{ord}(j)$$ does not match $$v_p(\mathrm{den}(j))$$ on ECQ.  There is also no note when the curve is semistable or count for the number of primes of bad reduction.

@JohnCremona, if any of these are things you think deserve an issue or a PR, feel free to open one!
*

roed314:
Thanks for the explanations!  I was just going through and recording the differences, not saying that they needed to be fixed.

JohnCremona:
I have created a feature request issue for adding Szpiro ratio (easy), Faltings height (not too hard) and possibly stable Faltings height (harder) at #6292 .  Of these, the Szpiro ratio could be done easily on the fly using what is already in the database.

Adding integral points is a lot more work, so not on the horizon and I have not made an issue even as a feature request.

There's now a PR for the BSD formula #6290 .

[...]

PR #6279, opened 2024-11-29
Better and more consistent display of invariants and generators for ECQ and ECNF

JohnCremona:
This deals with the points raised in #6275  and #6276

### DIFF
--- a/lmfdb/ecnf/WebEllipticCurve.py
+++ b/lmfdb/ecnf/WebEllipticCurve.py
@@ -483,6 +483,13 @@ class ECNF():
         except AttributeError:
             self.qc = "not determined"
 
+        # Mordell-Weil group
+        try:
+            invs = [0 for a in range(self.rank)] + list(self.torsion_structure)
+            self.mw_struct = "trivial" if len(invs) == 0 else r'\(' + r' \oplus '.join((r'\Z' if n == 0 else r'\Z/{%s}\Z' % n) for n in invs) + r'\)'
+        except AttributeError: # if self.rank not set
+            self.mw_struct = "unknown"
+
         # Torsion
         self.ntors = web_latex(self.torsion_order)
         self.tr = len(self.torsion_structure)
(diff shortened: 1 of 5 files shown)


JohnCremona:
I am finding and fixing some test failures now...

AndrewVSutherland:
How are you ordering generators?  It looks like infinite order generators come first, which seems fine, but among the infinite order generators it looks like on ECQ pages they are typically ordered by height (at least in the examples I looked at), but this is not true on the ECNF pages (see http://127.0.0.1:37777/EllipticCurve/2.0.3.1/44563.2/a/1 for example).  I think it would be better to always order them by height.

I could see an argument for ordering all the points by height (which would put the torsion points first, in which case we would probably want to change the way the MW group structure is written to match), but I don't feel strongly about this.

JohnCremona:
Over Q the generators of infinite order are indeed always in height order.  In detail, from any set of generators (mod torsion) I do a Minkowski reduction to get a new set, so the first generator has minimal height, etc.  After that, I try adding all torsion points to each of these minimal generators (recall tat this does not change the canonical height), picking whichever is smallest in a naive sense (literally, the shortest string representation).  The code for this is in https://github.com/JohnCremona/ecdata/blob/master/scripts/red_gens.py.  About 4 years ago I ran this on the elliptic curves over Q and changed the stored generators in the database.  So the situation over Q is good. (I should say that I did all this after prompting by Randall Rathbun, who had written his own code for "reducing" the generators in much the same way, but in the end I wrote my own.)

I could easily carry out the same process for ECNF, but have not done so.  Let's make a new issue for that.  It should not take ong as the code for handling generators over Q will work with trivial changes (if any), and there are fewer curves.

After this, the non-torsion generators over number fields would be ordered by height and minimal in the above sense, which is what we surely want.  I personally would not want to move the torsion generators first, even if they have height 0.  Instead, I had thought of having the torsion and non-torsion generators separated in the table, or in two tables, so the non-torsion ones would not need to have \infty in the order column and the torsion ones would not need to have 0 in the height column.  But I would also be happy to leave the table as it is (after improving the generators for ECNF).

Anyway, the proposed improvement of ECNF generators does not require any more changes to code or templates, so can we merge this now?

AndrewVSutherland:
Another option would be to just sort the generators when they are displayed on the page (rather than updating the data), but if you prefer to do that in the data, that's fine.  I don't think we need two tables and think that it's fine to list the infinite order generators first (but it would be nice to have these in height order).

JohnCremona:
OK, I'll add a sort now (and make a new PR) while I work on improving the data.

JohnCremona:
> OK, I'll add a sort now (and make a new PR) while I work on improving the data.

This is #6284 -- sorry not to have just added to this PR, as it was only a one-line addition.

Issue #5409, opened 2023-03-11
BSD normalisation issues for ECNF


JohnCremona:
It is very hard to get the BSD formula for elliptic curves over number fields exactly right.   There are several sources and they do not agree with each other regarding how the individual factors are normalised, though they do agree overall.  When computing the BSD data I went into this in great detail and even wrote 6 pages of notes on the issues, including how to compute the terms (as far as possible) in Sage, GP, Magma.  But some issues still remain:

    1. The regulator which appears in BSD is the usual height pairing determinant, but these heights must **not** be normalised as is usually done by dividing by the degree.  This is explained in the knowl for regulator, but the regulator we display is normalised, so anyone trying to recover the correct analytic Sha value will be out by a factor of $d^r$, where $d$ is the base field degree and $r$ the rank;
    2. The global period in BSD is a product of periods at each real and complex place.  At **complex** places an extra factor of 2 is needed.  This is stated correctly in the knowl **but** the global period does not have this factor in it; instead, an additional factor of $2^{r_2}$ was divided by in computing the analytic Sha (where $r_2$ is the number of complex places).

I don't think that 1 needs any action since the knowl ec.regulator explicitly says that in BSD one needs the non-normalised regulator.  But 2 is inconsistent (over imaginary quadratic fields and the mixed cubic 3.1.23.1 (all our other fields are totally real).  We should **either** change the data by multiplying the global period by $2^{r_2}$ to match the knowl (and the convention I fixed on); **or** keep the data the same, change the knowl, and change the BSD formula to include the factor $2^{r_2}$ explicitly.  I think the first of these is preferable.

There are plenty of other things which need to be got right, such as making sure that none f the definitons depend on the model of the curve (particularly important when there is no glbal minimal model) but I think I have those right...

JohnCremona:
I have been editing the knowls to make this consistent.  I will probably have to adjust code and/or data for item 2.  At the same time as any data change necessary for this I will also be dealing with #5308  and a second BSD data issue which means that for about a dozen curves the generators are unsaturated and the stored (and displayed) Sha value is 0.

JohnCremona:
The knowls are now updated and should be consistent with the conventions made explicit in https://github.com/JohnCremona/BSD.

Code and data changes to be made:
[...]

PR #6287, opened 2024-12-04
EC/ECNF consistency in local data display

JohnCremona:
This fixes some points raised at #6275, making the display of local data consistent between ECQ and ECNF:  table headings use the same notation, and ECNF now also has te text above the table saying whether or not the curve is semistable, and how many bad primes are.  (Note that, as before, the table has a row for any primes where the curve has good reduction but the model does not.  Such primes exist when there is no global minimal model; the model we then use has exactly one such prime.)

I also renamed "Period" --> "Global period" for ECNF a that is what the knowl defines, but ECQ still calls it "Real period" as that is what it is always called over Q (and there is a specific knowl too).

### DIFF
--- a/lmfdb/ecnf/WebEllipticCurve.py
+++ b/lmfdb/ecnf/WebEllipticCurve.py
@@ -359,6 +359,9 @@ class ECNF():
         self.cond_norm = web_latex(self.conductor_norm)
 
         Dnorm = self.normdisc
+        self.model_disc = self.disc.replace('w', Kgen).replace("*","").replace("(","").replace(")","")
+        if Kgen == 'phi':
+            self.model_disc = self.model_disc.replace(Kgen, r"\phi")
         self.disc = pretty_ideal(Kgen, self.disc)
 
         local_data = self.local_data
--- a/lmfdb/ecnf/templates/ecnf-curve.html
+++ b/lmfdb/ecnf/templates/ecnf-curve.html
@@ -276,7 +276,7 @@ cellpadding="5";
         </tr>
 
         <tr>
-        <td align='left'>{{ KNOWL('ec.period', title='Period') }}:</td>
+        <td align='left'>{{ KNOWL('ec.period', title='Global period') }}:</td>
 	<td>$\Omega(E/K)$</td><td>&approx;</td>
         <td> {{ ec.omega }}</td>
         </tr>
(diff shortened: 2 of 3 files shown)


JohnCremona:
I do not know why in the ECNF local data table the headings of the columns for the prime and Reduction type are not centred.  A fix for that would be welcome.

roed314:
All of the headings are left aligned (which is the standard for `ntdata` tables), it's just that those are the only ones where the width of the data is large enough that you can see it.  We have this in `style.css`:
```css
table.ntdata.centered th, table.ntdata.centered td {
  text-align: center;
}
```
so if you add `centered` as a class to the table it will do what you want.

Otherwise it looks good, so feel free to merge after you decide what to do about the centering!  Thanks for fixing these.

JohnCremona:
> All of the headings are left aligned (which is the standard for `ntdata` tables), it's just that those are the only ones where the width of the data is large enough that you can see it. We have this in `style.css`:
> 
> ```css
> table.ntdata.centered th, table.ntdata.centered td {
>   text-align: center;
> }
> ```
> 
> so if you add `centered` as a class to the table it will do what you want.
> 
> Otherwise it looks good, so feel free to merge after you decide what to do about the centering! Thanks for fixing these.

Thanks, I will try that.  CSS remains a dark art to me!

roed314:
If you change line 366 of ecnf-curve.html to `<table class="ntdata centered"><thead>` it centers everything.

I also noticed that $$\mathfrak{D}$$ is not defined (it looks like we use $$(\Delta)$$ above), and $$\mathfrak{N}$$ is capitalized in the local data table but not in the invariants section (where it is displayed as $$\mathfrak{n}$$).

JohnCremona:
Thanks for this!  I haven't yet mmade any changes to this as I have been working on displaying th BSD formula for CNF and have just got it working (including the adjustment to the global period at complex places as explained in #5409).

I will finish this one off and make sure the other one is good and make another PR for that tomorrow (Friday).

JohnCremona:
> If you change line 366 of ecnf-curve.html to `<table class="ntdata centered"><thead>` it centers everything.
> 
> I also noticed that D is not defined (it looks like we use ( Δ ) above), and N is capitalized in the local data table but not in the invariants section (where it is displayed as n ).

 - Centering now fixed.
 -  I changed the conductor ideal notation to upper case (i.e. \frak{N}) in all places rather the lower case.
 - Discriminants:  there is a distinction between the discriminant $\Delta$ of the model, which is a field element, and the minimal discrminant, which is an ideal.  We have been displaying the former in all cases,  as a principal ideal $(\Delta)$ rather than as an element so that te factorization which is also displayed does not have an uninteresting unit factor; and only also displaying the minimal discriminant ideal and its factorisation when that is different, which is precisely when there is no global minimal model.  And in such a case there is/was no line in the Invariants table with frak{D}_min, so its appearance in the local data table is mysterious, as you saw.
I tried different ways to deal with this.  The final version has an extra row in the Invariants table labelled "Discrimimant" which shows the discriminant of the model, with no factorization, with the knowl ec.discriminant defining that.  When there is a global minimal model there are two more lines, for the "Discriminant ideal" and its norm, showing the  principal ideal (Delta) and its factorization, and including " = D_min".  When there is no global minimal model there are four lines, two for the "Discriminant ideal" and its norm, with their factorizations, and two for the minimal discriminant ideal and its factorization.  In the former case the knowl for "Discriminant ideal" is ec.minimal_discrimimant; in the latter, the knowl for "Discriminant ideal" is ec.discriminant while that for "Minimal discriminant ideal" is ec.minimal_discriminant.  *And* I added to that knowl (ec.minimial_discriminant) to say when it is just (Delta) and how it differs when not.

Phew.  I have probably missed something, so another pairs of eyes would be welcome!  I do think it instructive to see and compare the ideal factorisations when the model is not a global minimal one, so you can see the extra 12th power.

PR update in a moment...

JohnCremona:
http://localhost:37777/EllipticCurve/4.4.19796.1/40.1/h/2
http://localhost:37777/EllipticCurve/2.2.229.1/1.1/a/1

are good samples to look at

roed314:
Looks good.

PR #6303, opened 2024-12-16
Catch more bits of data which may be missing

jwj61:
This adjusts python so it correctly catches if the tame and unramified degrees are missing.

See
https://beta.lmfdb.org/padicField/2.16.16.7
http://127.0.0.1:37777/padicField/2.16.16.7

### DIFF
--- a/lmfdb/local_fields/main.py
+++ b/lmfdb/local_fields/main.py
@@ -311,6 +311,11 @@ def formatbracketcol(blist):
         return 'not computed'
     return f'${blist}$'
 
+def intcol(j):
+    if not j:
+        return 'not computed'
+    return f'${j}$'
+
 lf_columns = SearchColumns([
     LinkCol("label", "lf.field.label", "Label", url_for_label),
     MathCol("n", "lf.degree", "$n$", short_title="degree", default=False),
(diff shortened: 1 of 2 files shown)


roed314:
Most of the invariants in the Galois section say "Not computed," but Wild Inertia Group is shown as "data not computed."  These should probably be consistent.

roed314:
Similarly, in search results like https://beta.lmfdb.org/padicField/?n=16&p=2&c=16&showcol=n.u.t.visible.unram.eisen.ind_of_insep.associated_inertia, some columns are explicitly shown as "not computed" and others left blank.

lmfdb/local_fields/main.py
```diff
@@ -304,6 +304,18 @@ def galcolresponse(n,t,cache):
...
 
+def formatbracketcol(blist):
+    if blist == []:
+        return r'$[\ ]$'
+    if not blist:
+        return 'not computed'
+    return f'${blist}$'
+
+def intcol(j):
+    if not j:
```
roed314, line 315: This should be `if j is None`, since this will display `not computed` for `j=0`.  I know that's not a possible value for `t` and `u`, but this code could eventually get copied and used elsewhere.

I'll go ahead and make this change.

roed314:
Hm, apparently `t` and `u` are empty strings not `None`, so my "fix" is wrong.  Working on it....

roed314:
Ok, should be fixed.  @jwj61, feel free to merge if you're happy with my changes.

PR #6305, opened 2024-12-17
Fix problems with decode_as_matrix when rep_type=GLFq

roed314:
Compare
* https://beta.lmfdb.org/GaloisGroup/15T34 (which gives a server error)
* https://localhost:37777/GaloisGroup/15T34

I also removed `^1` in the display of matrices over Fq and added the minimal polynomial of $$\alpha$$.

There will be compatibility problems between Sage and Magma if we end up outside the range where Conway polynomials are available (we can't just use `GF(q, modulus="primitive")` since we need to match the polynomial chosen when defining the field as the data is computed, and we don't currently store that polynomial).  Fortunately, the current set of qs is

```
4,8,9,16,25,27,32,49,64,81,121,125,128,169,243,256,289,343,361,529,625,729,841,961
```
and Conway polynomials are available in all of these cases.

### DIFF
--- a/lmfdb/groups/abstract/code.yaml
+++ b/lmfdb/groups/abstract/code.yaml
@@ -56,7 +56,7 @@ GLZq:
 
 GLFq:
   comment: Define the group as a matrix group with coefficients in GLFq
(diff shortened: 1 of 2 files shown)


jenpaulhus:
I tested a bunch of group pages which have GLFq representatives and they all seem to work.  Code looks good to me so I will go ahead and merge.

PR #6312, opened 2025-01-04
remove inaccessible history links

AndrewVSutherland:
Removes inaccessible code that references history knowls we no longer show or maintain (the content of these knowls will still be available, this PR just removes references to them that are not used)


roed314:
The test failure is from new hyperelliptic curve data, and is unrelated to this PR.  I'll make a new PR to fix it.

roed314:
See #6313 for the test failure.

Issue #6309, opened 2024-12-29
Possible wrong character value on newforms


jenpaulhus:
From the bug report:

The value of chi on https://www.lmfdb.org/ModularForm/GL2/Q/holomorphic/7/5/b/a/ should be -1 when evaluated at 3. The same issue is also on newform 7.3.b.a.

edgarcosta:
Where is the contradiction?

jenpaulhus:
I think they are referring to the table in the Character values section where it has a value of +1? I can get more info from the person who submitted the bug if needed.  I don't know newforms very well so if this is something that is obviously correct as stated on LMFDB, just let me know and I can convey to the person who submitted the bug report.

roed314:
The character is `3.b` which has order 2, so it's a bug that the image of 3 is 1 (this would be the trivial character).

edgarcosta:
Ups, never noticed that section in my life

saruulbao:
Hi! I am the one reporting the issue. I know that when I go to the homepage of the characters, I see the correct values there so it's not really a mistake. I spotted some other ones: Newform orbit 8.5.d.a. (same issue as in the 7.3.b.a one)  and Newform orbit 12.3.c.a (this one is even more weird as it tells you the value of the Nebentypus character at 7 is 0). I suspect there might be some systematic mistake in the display on the pages of these newforms. I really hope this issue can be addressed. Thank you!

edgarcosta:
The problem is here:
https://github.com/LMFDB/lmfdb/blob/main/lmfdb/classical_modular_forms/web_newform.py#L961

which calls:
https://github.com/LMFDB/lmfdb/blob/main/lmfdb/classical_modular_forms/web_newform.py#L961
perhaps under the wrong assumptions?

but just quickly looking at I don't see anything wrong, and the [data](https://www.lmfdb.org/ModularForm/GL2/Q/holomorphic/data/7.5.b.a) is correct, as we have:
```
'hecke_ring_character_values': [[3, [-1]]]
```


I hope this helps whoever has more time to dig into this.

wwylele:
I think I figured out what went wrong and posted a fix in #6318

roed314:
Great!  I probably won't have time to review it this week, but perhaps someone else will.  Thank you!

saruulbao:
> I think I figured out what went wrong and posted a fix in #6318

[...]

PR #6318, opened 2025-01-08
Fix Dirichlet character for newform orbit whose dimension is 1

wwylele:
This is my attempt of fixing #6309 . The `hecke_ring_character_values` field in this file seems to have double meaning: the generator value can be either the actual dirichlet character value (the complex e^{...}), or the log of it. Which way to interpret it [depends on the existence of non-0 `hecke_ring_cyclotomic_generator`](https://github.com/LMFDB/lmfdb/blob/main/lmfdb/classical_modular_forms/web_newform.py#L919-L926). Meanwhile, `ConreyCharacter.values_gens` always return the log value. This creates a discrepancy in this specific `dim==1` code block, where `hecke_ring_cyclotomic_generator` is set to 0 (meaning `hecke_ring_character_values` should be interpreted as the final values), but `hecke_ring_character_values` stores the log values. A small transformation is needed here

A similar transformation is done here: https://github.com/LMFDB/lmfdb/blob/65713a6fc29034e88eeb78e82106d495b1a4c4cd/lmfdb/classical_modular_forms/download.py#L27

### DIFF
--- a/lmfdb/classical_modular_forms/test_cmf.py
+++ b/lmfdb/classical_modular_forms/test_cmf.py
@@ -594,3 +594,58 @@ class CmfTest(LmfdbTest):
                 and 'mf_hecke_charpolys' in data and 'charpoly_factorization' in data
                 and 'mf_newform_portraits' in data and "data:image/png;base64" in data
                 and 'mf_hecke_traces' in data and 'trace_an' in data)
+
+    def test_character_values(self):
+        # A newform orbit of dimension 1
+        data = self.tc.get('/ModularForm/GL2/Q/holomorphic/12/3/c/a/').get_data(as_text=True)
+        character_values_table = r"""
+<table class="ntdata">
+  <tbody>
+        <tr>
+      <td class="dark border-right border-bottom">\(n\)</td>
+      <td class="light border-bottom">\(5\)</td>
+      <td class="dark border-bottom">\(7\)</td>    </tr>
+    <tr>
+      <td class="dark border-right">\(\chi(n)\)</td>
+      <td class="light">\(-1\)</td>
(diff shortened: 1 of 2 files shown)


edgarcosta:
Just to add some more info, according to the schema we have:
[columns.mf_hecke_nf.hecke_ring_character_values](https://www.lmfdb.org/knowledge/show/columns.mf_hecke_nf.hecke_ring_character_values)

> list of pairs [[m1,[a11,...,a1n]],...[mr,[a1r,...,arn]]] where mi are generators of (Z/NZ)* and [ai1,...,ain] is the value of chi(mi) expressed in terms of the Hecke ring basis or in cyclotomic representation [[c,e]] encoding c x zeta_m^e where m is hecke_ring_cyclotomic_generator`


while
[columns.mf_newforms.char_values](https://www.lmfdb.org/knowledge/show/columns.mf_newforms.char_values)

> quadruple where N is the level, n is the order of the character, u is a list of generators for the unit group of Z/NZ, and v is a corresponding list of integers for which chi(u[i]) = zeta_n^v[i]


@wwylele Would you mind adding some tests showing that the bug is fixed?

wwylele:
Sure, let me add some

wwylele:
I added some test. I am not quite happy that I did text matching over a whole html table. Is there a html parser in the testing framework that I am not aware of?

edgarcosta:
Unfortunately, there isn't an easy way. I think what you did is fine, it is fragile, but we will live with it.

One last request, before we merge. Would you mind adding a comment on top of your change? (this will help whoever looks at this again in the future)

wwylele:
Updated

PR #6324, opened 2025-01-14
Family pages

asiminah:
We created family pages for the families of modular curves on the lmfdb.


roed314:
This is now up on https://blue.lmfdb.xyz/ModularCurve/Q/ if anyone wants to take a look.

roed314:
The test failure now is probably due to a data change in local fields and not your fault.  @jwj61 are you able to take a look?

jwj61:
Actually, the problem is that including the Galois group is now broken for padic fields.  I noticed this earlier today and was going to look at it tomorrow anyway.

jwj61:
Yes, it is a data problem (and my fault).  Rewrite in progress.

jwj61:
Should be fixed now.

On a tangentially related note, we currently have columns gal and galT which are exactly the same.  I suggest deleting galT after removing it from the code (it is easy to search for the few instances of galT).  Objections?

roed314:
I certainly agree on dropping the column eventually, but I would wait until everything gets updated to not use the old column (the main thing that breaks [color branches](https://beta.lmfdb.org/static/color-branches.txt) is dropping columns).

AndrewVSutherland:
Apologies for the delay in reviewing this, I noticed a few issues.

Small things:

- bi-elliptic needs a knowl
- I think we should omit the remarks section whn both are unknown
- In the remarks section I think claims of the form "There are only * hyperelliptic/bi-elliptic curves." should instead say "There are only * hyperelliptic/bi-elliptic curves in this family."
- On http://127.0.0.1:37777/ModularCurve/Q/family/XS4 the $\ell$ in the first line should be $p$.
- On http://127.0.0.1:37777/ModularCurve/Q/family/Xarithpm1 I would move the +/-1 to the upper left corner to match X_{+/-1}(N); this will still define the same group.

Big things:

- The moduli descriptions of X_{arith,1}(M,MN) and X_{arith,+/-1}(M,MN) are incorrect, the current language descirbes  X_1(M,MN) and X_{+/-1}(M,MN);  the "arith" versions of these curves have canoncial field of definition Q for all M and N (that is the whole point of the "arith"), and are included in the LMFDB for all MN <= 70, e.g. X_{arith,1}(3,15)  is [15.576.9-15.a.2.1](https://beta.lmfdb.org/ModularCurve/Q/15.576.9-15.a.2.1/) (this isn't obvious from staring at the generators, but you can conjugate them to match).   For the moduli interpretation of the arith version, note that if you project to level M you get X_arith(M), so really you have point of order MN and an independent cyclic subgroup of order M. 
- X_{+/-1}(3,3N) also has canonical field of definition Q and these curves are in the database, e.g. X_{+/-1}(3,15)  is [15.576.9-15.a.2.4] (https://beta.lmfdb.org/ModularCurve/Q/15.576.9-15.a.2.4/) note this is not the same curve as X_{arith,1}(3,15) or X_{arith,+/-1)(3,15) (the later of which is [15.288.9.a.2](https://beta.lmfdb.org/ModularCurve/Q/15.288.9.a.2/).

We should decide whether we want to have separate families for X_1(M,MN) (which would only be relevant for M=2) and X_{arith,1}(M,MN), and similarly for X_{+/-1}(M,MN) (which would only be relevant for M =2,3) and X_{arith,+/-1}(M,MN).

Let's discuss this on the LMFDB Zulip.

AndrewVSutherland:
It looks like you give the correct moduli description for X_{arith,1}(M,MN) and X_{arith,+/-1}(M,MN) in your paper [arXiv:2501.10883](https://arxiv.org/abs/2501.10883), you should just make the pages in this PR match your paper :)

junbolau:
We addressed the minor issues above. We will wait for the Zulip discussion to mature before we make edits on the major issue.

AndrewVSutherland:
I have now added Xarith1(M,MN) and Xarithpm1(M,MN) names to all the relevant groups in the database, and I updated this PR accordingly.  Note that X_1(2,2N) will be displayed as X_1(2,2N) rather than X_{arith,1}(2,2N) but will be included in the X_{arith,1}(2,2N) family, and similarly for X_{+/-1}(2,2N).  If we decide we want to change this, it can be done in the data without changing the code, so it isn't really relevant to this PR, which I'm happy to merge once the two small issues below are fixed:
- on http://127.0.0.1:37777/ModularCurve/Q/family/Xarith1 you should change <NP> + C to <NP> \oplus C
- on http://127.0.0.1:37777/ModularCurve/Q/family/Xarithpm1 the canoncial field of definition is still wrong, it should be Q (and there should be no restriction on M)

asiminah:
Thank you for catching that - I have fixed the two small issues!

AndrewVSutherland:
Great! I will merge this once tests pass.

PR #6348, opened 2025-03-27
Add Eisenstein form for p-adic polynomials

jwj61:
Polynomials for p-adic fields have been updated.  This updates their display accordingly.

For unramified and totally ramified extensions, this makes no change.  In other cases, the polynomial is in "Eisenstein form", which essentially means that it can be written in a special form.  This uses that form for display, and the simpler for the "raw" version.  Once merged, the knowl for the defining polynomial will be updated to explain what this form is.

Sample page:
http://127.0.0.1:37777/padicField/3.9.9.7
http://beta.lmfdb.org/padicField/3.9.9.7

### DIFF
--- a/lmfdb/local_fields/main.py
+++ b/lmfdb/local_fields/main.py
@@ -125,9 +125,9 @@ def eisensteinformlatex(pol, unram):
     pol=R(pol)
     l=[]
     while pol != 0:
-      qr=pol.quo_rem(unram2)
-      l.append(qr[1])
-      pol=qr[0]
+        qr=pol.quo_rem(unram2)
+        l.append(qr[1])
+        pol=qr[0]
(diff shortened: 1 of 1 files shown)


jwj61:
Note, test failure is from groups

roed314:
This looks good, and I agree that the test failures are from groups (I'm hoping to look at that today, but may not get to it).

I think we should wait to merge this until @edgarcosta is finished upgrading the postgres server and we are able to update knowls again.

roed314:
I figured out where the new group errors are coming from: when we ported `gps_subgroups` to the `scratch` tablespace yesterday, `include_nones` accidentally got set to `false`.  I can't fix this until the database is not read-only.

roed314:
Fixed, so I'm merging this.

PR #6350, opened 2025-03-27
remove traces of python2 ; other ruff UP fixes

fchapoton:
This removes some checks that version == 3

and also replaces a few lists by iterators as suggested by ruff


roed314:
Lint failure, since `sys` is no longer used in two files.  Otherwise, looks fine to me.

fchapoton:
Voilà !

Issue #6112, opened 2024-07-11
Expand BMF and ECNF/IQF data to include more fields


JohnCremona:
Now, the data for BMF Bianchi modular forms (dimensions and rational newforms) and associated elliptic curves over imaginary quadratic fields includes 32 fields, all those of absolute discriminant up to 100 and also 163.

I have more data, for all fields of absolute discriminant up to 2000 (and going beyond that, in progress), and will upoad it.  I will use this Issue to record progress.

- [x] 1. For the 7 fields of abs disc between 100 and 120 I have data for all levels (conductors) of norm up to 1000, consisting of 5989 rational newforms and 12524 elliptic curves. These are now on beta.

- [x]  2. For the fields of abs disc between 123 and 700 (excluding 163 for which data was there already) I have data for level norms up to 100 only.  These will be next. That is about 270 fields. Now on beta.

 - [x]  3. For the fields of abs disc between 701 and 1000 I only have data for level norms up to 15, and will work on extending that to 100 if possible, though I may upload what I have already.  In this smaller range there are several fields with no rational newforms (or elliptic curves), but the dimension data is still available.
 
 - [x] 4. For the fields of abs disc between 1000 and 2500 (about 450 more fields) I have BMF data for levels of norm up to 15 or 11, dimensions and rational newforms and elliptic curve data.  This is ready for upload but is waiting until item 3 is finished and uploaded.

JohnCremona:
As I have uploaded data, I have edited the relevant knowls.  Later I'll also adjust the brows pages: on https://beta.lmfdb.org/ModularForm/GL2/ImaginaryQuadratic/ it would make sense to insert the last field for which the database has data after the three dots.  Similarly on https://beta.lmfdb.org/EllipticCurve/ (and not just for imaginary quadratic fields there).  This could be possibly automated, using a database query to find the largest discriminant of each base field signature.

JohnCremona:
Update on item 3: there are still 12 fields of abs disc between 700 and 1000 where I am still working on the levels of norm up to 100 (of which 2 are still in the 70s, 2 in the 80s, 7 in the 90s and one working on levels of norm 100).  That will be done by the end of 2024.

[...]

Issue #3058, opened 2019-05-08
Bianchi newforms (dimension 1) with no elliptic curve.


JohnCremona:
The table bmf_forms contains 185581 newforms of dimension 1 over the 5 fields 2.0.d.1 for d in {1,8,3,7,11}.  This includes 185541 with associated elliptic curves and 40 (listed below) where is no associated curve.  these are not just missing, they do not exist (sometimes these are known as "fake elliptic curves"); instead there is an abelian surface whose L-function which matches that of the elliptic curve.

Over these fields the table ec_nfcurves contains 186507 isogeny classes of elliptic curves, including 966 classes whose label includes "CM", which have CM by an order in the base field, which are not associated to cuspidal BMFs.  That leaves 185541 classes which match bijectively to the BMFs of the preceding paragraph.

The reason for this issue is (1) to record these numbers (as of 2019-05-08), (2) to list the exceptional BMFs, and (3) to put into the agenda the need to indicate on the home pages of the 40 exceptional BMFs that they are exceptional.  Currently their home pages just say, in the Friends box, "No elliptic curve" or (e.g.) "Isogeny class 2.0.3.1-67081.3-a missing" depending on the type of exception, not all of which are currently detected automatically.

JohnCremona:
Bianchi newforms with no elliptic curve (or "false elliptic curves", genuine or not)

www.lmfdb.org/ModularForm/GL2/ImaginaryQuadratic/2.0.4.1/34225.7/b
www.lmfdb.org/ModularForm/GL2/ImaginaryQuadratic/2.0.4.1/34225.7/a
www.lmfdb.org/ModularForm/GL2/ImaginaryQuadratic/2.0.4.1/34225.3/b
www.lmfdb.org/ModularForm/GL2/ImaginaryQuadratic/2.0.4.1/34225.3/a
www.lmfdb.org/ModularForm/GL2/ImaginaryQuadratic/2.0.8.1/5625.3/b
www.lmfdb.org/ModularForm/GL2/ImaginaryQuadratic/2.0.8.1/5625.1/b
www.lmfdb.org/ModularForm/GL2/ImaginaryQuadratic/2.0.8.1/6561.5/a
www.lmfdb.org/ModularForm/GL2/ImaginaryQuadratic/2.0.8.1/6561.5/d
www.lmfdb.org/ModularForm/GL2/ImaginaryQuadratic/2.0.8.1/21609.3/c
[...]

PR #6353, opened 2025-04-01
Bmf0curves

JohnCremona:
See #3058 for the associated Feature Request.

Some Bianchi newforms (weight 2, rational eigenvalues) do not have associated elliptic curves, though most do, and of those which should do we have curves in almost all cases but not quite all.  The new column 'curve_status' in the table bmf_forms allows this information to be displayed in a better way than just looking up on the fly whether or not there is a curve with some ad hoc method to tell whether it does not exist or is simply missing.  This column has been correctly filled for the BMFs currently in the database: 228896 forms, of which 228687 have curves, 193 do not have curves, and 16 should do but I have not yet found them; the value of 'curve_status' is 1, -1, 0 respectively:

`sage: [(n,db.bmf_forms.count({'curve_status':n})) for n in [-1,0,1]]`
`[(-1, 193), (0, 16), (1, 228687)]`

I added a section on the home page of a BMF headed "Associated elliptic curves" with one of three possible text templates.

Examples:

1. Normal (curve exists):

http://localhost:37777/ModularForm/GL2/ImaginaryQuadratic/2.0.3.1/106176.1/a/

(text says "This Bianchi newform is associated to the isogeny class 2.0.3.1-106176.1-a of elliptic curves.")

2. Elliptic curve should exist but is missing:

http://localhost:37777/ModularForm/GL2/ImaginaryQuadratic/2.0.511.1/100.5/a/

(text says: "This Bianchi newform is conjectured to be associated to an isogeny class of elliptic curves, but no matching elliptic curve is currently known.")

3. No elliptic curve exists:

http://localhost:37777/ModularForm/GL2/ImaginaryQuadratic/2.0.568.1/9.1/a/

(text says: "This Bianchi newform is not associated to any elliptic curves.")

I am open to suggestions for alternative wording, and for a knowl to be attached to the section heading.  Note that this is not primarily about modularity of elliptic curves over imaginary quadratic fields, which is now known (thanks to Newton and Caraiani)  for half the fields, but about whether the geometric object attached to a BMF is an elliptic curve or something of higher dimension; there are papers which the knowl should refer to.

I wrote the code so that if this is pushed to prod before the tables are then it should work, but I am also here asking for bmf_forms and ec_nfcurves both to be copied to prod. (The latter because I added a couple of doxen relevant curves recently, see #6112 .)

### DIFF
--- a/lmfdb/bianchi_modular_forms/templates/bmf-newform.html
+++ b/lmfdb/bianchi_modular_forms/templates/bmf-newform.html
@@ -39,6 +39,19 @@
 </table>
 </p>
 
+{% if data.dimension == 1 %}
+<h2>  Associated elliptic curves </h2>
+{% if data.ec_status == 1 %}
+This Bianchi newform is associated to the isogeny class <a href={{data.ec_url}}>{{data.label}}</a> of elliptic curves.
+{% endif %}
+{% if data.ec_status == 0 %}
+This Bianchi newform is conjectured to be associated to an isogeny class of elliptic curves, but no matching elliptic curve is currently known.
+{% endif %}
+{% if data.ec_status == -1 %}
+This Bianchi newform is not associated to any elliptic curves.
+{% endif %}
+{% endif %}
+
 <h2>  Atkin-Lehner eigenvalues  </h2>
 
 {% if data.have_AL %} 
--- a/lmfdb/bianchi_modular_forms/test_bmf.py
+++ b/lmfdb/bianchi_modular_forms/test_bmf.py
@@ -53,7 +53,7 @@ class BMFTest(LmfdbTest):
         Check that various search combinations work.
         """
         self.check_args(base_url+"?field_label=2.0.7.1&level_norm=322&count=10", 'Results (4 matches)')
-        self.check_args(base_url+"?start=0&include_base_change=off&include_cm=only&count=100", 'ModularForm/GL2/ImaginaryQuadratic/2.0.95.1/16.5/a/')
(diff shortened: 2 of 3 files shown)


JohnCremona:
Please hold off copying bmf_forms and ec_nfcurves until I have finished a new data upload (see #6112), then also bmf_dims will need copying.  Thanks.

JohnCremona:
Note that the extra BMF data uploaded causes one test in test_bmf to fail but that is fixed in PR #6356 .  These two PRs can be merged in either order.

roed314:
So this is ready to go, and doesn't need to wait on any more data being copied?

JohnCremona:
Yes: the code in this PR uses the new column if it exists but otherwise defaults to the old behaviour.  So it will certainly run fine on beta, and should also run fine on prod both before and after the data is copied over.

JohnCremona:
@roed314 Hold on -- I am testing that the code works OK when the curve_status column is absent by changing the one instance of self.curve_status in the code to self.curve_status_bogus, and it is not quite right.  To be fixed shortly...

JohnCremona:
OK now.  This code has now been tested to work whether or not the curve_status column is present in the table bmf_forms (which it is not yet on prod, though it is on beta).

AndrewVSutherland:
@JohnCremona Tests are failing due to the error
```
AssertionError: ModularForm/GL2/ImaginaryQuadratic/2.0.95.1/16.5/a/ not in the /ModularForm/GL2/ImaginaryQuadratic/?start=0&include_base_change=off&include_cm=only&count=100
```
and indeed this form does not appear in the first page of search results with count=100.  I think either you need to increase the count or check for a different field.

JohnCremona:
Sorry I missed that. Indeed, the data pushed that field off the first page.  I have changed what is looked for in the results page.

It's a long time since I made this PR, I had forgotten it was still pending.

PR #6356, opened 2025-04-03
updated BMF browse page field lists after new data added

JohnCremona:
This updates the fields displayed on the BMF main browse page to reflect the extra data recently added (now on beta, not yet on prod).  Instead of ending the lists with ...  I added at the end (after the ...) the last relevant field.  This is 2.0.2495.1 for dimension data but 2.0.2491.1 for newforms, since the actual last field (2.0.2495.1) has no rational newforms in the level range computed so far.

It would be possible to use db.bmf_dims.max('field_absdisc') and db.bmf_forms.min('field_disc') instead of hard-coding these in python (and until now I had not noticed the inconsistency in these column names and contents!) but I have no plans to compute any BMFs for fields of larger discriminant.  [Strictly speaking I  have put in dims for 2.0.2683.1 as well as that is the last field of class number 1-5, but there are no rational newforms there anyway in the small range so far computed.]

When this is merged I will update the relevant knowls.

This PR is independent of #6353 .

### DIFF
--- a/lmfdb/bianchi_modular_forms/bianchi_modular_form.py
+++ b/lmfdb/bianchi_modular_forms/bianchi_modular_form.py
@@ -64,13 +64,19 @@ def index():
     """
     info = to_dict(request.args, search_array=BMFSearchArray(), stats=BianchiStats())
     if not request.args:
-        gl2_fields = ["2.0.{}.1".format(d) for d in [4,8,3,20,24,7,40,11,52,56,15,68,19,84,88,23]]#,31,35,39,43,47,51,53,55,59,67,71,79,83,87,91,95,163]]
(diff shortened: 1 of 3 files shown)


JohnCremona:
I'll see about fixing the broken test shortly

JohnCremona:
I did list the 3 relevant tables on some issue or PR. bmf_firms, bmf_dims, ec_nfcurves.  All have a lot of extra data. Thanks.

AndrewVSutherland:
Done!

JohnCremona:
Great, thanks.  When this PR is on prod then things like this https://www.lmfdb.org/ModularForm/GL2/ImaginaryQuadratic/2.0.2199.1/4.2/a/
will be easier to find.

PR #6366, opened 2025-04-29
adding the group order to the top

edgarcosta:
Each time I was showing @emresertoz the page of a Galois group, he would look for the order at the top of the page, and I always had to point out to the lady gaga box.
It is unexpected to have an invariant on the box, but not at the top.
This PR proposes putting the order at the top.

@jwj61 and @roed314 , what do you think?

### DIFF
--- a/lmfdb/galois_groups/templates/gg-show-group.html
+++ b/lmfdb/galois_groups/templates/gg-show-group.html
@@ -10,9 +10,10 @@ table.reptable td, table.reptable th {
 }
 </style>
 
-<p> {{ place_code('gg') }}</p> 
+<p> {{ place_code('gg') }}</p>
   <p><h2>{{ KNOWL('gg.group_action_invariants', title='Group action invariants') }}</h2>
     <table>
+      <tr><td>{{KNOWL('group.order','Order')}}:<td>&nbsp;&nbsp;<td>{{info.ordermsg }}</td>
       <tr><td>{{KNOWL('gg.degree', 'Degree')}} $n$:<td>&nbsp;&nbsp;<td>${{info.n}}$</td>
       <td>{{ place_code('n') }}</td></tr>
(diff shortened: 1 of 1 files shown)


jwj61:
The top is supposed to be invariants of the group action while invariants of the group are lower down.  That was the logic for the current situation.  On one hand, I agree that it takes a little getting used to.  On the other hand, it is both in the gaga box and down below.

roed314:
It also bugs me.  I think that the group invariants are more basic, and more of them should be shown at the top of the page (including the group order).

I think the "Abstract group invariants" should be either first or second, and should include some additional information that's shown on the abstract group page.  In particular, I think it should have
* Order
* Exponent
* Composition factors
* All the boolean quantities shown on the abstract group page, in the same format

We may also want to show some special subgroups (center, commutator).

AndrewVSutherland:
Given that we show the abstract group name in the top section, it seems strange not to also show its order (these both seem like group invariants more than action invariants, but they are also both things you really want to know, so I think it makes sense to put them near the top).

I'm much less convinced about the exponent, composition factors, properties..., I think it's fine to put those further down.

I'm also curious what the plan is for when abstract groups are in production.  Presumably the abstract group name/label should then be an abstract group knowl (which would make group properties easy to find without scrolling down, in fact it might not even be necessary to list many group invariants on the Galois group pages if we had abstract group knowls).

AndrewVSutherland:
I could see an argument for putting the order after the t-number rather than before it (this PR effectively puts the order in the middle of the label).  But I can also see an argument against that (most people won't care about the t-number).

edgarcosta:
@roed314 , @assaferan , @SamSchiavone  discussed this on LMFDB Fridays and we are proposing moving most to the section of group invariants to the top. 
@roed314 will open a PR

PR #6367, opened 2025-05-02
Moving things around in galois group pages

roed314:
This is an alternative for #6366.  It does the following:
* Move the Group invariants section from the bottom to the top (with the exception of the Character Table, which becomes its own section).
* Change heading from "Group" to "Abstract Group," make the latex into an abstract group knowl, and have the label of the abstract group now shown implicitly in the knowl, rather than on a separate line
* Remove "Transitive number," since it's not a mathematical invariant and is already included in the label
* Fix a bug for code placement of `magma: NilpotencyClass(G);`
* Add closing `</td>` tags and adjust some spacing.


jwj61:
I think the transitive number should be put back in.  It isn't a counter made up by the LMFDB; it is a recognized standard identifier for the action.

roed314:
Ok; I put it back in.

PR #6370, opened 2025-05-14
New dealing with None in database entries

jwj61:
For cases where a database column entry is missing for a given row, it seems we now need to be more careful in testing for that.

http://beta.lmfdb.org/padicField/11.22.22.13

http://127.0.0.1:37777/padicField/11.22.22.13


roed314:
Sorry, I think these are my fault (I changed `include_nones` to be `True` for `lf_fields` to fix another issue).  I'll merge once tests pass.

PR #6372, opened 2025-05-16
Show Aut groups for fields

jwj61:
For both number fields and p-adic fields, this displays the automorphism group and not just the order.

http://127.0.0.1:37777/NumberField/16.0.3571460308796416.1
http://beta.lmfdb.org/NumberField/16.0.3571460308796416.1

http://127.0.0.1:37777/padicField/2.8.8.3
http://beta.lmfdb.org/padicField/2.8.8.3

This keys off of the Galois group.  For p-adic fields, we don't always have the Galois group, but sometimes the group is obvious from its order, so we give it even if we haven't computed the Galois group:
http://127.0.0.1:37777/padicField/7.21.21.10

As a last resort, we just display the order:
http://127.0.0.1:37777/padicField/2.22.22.274

When we have the group, we don't display its order separately since it is almost always (maybe always) completely obvious from the group name (most are cyclic).  In the rare case it is not already clear, the group is a knowl so clicking will reveal its order.

### DIFF
--- a/lmfdb/galois_groups/transitive_group.py
+++ b/lmfdb/galois_groups/transitive_group.py
@@ -155,7 +155,7 @@ class WebGaloisGroup:
         return gens
 
     def aut_knowl(self):
-        return abstract_group_display_knowl(f"{self._data['auts']}.{self._data['aut_sg_no']}")
+        return abstract_group_display_knowl(self._data['aut_label'])
 
     def gapgroupnt(self):
         if int(self.n()) == 1:
--- a/lmfdb/galois_groups/transitive_group.py
+++ b/lmfdb/galois_groups/transitive_group.py
@@ -154,6 +154,9 @@ class WebGaloisGroup:
         gens = ', '.join(gens)
         return gens
 
+    def aut_knowl(self):
+        return abstract_group_display_knowl(f"{self._data['auts']}.{self._data['aut_sg_no']}")
+
     def gapgroupnt(self):
         if int(self.n()) == 1:
             G = libgap.SmallGroup(1, 1)
--- a/lmfdb/number_fields/number_field.py
+++ b/lmfdb/number_fields/number_field.py
@@ -27,6 +27,7 @@ from lmfdb.galois_groups.transitive_group import (
     group_cclasses_knowl_guts, group_pretty_and_nTj, knowl_cache,
     group_character_table_knowl_guts, group_alias_table,
     dihedral_gal, dihedral_ngal, multiquad)
+from lmfdb.groups.abstract.main import abstract_group_display_knowl
 from lmfdb.number_fields import nf_page, nf_logger
 from lmfdb.number_fields.web_number_field import (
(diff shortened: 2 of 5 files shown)


lmfdb/galois_groups/transitive_group.py
```diff
@@ -154,6 +154,9 @@ def generator_string(self):
         gens = ', '.join(gens)
         return gens
 
+    def aut_knowl(self):
+        return abstract_group_display_knowl(f"{self._data['auts']}.{self._data['aut_sg_no']}")
```
roed314, line 158: Why not use `abstract_group_display_knowl(self._data['aut_label'])`?

lmfdb/number_fields/number_field.py
```diff
@@ -447,7 +448,8 @@ def render_field_webpage(args):
             factored_conductor = factor_base_factorization_latex(factored_conductor, cutoff=30)
             data['conductor'] = r"\(%s=%s\)" % (str(data['conductor']), factored_conductor)
     data['galois_group'] = group_pretty_and_nTj(n,t,True)
-    data['auts'] = db.gps_transitive.lookup(r'{}T{}'.format(n,t))['auts']
+    ggdata = db.gps_transitive.lookup(rf'{n}T{t}')
+    data['aut_gp_knowl'] = abstract_group_display_knowl(f"{ggdata['auts']}.{ggdata['aut_sg_no']}")
```
roed314, line 452: Similarly here, I think you can do
```python
data['aut_gp_knowl'] = abstract_group_display_knowl(db.gps_transitive.lookup(f'{n}T{t}', 'aut_label'))
```

roed314:
Some comments:
* I think Aut(K/Q) should have a knowl
* For number fields, I find it a bit strange that Aut(K/Q) is so far separated from Gal(K/Q).  What do you think about either duplicating the Galois group in the basic invariants (just the knowl, not the other information like conjugacy classes or character table), or moving Aut(K/Q) down into the Galois group section (and maybe renaming it)?
* For p-adic fields, I find it a bit less awkward, maybe because of the different section headings and (at least on some example pages) the fact that the Galois section is a bit closer to the top.  But I'd still be fine duplicating the Galois group knowl in the basic invariants.

jwj61:
On the database fields being used, I overlooked that aut_label was there -- changed to use aut_label.

I agree that Aut should be a knowl, and now it is for both number fields and p-adic fields.

In the case that the current field is Galois, it now says Aut(K/Q)=Gal(K/Q): instead of just Gal(K/Q), which was always awkward since the knowl describes a subgroup of S_n and it is given as an abstract group.  Adding Aut(K/Q) should make it more clear that we are trying to give the automorphism group.

I am not a fan of moving the Galois group up in general or duplicating it, so it deserves a wider discussion at least.  In any case, that seems orthogonal to the current changes.

jwj61:
Also changed the summary line on the index page to reflect new fields.

PR #6375, opened 2025-05-28
Updated p-adic section

roed314:
This PR is the result of several years work growing out of an AIM SQuaRE.  Details are described in [this paper](https://math.mit.edu/~roed/writings/research/padic_families.pdf), and the changes are testable on [olive](https://olive.lmfdb.xyz/padicField/).  Major changes include

* A factor of 55 increase in the number of p-adic fields in the LMFDB, which now includes all fields with degree less than 24 and p less than 200 (see the counts [here](https://olive.lmfdb.xyz/padicField/?search_type=Counts) and on the browse page).
* Grouping p-adic fields into families, and new search pages for both absolute and relative families.  There are far fewer families than fields, allowing information to be included up to degree 47 and relative extensions to be described.
* New invariants: rams, means, jump sets, roots of unity contained in the field, residual polynomials, associated inertia, indices of inseparability
* Family pages give generic polynomials (whose specialization produces fields in the family) along with an overview of the quantities that are constant and varying across the family.
* New field labels which better present the structure of the fields.
* Diagrams on family pages describing how the generic polynomial is constructed and how several invariants are related.
* Count search pages giving views on how the number of p-adic families/fields varies with certain inputs (p vs n, n vs e for fixed p, e vs c for fixed p and n).  These include counts of fields outside the range where all fields are stored, using formulas due to Monge.

@jwj61 was involved in this project, so it would be good to get an independent review.


edgarcosta:
I looked LMFDB Editors Coverage Sheet, and added Kiran accordingly, and removed myself.
I looked at some random pages and so far my comments are:
- the ramification polygon hard to read, e.g., https://olive.lmfdb.xyz/padicField/2.1.16.76b1.4971 
- the two tables in https://olive.lmfdb.xyz/padicField/ need to have a separate header, e.g. https://olive.lmfdb.xyz/ModularForm/GL2/Q/holomorphic/
- the sentence at top is quite long, I would skip the completeness part and leave that for the knowl.
- we should homogenize the language under the examples for search, but here are some things I would do:
  - remove "a range like "
  - the doubled "e.g."
  - drop the "[8,3]" for the groups
  - do we need e.g? (https://olive.lmfdb.xyz/ModularForm/GL2/Q/holomorphic/ doesn't have it for Projective image)

assaferan:
This looks great! 
In my opinion the only thing that is important to do before merging this is to update the [Source and Acknowledgments](https://olive.lmfdb.xyz/padicField/Source) and [Labels](https://olive.lmfdb.xyz/padicField/Labels) pages, as they still refer to the previous version.
I have some further comments and remarks below, but they are all minor and could also be done after merging - 
1. There are various related quantities (Galois groups, hidden slope content) that were not computed for some of the entries, and these appear in search results as "not computed" (see e.g. https://olive.lmfdb.xyz/padicField/?p=2&n=22). We should try to have a uniform handling of this - for example in https://www.lmfdb.org/ModularForm/GL2/Q/holomorphic/?level=100003 you can see that we simply do not display the data if it was not computed. Another option might be to have an agreed symbol such as N/A , -, or ? to indicate that something has not been computed.
2. Slopes - currently, it is very difficult to navigate between the different definitions of slopes. Here are some examples:
    a) There is no option to search by Swan slopes, by means or by rams
    b) The search field "Galois Artin slopes" has a knowl that refers to them as "hidden Artin slopes" and explains at the same time about Artin and Swan slopes. It would be better if the knowl had the same terminology as the search field.
    c) The knowl for visible slopes also includes data on the wild slopes, although these do not appear anywhere on the page.
    d) On the same note, the paper makes a good point towards using Swan slopes, and yet our default view uses Artin slopes. 
3. Subfamilies - this might be my misunderstanding, but it seems that the family [3.1.9.18b](https://olive.lmfdb.xyz/padicField/family/3.1.9.18b) should have four subfamilies (as in Table 4.1 of your paper, but also the equivalence class of residual polynomials seems to be {(x^3+a3, a3x+a10), (x^3-a3, a3x-a10)} for which there are four options) while the labels suggest only two. It would also help if the family page would include the relevant data, e.g, the residual polynomials.
4. The search box for indices of separability does not fit well on the main page (ind.    of), maybe not abbreviate.
5. Statistics - this might be a general problem with the statistics page, and not particular to p-adic fields, but it is very annoying here, as the [count table](https://olive.lmfdb.xyz/padicField/?search_type=Counts) goes beyond the page boundaries.  
6. In the knowl for ramification polynomial, the quotient by alpha^n looks like an ideal, consider changing to multiplication by alpha^{-n}.
7.  The Varying parameters in a family page could be a bit too much (see e.g. Galois groups and Hidden slopes in https://olive.lmfdb.xyz/padicField/family/2.1.16.79a). Consider showing only the first few, and expand when clicking. Also, here the data might be incomplete, which at the moment shows in the end. I think it would be better to present it near the name of the field (in the left hand side of the table). 
8. When viewing the column packet (e.g. [here](https://olive.lmfdb.xyz/padicField/family/2.1.16.79a?hst=None&one_per=packet)), we display the packet size. This is a good choice, but it should be reflected to the user, who might think this is some id number.

roed314:
We realized that the defining polynomials chosen are not specializations of the generic polynomials.  Closing this PR for now until this is addressed on the data side.

roed314:
The defining polynomials have been changed, so I'm reopening.

PR #6381, opened 2025-06-06
Draw spectrum of ring of integers on number field pages

havarddj:
Draw SVG picture of ring of integers in the property box and at the bottom of the page.

Screenshot 2025-06-06 at 22 16 00

Screenshot 2025-06-06 at 22 16 57


havarddj:
Thanks, I've fixed those now.

jwj61:
I think someone with proper access has to add svg to sage on legendre.

roed314:
I've done `sage -pip install svg.py` on both grace and prod.

roed314:
Ah, I was confused why you needed it there.  I've now installed it, but for future reference you can do something like `sage -pip install --user svg.py` to install it just for yourself on legendre (this is no longer needed since I installed it globally).

havarddj:
@edgarcosta It seems like the branch was merged already. I'm happy to open a new PR, but since it's just for a test function which is not used in prod, we could also just leave it. Preference?

roed314:
I think another PR would be fine.

PR #6384, opened 2025-06-09
use lazy series for Siegel dimensions

fchapoton:
as this simplifies things


fchapoton:
hmm, not clear to me where the failures can come from. Which version of sage do you use ?

edgarcosta:
> As a side note, I think all these formulas for the dimensions of Siegel modular forms should rather be inside sage.

I fully agree!

roed314:
I don't have time today to look at #6386, but I'll try to help sometime this week.

PR #6385, opened 2025-06-09
more checks and faster with ruff

fchapoton:
and fixing the few failures for these new checks


roed314:
I'll merge once tests pass

PR #6386, opened 2025-06-09
use a more recent sage

fchapoton:
let us try with sage 10.4

### DIFF
--- a/.environment.yml
+++ b/.environment.yml
@@ -4,15 +4,14 @@
 ##    conda env create -f environment.yml
 ##
 ## See https://towardsdatascience.com/a-guide-to-conda-environments-bc6180fc533
-## or    https://docs.conda.io/projects/conda/en/latest/user-guide/tasks/manage-environments.html
+## or  https://docs.conda.io/projects/conda/en/latest/user-guide/tasks/manage-environments.html
 name: lmfdb
-channel_priority: strict
-safety_checks: disabled
 channels:
   - conda-forge
+  - defaults
 dependencies:
-  - python=3.10.5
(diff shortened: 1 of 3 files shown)


fchapoton:
and it says

```
EnvironmentSectionNotValid: The following sections on '/home/runner/work/lmfdb/lmfdb/.environment.yml' are invalid and will be ignored:
 - channel_priority
 - safety_checks
Channels:
 - conda-forge
 - defaults
Platform: linux-64
Collecting package metadata (repodata.json): - \ | / - \ | / - \ | / - \ | / - \ done
Solving environment: / - \ failed
LibMambaUnsatisfiableError: Encountered problems while solving:
  - unsupported request
  - package python-3.11.13-h1a3bd86_0 is excluded by strict repo priority
```

fchapoton:
ok, I have made progress, but this may hurt the cacheing process

roed314:
Thanks, that looked like a lot of work.  Can you share why we need to set `use-only-tar-bz2` to false?

fchapoton:
I have now squashed all these experimental commits.

I have no clear idea why this `use-only-tar-bz2` is preventing things to work. Maybe some package provides only a `.conda` archive ? I am definitely a conda newbie.

edgarcosta:
I am not sure is worth to pin the python version. As we use whatever sage forces us.

fchapoton:
So let's keep the python version floating.

I have found a discussion about the caching : https://github.com/conda-incubator/setup-miniconda/issues/267

but the conclusions are not clear.

and see also https://github.com/embray/gappy/pull/27

PR #6394, opened 2025-06-11
some cleanup in Siegel code

fchapoton:
just slightly less terrible code, for the moment


roed314:
Thanks; that's certainly a big improvement.  I'll try to review later today.

PR #6396, opened 2025-06-12
Fix uncomputed Schur multiplier

roed314:
Compare https://beta.lmfdb.org/Groups/Abstract/20922789888000.a and http://localhost:37777/Groups/Abstract/20922789888000.a


jenpaulhus:
Looks good. Will merge when tests are done.

PR #6401, opened 2025-06-14
Search by family for groups

SamSchiavone:
1) Search and browse by family.
2) Clarify nonsplit products (where we know product is nonsplit) vs. possibly nonsplit products (where we just haven't computed whether there's a complement)
3) Add "Minimal" to transitive and permutation degree


SamSchiavone:
I tried running the tests locally, but they were taking forever. I think this is due to the current problems with the subgroup search.

roed314:
@SamSchiavone, if you're happy with my changes feel free to merge.  Otherwise we can talk during our meeting.

PR #6406, opened 2025-06-16
main -> dev

edgarcosta:


roed314:
There is a persistent error that only seems to occur when running against proddb (in the genus 2 curve tests), and @edgarcosta and I can't figure out where it's coming from.  Since it won't affect beta, I'm merging this PR (there are a bunch of improvements we want to see go live on beta), but we'll try to track it down soon.

Issue #6411, opened 2025-06-18
Broken groups family test


edgarcosta:
This test is failing:
https://github.com/LMFDB/lmfdb/blob/646963d28b4e823112bcaea8925ae8a846cddd4c/lmfdb/groups/abstract/test_browse_page.py#L242

see
https://github.com/LMFDB/lmfdb/actions/runs/15744974163/job/44379180526?pr=6409#step:11:199

and I can also see 16.11 here 
https://beta.lmfdb.org/Groups/Abstract/?family=any

PR #6412, opened 2025-06-18
Fix failing groups test

SamSchiavone:
Addressing https://github.com/LMFDB/lmfdb/issues/6411


edgarcosta:
Thank you! I simply didn't know how to fix the test quickly.

PR #6413, opened 2025-06-19
Family browse

roed314:
Reduces the number of groups show in the "browse by family section" and fixes some display issues.  The difference can be seen by comparing https://red.lmfdb.xyz/Groups/Abstract/ and https://beta.lmfdb.org/Groups/Abstract/, though red also includes some other changes that will not be merged here.


SamSchiavone:
Looks good to me. I didn't test all the new family searches, but the few that I tried seem to give the right results.

We should probably add more tests [here](https://github.com/roed314/lmfdb/blob/family_browse/lmfdb/groups/abstract/test_browse_page.py#L220). I wonder if there's a way to get ChatGPT or something to automatically make them.

Thanks for updating the [group families knowl](https://red.lmfdb.xyz/knowledge/show/group.families), too. At some point we may want to add more explanation for what the various families are, or refer to a reference whose conventions we follow. I know that for some of the groups of Lie type there are differing conventions on which quadratic form is used.

PR #6416, opened 2025-06-20
Major modular curves refactor

BarinderBanwait:
As a result of various backend changes on the tables (mainly `gps_gl2zhat_fine` and `gps_gl2zhat_coarse` being replaced with `gps_gl2zhat`), the modular curves code needs to be changed. This PR updates the repo accordingly. Some verification testing and some unit testing has been added (this testing is to be considered just a start).


AndrewVSutherland:
We should hold off merging this until `gps_gl2zhat` has been updated (which should fix the test failures).

PR #6419, opened 2025-06-20
improve asserts

edgarcosta:


edgarcosta:
Also resolves #6420

roed314:
Looks good to me; thanks!

PR #6430, opened 2025-06-24
Search based on hash

roed314:
This adds the ability to search by hash value.  Try http://localhost:37777/Groups/Abstract/?hash=5120%234714647875464396655.

### DIFF
--- a/lmfdb/groups/abstract/main.py
+++ b/lmfdb/groups/abstract/main.py
@@ -95,6 +95,8 @@ gap_group_label_regex = re.compile(r"^(\d+)\.(\d+)$")
 abstract_group_label_regex = re.compile(r"^(\d+)\.((\d+)|[a-z]+)$")
 # order_stats_regex = re.compile(r'^(\d+)(\^(\d+))?(,(\d+)\^(\d+))*')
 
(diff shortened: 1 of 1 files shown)


jenpaulhus:
If I'm not missing something, it looks like the Hash search box only shows up as a search option if you type in a URL with "?hash=blah" first. Is that  correct/intentional? Do we want to be able to add hashes as a search option for the Find box?

Might be worth mentioning the "order # hash" notation for searches.  Some folks (i.e. me when it's past my bedtime!) may just try the hash string itself before realizing you need to include the order also.

roed314:
> If I'm not missing something, it looks like the Hash search box only shows up as a search option if you type in a URL with "?hash=blah" first. Is that correct/intentional? Do we want to be able to add hashes as a search option for the Find box?

Yes, that's right.  This PR also makes it possible to enter a hash value in the Find box, with the format `N#hsh`.

> Might be worth mentioning the "order # hash" notation for searches. Some folks (i.e. me when it's past my bedtime!) may just try the hash string itself before realizing you need to include the order also.

Agreed; once we're ready to merge this PR I can update that knowl.

PR #6432, opened 2025-06-26
Add Family name column when family specified

roed314:
For many families, the chosen name of some members of that family is not that of the family (for example, PSL(2,4) and PSL(2,5) are both A5).  When a family is specified, this PR adds a column that displays which members of the family that row is.  It will be live shortly on red; see [here](https://red.lmfdb.xyz/Groups/Abstract/?family=any) for example.


jenpaulhus:
It looks like if we specify the Family in a search, then the Family name column you created no longer gives other names:

https://red.lmfdb.xyz/Groups/Abstract/?family=PSL

PSL(2,3) should have other names with it, for example.

roed314:
I'm not sure what you mean.  What would you like to see for PSL(2,3) in that case?

roed314:
The other option would be to show all the family names, and I think that's going to get ridiculous for some groups that show up at the beginning of search results (most notably S3 and A5).  Here's a screenshot showing it (I can't fit it all on my page): 
Screenshot 2025-06-26 at 14 11 30

Issue #6433, opened 2025-06-26
Error when number field Galois group not computed


roed314:
From the flasklog:
```
Traceback (most recent call last):
  File "/home/sage/sage-10.4/local/var/lib/sage/venv-python3.10/lib/python3.10/site-packages/flask/app.py", line 2073, in wsgi_app
    response = self.full_dispatch_request()
  File "/home/sage/sage-10.4/local/var/lib/sage/venv-python3.10/lib/python3.10/site-packages/flask/app.py", line 1519, in full_dispatch_request
    rv = self.handle_user_exception(e)
  File "/home/sage/sage-10.4/local/var/lib/sage/venv-python3.10/lib/python3.10/site-packages/flask/app.py", line 1517, in full_dispatch_request
    rv = self.dispatch_request()
  File "/home/sage/sage-10.4/local/var/lib/sage/venv-python3.10/lib/python3.10/site-packages/flask/app.py", line 1503, in dispatch_request
    return self.ensure_sync(self.view_functions[rule.endpoint])(**req.view_args)
  File "/home/lmfdb/lmfdb-git-dev/lmfdb/number_fields/number_field.py", line 746, in by_label
    return render_field_webpage({'label': nflabel})
  File "/home/lmfdb/lmfdb-git-dev/lmfdb/number_fields/number_field.py", line 473, in render_field_webpage
    ramified_algebras_data = nf.ramified_algebras_data()
  File "/home/lmfdb/lmfdb-git-dev/lmfdb/number_fields/web_number_field.py", line 997, in ramified_algebras_data
    loc_alg_dict = self.get_local_algebras()
  File "/home/lmfdb/lmfdb-git-dev/lmfdb/number_fields/web_number_field.py", line 986, in get_local_algebras
    transitive_group_display_knowl(LF['galois_label']),
  File "sage/misc/cachefunc.pyx", line 1015, in sage.misc.cachefunc.CachedFunction.__call__
    w = self.f(*args, **kwds)
  File "/home/lmfdb/lmfdb-git-dev/lmfdb/galois_groups/transitive_group.py", line 336, in transitive_group_display_knowl
    n, t = label.split("T")
AttributeError: 'NoneType' object has no attribute 'split'
[2025-06-26 04:59:41 UTC] 500 error on URL https://beta.lmfdb.org/NumberField/44.0.18038797319548302242232245613754598807607865052837142211957640472205046331907996909568.1
```

It seems to only occur on beta for the moment.

[...]

PR #6434, opened 2025-06-26
Deal with local algebras when Galois groups have not been computed

jwj61:
Some new local fields do not have Galois groups and slopes computed which affects number field pages.  This fixes that

http://127.0.0.1:37777/NumberField/44.0.18038797319548302242232245613754598807607865052837142211957640472205046331907996909568.1

The version on beta will just give a server error
http://beta.lmfdb.org/NumberField/44.0.18038797319548302242232245613754598807607865052837142211957640472205046331907996909568.1

### DIFF
--- a/lmfdb/number_fields/number_field.py
+++ b/lmfdb/number_fields/number_field.py
@@ -509,7 +509,11 @@ def render_field_webpage(args):
                         myurl = url_for('local_fields.by_label', label=lab)
                         if mm[3]*mm[2] == 1:
                             lab = r'$\Q_{%s}$' % str(p)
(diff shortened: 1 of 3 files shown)


jwj61:
Should add that this fixes issue #6433

roed314:
Sorry about the linting changes, which occurred in #6385.  I've turned the new linting off, and switched one of the lines back to having no space around the equal sign to check.

jwj61:
Thanks for the changes.  You didn't have to remove the spaces.  I don't care if they are there in the end; I just don't want to have to put them in myself.

Issue #6427, opened 2025-06-23
Parameterized families of number fields with a given Galois group


AndrewVSutherland:
On the Galois group pages, in addition to providing a link to number fields with this Galois group, in cases where a (not necessarily universal) parameterization is known, it would be nice to list known parameterizations on the Galois group page for the benefit of those who would like to be able to generate many examples.

E.g, for 21T2, there are 11 number fields in the LMFDB, but there is a parameterized family that can be found in the "Generic Polynomials" book by Jensen-Ledet-Yui `(https://books.google.com/books/about/Generic_Polynomials.html).

jwj61:
Wouldn't we also want examples of generic polynomials as well in cases where they are known?

As far as other parameterizations, many of these will come from Belyi maps.  Do we list these as well?

In all cases, there may be restrictions on the base field which would need to be noted (e.g. C_3 Belyi maps in our database don't give C_3 fields over Q, and some polynomials in that book work for say fields of characteristic different from 2).

If these are given for a relatively small percentage of the Galois groups, should it be regular database entries or done through bottom knowls?

roed314:
See also the [lecture notes](https://people.maths.bris.ac.uk/~matyd/InvGal/) from Tim Dokchitser's 2021 PCMI course, which includes a magma package for producing a bunch of regular families.

jwj61:
There are also the polynomials in at the end of the book by Malle and Matzat.  A problem would be typing them all in.  Dokchitser's package would certainly help with some of that but doesn't go as far.

jwj61:
Here is another variation.  For C_5, Jensen et. al. has a generic polynomial for characteristic 0.  There is a paper by Gene W Smith on cyclic groups of odd order which probably expands that, but I don't have access to the paper.

roed314:
> On None, I was thinking of [-1] to mean no primes (or the roles of None and [-1] could be switched).

We could also use `[]` to mean no primes and `None` to mean not computed.

roed314:
> Here is another variation. For C_5, Jensen et. al. has a generic polynomial for characteristic 0. There is a paper by Gene W Smith on cyclic groups of odd order which probably expands that, but I don't have access to the paper.

I tried to find this paper, and found [this](https://www.researchgate.net/publication/344497632_On_simple_families_of_cyclic_polynomials) on researchgate but then I couldn't find it on the AMS website.

roed314:
[...]

PR #6442, opened 2025-06-28
Regular families

jwj61:
This addresses issue #6427: it adds the ability to have polynomials for parameterized families of fields on Galois group pages.  There is currently data for degrees through 10, some in degree 11, and a few scattered groups after that.  To see an example,

http://127.0.0.1:37777/GaloisGroup/5T4

which is one of the few cases where it gives more than one family, and shows that individual polynomials can be marked as generic.

http://127.0.0.1:37777/GaloisGroup/4T3

shows that it can display that a family is generic for fields except in certain characteristics.

For a particularly complicated polynomial, view M24

http://127.0.0.1:37777/GaloisGroup/24T24680

In all cases, the user can use the raw/typeset feature to copy polynomial for a computer algebra system.

### DIFF
--- a/lmfdb/galois_groups/templates/gg-show-group.html
+++ b/lmfdb/galois_groups/templates/gg-show-group.html
@@ -118,7 +118,6 @@ table.reptable td, table.reptable th {
 
   <p>{{place_code('char_table')}}</p>
 
-
 {% if info.int_reps %}
 <h2>{{ KNOWL('gg.int_modules', title='Indecomposable integral representations') }}</h2>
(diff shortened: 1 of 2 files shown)


roed314:
Looks great to me!  I made a few code tweaks that won't affect how anything is displayed.  I'm happy to merge once tests pass, though perhaps it's worth giving other people a chance to take a look and make comments.

roed314:
FYI, I just copied `gps_regular_polynomials` to prod.

PR #6456, opened 2025-07-04
Improved Belyi plane equations

SamSchiavone:
Added new plane models for Belyi maps to the table `db.belyi_galmaps_new`. If everything looks okay, I can replace `db.belyi_galmaps` with this new table.

I also deactivated the embedded Belyi map pages for the moment. @roed314 pointed out that they are not rendering correctly. See [this one](https://beta.lmfdb.org/Belyi/4T5/4/4/3.1/a/(1,2,3,4),(1,3,4,2),(1,3,4)/) for instance; they all seem to have this problem. I think this bug probably started when we switched to using `raw_typeset` and `compress_expression` to display the equations.


AndrewVSutherland:
This branch is now running on https://purple.lmfdb.xyz/Belyi/ for anyone who wants to take a look.

SamSchiavone:
Great, thanks. I'll reload `db.belyi_galmaps` and update the PR.

PR #6468, opened 2025-07-05
Cyclic subgroups not in DB

jenpaulhus:
**Don't merge this until the groups data has been updated.**

This PR should fix errors for two subgroups which we have identified as cyclic but don't have that cyclic subgroup in the DB. 
```
6412032.a.528._.B 
6412032.a.1584._.BJ
```

My fix is pretty inelegant but it works!


jenpaulhus:
This should wait until the other PR is merged.   The two subgroups listed above should not throw errors anymore.

jenpaulhus:
Ok. This should be ready to be reviewed now.   

This is a workaround for 2 subgroups in the DB which we know are cyclic but for which the corresponding cyclic subgroup doesn't exist in our DB.   The beta version of these pages throw an error, and all other subgroup pages should be unaffected.

http://localhost:37777/Groups/Abstract/sub/6412032.a.528._.B
http://beta.lmfdb.org/Groups/Abstract/sub/6412032.a.528._.B

http://localhost:37777/Groups/Abstract/sub/6412032.a.1584._.BJ
http://beta.lmfdb.org/Groups/Abstract/sub/6412032.a.1584._.BJ

jenpaulhus:
Should be fixed now.  Also, I'm realizing it's probably better to check red.lmfdb.xyz and those two subgroups as the labels are still the old labels on beta.

PR #6482, opened 2025-07-09
add a find input above properties box

roed314:
An early attempt done live at ICERM; take a look and give us some feedback.


jwj61:
It is awkward that the spin-arrow is on top of the find button.

I thought this would only be on index pages for different areas, but it is showing up on search result and object homepages.

roed314:
Agreed; I decided that I didn't want to fix that live in front of the room.

Also, I need to make it so that this only shows up on browse pages.

roed314:
This is not ready; I'll work on it and resubmit.

Issue #6399, opened 2025-06-13
Seach by Sylow properties


roed314:
There are a couple improvements possible for searching by Sylow subgorup

1. On the subgroup search pages, it would be helpful to be able to require the result be a Sylow subgroup, and to show the prime in a column if it is.
2. On group search pages, there are a couple properties that might be relevant to search by: whether the Sylow subgroup at a particular prime is normal, and whether the complement is normal.

Squirrelcoding:
Hello, I’m a high school student with a strong interest in math (I don't even know how I stumbled upon this project), and I’d love to work on part 2. Right now, I'm currently getting SageMath set up (3+ hours into trying to build it) and digging into the relevant parts of the codebase. Sylow subgroups are one area I’m pretty comfortable with, so this seems like a great place to start contributing.

Just wanted to let you know I’m planning to take this on and am happy to coordinate if there are any existing plans!

PR #6484, opened 2025-07-09
Add Sylow subgroup column

MarkBcc168:
Allow search for Sylow subgroups and add column displaying which prime (part 1 of #6399).

### DIFF
--- a/lmfdb/groups/abstract/main.py
+++ b/lmfdb/groups/abstract/main.py
@@ -1254,6 +1254,7 @@ subgroup_columns = SearchColumns([
                           display_url,
                           short_title="Sub. name", apply_download=False),
         ProcessedCol("subgroup_order", "group.order", "Order", show_factor, align="center", short_title="Sub. order"),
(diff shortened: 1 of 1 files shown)


lmfdb/groups/abstract/main.py
```diff
@@ -1261,15 +1261,17 @@ def group_parse(info, query):
         CheckCol("solvable", "group.solvable", "solv", short_title="Sub. solvable"),
         CheckCol("maximal", "group.maximal_subgroup", "max", short_title="Sub. maximal"),
         CheckCol("perfect", "group.perfect", "perf", short_title="Sub. perfect"),
-        CheckCol("central", "group.central", "cent", short_title="Sub. central")]),
+        CheckCol("central", "group.central", "cent", short_title="Sub. central"),
+        ProcessedCol("sylow", "group.sylow_subgroup", "Sylow", lambda x: f"${latex(x)}$" if x > 1 else "", align="center", short_title="Sub. Sylow"),
```
jenpaulhus: Please move this line (which gives us the Sylow column) up a few rows so the Sylow column shows up immediately after the Order column.  Everything else looks great.

Issue #6056, opened 2024-05-18
Better checking on level_type


roed314:
Currently, http://www.lmfdb.org/ModularForm/GL2/Q/holomorphic/?level_type=powerful gives a server error (though this will be resolved once @AndrewVSutherland pushes sever modular forms tables to production this evening).  But even after that, I believe that http://www.lmfdb.org/ModularForm/GL2/Q/holomorphic/?level_type=banana will still give a server error.  We should check that the column name generated [here](https://github.com/LMFDB/lmfdb/blob/main/lmfdb/classical_modular_forms/main.py#L752) is valid.

PR #6490, opened 2025-07-09
Adding a check for level types 

lewismcombes:
Level types in search for classical modular forms now checks the string provided and gives an error if this doesn't match something in the table.


jenpaulhus:
This fixes #6056.

PR #6494, opened 2025-07-09
catch exception for invalid level_type

### LINKED ISSUE #6056: Better checking on level_type

StevenClontz:
Closes #6056

### DIFF
--- a/lmfdb/classical_modular_forms/main.py
+++ b/lmfdb/classical_modular_forms/main.py
@@ -169,7 +169,11 @@ def index():
         info['search_type'] = search_type = info.get('search_type', info.get('hst', ''))
 
         if search_type in ['List', '', 'Random']:
-            return newform_search(info)
+            try:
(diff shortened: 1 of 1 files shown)


jenpaulhus:
Im going to close this without merging because it is a duplicate of #6494.  But thank you!

StevenClontz:
Dupe of #6490 *

No worries. :-D

PR #6504, opened 2025-07-11
Few more abstract group fixes

jenpaulhus:
This fixes an issue with downloads where if a boolean was "none", it would record it as such in the download instead of disregarding.
https://red.lmfdb.xyz/Groups/Abstract/100000000.bg/download/magma
will fail to load in magma, but this works:
http://localhost:37777/Groups/Abstract/100000000.bg/download/magma

It adds auto_gens pages back for all Lie groups except those whose names start with a "P"  (i.e. the ones that are represented by quotients in the data currently), and for all other matrix groups.
http://localhost:37777/Groups/Abstract/4080.a
https://beta.lmfdb.org/Groups/Abstract/4080.a

It adds generators of subgroups when the ambient group was Lie type with a name that started with "P". 
http://localhost:37771/Groups/Abstract/sub/6072.a.506.b1.a2
http://red.lmfdb.xyz/Groups/Abstract/sub/6072.a.506.b1.a2

It removes magma downloads of character tables from Lie groups since we don't yet have functionality in decode() to create elements of a Lie group.

### DIFF
--- a/lmfdb/groups/abstract/main.py
+++ b/lmfdb/groups/abstract/main.py
@@ -2096,20 +2096,12 @@ def download_boolean_string(G,dltype,ul_label):
     else:
         return ""
 
-    s += "Agroup := " + str(G.Agroup).lower() + ", \n"
-    s += "Zgroup := " +	str(G.Zgroup).lower() + ", \n"
-    s += "abelian := " + str(G.abelian).lower() + ", \n"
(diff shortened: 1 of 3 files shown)


jwj61:
I don't see the difference for generators of a subgroup.  They seem to be on both pages.

jenpaulhus:
We're getting that inner by out of Gap so I'm guessing we need to change the output from GAP to display it like the others.

PR #6505, opened 2025-07-11
Add lady Gaga pictures for Galois groups

jwj61:
For some groups we get the new picture

http://beta.lmfdb.org/GaloisGroup/25T50
http://127.0.0.1:37777/GaloisGroup/25T50

There is an entry in learnmore for a Galois group home page explaining the picture.

In some cases, we have not computed the picture (yet).  Then the user gets a picture of Galois the person (taken from wikipedia which lists it as public domain) instead.

http://127.0.0.1:37777/GaloisGroup/30T7
http://beta.lmfdb.org/GaloisGroup/30T7


AndrewVSutherland:
I love this!.  But I think the picture should be centered in the properties box.

jwj61:
Tried to center diagrams and "photo".  I think diagrams works reasonably well.  The photo doesn't move in the one I was checking, but probably won't without some tricks.  It sits in a table inside that portion of the properties box.  It is centered in the table, but another part of the properties box is pushing it wider.

PR #6509, opened 2025-07-11
Add toy example of Integers for LMFDB novices

StevenClontz:
Ready for comment.

I think there's a wide audience of people, students, non-mathematicians, and even some mathematicians, who will not appreciate what LMFDB does in the abstract from the current homepage. And I definitely think that having a sandbox with negligible mathematical prerequisties would be helpful for the community's tutorials and trainings.

This PR implements a toy "Integers" page, along with a few toy features. I'd like to elaborate on this or something similar for a future chapter of https://g4m.code4math.org/g4m.html While @roed314 proposed having this on a branch, which could be updated from main every so often, I think having this on main would reduce friction from its use.

To keep this from appearing in production, I implemented the requirement of setting the environment variable of `LMFDB_TUTORIAL` to `on` for the Integers to appear in the LMFDB, and it's not included in the allow-list of URLs (which seems to be ignored in debug mode).

### DIFF
--- a/lmfdb/app.py
+++ b/lmfdb/app.py
@@ -714,7 +714,7 @@ def sitemap():
 
 @cached_function
 def WhiteListedRoutes():
-    routes = [
+    return [
         'ArtinRepresentation',
         'Character/Dirichlet',
         'Character/calc-gauss/Dirichlet',
@@ -786,7 +786,6 @@ def WhiteListedRoutes():
         'whitelistedsitemap',
         'zeros/zeta'
     ]
-    return routes
 
 
 @cached_function


lmfdb/app.py
```diff
@@ -710,7 +714,7 @@ def sitemap():
 
 @cached_function
 def WhiteListedRoutes():
-    return [
+    routes = [
```
StevenClontz:
```suggestion
    return [
```

lmfdb/app.py
```diff
@@ -782,6 +786,7 @@ def WhiteListedRoutes():
         'whitelistedsitemap',
         'zeros/zeta'
     ]
+    return routes
```
StevenClontz:
```suggestion
```

Issue #6060, opened 2024-05-20
Add code snippet/download to select the correct newform to CMF pages


AndrewVSutherland:
The code snippets/downloads for PariGP/SageMath/Magma that appear on newform pages currently split the relevant newspace and then include a line that simply grabs the first newform in the list returned (with a comment that the index might be incorrect).  The ordering of the newforms returned doesn't necessarily match the ordering in the LMFDB and there is no way for us to know ahead of time what the correct index is (indeed, in some cases it is non-deterministic).

However, for each newspace we know exactly how many traces are needed to distinguish all the newforms in the space, and the traces of the newform in the page, so it would be easy to add a short block of code that selects the correct newform by checking traces.  In many cases, the dimension (the trace of a_1(f)) is already enough and in any case this should be very quick and will save the user the hassle of doing this themselves (and the confusion of potentially not realizing they need to do it).

For example, for the newform [80.5.h.b](https://www.lmfdb.org/ModularForm/GL2/Q/holomorphic/80/5/h/b/) mentioned in #6059, the following Magma code can be used to determine the correct index:

```
traces := [2,0,0,0,14]; // traces for 80.5.h.b up to the trace_bound of the newspace 80.5.h
chi := DirichletCharacter("80.79");
S:= CuspForms(chi, 5);
N := Newforms(S);
i := [i:i in [1..#N]|[ZZ|Trace(Coefficient(N[i][1],j)):j in [1..#traces]] eq traces];
f := N[i]; // Galois orbit 80.5.h
```

AndrewVSutherland:
It would be great if someone who knows PariGP (e.g. @pascalmolin) and SageMath (e.g. @roed314) could give similar snippets for PariGP and SageMath.

These should  be added to the code downloads, but might be too much to put into the snippets.  We could just have the snippet be a comment that says "click the Download to blah link to get code that selects this newform".  I think that would be better than having a line that uses what is likely the wrong index (this would avoid the confusion that led to #6059).

Issue #5403, opened 2023-03-10
CMF code snippets don't choose the correct form


AndrewVSutherland:
The Sage and Pari code snippets for CMF pages currently just pick the first newform in the space (in whatever order Sage or Pari decide to use).  They should instead pick the form with the correct traces (and the snippets can use the trace bound for the newspace which is stored in the database to figure out the right number of traces to check, which will typically be less than 10).  This could be done in list comprehension to keep the snippet short.

We could also do this in the Magma as an alternative to using Hecke kernels (sadly, using the Hecke kernel can be slower than decomposing the whole space and picking the form with the right traces).

roed314:
This was done for Sage in #6515, and now needs to be done for pari and magma.

PR #6515, opened 2025-07-11
Added Sage code to select the correct newform on CMF pages

PAMarcus:
Addresses #6060 and #5403 for Sage. Not fully tested for large spaces which take long to compute the newforms.


roed314:
Looks good to me.

PR #6519, opened 2025-07-14
Fix trivial download

jenpaulhus:
The trivial group needed its own special treatment for magma and gap downloads.

http://localhost:37777/Groups/Abstract/1.1/download/magma
http://beta.lmfdb.org/Groups/Abstract/1.1/download/magma

http://localhost:37777/Groups/Abstract/1.1/download/gap
http://beta.lmfdb.org/Groups/Abstract/1.1/download/gap

### DIFF
--- a/lmfdb/groups/abstract/main.py
+++ b/lmfdb/groups/abstract/main.py
@@ -2255,6 +2255,18 @@ def download_char_table(G,dltype,ul_label):  # G is web abstract group
         return ""
 
 
+def download_trivial_construction(dltype):  #trival gp construction is different
+    if dltype == "gap":
+        s = "GPC := TrivialGroup(); \n"
+        s += "GPerm := SymmetricGroup(1); \n"
+    elif dltype == "magma":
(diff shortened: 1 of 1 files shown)


lmfdb/groups/abstract/main.py
```diff
@@ -2283,7 +2295,10 @@ def download_group(**args):
     s = com1 + " Group " + label + " downloaded from the LMFDB on %s." % (mydate) + " " + com2
     s += "\n \n"
 
-    if wag.complex_characters_known is False:
+
+    if label == "1.1":
+        cc_known = False
+    elif wag.complex_characters_known is (False or None):
```
roed314: This won't work the way you expect: `False or None` is `None`.  You might want `in [False, None]` or `wag.complex_characters_known is False or wag.complex_characters_known is None`

roed314 (changes requested):

lmfdb/groups/abstract/main.py
jenpaulhus: Fixed.

PR #6141, opened 2024-08-13
Update code snippets on group pages + add copy functionality

jenpaulhus:
We implement code snippets for matrix groups over Fq (and over Fp for GAP).

For example:
12240.b now has a code snippet in Magma and GAP for the matrix group construction for this group.
Same for 58806.h

12.4 now has a code snippet in GAP for matrix group / F3

Also, thanks to @edgarcosta we add a copy button at the beginning of all the snippets, which could be implemented on other pages as well.

PR #6532, opened 2025-07-14
New codeblocks for all pages

havarddj:
Work in progress, feedback appreciated! 

All the "place_code" functionality is moved from the jinja macro in homepage.html to a new class in utils/place_code.py. This includes copying functionality, and more reasonable defaults for selecting the code.

Here's a preview: 
Screenshot 2025-07-14 at 18 01 48

A lot of the code is inspired by the version for the groups page by @jenpaulhus and @edgarcosta in #6141.

### DIFF
--- a/lmfdb/utils/place_code.py
+++ b/lmfdb/utils/place_code.py
@@ -49,7 +49,7 @@ class CodeSnippet():
 
     def show_commands_box(self):
         """Display 'Show commands' box and corresponding logic"""
-        lang_names = {"pari": "PariGP", "sage": "SageMath", "magma": "Magma", "oscar": "Oscar"}
+        lang_names = {"pari": "PariGP", "sage": "SageMath", "magma": "Magma", "oscar": "Oscar", "gap": "Gap"}
         box_str = r"""<div align="right" style="float: right; margin-top:2px;">""" + "Show commands: "
         lang_strs = []
         for lang in self.langs:
--- a/lmfdb/utils/place_code.py
+++ b/lmfdb/utils/place_code.py
@@ -18,19 +18,19 @@ class CodeSnippet():
         """
         self.code = code
         self.item = item
-        if 'prompt' in code.keys():
-            self.langs = sorted(list(code['prompt'].keys()))
-        elif 'show' in code.keys():
-            self.langs = sorted(list(code['show'].keys()))
-
+        if 'prompt' in code:
+            self.langs = sorted(code['prompt'])
+        elif 'show' in code:
+            self.langs = sorted(code['show'])
         else:
+            # for backwards compatibility
             self.langs = []
         
         self.pre, self.post = pre, post
 
         # edit these when adding support for more languages 
         self.comments = {'magma': '//', 'sage': '#',
-                         'gp': '\\\\', 'pari': '\\\\', 'oscar': '#'}
+                         'gp': '\\\\', 'pari': '\\\\', 'oscar': '#', 'gap': '#'}
         self.full_names = {"pari": "Pari/GP", "sage": "SageMath", "magma": "Magma", "oscar": "Oscar", "gap": "Gap"}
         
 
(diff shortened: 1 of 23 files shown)


edgarcosta:
it is now live on: https://blue.lmfdb.xyz/

jenpaulhus:
This looks good to me!

havarddj:
Here are some notes on the features of this PR: 
- It moves most of the functionality for the code blocks and "Code to LANG" in the properties box to the utils file `place_code.py` (this could perhaps have a better name). Code blocks were originally defined by a jinja macro, and quite basic. Now they're more consistent across pages, and can be modified more easily.
- For classical modular forms, exporting code looks quite complicated, so it is not touched by the PR. For (most) of the other pages:
- Code for the "frontmatter" of the "Code to LANG" file generated is now moved to the relevant yaml file, instead of having each main.py implement its own frontmatter builder.
- In a similar vein, previously function descriptions (placed in a comment above each command in the "Code to lang" file) lived in a variable "code_names"; they are now moved to a "comment:" tag in the yaml. This had already been done in abstract groups.

We (@jenpaulhus, @roed314, @jwj61 and I) also discussed adding tests of code snippets as individual sections of the yaml files, and integrating it into `test.sh`. I'll do this in a separate PR.

lmfdb/utils/place_code.py
```diff
@@ -0,0 +1,119 @@
...
+    """ Utility class for displaying code snippets on lmfdb pages
+    """
+    def __init__(self, code, item=None, pre="", post=""):
+        """NB: Pre and post are used exclusively for placing code
+        and is useful for displaying codes in table, for example.
+        """
+        self.code = code
+        self.item = item
+        if 'prompt' in code.keys():
+            self.langs = sorted(list(code['prompt'].keys()))
```
roed314: You can replace this with just `sorted(code['prompt'])`, and the same below.

lmfdb/utils/place_code.py
```diff
@@ -0,0 +1,119 @@
...
+class CodeSnippet():
+    """ Utility class for displaying code snippets on lmfdb pages
+    """
+    def __init__(self, code, item=None, pre="", post=""):
+        """NB: Pre and post are used exclusively for placing code
+        and is useful for displaying codes in table, for example.
+        """
+        self.code = code
+        self.item = item
+        if 'prompt' in code.keys():
```
roed314: This can be `if 'prompt' in code`, and the same for the `elif` below.

lmfdb/utils/place_code.py
```diff
@@ -0,0 +1,119 @@
...
+        }
+        </script>
+        """
+        return js_str + box_str
+    
+    def build_frontmatter(self, label, lang):
+        """ Build frontmatter for export_code
+        """
+        codefrmt = self.code['frontmatter']
+        keys = list(codefrmt.keys())
```
roed314: This can just be `set(codefrmt)` rather than `list(codefrmt.keys())`.

roed314: Actually, you don't need it: just use `key in codefrmt` below rather than `key in keys`.

lmfdb/utils/place_code.py
```diff
@@ -0,0 +1,119 @@
...
+        return frmt.format(lang=self.full_names[lang], label=label) + "\n"
+    
+    def export_code(self, label, lang, sorted_code_names):
+        """ Export code via 'Code to LANG' button for properties box
+        """
+        if lang == 'gp':
+            lang = 'pari'
+        code = self.build_frontmatter(label,lang)
+        for key in sorted_code_names:
+            assert key in self.code.keys()
```
roed314: Can use `key in self.code`

lmfdb/utils/place_code.py
```diff
@@ -0,0 +1,119 @@
...
+            if not self.code[key] is None and lang in self.code[key]:
```
roed314: Can use `self.code[key] is not None`

roed314:
Looking at the Magma commands on elliptic curves, there's a `%s` that didn't get substituted (`E := EllipticCurve(%s)` [here](http://localhost:37777/EllipticCurve/Q/127050/gr/2/download/magma?label=127050.gr2) for example).  The same problem exists for Sage.  This section seems to be new with this PR (I don't see it on beta).

PR #6533, opened 2025-07-15
Fix Dynamic Statistics bug

### LINKED ISSUE #4898: Dynamic stats bug in abvar
Something seems to be broken on

https://beta.lmfdb.org/Variety/Abelian/Fq/dynamic_stats?count=None&hst=DynStats&col1=g&proportions=recurse&col2=angle_rank&search_type=DynStats

which only lists g=4.  The same page throws a server error on production:

https://www.lmfdb.org/Variety/Abelian/Fq/dynamic_stats?count=None&hst=DynStats&col1=g&proportions=recurse&col2=angle_rank&search_type=DynStats

jenpaulhus:
This PR fixes an error where the "create your own statistics" links weren't working for: abvar, groups, p-adics, and CMF.  
It also puts a lower case "s" in "Dynamic Statistics" in the breadcrumbs (I believe this is our convention to only capitalize the first word)

http://localhost:37777/ModularForm/GL2/Q/holomorphic/dynamic_stats
http://localhost:37777/padicField/dynamic_stats
http://localhost:37777/Groups/Abstract/dynamic_stats
http://localhost:37777/Variety/Abelian/Fq/dynamic_stats
should all work, but return errors on beta.

I am not sure that it actually fixes #4898 as I am not sure what the data looks like for Ab var.  And it may be worth having folks who work on these different sections confirm that the dynamic stats are working as intended.

For groups, I created an abstract-refine-search.html to match the other 3 sections.  Directly calling refine_search_form.html caused issues because it was not extending homepage.html.  I can change that if needed.

### DIFF
--- a/lmfdb/groups/abstract/stats.py
+++ b/lmfdb/groups/abstract/stats.py
@@ -207,7 +207,8 @@ class GroupStats(StatsDisplay):
         from .main import group_parse
         group_parse(info, query)
 
-    dynamic_parent_page = "abstract-search.html"
+#dynamic_parent_page = "abstract-search.html"
+    dynamic_parent_page = "abstract-refine-search.html"
     dynamic_cols = ["order", "order_factorization_type", "abelian"]
 
     @lazy_attribute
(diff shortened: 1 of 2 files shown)


AndrewVSutherland:
@jenpaulhus @roed314 @jwj61 It seems strange to me that the dynamic statisits for p-adic fields only offer two variables (galois group and slopes).  Shouldn't things like ramification index, discriminant exponent, residue field degree, residue field characteristic, ... also be options?

Issue #6331, opened 2025-02-26
Change error message


jenpaulhus:
(Based on a bugreport filled in for a slow query)

We currently have a message that says ""Error: The search query took longer than expected! Please help us improve by reporting this error here"

But maybe best to just suggest people try again later, or go to the beta page, instead of having people report right away.

AndrewVSutherland:
Just to add some context, it is not that uncommon for the https://www.lmfdb.org to experience temporary slow downs due to an excessive number of queries from bots (we try to strike a balance between minimizing captcha's and managing the server load induced by bots; we tend to err on the side of assuming queries are coming from humans, but this means we occasionally get overwhelmed by bots).  The vast majority of slow queries that get reported as bugs are not reproducible for this reason; indeed, it will very often be the case that simply repeating the same query will return the desired response almost immediately because the query that timed out has brought all of the relevant indices into cache.

PR #6548, opened 2025-07-15
Change error message

jwj61:
Addresses issue #6331 

This changes the error message on timeout to read

The search query took longer than expected! Please try again later, or use https://beta.lmfdb.org. If your search still times out, please help us improve by reporting this error <a href="%s" target=_blank>here</a>.

Good news, it is running on https://pink.lmfdb.xyz

Bad news, I can't find a query which times out


AndrewVSutherland:
image

PR #6550, opened 2025-07-15
Remove negative space

### LINKED ISSUE #5786: Confusing display for fundamental units
We use raw/typeset mode to show fundamental units on number field pages, which is great.  But the ellipses that get added to truncate long integers sometimes get covered up and look like multiplication:

Screenshot 2023-12-20 at 11 11 14

The first numerator looks like `26*54`.  I was trying to figure out how we were choosing that factorization, but eventually realized that it was a spacing issue.

jwj61:
Fix #5786 

In units of a number field, some coefficients are being compressed, but the ... was not always visible.  Turns out, the compression function puts negative space around the ellipsis, so this just removes it.

http://127.0.0.1:37777/NumberField/6.0.4920750000000.7
https://beta.lmfdb.org/NumberField/6.0.4920750000000.7

### DIFF
--- a/lmfdb/utils/web_display.py
+++ b/lmfdb/utils/web_display.py
@@ -687,7 +687,7 @@ def compress_poly_Q(rawpoly,
 
     def frac_string(frac):
         if frac.denominator() == 1:
-            return compress_int(frac.numerator(), negative_space=False)[0]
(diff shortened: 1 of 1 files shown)


jwj61:
Adjusted as suggested

PR #6566, opened 2025-07-16
minor details done by ruff --select=FURB --fix

fchapoton:
about using min, etc


lmfdb/zeros/zeta/platt_zeros.py
```diff
@@ -8,7 +8,7 @@
 import sqlite3
 import struct
 import sys
-from math import log
+from math import log, log2
```
edgarcosta:
```suggestion
from math import log2
```

Issue #4535, opened 2021-04-23
L-functions connection to abelian varieties over finite fields


edgarcosta:
For rational L-functions of motivic weight one, we should link the local factors for small p with abelian varieties over finite fields.

We should also then provide this infrastructure for elliptic curves over Q/NF and genus 2 curves.

PR #6573, opened 2025-07-17
L-functions connection to abelian varieties over finite fields

HannekeWiersema:
Partially resolved issue #4535.

For rational L-functions of motivic weight one, we have linked the local factors for small p with abelian varieties over finite fields. We have only provided a column with labels if at least one of these has a corresponding variety on LMFDB.
 
This was then also done for isogeny classes of elliptic curves over Q and genus 2 curves. For elliptic curves over Q this involved adding a new table with L-function data, for genus 2 curves we just added a column to the existing L-function data table.

Elliptic curves over number fields still outstanding.


roed314:
I get an error on http://localhost:37777/EllipticCurve/Q/52246781/a/ because the L-function is not in the database.

roed314:
Let me know when you guys are ready for me to take a look again.

PR #6575, opened 2025-07-17
Only show Remarks section when needed for modular curve families

jwj61:
On modular curve families page, there is a section for remarks which is sometimes empty.  This does not display the Remarks section when there is no content.

There is no actual remarks:
https://beta.lmfdb.org/ModularCurve/Q/family/Xarithpm1
http://127.0.0.1:37777/ModularCurve/Q/family/Xarithpm1

There are remarks:
https://beta.lmfdb.org/ModularCurve/Q/family/X1
http://127.0.0.1:37777/ModularCurve/Q/family/X1

### DIFF
--- a/lmfdb/modular_curves/templates/modcurve_family.html
+++ b/lmfdb/modular_curves/templates/modcurve_family.html
@@ -33,9 +33,10 @@
     </tr>
 </table>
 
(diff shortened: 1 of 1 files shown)


AndrewVSutherland:
I'm in the process of adding some remarks now, but I see no reason not to merge this PR.

PR #6579, opened 2025-07-18
Added more code snippets for the abstract group pages

rvisser7:
This adds further code snippets for the abstract group pages, and in particular also adds code in SageMath for groups (in addition to Magma and Gap).  In order for the user to run the snippets in a linear order without errors, we've added code at the top of each group page which provides the (currently known) somewhat "best" way to define the group G in each language.  If G is in a particular group family (e.g. Cyclic, Symmetric, Dihedral, Alternating,...), then we usually provide the code specific to that family, otherwise we default to the permutation group code (at least for now).

Example of code snippets for the dihedral group of order 40:
http://localhost:37777/Groups/Abstract/40.6
https://beta.lmfdb.org/Groups/Abstract/40.6

Gap:
image

Magma:
image

Sage:
image

Example where we default to the permutation group code at the top:
http://localhost:37777/Groups/Abstract/16.5
https://beta.lmfdb.org/Groups/Abstract/16.5

Feedback is very much appreciated!  Certainly many things can still be improved, and many more group code snippets can still be added for Magma/Gap/Sage (maybe Pari can also be added, although I'm not sure how much group theory it has?)

One still annoying thing is that when the permutation code is very long, this pushes everything on the same page further down, below the properties box on the right.  We should probably deal with this behaviour in a better way?

This was done at the [LMFDB workshop](https://math.mit.edu/~roed/conferences/LMFDB25/) hosted at MIT.


edgarcosta:
> One concern I have with offering the character tables and conjugacy classes as a code snippet is that the outputs will often not match the character tables/conjugacy class reps we produce since our groups are often stored with a different element_repr_type. I think this is going to be confusing to new users.

We could have comment pointing that out, but overall we should have code snippets even if it creates confusion

rvisser7:
Many thanks for the feedback!  I'd be happy to take out the character table and conjugacy class code snippets (at least for now) if this would look better.  Though I like the idea of commented snippets; e.g. for Sage we could put:

`G.character_table()  # Output not guaranteed to exactly match the LMFDB table`

if this looks ok?

rvisser7:
Here's a preview of version 2 of the scrollable code snippets: [in particular, we've made the scroll bars slightly narrower and only applied the scrollable part to the code itself (so the copy-code picture and `magma:` does not scroll)]

image

havarddj:
I like this a lot!

edgarcosta:
Screenshot 2025-07-18 at 15 31 40
there might be a way to avoid this table overflow

rvisser7:
Great, thank you!  I've pushed the changes.

I'll just also note that, at the moment, the logic which goes through the various cases to generate the top code snippet for each group is a rather hacky 80-line lump of code in the `groups/abstract/web_groups.py` file.  I'm sure there's a far better way to do this, so I'm happy to refactor this in a better way if anyone has suggestions :)

jenpaulhus:
With the character table and some of the other computations, for very large groups, do we want to have a warning about time to compute?  I mean, you can ask magma to compute the character table of A47 but  it's 62494 x 62494 and I don't think it's gonna finish in any reasonable time.

Maybe there are some other computations that are also extremely time consuming for large groups.

jwj61:
I don't think warnings are necessary.

jenpaulhus:
It seems a little silly to have snippets on our pages that will never run.

rvisser7:
I've added a few more code snippets on the groups pages, and have also included scroll bars to any code snippet in the LMFDB with enough characters (at present, it's set to include scrollbars if more than 90 characters, if it's at the top of the page, or more than 160 characters, if it's in the body; these values are mostly arbitrary and can be fine-tuned)

As always, feedback is much appreciated!  (including tips on how to make the code look better)!  It should be available on https://pink.lmfdb.xyz/ (thank you @edgarcosta !)  (I'll probably be offline for most of the weekend, as I'll be travelling, but can continue to work on this once I get back to Prague)

roed314:
@rvisser7, I'm going to work on fixing the last of the data issues over the next day or two, and it would be great to have this PR included when we add groups to production after that.  Let us know if there's any feedback you need to finalize this!

havarddj:
Sorry if this is slightly off-topic, but there's some weird font difference in some of the group pages which you might want to fix while you're at it: (firefox on macos)
image
The reason is that the function `show_outer_group` in web_groups.py returns a mathrm-ed latex string. I guess the easiest fix is to change it to regular text.

havarddj:
Sorry about that! Great that it's fixed now. I'll have a look at the PR tomorrow.

rvisser7:
There are still a few further visual bugs to iron out:

1.  E.g. for this group: https://pink.lmfdb.xyz/Groups/Abstract/16.6 , using a large enough monitor, the sage code at the top seems to appears fine:
image

but with a smaller screen, the scrollbar automatically appears (as it should), but now the language prompt is misaligned with the code itself:
image


2.  It also appears that the copy-code button seems to render slightly differently between Chrome/Edge, and Firefox, especially for multiline code snippets. E.g. for this number field: https://pink.lmfdb.xyz/NumberField/2.0.4.1 , the sage code for the class number formula seems to render well on Chrome and Edge:

image

but on Firefox, there's a bit too much bottom spacing between the copy-code image and its border:

image

Any feedback on how to fix these would be much appreciated!

jenpaulhus:
On Safari on a Mac on pink.lmfdb.xyz, I'm still getting issues where the scroll bar cuts off the command itself so viewers can't see the command. I'm attaching 2 screenshots.


Screenshot 2025-07-23 at 11 13 37
Screenshot 2025-07-23 at 11 13 26

havarddj:
I tried on safari as well. Does the bar disappear after a couple of seconds when you stop scrolling, or does it stay there?

rvisser7:
> On Safari on a Mac on pink.lmfdb.xyz, I'm still getting issues where the scroll bar cuts off the command itself so viewers can't see the command. I'm attaching 2 screenshots.
> 

Thank you very much for letting me know about this and for the screenshots! 👍 (this stresses the importance to test on all browsers!) 

@havarddj and I did some further testing with the scrollbars.  We've added `scrollbar-gutter: stable;` and `white-space: preserve nowrap;` to the `.code` and  `.code, .prompt` parts of the `style.css` file, respectively.  At least for us, this seems to display scrollbars a bit better (although probably still not perfect).

@jenpaulhus:  Does this make any difference for you; does it make things look better or worse?  (I'm afraid I don't have easy access to a Mac, so I can't easily test this myself)

jenpaulhus:
Still was a problem on one computer so I dug around a bit.  It isn't working for me on Safari on an old Intel Mac that is running Ventura (to @havarddj  question, in this case the scroll bar never disappears). 

It DOES work on Safari on my college computer which is one of the M3 chips and running Sequoia. 

So looks like the only issue is with older versions of Safari?  (I mean, my Safari is up to date but up to date for an older OS.)

jenpaulhus:
One other small thing I noticed when testing the scroll bar on random pages.   If you go to https://pink.lmfdb.xyz/Groups/Abstract/2992896.d

this is one of the few groups that we don't have as a permutation representation so you get a "bad" top line defining the group as something like ```G:=PermutationGroup(none)```.

Thank you both for your work on this!!

rvisser7:
> One other small thing I noticed when testing the scroll bar on random pages. If you go to https://pink.lmfdb.xyz/Groups/Abstract/2992896.d
> 
> this is one of the few groups that we don't have as a permutation representation so you get a "bad" top line defining the group as something like `G:=PermutationGroup(none)`.

Many thanks for catching this example!  I've updated the code for generating the top code snippet to also display code for either the polycyclic presentation or a matrix group presentation for the group G (if this exists and is implemented in the relevant language)

So the top code snippet for https://pink.lmfdb.xyz/Groups/Abstract/2992896.d now gives a `PCGroup` implementation for Magma and Gap, and gives a `MatrixGroup` representation for Sage.  (at the moment, I'm not sure if Sage has something similar to  `PCGroup` ?)  However, it seems giving a MatrixGroup implementation at the top isn't really ideal, as some of the code snippets unfortunately don't work for matrix groups  (e.g. `G.composition_series()` doesn't work if G is a matrix group in Sage -  similarly, in Magma, `Socle(G)` doesn't seem to work for groups G of type `GrpMat`).  I'm not sure how best to deal with this?

There do also seem to be groups where we don't have either a permutation or a matrix representation, e.g. https://pink.lmfdb.xyz/Groups/Abstract/5832.jd .   So while the top code snippet for this group now gives the `PCGroup` implementation for Magma and Gap, it now unfortunately displays nothing for Sage.

At least in principle, Sage does have support for finitely presented groups (e.g. doing something like `F.<a,b> = FreeGroup(); G = F / [a^2, b^4]`), except I think this is *very* inefficient, and we probably don't want to have this as our top code snippet.  Any suggestions on how best to deal with this would be very welcome!

rvisser7:
Thank you very much all for the help and feedback!  I've now added the `sage_gap` language to the `code.yaml` groups file, which implements the group G in SageMath, but instead using the `gap` interface.  This is now displayed as our construction for the top code snippet, if there's no other (easy) way to natively define G in Sage.  Furthermore, if we're using either `MatrixGroup` or `AbelianGroup` to define G in our top code snippet, then we now hide all the code snippets which don't work for these types of groups.

I think the displayed code within the code snippets themselves (and the logic for the top code snippet) should now be mostly done, except for very possible typos! :)

All that remains is to ensure the code snippets are rendering properly on all browsers (in particular, checking Safari on Mac)!

For testing purposes, below are examples of groups which (hopefully) cover all the possible cases for our top code snippet:

 - Example of cyclic group:

http://localhost:37777/Groups/Abstract/10.2
https://pink.lmfdb.xyz/Groups/Abstract/10.2

 - Example of symmetric group:

http://localhost:37777/Groups/Abstract/24.12
https://pink.lmfdb.xyz/Groups/Abstract/24.12

 - Example of alternating group:

http://localhost:37777/Groups/Abstract/12.3
https://pink.lmfdb.xyz/Groups/Abstract/12.3

 - Example of dihedral group:

http://localhost:37777/Groups/Abstract/24.6
https://pink.lmfdb.xyz/Groups/Abstract/24.6

 - Example of dicyclic group:

http://localhost:37777/Groups/Abstract/24.4
https://pink.lmfdb.xyz/Groups/Abstract/24.4

 - Example of group in the Gap database (but not in a special family)

http://localhost:37777/Groups/Abstract/256.123
https://pink.lmfdb.xyz/Groups/Abstract/256.123

 - Example of abelian group not in Gap database:

http://localhost:37777/Groups/Abstract/1024.yw
https://pink.lmfdb.xyz/Groups/Abstract/1024.yw

 - Example of a special matrix group (implemented in Gap, Magma, and Sage)

http://localhost:37777/Groups/Abstract/115379712.a
https://pink.lmfdb.xyz/Groups/Abstract/115379712.a

 - Example of special matrix group (only implemented in Magma, for now):

http://localhost:37777/Groups/Abstract/4769856.a
https://pink.lmfdb.xyz/Groups/Abstract/4769856.a

 - Example of group which only has a permutation representation:

http://localhost:37777/Groups/Abstract/1006632960.k
https://pink.lmfdb.xyz/Groups/Abstract/1006632960.k

 - Example of group which only has a polycyclic representation:
(and thus Sage uses the `gap` interface for top code snippet)

http://localhost:37777/Groups/Abstract/5832.jd
https://pink.lmfdb.xyz/Groups/Abstract/5832.jd

 - Example of group which only has a matrix representation (and not in special family):

http://localhost:37777/Groups/Abstract/7228212648.a
https://pink.lmfdb.xyz/Groups/Abstract/7228212648.a

 - Example of Chevalley group:

http://localhost:37777/Groups/Abstract/214841575522005575270400.a
https://pink.lmfdb.xyz/Groups/Abstract/214841575522005575270400.a

 - Example of Twisted Chevalley group:

http://localhost:37777/Groups/Abstract/211341312.a
https://pink.lmfdb.xyz/Groups/Abstract/211341312.a

roed314:
This is what it looks like on Safari: 
Screenshot 2025-07-25 at 22 31 14

I don't know why it wraps to multiple lines, or why the scroll bar covers up the bottom shading, but I think it's fine.  And Firefox and Chrome look fine on a Mac.

roed314:
Thanks @rvisser7!

PR #6595, opened 2025-07-18
Add subgroup boolean columns to search result

### LINKED ISSUE #6478: Add new booleans for subgroups to search results
We now have a lot more booleans computed for subgroups (e.g., metabelian, metacyclic, quotient_metabelian, quotient_nilpotent, quotient_supersolvable) which should show up in search results for subgroups.

jenpaulhus:
Fixes #6478.

This adds a bunch of boolean columns to subgroup search result.  I just threw them all in, but we may want to make some of them not show up by default.  Also happy to hear thoughts on shortened names or the order of columns. 

http://localhost:37777/Groups/Abstract/?search_type=Subgroups
http://beta.lmfdb.org/Groups/Abstract/?search_type=Subgroups

### DIFF
--- a/lmfdb/groups/abstract/main.py
+++ b/lmfdb/groups/abstract/main.py
@@ -1314,6 +1314,13 @@ subgroup_columns = SearchColumns([
         CheckCol("maximal", "group.maximal_subgroup", "max", short_title="Sub. maximal"),
         CheckCol("perfect", "group.perfect", "perf", short_title="Sub. perfect"),
         CheckCol("central", "group.central", "cent", short_title="Sub. central"),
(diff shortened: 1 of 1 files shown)


jwj61:
I would make many of them not show up by default, e.g., cyclic, solvable, perfect, simple, nilpotent, Agp, Zgp, metabelian, metacyclic, super solvable, and for the quotient part, I would not display cyclic, solvable, simple, nilpotent, Agp, metabelian.

roed314:
After some waffling, I agree with John about which columns should be shown by default.  I've made that change, and reordered the columns a bit.

Issue #4897, opened 2021-12-10
Show elliptic curve equations on abvar pages with g=1


AndrewVSutherland:
On abvar pages for dimension 2 isogeny classes that contain a Jacobian we list curves but not for dimension 1.  Compare

https://www.lmfdb.org/Variety/Abelian/Fq/1/17/g
https://www.lmfdb.org/Variety/Abelian/Fq/2/17/al_cl

We should show the  elliptic curves in the dimension 1 isogeny classes (or at least one example).

tomgrubbmath:
For the example: https://www.lmfdb.org/Variety/Abelian/Fq/1/17/g , the underlying class is missing a "curves" attribute . Is there a way of seeing if this is a data issue? The database storing the isogeny classes is "av_fq_isog"

roed314:
Yep, this is a data issue.  It shouldn't be too hard to compute this, but I'm busy with other things.  If you'd like to give it a stab, if you send me a file in the following format I can add it to the database.  I've given a few example lines; obviously you should compute the curves for every `g=1` isogeny class.

```
label|curves
text|text[]

1.2.ac|{y^2+y=x^3+x^2+1}
1.2.ab|{y^2+xy=x^3+x^2+1}
```

tomgrubbmath:
For computing this, would the relevant resource be Section 3.7 here: https://arxiv.org/abs/2003.05380 (and, following the trail, going to work of Howe/Lauter)? I can take a stab at it but I'll need to get up to speed so it will just take a bit of spin up time.

roed314:
The work of Howe and Lauter is aimed at determining _whether_ an isogeny class contains a Jacobian, not enumerating curves whose Jacobian lies in the class.

Instead, the best strategy will follow what Drew did in higher dimension.  For each q in the database (prime powers q < 500 together with 512, 625, 729, 1024), numerate all elliptic curves over Fq up to isomorphism.  In Magma you can do this using the `EllipticCurveFromjInvariant` and `Twists` functions for example: the first gives you one example for each j, then the second gives you all twists of that (isomorphism classes over Fq).  Then you count points to get the Frobenius polynomial and group them into isogeny classes.

tomgrubbmath:
Got it. Thanks for clarifying David! I'll take a crack.

tomgrubbmath:
[...]

PR #6599, opened 2025-07-20
Elliptic curve equations on abvar pages for g=1

JaneShi99:
This is a pull request for issue #4897.

It generates data of the following form:
label | {eq_1, eq_2, eq_3}
where each label is an isogeny class label of abelian variety of genus 1 over F_q, with q a prime power in [2..499] + {512, 625, 729, 1024}, and each eq is the equation of a distinct isomorphism class of curves with the corresponding isogeny class.

For example:
```
1.4.ab|{y^2+x=x^3+a,y^2+x=x^3+a^2}
1.4.b|{y^2+x=x^3+a*x^2+a^2,y^2+x=x^3+a^2*x^2+a}
```

I sent the generated data to @roed314 

Step 1: Running `magma g1.m` generates all the data (writes to `output.txt`) with either (j-invariant not 0, 1728) or (F_q is not characteristic 2, 3). 
Note that 
- For F_q with characteristic not equal to 2 or 3, it displays a canonical representative for each isomorphism class of elliptic curves. 
- For F_q with characteristic 2, 3, with a nonzero j-invariant, it displays a (not necessarily canonical) representative for each isomorphism class.

Step 2: Running `load("g1_char23_j_0.sage")` in the sage shell generates all data (writes to `output23_0_1728.txt`) for fields of characteristic 2, 3 and j-invariant equal to 0, thanks to @JohnCremona 's database `curves_with_j_0_char2, curves_with_j_0_char3` in sage. A canonical curve is displayed for each isomorphism class.

When running the commands in Step 1 and Step 2, it also generate a file that counts the number of isogeny labels for each F_q. Running `load("check_correctness.sage")` in the sage shell adds the total number of isogeny label for each characteristic q, (it adds number of labels over `output.txt` and `output23_0_1728.txt`, since both has data for characteristic 2 or 3) and compare it to the number of total isogeny labels on the lmfdb database. It checks that these numbers are equal. When I ran this command, it confirmed that the numbers match.


scripts/abvar/check_correctness.sage
```diff
@@ -0,0 +1,38 @@
...
+Reads files from counts_filenames, and for each prime power q,
+add the number of isogeny labels of genus 1 over F_q
+from each of the files. Then we check if it matches the LMFDB data.
+'''
+
+counts_filenames = ["counts23_0_1728.txt", "counts.txt"]
+
+counts  = defaultdict(int)
+
+def is_int(s):
```
roed314: It's better to use `s.isdigit()` rather than this function.

scripts/abvar/check_correctness.sage
```diff
@@ -0,0 +1,38 @@
...
+            if not line: 
+                continue
+            tokens = line.split(",")
+            
+            if all(is_int(token) for token in tokens) and len(tokens)>1:
+                counts[int(tokens[0])] += int(tokens[1])
+                
+lmfdb_counts = db.av_fq_isog.stats.column_counts(["g", "q"])
+
+for key in sorted(counts.keys()):
```
roed314: You can use `sorted(counts)` rather than `sorted(counts.keys)` because iterating over a dictionary iterates over keys by default.

scripts/abvar/g1_char23_j_0.sage
```diff
@@ -0,0 +1,56 @@
...
+    else:
+        return 'a' + cremona_letter_code(-m)
+    
+def get_label(f,q):
+    g=1
+    return '%s.%s.%s' % (g, q, '_'.join(signed_cremona_letter_code(f[i]) for i in range(1,g+1)))
+    
+    
+def output_data():
+    with open(output_filename, "w") as file:
```
roed314, line 22: You don't need to do this: writing to a file that doesn't exist with mode "a" will create it automatically.

scripts/abvar/g1_char23_j_0.sage
```diff
@@ -0,0 +1,56 @@
...
+            curves = curves_with_j_0_char2(F)
+        else:
+            curves = curves_with_j_0_char3(F)
+            
+        for curve in curves:
+            Lpoly = curve.frobenius_polynomial()
+            to_output = (curve._equation_string()).replace(" ", "")
+            label_to_curves[get_label(Lpoly,q)].add(to_output)
+        
+        for label in label_to_curves.keys():
```
roed314: You can skip the `.keys()` here.  Even better, use `for label, equations in label_to_curves.items():`  and then `','.join(equations)` below.

roed314:
I've uploaded the data.  It looks good, but uncovered an error in how we're displaying curves (see #6600).  Once this has been merged to production, I'll push the data there and then we can close #4897.

The scripts on this PR are independent; I'm happy to merge once you address the comments.  Thanks for this, it's been a long time coming!

edgarcosta:
Thank you @JaneShi99 

Could you clarify for me what is `a` in https://beta.lmfdb.org/Variety/Abelian/Fq/1/1024/cl ?

roed314:
@edgarcosta, I'm sure that $a$ is the generator of $\mathbb{F}_q$ (so a root of the Conway polynomial $z^10 + z^6 + z^5 + z^3 + z^2 + z + 1$ in the case that $q=1024$).

roed314:
Basically because $a^{597}$ is much shorter than $a^9 + a^7 + a^6 + a^5 + a^4 + a^2 + a$.  We make this choice in displaying elements of finite fields in matrices for group constructions, and I think it's a reasonable convention for non-prime fields elsewhere in the LMFDB as well (though if an element is actually in the prime field, displaying it as an integer is better).

JaneShi99:
Hi @roed314 and @edgarcosta, I figured out the error:

I used the following code to convert the elliptic curve to string:

```
    f, yCoeff := HyperellipticPolynomials(E);
    if yCoeff eq 0 then 
        return "y^2 = " cat Sprint(f);
    else 
        return &cat["y^2 + ", Sprint(yCoeff), " = ", Sprint(f)];
    end if;
```
HyperellipticPolynomials of y^2 + h(x)y = f(x) returns f(x) and h(x).

So, for example, if E is defined by the following, 
```y^2 + x*y = x^3 + a^49```
We would have yCoeff = x, then my wrong code would return
```y^2 + x = x^3 + a^49```

So I changed it to the following instead:
```        
        if yCoeff eq 0 then
            return "y^2 = " cat Sprint(f);
        elif yCoeff eq x then 
            return &cat["y^2 + x*y = ", Sprint(f)];
        else 
            return &cat["y^2 + (", yCoeff ,")*y = ", Sprint(f)];
        end if;
```

I made a new PR and emailed @roed314 the new file. Thanks @edgarcosta again for spotting this!!

JohnCremona:
There already exists code for creating a typeset elliptic curve equation

roed314:
The data is now live on [beta](https://beta.lmfdb.org/Variety/Abelian/Fq/1/61/c) and [production](https://beta.lmfdb.org/Variety/Abelian/Fq/1/61/c).  Thanks @JaneShi99!  Once we figure out the best version of the code to include in the scripts folder, I'm happy to merge this PR for future reference.

roed314:
If there's a way to improve the generation for typesetting the curves, we can update the script in a future PR.

JohnCremona:
> > There already exists code for creating a typeset elliptic curve equation
> 
> Thank you for pointing this out! I also thought that it would exist, but I'm not quite sure how to find it. What I have now is the closest I could get. Would you mind pointing me to where the magma code is, or what function I could call?

The function latex_equation() in lmfdb/ecnf/WebEllipticCurve.py should do what you want.  That was used forr equations of curves over number fields before we strored the strings in the database.  There's a simpler version in lmfdb/web_ec.py which was for curves over QQ.  There, there are also functions for homogeneous and short equations.

scripts/abvar/g1_char23_j_0.sage
roed314: For the record, Jane told me that these lines are intended to delete the files if they were created in a previous run, which seems reasonable.

Issue #6611, opened 2025-07-24
Display of Jacobians over power prime fields


edgarcosta:
When we display equations over non-prime fields we use $a$, however, $a$ is ambiguous, e.g. in https://beta.lmfdb.org/Variety/Abelian/Fq/1/1024/cl

If by some chance $a$ is primitive, for example, we are using conway polynomials, which I think we are, we could display $\zeta$ instead and get rid of the ambiguity.

See the discussion following https://github.com/LMFDB/lmfdb/pull/6599#issuecomment-3097321577

PR #6613, opened 2025-07-25
For isogeny class of abvar of dimension 1, displays an explanation for the variable in the equation

JaneShi99:
This PR resolves issue #6611 

For isog classes of abvars of dimension 1, it already displays some equations. If 'a' shows up in any of the equation, we display a knowl that explains what 'a' is.


roed314:
I updated it to better handle the case that `curves` is null.  I'll merge once tests pass.

PR #6618, opened 2025-07-28
another UTC fix

AndrewVSutherland:
Looks like this one was missed in https://github.com/LMFDB/lmfdb/pull/6564


roed314:
Merged #6619 instead.

PR #6642, opened 2025-08-11
Fix family names in dropdown menu

jenpaulhus:
This changes the names of groups in the Family dropdown menu to match our new convention.    So GO in Magma is now Orth and GU is Unitary and the C* are now G*.  

Go to the main page and look at the dropdown:
http://localhost:37777/Groups/Abstract/
vs
https://beta.lmfdb.org/Groups/Abstract/

Also fixes a tiny bug for GOrth.


jenpaulhus:
Closing this without merging since it will be subsumed by #6643.

PR #6643, opened 2025-08-12
Cox search

roed314:
I made some underlying data changes to the `gps_families` and `gps_special_names` tables to adapt to our new conventions on G vs C and to support Coxeter subfamilies better.  This PR subsumes both #6632 and #6642.


jenpaulhus:
Overall looks good but wondering if we can fix up the data even more? (This relates to @rvisser7 good point about the "Lie type" section under "Construction")

Take http://localhost:37777/Groups/Abstract/24261120.b  which we now call O^+(6,3).  Except on its webpage we still call it GO^+(6,3) at the top and that is the Lie type we give it under "Constructions".

If you instead search for the family with GO^+(6,3), you would go to this page instead:
http://localhost:37771/Groups/Abstract/48522240.a
which we also call GO^+(6,3).

roed314:
Alright, this PR should be backward compatible with latex macros now.  The indented ones in `base.html` are deprecated and should be deleted once we have all the data and knowls updated to new versions.  Note that this does mean that there will be some unintuitive macros: for example, `\GU` and `\GO` will not give GU and GO respectively, you have to use `\GUnitary` and `GOrth`.

roed314:
Let me know if you think we should merge this now or wait until I fix the data problems in `gps_groups`.  If you notice any data issues in `gps_families` or `gps_special_names`, let me know.

rvisser7:
Great, thank you! Also thanks for improving the implementation of the Coxeter families 🙂.  I think the ``gps_families`` and ``gps_special_names`` tables look good to me!

I can have a go at updating some of the tex macros in the group families knowls after this is merged (once I have some more time).  I'll defer to you and @jenpaulhus if you'd like to merge now or after correcting the names and tex_names in ``gps_groups`` - I'm happy with either.

PR #6660, opened 2025-08-29
Check that there is enough space on grace and devmirror

roed314:


roed314:
This needs to wait until https://github.com/roed314/psycodict/pull/39 is merged, at which point we should update psycodict on servers and computers where people will be updating data.

PR #6666, opened 2025-08-30
details for mypy/ruff/pycodestyle in web_number_field

fchapoton:
some ruff and pycodestyle fixes

and some typing annotations

### DIFF
--- a/lmfdb/number_fields/web_number_field.py
+++ b/lmfdb/number_fields/web_number_field.py
@@ -168,7 +168,7 @@ rcycloinfo = {'3.3.49.1': 7, '3.3.81.1': 9, '5.5.14641.1': 11,
   '44.44.860115008245742907292219227824365111518443501386754869093198564719785672888549376.1':276,
   '40.40.32473210254684090614318847656250000000000000000000000000000000000000000.1':300}
 
-cyclolookup = {n:label for label,n in cycloinfo.items()}
+cyclolookup = {n: label for label, n in cycloinfo.items()}
 cyclolookup[1] = '1.1.1.1'
 cyclolookup[3] = '2.0.3.1'
 cyclolookup[4] = '2.0.4.1'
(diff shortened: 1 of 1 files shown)


lmfdb/number_fields/web_number_field.py
```diff
@@ -793,10 +807,11 @@ def regulator(self):
         return na_text()
 
     def units(self):  # fundamental units
-        res = None
         if self.haskey('units'):
             return self._data['units']
-        elif self.unit_rank() == 0:
+        # the next few lines looks very dubious
```
roed314: Yeah, I agree

jwj61: Partly this goes back to the switch to postgres which was doubling backslashes, so lines 816 and 817 were undoing that.  The lines before that look like they have to distinguish the cases where we know that there are no fund. units, and where we haven't computed them

fchapoton: should I remove all of that ?

jwj61: Remove what?  The code which distinguishes when there are no units from when we just haven't computed them?  Why would you delete that?

Deleting the substitution on double backslashes is probably ok, but I can't readily inspect the data to be sure it is ok since it is in an extras table.

fchapoton: well, the only return which can really happen is not using ``res`` at all

jwj61: Fair enough.  I see now that the other cases are dealt with downstream.

roed314: I'm fine changing this to something equivalent.

roed314:
The managing editors are going to talk about mypy type annotations at our next meeting (currently scheduled for September 22).  We won't be merging those parts of this PR until then.  If you're happy to wait, feel free to convert this to a draft PR; you're also welcome to extract the other parts of this PR into something with no type annotations that we can review separately.

fchapoton:
so, any news about your meeting and the policy for annotations ?

roed314:
For now, we're not going to accept type annotations as PRs to the LMFDB.  If someone wants to make an argument for why they would be helpful, we can revisit this.

For this PR, I'll go ahead and remove the type annotations and merge, since the rest is fine.

PR #6669, opened 2025-08-31
Add groups to production

roed314:
Removes the "beta-only" status for the abstract groups link in the sidebar.


roed314:
To test, explore https://www.lmfdb.org/Groups/Abstract/

AndrewVSutherland:
Reviewing this now...

AndrewVSutherland:
The number of divisions does not appear to constrain the search at all, for any value I put in (including 0) the search results include all groups  (this is true both in the current PR and on beta):

https://beta.lmfdb.org/Groups/Abstract/?number_divisions+=0

roed314:
Thanks for catching that; fixed in #6670.

PR #6671, opened 2025-08-31
Mypy api2 motives belyi

fchapoton:
just some minor changes to satisfy `mypy` on these three folders


roed314:
See [my comment on #6666](https://github.com/LMFDB/lmfdb/pull/6666#issuecomment-3246552634); we won't be merging this for a few weeks until we have a chance to talk about type annotations in the LMFDB.

PR #6674, opened 2025-09-02
Insert cc_highlight_i if missing

roed314:
We're getting a bunch of flasklog errors for URLs like https://www.lmfdb.org/Groups/Abstract/char_table/1192464.a?cc_highlight=3C.  With this PR, http://localhost:37777/Groups/Abstract/char_table/1192464.a?cc_highlight=3C works.


jenpaulhus:
Thanks for fixing this!

PR #6694, opened 2025-09-16
Add code snippets for Lie type group matrix representations

rvisser7:
This adds some more code snippets for the Lie type representations for each group.  The current implementation simply gives a sequence of code snippet blocks (one after the other) for each Lie type construction.  (any feedback on better ways to display this would be very welcome!)

Example of group with six different Lie type constructions:
https://beta.lmfdb.org/Groups/Abstract/12.3
http://localhost:37777/Groups/Abstract/12.3

image

Example of group with a lot of Lie type constructions (24 constructions):
https://beta.lmfdb.org/Groups/Abstract/6.1
http://localhost:37777/Groups/Abstract/6.1
(displaying all the Magma Lie type constructions does take a lot of horizontal space here)

Example of group where the only stored representation is of Lie type:
https://beta.lmfdb.org/Groups/Abstract/79234877030400.a
http://localhost:37777/Groups/Abstract/79234877030400.a

The Lie type constructions can also be displayed as the top code snippet (if displayed, the lowest priority number is chosen).  Now every group stored in the database should have at least a Magma top code snippet! :) .  If generators are also stored with the construction, then this is also considered as a possible GAP and Sage top code snippet, even if the Lie type is not natively implemented in GAP/Sage.

Ideally we should probably update the families data in `gps_groups` to use the new family names before merging this.  Though as a temporary solution, and to test this in the mean time, the code defines an `old_to_new_family_name` dictionary which converts the old family names to the new names.  Once the data is updated, this line can of course be deleted.

As always, any feedback is very welcome! 🙂


rvisser7:
I've made a few updates to this PR.  The code which fetches the Lie-type representation data is now implemented as a lazy attribute `lie_representations` method (so if the representations become later stored in a different table, only that part of the code will need to be updated - the code snippet display logic will remain unchanged).

I'll briefly summarise the logic here:  For any abstract group $G$, we let $L$  be an ordered list of all its Lie-type representations stored in the database.  Then to display the Lie-type code snippets for $G$ in the language $M$ (where $M$ is either GAP, Magma, or Sage), we do as follows:

1.  If there exists some $r \in L$ which is natively implemented in $M$, then we display a list of code snippets for every Lie-type representation $r$ which is natively implemented in $M$, in the same order as stored in `G['representations']['Lie']` in the database (in particular, no code snippets are displayed for any Lie type representation $r$ not natively implemented in $M$).

2. Otherwise, if none of the $r \in L$ is natively implemented in $M$, **but** there exists some $r \in L$ which has an explicit set of generators, then we display exactly one code snippet corresponding to the *first* listed $r$ with stored generators, implemented via a generic matrix group construction.

3.  Otherwise, we display nothing (i.e if no $r \in L$ is natively implemented in $M$ and none have generators stored)

I feel like this is probably a decent middle-ground option with having code snippets for the Lie-type representations, but avoids them taking up too much horizontal space when many representations aren’t natively implemented in $M$.

In both (1) and (2), the first displayed code snippet is considered as a potential top code snippet.   In particular, we now prioritise the Lie type snippets according to their given order in the database (and not necessarily by the priority number given in `gps_families`).   

At the moment, the representations data is still fetched from the `gps_groups` table, as we need to know the order of the Lie-type representations to deduce the priority and order that we display them in.  But once information about the priority is added to `gps_special_names`, then we can update the `lie_representations` method to fetch data from there instead.

Note that if $M$ is Magma, then we'll always be in case (1) above whenever $L \not = \emptyset$.  Another interesting side-effect of this is that now every group $G$ currently in the database will have a Sage top code snippet, with the sole exception of the group [896.19347](https://beta.lmfdb.org/Groups/Abstract/896.19347)  (i.e. this is the only group with no permutation representation, no matrix representation, and none of its Lie-type representations have generators stored!) :)

As always, any comments, feedback or other suggestions are very welcome! 🙂

PR #6721, opened 2025-10-07
Allow definitions of terms to reference external definitions

roed314:
Once this PR is merged, you can easily add external links to 
* wikipedia
* mathlib
* groupprops
* mathworld
* nlab
* wikidata
* stacks project
* zbl
* arxiv
* href
* mr

There are two methods: via a `\cite` command (as has existed previously) or with the new `DEFINES` macro, which is intended for definitions of terms.

To test, start creating/editing a knowl, and change a definition (currently just bolded using markdown's `**`) to a macro like
```
{{DEFINES("group", wikipedia="Group_(mathematics)", mathlib="Algebra/Group/Defs.html#Group", groupprops="Group", mathworld="Group", nlab="group", wikidata="Q83478")}}
```
You can also specify a custom knowl to use on the term, rather than the autogenerated links currently used (or the direct link used if there is only one external definition).

This PR also adds the ability to include parenthetical notes (e.g. `(Thm 1.1)`) by adding them after an `@` symbol.

Once this is merged, I'll update doc.knowl.guidelines with instructions on using it and send an announcement on Zulip.

### DIFF
--- a/lmfdb/knowledge/main.py
+++ b/lmfdb/knowledge/main.py
@@ -195,7 +195,7 @@ def ref_to_link(txt):
         link = f'{{{{ LINK_EXT("{disp}", "{url}") | safe}}}}'
         if fragment:
             link += f" ({fragment})"
-        ans.append()
+        ans.append(link)
     return "[" + ", ".join(ans) + "]" + everythingelse
 
 def md_latex_accents(text):


roed314:
@AndrewVSutherland, I think this is ready for review now.

PR #6741, opened 2025-10-28
Use content.announcements knowl for announcements box

roed314:
This PR changes the LMFDB homepage to update its Announcements box from a knowl ([content.announcements](https://beta.lmfdb.org/knowledge/show/content.announcements)) instead of being static. This will hopefully keep the announcements more up to date, and since it only shows the top 4 announcements, allows for keeping old announcements around.  I've populated the knowl with some exciting LMFDB features going back to 2019, but other LMFDB editors should feel free to add more!

It also links to an [ongoing projects](https://beta.lmfdb.org/knowledge/show/content.ongoing) knowl, describing major projects underway within the LMFDB.  Let me know if I've missed anything, and feel free to expand on it.

Finally, I change the color of the Announce box if one of the announcements is the same month or an adjacent month to the current one (I used the same colors as when hovering in the sidebar).

You can see this live on https://blue.lmfdb.xyz.

### DIFF
--- a/lmfdb/homepage/boxes.py
+++ b/lmfdb/homepage/boxes.py
@@ -12,7 +12,7 @@ class Box():
             # Dynamic content generated from knowl
             from lmfdb.knowledge.knowl import knowldb
             from lmfdb.knowledge.main import md
-            max_entries = 4
+            max_entries = 3
             content = [line.strip() for line in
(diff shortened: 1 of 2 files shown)


jenpaulhus:
LGTM

PR #6744, opened 2025-10-31
Fix #6742 - use flexbox for front page layout

### LINKED ISSUE #6742: Use flexbox rather than tables
The LMFDB makes heavy use of tables.  These are not very mobile friendly, since they don't respond well to large changes in the width of the window.

A better approach is to use CSS flexbox layout, which can reflow the content in a narrow window.  This could be done iteratively across the LMFDB.

havarddj:
This PR changes the structure of the boxes on the home page to use css flexbox instead of styling html tables. This makes it much easier to add content without messing up the layout, or even adding more boxes down the line. It also improves the user experience on small screens.

**Before:**
Screenshot 2025-10-31 at 20 44 30

Screenshot 2025-10-31 at 20 45 06

**After:**

Screenshot 2025-10-31 at 20 44 42
Screenshot 2025-10-31 at 20 45 29

(Also fixes some unclosed \<a\>-tags in the yaml)

### DIFF
--- a/Postgres_FAQ.md
+++ b/Postgres_FAQ.md
@@ -727,7 +727,7 @@ Data Validation
    You can use the `lmfdb.verify` module.  Create a python file in the
    `lmfdb/verify` folder with the same name as your postgres table,
    and a class in that file that inherits from `TableChecker` in
-   `lmdfb.verify.verification`.  You can then add tests, which should
+   `lmfdb.verify.verification`.  You can then add tests, which should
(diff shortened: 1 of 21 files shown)


havarddj:
Good call, I've fixed that and made a couple more tweaks. @assaferan, did you want to review it? Let me know if you have any suggestions! That said, I feel like I've personally reached the point of diminishing returns (~partly~ mostly a skill issue)

PR #6748, opened 2025-11-06
Update code-11.1-a1-gp.log

hgu2699:


roed314:
See [this comment](https://github.com/LMFDB/lmfdb/pull/6749#issuecomment-3494856636), which also applies here.

PR #6757, opened 2025-11-06
Update code.yaml

hgu2699:
ECNF/GP: fix conductor-norm snippet by calling idealnorm(K, ellglobalred(E)[1]) (missing K), resolving “too few arguments” in code-11.1-a1-gp.log and code-81.1-CMa1-gp.log.


assaferan:
This is supposed to be fixing #6761

Issue #6696, opened 2025-09-19
cyclic isogeny classes of AV/Fq


AndrewVSutherland:
Theorem 2.2 of [this paper](https://arxiv.org/abs/1806.10842) gives a very simple criterion that allows one to use the Weil polynomial to determine whether the corresponding isogeny class contains a non-cyclic Jacobian or not, which would make it very easy to add a `cyclic` boolean to the  `av_fq_isog` table, as well as a column listing `noncyclic_primes` (the criterion can be applied to each ell-Sylow subgroup, and we can call a prime ell noncyclic if there is member of the isogeny class whose group of rational points has ell-rank > 1).  One could then add (advanced) search options for both columns.

hgu2699:
> Theorem 2.2 of [this paper](https://arxiv.org/abs/1806.10842) gives a very simple criterion that allows one to use the Weil polynomial to determine whether the corresponding isogeny class contains a non-cyclic Jacobian or not, which would make it very easy to add a `cyclic` boolean to the `av_fq_isog` table, as well as a column listing `noncyclic_primes` (the criterion can be applied to each ell-Sylow subgroup, and we can call a prime ell noncyclic if there is member of the isogeny class whose group of rational points has ell-rank > 1). One could then add (advanced) search options for both columns.

Hi, Dr. Sutherland. Where could I find av_fq_isog table? 

<img width="1456" height="699" alt="Image" src="https://github.com/user-attachments/assets/2b7b8f7a-219d-43ae-98dc-68c8a9683fcf" />

PR #6772, opened 2025-11-16
Add cyclic isogeny invariants, search options, and tests for AV/Fq Issue #6696

hgu2699:
This PR adds support for viewing and searching the cyclicity of the group of points for isogeny classes of abelian varieties over finite fields.

**Main changes**

* `lmfdb/abvar/fq/isog_class.py`
  * Extend `AbvarFq_isoclass` to include:
    * `is_cyclic` (whether the group of points is cyclic)
    * `noncyclic_primes` (primes where the isotypic components are non-cyclic)
  * Expose this information on the AV/Fq isogeny class page via `properties()`, under:
    * “Cyclic group of points”
    * “Non-cyclic primes”

* `lmfdb/abvar/fq/main.py`
  * (If applicable in your branch) add/adjust search options and columns so that the new invariants can be queried and, where relevant, shown in the browse/search results.

* `lmfdb/abvar/fq/test_av.py`
  * Update `test_cyclic_group_of_points_display` so it checks for the new “Cyclic group of points” text and passes with the updated display.

* `lmfdb/abvar/fq/test_browse_page.py`
  * (If you modified it) add/adjust tests to ensure the new search options and columns behave as expected.

**Testing**

All tests in the AV/Fq module pass locally:

```bash
sage -python -m pytest lmfdb/abvar/fq

### DIFF
--- a/lmfdb/abvar/fq/templates/show-abvarfq.html
+++ b/lmfdb/abvar/fq/templates/show-abvarfq.html
@@ -33,19 +33,18 @@
     {% endif %}
     {# --- NEW: cyclicity invariants --- #}
     {% if cl.is_cyclic is not none %}
-    <tr>
-      <td>{{ KNOWL('av.fq.cyclic_group_points', title='Cyclic group of points') }}:</td>
-      <td>&nbsp;&nbsp;
-        {% if cl.is_cyclic == 1 or cl.is_cyclic is true %}
-          yes
-        {% elif cl.is_cyclic == 0 or cl.is_cyclic is false %}
-          no
-        {% else %}
-          unknown
-        {% endif %}
-      </td>
-    </tr>
+      <tr>
+        <td>{{ KNOWL('av.fq.cyclic_group_points', title='Cyclic group of points') }}:</td>
+        <td>&nbsp;&nbsp;
+          {% if cl.is_cyclic %}
+            yes
+          {% else %}
+            no
+          {% endif %}
+        </td>
+      </tr>
     {% endif %}
+
 
     {% if cl.noncyclic_primes %}
     <tr>

### CODE REVIEW

File: lmfdb/abvar/fq/isog_class.py [hunk 1]
```diff
@@ -231,6 +236,20 @@ def properties(self):
             ("Geometrically simple", "yes" if self.is_geometrically_simple else "no"),
             ("Primitive", "yes" if self.is_primitive else "no"),
         ]
+        # Cyclicity information (new)
+        ic = getattr(self, "is_cyclic", None)
```
roed314: Because you set `dbdata["is_cyclic"] = None` if not present in `__init__`, you can just use `ic = self.is_cyclic` here.

lmfdb/abvar/fq/isog_class.py
```diff
@@ -231,6 +236,20 @@ def properties(self):
...
+        if ic is True:
+            cyclic_val = "yes"
+        elif ic is False:
+            cyclic_val = "no"
+        else:
+            cyclic_val = "unknown"
+        props.append(("Cyclic group of points", cyclic_val))
```
roed314: Instead of adding these to the properties box, you should add them to the list of invariants in the main page.  In general, everything in the properties box should be echoed on the main page.  We also choose to only show a subset of the data in the properties box, and I don't think these two pieces of information are central enough to include.

lmfdb/abvar/fq/isog_class.py
```diff
@@ -231,6 +236,20 @@ def properties(self):
...
+
+        primes = getattr(self, "noncyclic_primes", None)
```
roed314: Same comments as above: you can use `self.noncyclic_primes` and this should go in the invariants section not properties box.

lmfdb/abvar/fq/isog_class.py
roed314: (to add them to the list of invariants, you'll need to edit the template `show-abvarfq.html`)

lmfdb/abvar/fq/main.py
```diff
@@ -401,6 +402,16 @@ def nbsp(knowl, label):
             label="Jacobian",
             knowl="ag.jacobian"
         )
+        cyclic = YesNoBox(
+            "cyclic",
```
roed314, line 408: You need to write a knowl for this.

lmfdb/abvar/fq/main.py
```diff
@@ -401,6 +402,16 @@ def nbsp(knowl, label):
...
+            label="Cyclic group of points",
+        )
+        noncyclic_primes = TextBox(
+            "noncyclic_primes",
```
roed314, line 420: And a knowl here as well.

lmfdb/abvar/fq/main.py
roed314: I also think this should be an advanced search box.

lmfdb/abvar/fq/main.py
```diff
@@ -515,7 +527,7 @@ def short_label(d):
             [p, simple],
             [g, geom_simple],
             [initial_coefficients, polarizable],
-            [p_rank, jacobian],
+            [p_rank, jacobian, cyclic],
```
roed314: This needs its own row, and you should have `noncyclic_primes` here as well (you shouldn't usually have search inputs that are available on the search results but not on the browse page)

lmfdb/abvar/fq/main.py
roed314: You should add a mode selector ("include", "exactly", "subset").  See how this is done in number fields for example.

lmfdb/abvar/fq/main.py
```diff
@@ -561,6 +574,7 @@ def common_parse(info, query):
     parse_ints(info, query, "hyp_cnt", qfield="hyp_count", name="Number of Hyperelliptic Jacobians")
     parse_ints(info, query, "twist_count")
     parse_ints(info, query, "max_twist_degree")
+    parse_primes(info, query, "noncyclic_primes", qfield="noncyclic_primes")
```
roed314: Once you add the mode selector, it should get passed in to `parse_primes` here.

lmfdb/abvar/fq/test_browse_page.py
```diff
@@ -279,3 +280,30 @@ def test_search_combos(self):
...
         self.check_args("/Variety/Abelian/Fq/?g=3&jacobian=no&polarizable=yes", "3.2.a_ac_a")
+    
+    def test_search_cyclic_group(self):
+        r"""
+        Check that we can restrict to cyclic or non-cyclic groups of points
+        using the is_cyclic search parameter.
+        """
+        # We only check that the search runs and returns the usual search
+        # results page; we do not assert specific labels since the data
+        # on devmirror can change.
```
roed314: I'm not sure what you mean by data changing; I think including specific labels in this test makes sense.

lmfdb/abvar/fq/test_browse_page.py
```diff
@@ -279,3 +280,30 @@ def test_search_combos(self):
...
+            "/Variety/Abelian/Fq/?is_cyclic=no",
+            "Abelian variety search results"
+        )
+
+    def test_search_noncyclic_primes(self):
+        r"""
+        Check that the noncyclic_primes search parameter is accepted.
+        """
+        self.check_args(
+            "/Variety/Abelian/Fq/?noncyclic_primes=2",
```
roed314, line 338: Same here: I would include specific labels in the test.

roed314 (changes requested):
Good start!  Let me know if you have any questions about these suggestions.

lmfdb/abvar/fq/test_browse_page.py
hgu2699: > Same here: I would include specific labels in the test.

changed all these files. review requested

lmfdb/abvar/fq/templates/show-abvarfq.html
```diff
@@ -31,6 +31,28 @@ <h2>Invariants</h2>
     {% if cl.size is not none %}
     <tr><td>{{KNOWL('av.fq.isogeny_class_size',title='Isomorphism classes')}}:</td><td>&nbsp;&nbsp;{{cl.size}}</td></tr>
     {% endif %}
+    {# --- NEW: cyclicity invariants --- #}
+    {% if cl.is_cyclic is not none %}
+    <tr>
+      <td>{{ KNOWL('av.fq.cyclic_group_points', title='Cyclic group of points') }}:</td>
```
roed314: The knowl av.fq.cyclic_group_points doesn't exist.  Send me a message on Zulip if you need an account on beta to be able to create knowls.

lmfdb/abvar/fq/templates/show-abvarfq.html
```diff
@@ -31,6 +31,28 @@ <h2>Invariants</h2>
...
+      <td>&nbsp;&nbsp;
+        {% if cl.is_cyclic == 1 or cl.is_cyclic is true %}
```
roed314: `is_cyclic` is a boolean column, so you don't need to compare with 1 and 0.

lmfdb/abvar/fq/templates/show-abvarfq.html
```diff
@@ -31,6 +31,28 @@ <h2>Invariants</h2>
...
+        {% else %}
+          unknown
+        {% endif %}
+      </td>
+    </tr>
+    {% endif %}
+
+    {% if cl.noncyclic_primes %}
+    <tr>
+      <td>{{ KNOWL('av.fq.noncyclic_primes', title='Non-cyclic primes') }}:</td>
```
roed314, line 51: This knowl also needs to be created.

lmfdb/abvar/fq/templates/show-abvarfq.html
hgu2699: > `is_cyclic` is a boolean column, so you don't need to compare with 1 and 0.

fixed this one. I will message you on Zulip about beta account

lmfdb/abvar/fq/templates/show-abvarfq.html
hgu2699: > This knowl also needs to be created.

The failing jobs (classical modular forms, groups/abstract, modular curves, Hilbert modular forms) all show psycopg2.errors.DiskFull / OutOfMemory errors from devmirror Postgres. The abvar/fq tests pass. This looks like an infrastructure issue rather than something caused by this PR.

AndrewVSutherland: The knowl has been created at https://beta.lmfdb.org/knowledge/show/av.fq.noncyclic_primes but I think it needs to add a sentence clarifying what this means for an isogeny class.  I think it means that ell is a non-cyclic prime for *some* A in the isogeny class, in other words not every A(Fq) has a cyclic ell-Sylow subgroup, but this should be checked.

lmfdb/abvar/fq/templates/show-abvarfq.html
AndrewVSutherland: https://beta.lmfdb.org/knowledge/show/av.fq.cyclic_group_points is there now, but I think it needs an additional sentence explaining what this means for an isogeny class (I think it means that A(Fq) is cyclic for *every* A in the isogeny class, but this should be checked).

lmfdb/abvar/fq/templates/show-abvarfq.html
roed314: Yep, I just changed it.

PR #6773, opened 2025-11-20
Completeness testing

roed314:
This PR adds messages when search results are complete.

See [this search on blue.lmfdb.xyz](https://blue.lmfdb.xyz/NumberField/?signature=[0%2C1]&class_number=17) for example.  There are a bunch of tests of the new functionality in the `test_utils.py` file.

Most of the completeness criteria were drawn from completeness knowls, but some were adapted from code that used to run on @jwj61's hobbes server with a similar goal, which he kindly provided.


jwj61:
Works for me.

Note, there is so much here that I didn't scrutinize the code.

jenpaulhus:
I checked higher genus and groups carefully, and also spot checked a few other sections.  This looks good to me.  Is this about ready to merge?

AndrewVSutherland:
This looks great!  I haven't checked everything one could check but was happy with what I checked.  I'll note that there are many cases where we have complete data but this is not noted (e.g. we have every elliptic curve over Q with good reduction away from 2).  We might want to think about ways to easily add additional completeness criteria for special cases of interest.

PR #6777, opened 2025-11-23
Fix incomplete message in higher genus pages

jenpaulhus:
This fixes some incorrect messages in the higher genus page due to the way I stored data. 

Compare 
https://www.lmfdb.org/HigherGenus/C/Aut/?genus=4&inc_hyper=only
which claims at the top that this may be an incomplete search and 

https://localhost:37777/HigherGenus/C/Aut/?genus=4&inc_hyper=only
where the message is no longer displayed.


jenpaulhus:
@roed314 Once this is ok to merge, can you also then merge it to the branch feeding the blue server so I can confirm this fixes the issue I was seeing there?

Issue #6806, opened 2025-12-22
Improve downloaded search results for elliptic curves over number fields


rvisser7:
When downloading the search results for a set of elliptic curves over number fields $\mathbb{Q}(\alpha)$ (for any language Sage, Magma, Pari, or Oscar) , the data unfortunately doesn't include an easy way to actually use the elliptic curves as objects in the respective language.  Currently, it only contains a string of the latexed Weierstrass equation, which isn't particularly useful when trying to run your own computations on the data.

Example: https://beta.lmfdb.org/EllipticCurve/?field=2.0.111.1&conductor_norm=1
When downloading to Sage all elliptic curves with conductor norm 1 over the base field 2.0.111.1, the file contains the following:

```
columns = ["short_label", "short_class_label", "field_label", "conductor_norm", "rank", "torsion_structure", "cm", "sato_tate_group", "equation"]
data = [
["1.1-a1", "1.1-a", "2.0.111.1", 1, 1, [], 0, "1.2.A.1.1a", "{y}^2+{y}={x}^3+\\left(-a+1\\right){x}^2+\\left(-3a-16\\right){x}+7a-31"],
["1.1-a2", "1.1-a", "2.0.111.1", 1, 1, [], 0, "1.2.A.1.1a", "{y}^2+{y}={x}^3+\\left(-a+1\\right){x}^2+\\left(-563a-1486\\right){x}-14287a+263"],
["1.1-b1", "1.1-b", "2.0.111.1", 1, 1, [], 0, "1.2.A.1.1a", "{y}^2+{y}={x}^3+a{x}^2+\\left(3a-19\\right){x}-7a-24"],
["1.1-b2", "1.1-b", "2.0.111.1", 1, 1, [], 0, "1.2.A.1.1a", "{y}^2+{y}={x}^3+a{x}^2+\\left(563a-2049\\right){x}+14287a-14024"]
]

def create_record(row):
    out = {col: val for col, val in zip(columns, row)}
    return out
```

I think it would definitely be a lot more helpful if the downloaded search results could create an `EllipticCurve` object, as is currently done for elliptic curves over Q! E.g. It would be great to have something similar to what is currently implemented for downloading elliptic curves over Q to Sage:

```
columns = ["lmfdb_label", "lmfdb_iso", "class_size", "conductor", "rank", "torsion_structure", "cm_discriminant", "ainvs"]
data = [
[...]

Issue #4810, opened 2021-10-17
Make elliptic curve search results downloads more useful


AndrewVSutherland:
It would be nice if the search result downloads for elliptic curves included a corresponding list of labels and a make_data function that would create the elliptic curves, as is done for the genus 2 curve search result downloads.

roed314:
See also #4933 for some suggestions on changing how downloads of search results work.

JohnCremona:
I just tried out the Sage version of this and have two suggestions:
   1.  there is a comment in the file 'To create a list of curves, type "curves = make_data()"'.  I suggest that instead we have print( 'To create a list of curves, type "curves = make_data()"') so that the user sees the instruction when they read the file into sage.
   2.  As well as creating the curves from their a-invariants you can set the curve's _lmfdb_label attribute, by adding lines like
`for lab, E in zip(labels, curves): E._lmfdb_label = lab`
Then, for any curve E in the list created, the user can see its label easily, and can even run E.lmfdb_page() to see its LMFDB home page.

I suspect that people using this will ask for more of the data stored for the curve in the LMFDB to be included in the file (rank, generators, etc), which could be done in a follow-up.  Again, these extra properties can be stored in the Sage elliptic curve object.

I will try the Magma and gp version next.

JohnCremona:
Magma and gp version work fine -- but again I suggest having the hint about make_data display when the file is read.  In the case of gp I would suggest  "curves = make_data();" with a semicolon, as otherwise the user sees the list (which ir not pretty).

I'll be happy to review a version with these minor changes.

edgarcosta:
I would not print anything, instead, I would move the `To create` higher up, e.g.:  

```
# Elliptic curves downloaded from the LMFDB on 23 January 2022.
# Query "{'conductor': 11}" returned 3 curves.
#
# To create a list of curves, type "curves = make_data()"
#
#
# Below are two lists, one called labels, and one called data (in matching order).
# Each entry in the data list has the form:
#    [[a1, a2, a3, a4, a6] Weierstrass coefficients]
# defining the elliptic curve y^2 + a1xy + a3y = x^3 + a2x^2 + a4x + a6.

roed314:
Resolved by #6821.

PR #6821, opened 2026-01-16
Add elliptic curve construction for downloaded ECNF search results

rvisser7:
This PR solves issue #6806  (and also the related issue #4810).  When downloading ECNF search results for Sage, Magma, or Pari/GP, each elliptic curve is now explicitly constructed as an EllipticCurve object in the respective language.  This is implemented similarly to elliptic curves over Q, using an `ECNFDownloader` class.

In order to construct each elliptic curve, I've added a hidden `"field_coeffs"` search column which is only shown in the downloaded files, which gives the coefficients of the defining polynomial of the base field $K$.  This column is then used in the function `create_record(row)` to construct each elliptic curve over its respective base field.

As always, any comments or feedback is very welcome! 🙂 

An example when searching for elliptic curves over base field 2.0.111.1 with conductor norm 1:
https://beta.lmfdb.org/EllipticCurve/?field=2.0.111.1&conductor_norm=1
http://localhost:37777/EllipticCurve/?field=2.0.111.1&conductor_norm=1

Sage download:
http://localhost:37777/EllipticCurve/?download=1&query=%7B%27field_label%27%3A+%272.0.111.1%27%2C+%27conductor_norm%27%3A+1%7D&field=2.0.111.1&conductor_norm=1&Submit=sage

Magma download:
http://localhost:37777/EllipticCurve/?download=1&query=%7B%27field_label%27%3A+%272.0.111.1%27%2C+%27conductor_norm%27%3A+1%7D&field=2.0.111.1&conductor_norm=1&Submit=magma

Pari/GP download:
http://localhost:37777/EllipticCurve/?download=1&query=%7B%27field_label%27%3A+%272.0.111.1%27%2C+%27conductor_norm%27%3A+1%7D&field=2.0.111.1&conductor_norm=1&Submit=gp

### DIFF
--- a/lmfdb/ecnf/main.py
+++ b/lmfdb/ecnf/main.py
@@ -414,6 +414,7 @@ modell_image_label_regex = re.compile(r'(\d+)(G|B|Cs|Cn|Ns|Nn|A4|S4|A5)(\.\d+)*(
 class ECNFDownloader(Downloader):
     table = db.ec_nfcurves
     title = "Elliptic curves over number fields"
+    short_name = "curves"
 
     def postprocess(self, row, info, query):
         # Look up the defining polynomial coefficients for the base field of the curve


JohnCremona:
I have tried this out on some random curves, testing all of magma, gp, sage and text, csv.  For the first three I tested that the commands worked.  I did not test Oscar since I don't use that, and someone else should.

 - The CSV option seemed to produce an empty file. Presumably that is not right?

 - magma, gp, sage, oscar: the instructions says something like 
`To create a list of fields, type "fields:= make_data();"`
so the word 'fields' needs to be changed to 'curves' twice.  I think this is done by adding `short_name = curves` after setting `table` and `title`.

 - In the gp case, there is no semicolon after make_data() but I think there should be, as gp echos the output by default and to avoid this one has to put a semicolon.  It is puzzling that this happens, since in downloader.py I see that within GPLanguage we have `line_end = ';'`.

 - The explanatory text includes
> {# Note:  I left out the knowl for degree here, since it is technically defining the degree of a number field, rather than a general field extension (which is suggested by this definition)... degree #}

 which I think it would be better to remove.  It comes from the knowl 'nf' having a comment in it which is copied here.   Can we fix whatever part of downloader.py gets the knowl text to delete any commented out text?

rvisser7:
Many thanks @JohnCremona for reviewing this!  I've implemented your point 2 above.  Given that points 1, 3, 4 all deal with issues particular to the Downloader utils class, I've opened a separate PR #6860 for this.  Perhaps that PR could be reviewed and merged first, and then this one can be merged (though the order doesn't strictly matter).

I haven't implemented anything in the downloads yet for Oscar, as I don't really use Oscar, so I'm not too familiar with the language (and there also doesn't seem to be any Oscar code snippets implemented for ECNF).

PR #6823, opened 2026-01-18
First pass at removing extras

jwj61:
This makes changes corresponding to the change from an extras table and its replacement as a regular table.  To properly test this, we need to delete the old extras table (or we will be getting its data when we pull data for a number field).  I think the steps should be:

 1. copy nf_fields and nf_fields_extra to the cloud version of the database
 2. merge this pull request
 3. delete the extras table for number fields for beta
 4. test this


edgarcosta:
> copy nf_fields and nf_fields_extra to the cloud version of the database
done

Issue #6833, opened 2026-01-27
ECQ: show which of a curve's base-changes are in the database, if any


JohnCremona:
See #3837 for related discussion.

On each home page of an elliptic curve over Q in the Related Objects box we should list the labels of (as links) all of its base-change curves in the ECNF database.   This requires no database changes, as the ec_nfcurves table has a column 'base_change' which lists the labels of any curves /Q of which it is the base change, and these are listed in its Related Objects box.

PR #6854, opened 2026-02-02
ECQ: show base-changed curves in database on curve home pages

JohnCremona:
This is a draft PR towards satisfying the feature request #6833 .  It implements both options (1) and (2) defined there, though the implementation of (2) does not work properly.  My intention is that someone fixes (2) and then we remove option (1).


AndrewVSutherland:
Rerunning the test fixed the problem (presumably a different random curve).

AndrewVSutherland:
Should we close this in view of #6878?

Issue #6190, opened 2024-09-26
Add ell-adic index to Galois rep tables on ECQ pages


AndrewVSutherland:
Currently the "Galois representations" section of home pages for elliptic curves over Q includes a table that gives the ell-adic image for each prime ell where the image is not [maximal](https://www.lmfdb.org/knowledge/show/ec.maximal_elladic_galois_rep); note that this applies to CM curves as well as non-CM curves.

In the non-CM case the index of the ell-adic representation in GL_2(Z_ell) is part of the label that identifies the image, but in the CM case it is not obvious what the index is (relative to the maximal image for an elliptic curve with CM by the same order, as defined in section 12 of [RSZB](https://arxiv.org/pdf/2106.11141), and even in the non-CM case it is only obvious if one knows what the label format is.

I think it would be helpful to list the index explicitly as a separate column in the table of non-maximal ell-adic representations.  For non-CM curves this can be extracted directly from the second component of the image label, and for CM curves with j(E) != 0,1728 it will always be 2 (it is either 1 or 2, and if it is 1 the image is maximal and will not be listed).  To handle j(E) = 0,1728 it would suffice to add an additional table which gives the index of the maximal image at prime power levels for all the relevant prime powers (there won't be more than a few hundred of these).

alozanoroble:
Agree with Drew here that it would be useful to list the index explicitly, just like the adelic index is listed explicitly.

PR #6863, opened 2026-02-06
display ell-adic index for j=0,1728 curves

AndrewVSutherland:
The final step of addressing #6190.  @alozanoroble if you have time it would be great if you could take a look at this on https://blue.lmfdb.xyz/EllipticCurve/Q/, especially the j=0,1728 cases.

### DIFF
--- a/lmfdb/elliptic_curves/web_ec.py
+++ b/lmfdb/elliptic_curves/web_ec.py
@@ -399,14 +399,15 @@ class WebEC():
             if self.cm:
                 if self.cm == -3:
                     # for j=0 nonmaximal elladic-images have index 3 for ell > 3 (see https://arxiv.org/abs/1809.02584v3 Theorem 1.4)
-                    # for ell=2,3 the index can very but is always determined by the image mod 16,27, respectively
+                    # for ell=2,3 the index can vary but is always determined by the image mod 16,27, respectively
                     # the dictionary below lists the index of the maximal image for j=0 modulo relevant powers of 2,3
                     j0_elladic_indices = { 2:1, 3:4, 4:4, 8:16, 9:36, 16:64, 27:324 }
-                    gd['elladic_index'] = 3 if gd["prime"] > 3 else int(s[1]) // j0_elladic_indices[int(s[0])]
+                    gd['elladic_index'] = 3 if gd['prime'] > 3 else int(s[1]) // j0_elladic_indices[int(s[0])]
                 elif self.cm == -4:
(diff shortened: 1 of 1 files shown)


lmfdb/elliptic_curves/web_ec.py
```diff
@@ -451,13 +395,24 @@ def red(p):
         galois_data = list(db.ec_galrep.search({'lmfdb_label': lmfdb_label}))
         data['galois_data'] = [r for r in galois_data if r["prime"] > 0]
         for gd in data['galois_data']:
+            s = gd['elladic_image'].split('.')
             if self.cm:
-                if self.cm in [-3,-4]:
-                    gd['elladic_index'] = '?'
+                if self.cm == -3:
+                    # for j=0 nonmaximal elladic-images have index 3 for ell > 3 (see https://arxiv.org/abs/1809.02584v3 Theorem 1.4)
+                    # for ell=2,3 the index can very but is always determined by the image mod 16,27, respectively
```
roed314: This should be "vary" not "very"

lmfdb/elliptic_curves/web_ec.py
```diff
@@ -451,13 +395,24 @@ def red(p):
...
+                    # the dictionary below lists the index of the maximal image for j=0 modulo relevant powers of 2,3
+                    j0_elladic_indices = { 2:1, 3:4, 4:4, 8:16, 9:36, 16:64, 27:324 }
+                    gd['elladic_index'] = 3 if gd["prime"] > 3 else int(s[1]) // j0_elladic_indices[int(s[0])]
+                elif self.cm == -4:
+                    # for j=1728 the elladic image is maximal except possibly at 2 (see https://arxiv.org/abs/1809.02584v3 Theorem 1.2(4))
+                    # for ell=2 the index can vary but is always determined by the image mod 16
+                    # the dictionary below lists the index of the maximal image for j=1728 modulo relevant powers of 2
+                    j1728_elladic_indices = { 2:3, 4:6, 8:24, 16:96 }
+                    gd['elladic_index'] = int(s[1]) // j1728_elladic_indices[int(s[0])]
```
roed314, line 412: Comparing with the case above, shouldn't this be something like `1 if gd["prime"] > 2 else int(s[1]) // j1728_elladic_indices[int(s[0])]`?

roed314: Or will other primes not show up because they're automatically maximal?

PR #6865, opened 2026-02-10
Add Oscar code snippets for elliptic curves over number fields

rvisser7:
This PR adds some basic code snippets in Oscar, for elliptic curves over number fields.  When downloading the Oscar code snippets to file, a note about running `using Oscar;` is also added at the top, both for ECQ and ECNF  (as is currently done for number fields).

**Update**:  The frontmatter message to run `using Oscar;` is now added to all Oscar download files.

Comments/feedback very welcome :)

E.g. Elliptic curve 11.1-a1  
https://beta.lmfdb.org/EllipticCurve/2.0.11.1/11.1/a/1
http://localhost:37777/EllipticCurve/2.0.11.1/11.1/a/1

E.g. Elliptic curve 81.1-CMa1
https://beta.lmfdb.org/EllipticCurve/2.0.3.1/81.1/CMa/1
http://localhost:37777/EllipticCurve/2.0.3.1/81.1/CMa/1

### DIFF
--- a/lmfdb/elliptic_curves/code.yaml
+++ b/lmfdb/elliptic_curves/code.yaml
@@ -33,9 +33,6 @@ not-implemented:
 frontmatter:
   all: |
     {lang} code for working with elliptic curve {label}
-  oscar: |
-    If you have not already loaded the Oscar package, you should type "using Oscar;" before running the code below.
-
 
 curve:
   comment: Define the curve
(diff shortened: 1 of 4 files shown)


lmfdb/elliptic_curves/code.yaml
```diff
@@ -33,6 +33,9 @@ not-implemented:
 frontmatter:
   all: |
     {lang} code for working with elliptic curve {label}
+  oscar: |
+    If you have not already loaded the Oscar package, you should type "using Oscar;" before running the code below.
```
edgarcosta: we should perhaps do this globally?

rvisser7: thanks, I agree!  I've now put the message to load Oscar in the `utils/place_code.py` file  (and removed it from the code.yaml files).

Issue #6686, opened 2025-09-12
The spacing for the isogeny graph is too large on the isogeny class page


JaneShi99:
For example, on this page, the isogeny graph is only two nodes and an edge, but it takes a large part of the page.

https://www.lmfdb.org/EllipticCurve/4.4.14013.1/29.1/c/
https://www.lmfdb.org/EllipticCurve/Q/113/a/

rvisser7:
I've had a go at looking into this issue.  It does seem that the excess whitespace is coming from Sage's default graph plotting (rather than a consequence of any frontend on our side, I think).  E.g. checking `lmfdb/elliptic_curves/isog_class.py`, the following Sage code should essentially be the same as how we produce the isogeny graph for the [113.a](https://www.lmfdb.org/EllipticCurve/Q/113/a/) isogeny class:

```
G = Graph()
G.add_edge(("a1", "a2", 2))
G.set_pos(pos={"a1":[-0.5,0],"a2":[0.5,0]})
P = G.plot(edge_labels=True, vertex_size=1000)
P.show(axes_pad=0.1)
```

Running this in Sage v10.6, there's already a lot of excess whitespace above and below the graph.  I would say there's essentially two ways to fix this, i.e. running a fix either before or after the png is rendered:

 - **Option 1** (pre-render fix):  We could manually set the x and y-axis limits for `self.graph`, e.g. running `P.set_axes_range(...)` , which seems to fix the issue.  Though I haven't quite figured out yet if there's a simple solution which will globally handle all possible isogeny classes in a nice way.  The issue is that some isogeny graphs will be naturally take up more vertical spacing and some will take up more horizontal spacing, so I imagine we probably wouldn't want to set a fixed width or height (or a fixed aspect ratio) for all possible graphs?  Though since we're already hardcoding the isogeny graphs for small cases (which are exactly the cases causing the most excess whitespace), it's certainly doable to also just hardcode the axes limits for small cases.

 - **Option 2** (post-render fix):  Alternatively, we could just use the Python Imaging Library (PIL) to crop the surrounding whitespace, after Sage has rendered a png for the graph.  E.g. one could implement something to similar to [this post](https://stackoverflow.com/questions/10615901/trim-whitespace-using-pil).  I've tested this on some sample isogeny graphs and it seems to work quite well!

[...]

Issue #2801, opened 2019-03-19
Isogeny graphs for elliptic curves/Q


davidfarmer:
Currently the isogeny graph is a static image.

If it were stored as data, then it could be displayed as an svg that shows more
information and also lets people interact with it.

For example, the \Gamma_0(N)-optimal curve is currently not distinguished in the graph.

Currently only the final number on the label is displayed.

Other information, such as the torsion order or the modular degree, could be
encoded in the image.

And the curves in the image could be clickable.

Also, the current image is not clopped closely, which wastes a lot of space
(particularly when the graph is small).

JohnCremona:
Excellent points which I have also been discussing with @roed314 .
Having fuller labels would probably be easy, also possibly marking the optimal curve.  The rest is harder but a good project for someone.  I once spent half a day trying to crop the blank space around the images but was completely unsuccessful.  I'm sure that someone with more expertise would manage that easily.

davidfarmer:
I am working on this

JohnCremona:
> I am working on this

Good!   If all goes well for EC/Q please also try your hand at the larger graphs over number fields.  There's a gallery of interesting cases at http://homepages.warwick.ac.uk/staff/J.E.Cremona/isogeny_graphs.html

jvoight:
Another thing to tack on to this (which should be easy): a speaker at Quebec-Maine (Garen Chiloyan at University of Connecticut) just complained that when the isogeny graph is trivial (single vertex), it does not display.  I basically agree, can we display that it is trivial?  That is (redundant) information that the user may reasonably expect to appear.

JohnCremona:
Of course we can display the trivial graph.  I seem to remember that the decision not to was based on a previous user's comment that it was superfluous.  Just delete the if... test in the templae for those pages.

edgarcosta:
The just deleting the if statement is a good first issue, hence, the label.

JohnCremona:
> The just deleting the if statement to display the trivial graph is a good first issue, hence, the label.

-- but the part which @davidfarmer is working on is not at all trivial.  So the trivial first issue would have to be moved to another issue.

davidfarmer:
Here is a screen shot of what I am working on:

https://aimath.org/~farmer/tmp/isograph.png

JohnCremona:
> Here is a screen shot of what I am working on:
> 
> https://aimath.org/~farmer/tmp/isograph.png

[...]

PR #6874, opened 2026-02-13
Replace static PNG isogeny graphs with interactive Cytoscape.js

edgarcosta:
A draft attempt to resolve excessive whitespace in isogeny graphs:  https://github.com/LMFDB/lmfdb/issues/6686  and https://github.com/LMFDB/lmfdb/issues/2801

There is some cleanup to do, but here are some samples:
⚫  : https://olive.lmfdb.xyz/EllipticCurve/Q/37/a/ (to be improved, but how?)
`*-*` : https://olive.lmfdb.xyz/EllipticCurve/Q/26/b/
`*-*-*`: https://olive.lmfdb.xyz/EllipticCurve/Q/11/a/
`*-*-*-*`: https://olive.lmfdb.xyz/EllipticCurve/Q/27/a/ 
T : https://olive.lmfdb.xyz/EllipticCurve/Q/17/a/
🔲 : https://olive.lmfdb.xyz/EllipticCurve/Q/20/a/
🔲 🔲 : https://olive.lmfdb.xyz/EllipticCurve/Q/14/a/
`>-<` : https://olive.lmfdb.xyz/EllipticCurve/Q/21/a/
🐙 : https://olive.lmfdb.xyz/EllipticCurve/Q/15/a/
🐙 : https://olive.lmfdb.xyz/EllipticCurve/Q/30/a/ (something went wrong)

💥 : https://olive.lmfdb.xyz/EllipticCurve/4.4.2304.1/1.1/a/

This branch is being tracked in olive.lmfdb.xyz 

@havarddj ready for your comments!

### DIFF
--- a/lmfdb/elliptic_curves/isog_class.py
+++ b/lmfdb/elliptic_curves/isog_class.py
@@ -95,6 +95,7 @@ class ECisog_class():
                 c['optimal'] = None
                 c['optimality_known'] = False
         for c in self.curves:
+            c.setdefault('degree', 0)
             c['ai'] = c['ainvs']
             c['curve_url_lmfdb'] = url_for(".by_ec_label", label=c['lmfdb_label'])
             c['curve_url_cremona'] = url_for(".by_ec_label", label=c['Clabel']) if self.conductor < CREMONA_BOUND else "N/A"


roed314:
These are cool @edgarcosta!  Following up on a few of Robin's comments:
* I would use just the short label (like `a1`) and omit the conductor part.
* I agree that we probably don't need draggable notes in this application, but there are other cases where it would be helpful.
* the popups with info are great!

As we refine this, it would be nice to see how the subgroup diagrams look in this system.

JohnCremona:
This looks very good indeed -- thanks, @edgarcosta !

I don't see the need for having individual vertices draggable.  For the larger diagrams (such as the last example shown) I would like the ability to rotate the whole diagram in 3D-graphics style -- this is what you get when you create an isogeny class in Sage directly (implemented by me a while ago) but I never worked out how to store a 3D graphics object which we could display in the LMFDB.  This is absolutely not essential.

I agree that we should be consistent in how much of the labels we use to decorate the graph nodes.  In fact only the number at the end of the label is varying across the class so a minimal version would be to just have that (perhaps omitting the single '1' if the class is trivial).

Having the pop-up info for the curves in the class is great.

havarddj:
Looks great! Here are a couple of comments: 
1. The max-width for the tooltips should be a bit bigger; also, currently the tooltips are not allowed to expand beyond the parent div, so the left-most tooltip box is often too narrow. I'm sure your LLM can fix that!
Screenshot 2026-02-20 at 16 09 36
2. When there's no torsion, perhaps it should say "[0]" or "None" instead of "[]", e.g. in https://olive.lmfdb.xyz/EllipticCurve/Q/37/a/
3. A really cool feature would be if hovering over a node would also highlight the corresponding entry in the table below. It shouldn't be too hard, but perhaps it's not worth the extra complexity.

I'll add some comments in the code as well.

lmfdb/static/isogeny_graph.js
```diff
@@ -0,0 +1,286 @@
...
+
+    } // end nNodes > 1
+
+    // Create tooltip element
+    var tooltip = document.createElement('div');
+    tooltip.className = 'isogeny-tooltip';
+    tooltip.style.cssText = 'display:none; position:absolute; background:#fff; ' +
+        'border:1px solid #ccc; border-radius:4px; padding:8px 12px; ' +
+        'font-size:13px; line-height:1.5; box-shadow:0 2px 8px rgba(0,0,0,0.15); ' +
+        'z-index:1000; pointer-events:none; max-width:280px;';
```
havarddj: I don't think it's necessary to set max-width (or at least, you can set it quite big without too much harm; I would always prefer the data to be on a single line)

lmfdb/static/isogeny_graph.js
```diff
@@ -0,0 +1,286 @@
...
+                    'background-color': '#e8f0fe'
+                }
+            },
+            {
+                selector: 'edge',
+                style: {
+                    'label': 'data(label)',
+                    'font-size': '11px',
+                    'text-background-color': '#fff',
+                    'text-background-opacity': 0.85,
```
havarddj: I prefer 'text-background-opacity: 1', so you don't get the ghost lines

lmfdb/static/isogeny_graph.js
```diff
@@ -0,0 +1,286 @@
...
+                    'font-size': '12px',
+                    'font-family': 'sans-serif',
+                    'background-color': '#fff',
+                    'border-width': 2,
+                    'border-color': '#555',
+                    'width': 60,
+                    'height': 30,
+                    'shape': 'round-rectangle',
+                    'color': '#333',
+                    'cursor': 'pointer'
```
havarddj: This line doesn't seem to do anything. Instead, you should change pointer on mouseover events, I'll comment below.

lmfdb/static/isogeny_graph.js
```diff
@@ -0,0 +1,286 @@
...
+        if (d.faltings_height !== undefined) _isoTooltipLine(tooltip, 'Faltings height: ' + d.faltings_height);
+        if (d.cm !== undefined) _isoTooltipLine(tooltip, 'CM: ' + d.cm);
+        if (d.optimal) _isoTooltipEmphasis(tooltip, 'Optimal curve');
+
+        tooltip.style.display = 'block';
+
+        // Position tooltip near the node
+        var pos = node.renderedPosition();
+        tooltip.style.left = (pos.x + 15) + 'px';
+        tooltip.style.top = (pos.y - 10) + 'px';
```
havarddj: add `$('html,body').css('cursor', 'pointer');` here

lmfdb/static/isogeny_graph.js
```diff
@@ -0,0 +1,286 @@
...
+    });
+
+    cy.on('mouseout', 'node', function() {
+        tooltip.style.display = 'none';
```
havarddj, line 269: Add `$('html,body').css('cursor', 'default');` here

lmfdb/utils/utilities.py
```diff
@@ -951,6 +951,112 @@ def encode_plot(P, pad=None, pad_inches=0.1, remove_axes=False, axes_pad=None, f
     return "data:image/png;base64," + quote(b64encode(buf))
 
 
+def graph_to_cytoscape_json(G):
```
havarddj: Perhaps it makes sense to put this in a separate file?

lmfdb/templates/cytoscape_scripts.html
```diff
@@ -0,0 +1,23 @@
+{# Shared Cytoscape.js scripts for interactive isogeny graphs.
```
havarddj, line 1: While it's nice to have the options, perhaps all of these are a bit overkill? We could restrict to 5, and give them more meaningful names.

lmfdb/static/isogeny_graph.js
```diff
@@ -0,0 +1,286 @@
...
+    });
+
+    // Prevent dragging nodes outside the visible area
+    cy.on('drag', 'node', function(evt) {
+        var node = evt.target;
+        var pos = node.position();
+        var ext = cy.extent();
+        var hw = node.width() / 2;
+        var hh = node.height() / 2;
```
havarddj: add +10 to these to make account for the shadow around the box

Issue #965, opened 2016-03-31
base-change of elliptic curves over Q


AurelPage:
From /roadmap: Base change: the user could select an extension field and see the page for the base-changed curve.  This would be easy for cases where the base-change curve is already in the database.

Currently, from elliptic curves over number fields that are base-change, there is a link to the curve over Q. This asks for a link in the opposite direction.

JohnCremona:
We don't want to allow users to create arbitrary base-changes, and I think it would be too expensive to do anything like this on the fly.  One possibility is to do a one-off computation which went through all curves E/Q in the database, and all number fields F for which we have any curves, compute the base-change, see if that is in the database and if so store its label.  Or use the fact that every E/F already knows if it s base-change (see example below).  Then every E/Q would have stored a list of labels of its base-changes in the database.  To keep this up-to-date, two things need to be done: (1) when extra curves /Q are added, this procedure needs to be done for them; (2) when extra curves over fields F/Q are added which are base-change (something which is known at upload time) the appropriate ec/Q records could be updated.

This would be quite a lot of work.

Then the home page of any E/Q which has any base-changes in the database could show them.  For example, there are 44 different base-changes of 11a1:

> sage: len(list(db.ec_nfcurves.search({'base_change':{'$contains':'11.a1'}})))
44

JohnCremona:
In fact it turns out to be rather easy to get a list, for each EC/Q, of which EC/F are its base-changes in the database.  Doing this for 1 curve from each isogeny class reveals that out of 1.74 million isogeny classes, only 5221 have base-changes present.  A simple rewrite (after creation of a new column) would record these lists for all curves.

PR #6878, opened 2026-02-14
add base_change_label search option to ecnf

AndrewVSutherland:
This PR adds the ability to seach for elliptic curves over number fields that are base changes of a specific curve (which could be defined over Q or any number field, it simply checks whether the specified label is one of the labels listed in the `base_change` column).  This is meant to facilitate the implemenation of #965 via #6854, as well as providing a way to access this capabilitiy directly from the ECNF search page.

@JohnCremona I think this will let you do a cleaner implementation of (2) in #6854 that avoids searching the ECNF table preemptively; the search for base changes will only get triggered if the user clicks the base change link in the related objects box.

### DIFF
--- a/lmfdb/ecnf/main.py
+++ b/lmfdb/ecnf/main.py
@@ -27,6 +27,7 @@ from lmfdb.number_fields.web_number_field import nf_display_knowl
 from lmfdb.sato_tate_groups.main import st_display_knowl
 from lmfdb.ecnf import ecnf_page
 from lmfdb.ecnf.ecnf_stats import ECNF_stats
+from lmfdb.elliptic_curves.web_ec import cremona_label_to_lmfdb_label
 
 from lmfdb.ecnf.WebEllipticCurve import (ECNF, web_ainvs, LABEL_RE,
                                          CLASS_LABEL_RE,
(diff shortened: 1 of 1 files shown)


AndrewVSutherland:
@JohnCremona Sorry if I wasn't clear in what I wrote above, I went ahead and added a "Base changes" link to the ECQ related objects pages in this PR so you can better understand what I meant.  The link will only appear for curves that have a base change in the ECNF database -- this does require one DB lookup, but it is just checking existence, and is quite quick (especially because it includes the constraint q_curve=True, which as you surmised, speeds up the $contains search by reducing the number of rows it needs to check).

If you are happy with this you could just this PR and close #8654 (but feel free to modify either or tell me what to change in this one).

lmfdb/ecnf/main.py
```diff
@@ -534,6 +533,13 @@ def elliptic_curve_search(info, query):
         elif info['Qcurves'] == 'non-base-change-Q-curve':
             query['q_curve'] = True
             query['base_change'] = []
+    if 'base_change_label' in info:
+        if "Qcurves" in info and info['Qcurves'][:3] == 'non':
+            err = "Incompatible base change query options, you cannot simultaneous exclude base changes and specify a base change label."
+            flash_error(err)
+            raise ValueError(err)
+        query['q_curve'] = True
+        query['base_change'] = { '$contains': info['base_change_label'] }
```
JohnCremona: At this point we chould convert any Cremona labels to LMFDB labels since (after tidying up the base_change column of ec_nfcurves) that column will only contain Clabels, but users may choose to enter Clabels.  The function `cremona_label_to_lmfdb_label()` in `web_ec.py` will do that (and will leave unchanged any LFMDB labels).

AndrewVSutherland: done

PR #6879, opened 2026-02-14
Catch errors in completeness framework

roed314:
With this PR, errors in the completeness framework print a warning message and log an error to the flasklog, instead of causing a server error.  We can find and fix the underlying problems by examining the flasklog.

Compare:
* https://www.lmfdb.org/EllipticCurve/Q/?conductor_type=multiple&conductor=7
* http://localhost:37777/EllipticCurve/Q/?conductor_type=multiple&conductor=7

* https://www.lmfdb.org/ModularForm/GL2/Q/holomorphic/?level_type=multiple&level=23&weight=1
* http://localhost:37777/ModularForm/GL2/Q/holomorphic/?level_type=multiple&level=23&weight=1

### DIFF
--- a/lmfdb/utils/search_wrapper.py
+++ b/lmfdb/utils/search_wrapper.py
@@ -5,7 +5,7 @@ from psycopg2.errors import NumericValueOutOfRange
 from sage.misc.decorators import decorator_keywords
 from sage.misc.cachefunc import cached_function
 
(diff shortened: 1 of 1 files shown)


AndrewVSutherland:
This works, but I note that there are no timestamps in the flasklog.  I don't think it's worth delaying this to fix/change that so I am merging.

PR #6904, opened 2026-02-27
Diagram search (using d3.js)

havarddj:
This PR aims to implement searching via interactive pictures using the LMFDB, implemented using [d3.js](https://d3js.org/). 

TLDR: 
Screenshot 2026-02-27 at 18 05 16

This is still work in progress, please report any bugs you find here! 

## Examples
Examples to play with (not implemented for all pages, but many more!):

- https://pink.lmfdb.xyz/Genus2Curve/Q/?search_type=Diagram
- https://pink.lmfdb.xyz/EllipticCurve/Q/?search_type=Diagram
- https://pink.lmfdb.xyz/L/?hst=Diagram&origin=ECQ&x-axis=conductor&y-axis=z1&color=root_angle&count=1000
- https://pink.lmfdb.xyz/Groups/Abstract/?hst=Diagram&x-axis=order&y-axis=exponent&color=rank&count=1000
- https://pink.lmfdb.xyz/NumberField/?hst=Diagram&x-axis=rd&y-axis=regulator&color=num_ram&count=1000

If anyone want to get involved, that would be awesome! There are a number of wrinkles to iron out.

## How it's made
The html and js can be found in `lmfdb/templates/d3_diagram.html`. This includes the `refine_search_form.html` template to avoid reimplementing search from scratch, then displays the diagram with the relevant javascript (somewhat vibe-coded).

To generate searches, we add a new search type in `lmfdb/utils/search_boxes.py`, and add an extra method `_diagram_search` to `SearchWrapper` in `lmfdb/utils/search_wrapper.py`. This does a custom search and sets the `d3-data` field of `info` with the numerical data, which is then passed to javascript (via flask's `tojson`) in the template file.

One of my goals when making this was to make it as general-purpose as possible. Do a ctrl+f for "diagram" in `lmfdb/lfunctions/main.py` to get an idea of how it works.

## Known issues and TODOs
- [x] Currently no way to get to diagram search from other pages (requires some discussion)
- [x] Inconsistent naming of routes ("diagram_search" vs "diagrams") - this will be fixed shortly
- [x] Overlapping points - no way to see multiplicities. 
- [ ] Javascript should be moved from `d3_diagrams.js` to a separate file
- [x] The number of search results is currently limited to however many html elements your browser can handle. This could perhaps be improved by using a canvas, but there is a limitation here (for example, drawing all elliptic curves in the database will probably not work). 
- [x] More space above diagram
- [x] GL2-type in color on ECQ is weird
- [x] y-axis = torsion_order in genus2 crashes
- [ ] Consider if setting x/y-axes & color should be hidden behind advanced options (what are good defaults?)
- [x] "hash" field for groups is strange (crashes and shouldn't be there)
- [x] adelic image key error for ECQ
- [ ] adelic level/index/genus have weird dots at the top of the diagram
- [ ] Tooltips should be contained in diagram for dots "high up"
- [ ] Make sure we don't pull more columns from the relevant table than those required for the diagram (minor speedup?)
- [ ] Fix aligment of boxes below diagram

**Future features:**
- [ ] Add (optional) gridlines
- [ ] Allow for different dots, relative size (e.g for multiplicities)
- [ ] Double click to full-screen (or almost full-screen)
- [ ] Select multiplicity as color option
- [ ] Searching for range(s) from home page should make those the x/y-axes by default

### DIFF
--- a/lmfdb/utils/search_wrapper.py
+++ b/lmfdb/utils/search_wrapper.py
@@ -359,6 +359,13 @@ class SearchWrapper(Wrapper):
     def _diagram_search(self, info):
         """
         Handle diagram search mode, displaying results in d3.js visualization.
+
+        Diagram search will be available automatically on pages by default, but can be
+        disabled by adding 'has_diagram = False' to the relevant SearchArray subclass which handles search.
+
+        Additionally, one should pass 'diagram_opts = {}' as keyword argument to the @search_wrap macro, 
+        with options specifying the title, breadcrumbs, and default x/y-axes and color keys, matching numerical
+        (or in the case of 'color_default', boolean) columns in the database.
         """
         # Get diagram options with defaults
         opts = self.diagram_opts or {}
--- a/lmfdb/belyi/main.py
+++ b/lmfdb/belyi/main.py
@@ -888,6 +888,7 @@ def labels_page():
 class BelyiCommonSearchArray(SearchArray):
     jump_knowl = "belyi.search_input"
     jump_label = "Label"
+    has_diagram = False
 
     def __init__(self):
         self.deg = TextBox(
--- a/lmfdb/higher_genus_w_automorphisms/main.py
+++ b/lmfdb/higher_genus_w_automorphisms/main.py
@@ -1281,6 +1281,7 @@ class HGCWASearchArray(SearchArray):
         'full_label': False,
         'full_auto': False,
     }
+    has_diagram = False
 
     def __init__(self):
         genus = TextBox(
--- a/lmfdb/number_fields/test_numberfield.py
+++ b/lmfdb/number_fields/test_numberfield.py
@@ -130,13 +130,3 @@ class NumberFieldTest(LmfdbTest):
         # Also ensure we're not getting quoted strings
         assert '"[0, 1]"' not in page
         assert '"[2, 0]"' not in page
-
-    def test_diagram_search(self):
-        """
-        Check that diagram search page loads correctly
-        """
-        L = self.tc.get('/NumberField/?degree=2&search_type=Diagram&count=100')
-        data = L.get_data(as_text=True)
-        # Check that D3 diagram template elements are present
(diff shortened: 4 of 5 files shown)


lmfdb/utils/search_wrapper.py
```diff
@@ -374,7 +411,9 @@ def __call__(self, info):
                                     res[row, col] = 0
                                 else:
                                     res[row, col] = None
-                info['count'] = 50 # put count back in so that it doesn't show up as none in url
+                info["count"] = (
+                    50  # put count back in so that it doesn't show up as none in url
+                )
```
roed314: I'm not a fan of the line break here; just leave this on one line. 😄

lmfdb/utils/search_wrapper.py
```diff
@@ -539,18 +587,222 @@ def __call__(self, info):
...
+        # TODO: make sure we only search for fields that we actually need!
+
+        # It's fairly common to add virtual columns in postprocessing that are then used in MultiProcessedCols.
+        # These virtual columns won't be present in the database, so we just strip them out
+        # We have to do this here since we didn't have access to the table in __init__
+        proj = query.pop("__projection__", self.projection)
+        if isinstance(proj, list):
+            proj = [col for col in proj if col in table.search_cols]
+
+        count = parse_count(info, self.result_count_default)
```
roed314: We should probably enforce an upper limit here to avoid people trying to render 1,000,000 HTML elements, but I'm not sure what the limit should be.

havarddj: I think 20 000 is reasonable; my computer struggled with 50 000 a while ago when I tried. Going a little under that is probably not a bad idea.

lmfdb/utils/search_wrapper.py
havarddj: Nitpicking! But I agree.

AndrewVSutherland:
Looks good!

PR #6919, opened 2026-03-05
Adding messages explaining why these are in alpha

assaferan:
For Hilbert modular surfaces, Eisenstein classical modular forms and Siegel modular forms


roed314:
Looks good.  I'm confused why tests didn't run though...

Issue #2479, opened 2018-06-07
Uniform breadcrumbs


jenpaulhus:
Some issues left unresolved from pull request #2466. 

Maass Waveforms is a breadcrumb on the following pages which goes to the Modular Forms page, not to the current page. (http://127.0.0.1:37777/ModularForm/GL2/Q/Maass/BrowseGraph/1/10/0/10/ http://127.0.0.1:37777/ModularForm/GL2/Q/Maass/4f5558f688aece2646000023)

"Elliptic Curves" bread in "Elliptic Curves over Q" links to "Elliptic Curves over Number Fields" (http://127.0.0.1:37777/EllipticCurve/) but so does the "Elliptic Curves" bread in "Elliptic Curves over Number Fields". Do we want to add an Elliptic Curves page like the Modular Forms page?

There were a few places where we thought the breadcrumb should be plural, like: Lattice(s) Elliptic Curve(s). For example, Lattice is on most of the breadcrumbs, but the search results have the breadcrumb as "Lattices". We did not change these yet, although we have the locations marked if you would like this changed implemented.

Two pages still under construction: Hecke Characters don't have any breadcrumbs visible (they are listed as "HeckeAlgebra" in the code and we did not add a space between the two words yet) and Hypergeometric does not have breadcrumbs for Completeness, Labels, and Source in Learn more about.

AndrewVSutherland:
I believe this issue has been addressed for Maass forms, and Elliptic Curves is plural.  Things that remain to be done are:
- Top level page for elliptic curves with links to EC/Q and EC/NF that could be used in elliptic curve breadcrumbs.
- Lattices should be plural in bread crumbs
- Hypergeometric RCS pages now have bread crumbs, but the last crumb is incomplete (C/R/S should be Completeness/Reliability/Source).
It might make more sense to create separate PRs for each of these but I won't do that now.

jwj61:
I think the only remaining issue is that with elliptic curves over Q, clicking on the "elliptic curve" bread takes you to elliptic curves over number fields.  Maybe that page should have a link in the browse area to take the user to the elliptic curves over Q page.

PR #6941, opened 2026-03-20
Elliptic Curve landing page

jenpaulhus:
This PR adds a landing page for elliptic curves (similar to the landing page for modular forms) and finishes fixing issue #2479 

http://localhost:37777/EllipticCurves/

It could use some feedback before it is merged.

For EC/Q things work pretty smoothly.  The "Elliptic curves" bread goes to the page above (click on "Elliptic curves" in the bread on any EC page).

For ECNF, this does change the behavior of the bread a bit.  It adds a Q(\alpha) bread at the top of the main page and the RCS (RSC? CRS?) knowls:
http://localhost:37777/EllipticCurve/

But then for individual pages, it does not currently add a Q(\alpha) bread as there is already a Q(blah) bread added for that particular \alpha.  I could add Q(\alpha) also, but I wasn't sure if that made the bread too long.

I can easily change this behavior, just wanted to get feedback before doing so.

Also, I have commented out some lines of code that I will remove once this is ready to be merged, but I wanted to keep them in to easily undo parts as needed (adding this note as much to remind myself later as for anyone else).

### DIFF
--- a/lmfdb/app.py
+++ b/lmfdb/app.py
@@ -473,6 +473,14 @@ def modular_forms():
     return render_template('single.html', title=t, kid='mf.about', bread=b)
 
 
+@app.route('/EllipticCurves')
+@app.route('/EllipticCurves/')
+def elliptic_curves():
+    t = 'Elliptic curves'
+    b = [(t, url_for('elliptic_curves'))]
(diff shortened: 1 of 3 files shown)


jenpaulhus:
#7009 supersedes this now.

PR #6961, opened 2026-04-06
Add max ramified prime to searches

jwj61:
Visible on the main number field page

http://beta.lmfdb.org/NumberField/

Refined search page with modified sort order
http://127.0.0.1:37777/NumberField/?start=50&degree=4&maxp=2-5&sort_order=max+p&sort_order=max+p

Note, this allows the search but doesn't deal with completeness checking yet.


lmfdb/number_fields/number_field.py
```diff
@@ -1334,27 +1342,23 @@ def __init__(self):
...
             [count]]
 
         self.refine_array = [
-            [degree, signature, num_ram, ram_primes, ur_primes ],
-            [gal, field_is, subfield, class_group, class_number],
-            [discriminant, rd, grd, narrow_class_group, narrow_class_number],
-            [cm_field, relative_class_number, regulator, completion, monogenic],
-            [index, inessentialprimes, is_minimal_sibling, unit_signature_rank]]
-
-            #[degree, signature, class_number, class_group, cm_field],
```
roed314, line 1356: I would delete these comments.

jwj61:
I think you are looking at the wrong side of the diff.  Part of my changes is to delete those comments (they are on the left, but not on the right).

roed314:
Yes, I was tired.

Issue #6960, opened 2026-04-05
Auto Group Knowl


jenpaulhus:
The knowl for the automorphism group of 512.25 returns an error

 https://www.lmfdb.org/Groups/Abstract/512.25 

Error message says "ERROR in the template: list.remove(x): x not in list. Please edit it to resolve the problem"

I have not had a chance to debug.  Lots of groups have knowls that work fine but  lots of groups of order 512 fail as well as 384.17773

jwj61:
Tracing this, it is autknow_data which calls create_boolean_aut_string, which creates a list of "unknown" attributes, and then removes some based on other attributes.  Sometimes, an attribute is already removed from this list and it tries to remove it again.

I have a fix which does the removal safely, but I accidentally made a pull request from my main branch.  Once that clears, I can submit this fix.

jwj61:
Was able to make a pull request for just this.

PR #6962, opened 2026-04-08
Fix aut group knowl

jwj61:
Fix issue #6960 

Open the knowl for the automorphism group

http://beta.lmfdb.org/Groups/Abstract/512.25
http://127.0.0.1:37777/Groups/Abstract/512.25

### DIFF
--- a/lmfdb/groups/abstract/main.py
+++ b/lmfdb/groups/abstract/main.py
@@ -738,9 +738,7 @@ def create_boolean_aut_string(gp, prefix="aut_", type="normal", name="automorphi
         unknown.remove('nonsolvable')
     prop = 'pgroup'  # if p-group, we know it is nilpotent, solvable, and supersolvable
     if getattr(gp,prefix+prop,None) > 1:
-        unknown.remove('nilpotent')
-        unknown.remove('solvable')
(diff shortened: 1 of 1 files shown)


lmfdb/groups/abstract/main.py
```diff
@@ -738,9 +738,7 @@ def create_boolean_aut_string(gp, prefix="aut_", type="normal", name="automorphi
         unknown.remove('nonsolvable')
     prop = 'pgroup'  # if p-group, we know it is nilpotent, solvable, and supersolvable
     if getattr(gp,prefix+prop,None) > 1:
-        unknown.remove('nilpotent')
-        unknown.remove('solvable')
-        unknown.remove('supersolvable')
+        unknown=list(set(unknown)-{'nilpotent','solvable','supersolvable'})
```
roed314: This doesn't preserve the order.  I think better would be
```
for col in ['nilpotent', 'solvable', 'supersolvable']:
    if col in unknown:
        unknown.remove(col)
```

jwj61:
I wondered after submitting that maybe order mattered, so changed it so that the order is preserved.

roed314:
Looks good!

PR #6968, opened 2026-04-14
Mf hecke cc

roed314:
Switch from using `mf_hecke_cc` to files on beta for getting Hecke coefficients beyond $n=100$.  Always display 100 coefficients on embedded newform pages, and limit the displayed coefficients on non-embedded newform pages (changing the previous limit of 1000 to 100).

Here we set up a mechanism for redirecting file requests to beta.lmfdb.org.  This can't be tested yet, since this code is not yet running on beta.  I propose merging this to main and dev, then testing that code before merging to web.  I've added tests; the one checking that we can download the file from beta will fail for the moment.

Once this is merged, we can reload `mf_hecke_cc` with a smaller version.


roed314:
This is ready for review, despite the failing test (see the comment above).

AndrewVSutherland:
LGTM

PR #6973, opened 2026-04-19
main -> web

roed314:


AndrewVSutherland:
Looks like test_mf_hecke_cc_dataset is failing, which I guess is expected (on both devmirror and proddb)?

roed314:
The problem is actually kind of annoying.  The links https://beta.lmfdb.org/ModularForm/GL2/Q/holomorphic/mf_hecke_cc/?N=11&k=2 and https://beta.lmfdb.org/ModularForm/GL2/Q/holomorphic/mf_hecke_cc/?label=21.2.e.a are functional, but we can't test them from CI because we're blocking connections with the javascript cookie requirement.  I'm not sure what the easiest way around this is.  I suspect that the best option is to manually add the required cookie to the request, but I didn't manage to do that over the weekend.

PR #6980, opened 2026-04-24
Add more latex names for number fields

rvisser7:
For some families of number fields $K$, we give a "pretty" latexed name for $K$ at the top of the field homepage, currently implemented for quadratic fields $\mathbb{Q}(\sqrt{D})$, biquadratic fields $\mathbb{Q}(\sqrt{A}, \sqrt{B})$, and cyclotomic fields $\mathbb{Q}(\zeta_n)$.  This PR extends the `field_pretty` function to give latexed names for some additional families of number fields. :)

The following three cases have been added:

 - **General multi-quadratic fields** $\mathbb{Q}(\sqrt{D_1}, \sqrt{D_2}, \dots, \sqrt{D_k})$ :  For fields $K$ of degree $2^k$ with $2^k - 1$ quadratic subfields, this now selects the $k$ smallest quadratic subfields $Q(\sqrt{D_1}), \dots, Q(\sqrt{D_k})$ which generate $K$.
  
   http://localhost:37777/NumberField/8.0.12960000.1
   http://localhost:37777/NumberField/16.0.63456228123711897600000000.10
   http://localhost:37777/NumberField/32.0.4026692887688564776141139207792885760000000000000000.1

 - **Imprimitive (non-biquadratic) quartic fields** $\mathbb{Q}(√ (A + B \sqrt{D}))$:  This checks if a quartic field $K$ has exactly one quadratic subfield $\mathbb{Q}(\sqrt{D})$, and then computes a generator for $K$ of the form $√(A + B \sqrt{D})$ by factoring the defining polynomial for $K$ over $\mathbb{Q}(\sqrt{D})$.  If $A = 0$, then returns the latexed form as a pure quartic field $\mathbb{Q}(\sqrt[4]{n})$.

   http://localhost:37777/NumberField/4.0.512.1
   http://localhost:37777/NumberField/4.2.2048.1
   http://localhost:37777/NumberField/4.4.2048.1

 - **Pure cubic fields**: $\mathbb{Q}(\sqrt[3]{n})$: Detects if a cubic field K is of the form $\mathbb{Q}(\sqrt[3]{n})$, by computing a real root of the defining polynomial for $K$.

   http://localhost:37777/NumberField/3.1.108.1
   http://localhost:37777/NumberField/3.1.243.1
   http://localhost:37777/NumberField/3.1.300.1
   

As always, any comments/feedback very welcome! 🙂

### DIFF
--- a/lmfdb/number_fields/web_number_field.py
+++ b/lmfdb/number_fields/web_number_field.py
@@ -261,16 +261,21 @@ def field_pretty(label):
             
         # Case 4b: Imprimitive quartic fields of type Q(sqrt(A + Bsqrt(D)))
         if len(subs) == 1:
-            # Factorise defining polynomial for K over Q(sqrt(D))
-            D = subs[0]
-            Ksub = QuadraticField(D)
-            relative_poly = poly.factor()[0]
-            rel_ceoffs = relative_poly.coefficients(sparse=False)
-            ans = rel_coeffs[1]**2 - 4*rel_coeffs[0]*rel_coeffs[2]
+            quad_sub = wnf.from_coeffs(string2list(str(subs[0][0])))
+            if not quad_sub._data is None:
+                # Factorise defining polynomial for K over Q(sqrt(D))
+                quad_label = str(quad_sub.get_label())
+                D = quad_label.split('.')[2]
+                Ksub = NumberField(quad_sub.poly(), 'b')
+                Rsub = PolynomialRing(Ksub, 'x')
+                relative_poly = Rsub(wnf.poly())
+                factors = relative_poly.factor()
 
-            return r'\(\Q(\sqrt{%s}\)' % ans
+                # Can extract the first quadratic factor
+                rel_coeffs = factors.coefficients(sparse=False)
+                ans = rel_coeffs[1]**2 - 4*rel_coeffs[0]*rel_coeffs[2]
+                return r'\(\Q(\sqrt{%s})\)' % latex(ans)
 
-            
     # Case 5: Maximal real subfields of cyclotomic fields Q(\zeta_N)^+
     if label in rcycloinfo:
         return r'\(\Q(\zeta_{%d})^+\)' % rcycloinfo[label]
--- a/lmfdb/number_fields/web_number_field.py
+++ b/lmfdb/number_fields/web_number_field.py
@@ -212,10 +212,21 @@ def is_fundamental_discriminant(d):
 
 @cached_function
 def field_pretty(label):
+    """
+    Given an LMFDB number field label, returns a "pretty" latexed representation of this field (if this exists)
+    Otherwise, simply returns back the label itself if unable to find a latex represntation
+
+    Cases implemented:  Quadratic fields, Pure cubic fields, Impritive quartic fields, (real) cyclotomic fields, and multi-quadratic fields
+    """
+
     d, r, D, _ = label.split('.')
+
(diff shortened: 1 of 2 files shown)


lmfdb/number_fields/web_number_field.py
```diff
@@ -238,11 +268,105 @@ def field_pretty(label):
...
+            
+        # Case 4b: Imprimitive quartic fields of type Q(\sqrt(A + B*\sqrt(D)))
+        if len(subs) == 1:
+            quad_sub = wnf.from_coeffs(string2list(str(subs[0][0])))
+            if not quad_sub._data is None:
+                # Get unique quadratic subfield Q(sqrt(D))
+                quad_label = str(quad_sub.get_label())
+                D = _quad_label_to_D(quad_label)
+                Ksub = QuadraticField(D, 'sqrtD')
+                sqrtD = Ksub._first_ngens(1)[0]
```
roed314: This should just be `Ksub.gen(0)`.

lmfdb/number_fields/web_number_field.py
```diff
@@ -238,11 +268,105 @@ def field_pretty(label):
...
+
+                # Return final pretty latex
+                if A == 0:
+                    # Case: Pure quartic field
+                    return r'\(\Q(\sqrt[4]{%d})\)' % (D*B**2)
+                else:
+                    B_str = "+" if B == 1 else "-" if B == -1 else f"{B:+d}"
+                    return r'\(\Q(\sqrt{%d %s %s})\)' % (A, B_str, _sqrt_symbol(D))
+
+    # Case 5: Maximal real subfields of cyclotomic fields Q(\zeta_N)^+
```
roed314: I would put this before Case 4.

lmfdb/number_fields/web_number_field.py
```diff
@@ -238,11 +268,105 @@ def field_pretty(label):
...
+            # Compute set of all primes dividing the Ds
+            primes = sorted({int(p) for D in all_Ds for p in ZZ(abs(D)).prime_divisors()})
+
+            # Keep track of prime exponents used so far
+            all_prime_exponents = []
+
+            for D in sorted_Ds:
+                # Convert D to a vector of prime exponents mod 2 (including sign)
+                prime_exp = [int(D < 0)]+[D.valuation(p)%2 for p in primes]
+                if vector(prime_exp) not in matrix(GF(2), all_prime_exponents).row_space():
```
roed314: You probably want to keep the space around as you vary `D`, and not recompute it each time.

lmfdb/number_fields/web_number_field.py
rvisser7: thanks, fixed!

lmfdb/number_fields/web_number_field.py
rvisser7: thanks, done!  Though this does mean that fields like [4.4.2304.1](https://beta.lmfdb.org/NumberField/4.4.2304.1) now have the latex label $\mathbb{Q}(\zeta_{24})^+$ instead of $\mathbb{Q}(\sqrt{2}, \sqrt{3})$.

lmfdb/number_fields/web_number_field.py
rvisser7: Thanks, the row_space is now only computed again if `all_prime_exponents` gets updated  (i.e. this will now only get called at most $k$ times for a degree $2^k$ multiquadratic field).

PR #6982, opened 2026-04-24
An alternative to 6941

jenpaulhus:
Here's a possible alternative if we don't want a landing page for elliptic curves as is proposed in #6941.

This is NOT ready to merge yet, and maybe worth putting on a color server so folks can look at.

It uses two tabs (like on alpha lattices) on the EC and ECNF index pages that people can toggle between to get to these two sections.  The proposal is also to reduce the sidebar entries to only have one link saying "Elliptic curves" which goes to the rational EC page.  This would require one extra click if you want ECNF.

Easy to change the names of the tabs too!

Feedback welcome!

### DIFF
--- a/lmfdb/ecnf/templates/ecnf-index.html
+++ b/lmfdb/ecnf/templates/ecnf-index.html
@@ -3,7 +3,7 @@
 {% block content %}
 
 
-{{ tabs(("Q", "ec.rational_elliptic_curves", false), ("Number Fields", ".index", true)) }}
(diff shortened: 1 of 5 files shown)


roed314:
This is up on [blue](https://blue.lmfdb.xyz/EllipticCurve/Q/)

roed314:
We'd talked about adding a submenu below elliptic curves on the sidebar that would open up automatically if you click on elliptic curves, or if you click on a little caret.  I think we probably want to add something like that before merging this.  Perhaps a good exercise for Claude.  :smile:

jenpaulhus:
#7009 supersedes this.

PR #6987, opened 2026-05-05
Update lattices on alpha

edgarcosta:


edgarcosta:
@roed314 are we expecting all the tests to pass, correct?

Issue #7004, opened 2026-05-13
Error when downloading search results for elliptic curves over number fields


rvisser7:
It seems that downloading the search results for elliptic curves over number fields fails for certain queries.  E.g.  trying to download the search results when searching for conductor norm 1 or conductor norm 2 curves fails to produce a valid download file:

https://beta.lmfdb.org/EllipticCurve/?download=1&query=%7B%27conductor_norm%27%3A+1%7D&conductor_norm=1&Submit=text
https://beta.lmfdb.org/EllipticCurve/?download=1&query=%7B%27conductor_norm%27%3A+2%7D&conductor_norm=2&Submit=text

Another (possibly related?) error occurs when trying to download the search results for elliptic curves over the base field 4.4.7225.1.  It returns "Error: unable to convert '-3/2' to an integer".

https://beta.lmfdb.org/EllipticCurve/?download=1&query=%7B%27field_label%27%3A+%274.4.7225.1%27%7D&field=4.4.7225.1&Submit=text

PR #7005, opened 2026-05-13
Bug fix for ECNF downloader

rvisser7:
This fixes issue #7004 - the ECNF downloader now parses the coefficients of elliptic curves as rational linear combinations of $a^i$, where previously it wrongly assumed it was integer linear combinations of $a^i$  (this is only true if the base field $K$ is monogenic where $O_K = \mathbb{Z}[a]$).

https://beta.lmfdb.org/EllipticCurve/?download=1&query=%7B%27conductor_norm%27%3A+1%7D&conductor_norm=1&Submit=text
http://localhost:37777/EllipticCurve/?download=1&query=%7B%27conductor_norm%27%3A+1%7D&conductor_norm=1&Submit=text

### DIFF
--- a/lmfdb/ecnf/main.py
+++ b/lmfdb/ecnf/main.py
@@ -454,7 +454,7 @@ class ECNFDownloader(Downloader):
         row["field_coeffs"] = poly
 
         # Convert Weierstrass coefficients from string to a list of list of integers
-        row['ainvs'] = [[ZZ(aj) for aj in ai.split(",")] for ai in row['ainvs'].split(";")]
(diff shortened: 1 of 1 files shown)


lmfdb/ecnf/main.py
```diff
@@ -454,7 +454,7 @@ def postprocess(self, row, info, query):
         row["field_coeffs"] = poly
 
         # Convert Weierstrass coefficients from string to a list of list of integers
```
edgarcosta:
```suggestion
        # Convert Weierstrass coefficients from string to a list of list of rationals
```

PR #7014, opened 2026-05-16
Expand completeness checker for elliptic curves and number fields

rvisser7:
This PR extends the completeness checker (``lmfdb/utils/completeness.py``) with a few more cases for elliptic curves, genus 2 curves, and number fields.  

As always, any comments/feedback much appreciated! 🙂 

 - **Elliptic curves**:  Since the LMFDB has all elliptic curves over Q with 7-smooth conductor, we can thus guarantee completeness when searching for elliptic curves over Q whose bad primes are contained in {2,3,5,7}.  This hence addresses Drew's comment in [#6773](https://github.com/LMFDB/lmfdb/pull/6773#issuecomment-3582742829).  

    https://beta.lmfdb.org/EllipticCurve/Q/?bad_quantifier=subset&bad_primes=2%2C3%2C5%2C7
    http://localhost:37777/EllipticCurve/Q/?bad_quantifier=subset&bad_primes=2%2C3%2C5%2C7

 - Further to the above, we can also guarantee completeness when searching for elliptic curves over Q whose bad primes are contained in some fixed set $S$, when the maximum possible conductor $M = \prod_{p \in S} p^{e_p}$, with $e_2 = 8$, $e_3 = 5$, and $e_p = 2$ for $p > 3$, is at most 500 000.  E.g.  we can guarantee completeness for elliptic curves E/Q good away from $S =$ {23, 29} or $S =$ {701}.

    https://beta.lmfdb.org/EllipticCurve/Q/?bad_quantifier=subset&bad_primes=23%2C29
    http://localhost:37777/EllipticCurve/Q/?bad_quantifier=subset&bad_primes=23%2C29

 - **Genus 2 curves**:  We can (rather trivially) guarantee completeness for genus 2 curves over Q with conductor less than 121, since no such curves exist!  Though (as far as I know) this requires assuming paramodularity.

    https://beta.lmfdb.org/Genus2Curve/Q/?cond=1-120
    http://localhost:37777/Genus2Curve/Q/?cond=1-120

 - **Regulator bounds for number fields**:  For non-CM number fields $K$, the regulator $R$ satisfies a lower bound of the form $R \geq A (log \Delta_K)^B$ for some effective constants $A$ and $B$ depending on the signature of $K$.  Combined with our existing discriminant completeness bounds, this yields completeness guarantees up to an explicit regulator bound for certain signatures (though this would be a somewhat small bound for most signatures).

    To make this work within the `NFBound` class, I've added the regulator to `display_reason`, and have added a method `clear_regulators` which does essentially the same thing as `clear_signatures`, except for the regulator $R$ instead of the discriminant $D$.   We then call `self.clear_regulator` as part of Completeness 1 in `_one_n` and return True if all signatures have been certified.

    https://beta.lmfdb.org/NumberField/?signature=%282%2C0%29&regulator=0-7
    http://localhost:37777/NumberField/?signature=%282%2C0%29&regulator=0-7

    https://beta.lmfdb.org/NumberField/?signature=%282%2C1%29&regulator=0-0.5
    http://localhost:37777/NumberField/?signature=%282%2C1%29&regulator=0-0.5

    https://beta.lmfdb.org/NumberField/?signature=%285%2C1%29&regulator=0-6
    http://localhost:37777/NumberField/?signature=%285%2C1%29&regulator=0-6

I've also been adding various extra comments and docstrings throughout the ``completeness.py`` file, mainly for my own understanding of how everything works.  I thought I'll keep them in as it might benefit others (but can also take them out if you'd think they're unnecessary).


roed314:
I need to focus on final exams for a few days, but I'll be happy to look this once I'm done!

rvisser7:
Thanks - of course no rush!  I've realised there are many more cases we could still add, though I think the above is probably enough for this PR. :)

I've opened a separate issue #7017 to help track more cases we could add to the completeness checker in the future.

lmfdb/utils/completeness.py
```diff
@@ -769,16 +799,104 @@ def is_smooth(n):
...
+        elif isinstance(query_val, list):
+            # Exact match case
+            return set(query_val).issubset(self.allowed_set)
+        return False
+
+
+class RestrictedBadPrimesConductor(ColTest):
+    """
+    Checks completeness based on the maximum conductor for a given set of restricted bad primes.
+    E.g. can be used for elliptic curves, genus 2 curves, or general abelian varieties (e.g. see Brumer-Kramer paper).
```
roed314: Can you add a more specific reference here?

lmfdb/utils/completeness.py
```diff
@@ -769,16 +799,104 @@ def is_smooth(n):
...
+        query_val = Ds[0]
+        bad_primes = None
+        
+        if isinstance(query_val, dict):
+            if '$containedin' in query_val:
+                bad_primes = query_val['$containedin']
+        elif isinstance(query_val, list):
+            bad_primes = query_val
+        
+        if bad_primes is None or not bad_primes:
```
roed314: If `bad_primes = []`, shouldn't this return True, since if there are no bad primes the conductor is 1 (and thus at most the `conductor_bound`, which I guess we should check is positive)?

If so , we can just move this to an else statement and remove the `bad_primes = None` setup step.

lmfdb/utils/completeness.py
```diff
@@ -1822,25 +2046,50 @@ def describe(tups):
...
+        and restrict the remaining regulator range accordingly.
+        """
+
+        if 2 <= n < len(self._maxReg):
+            m = infinity
+            for r2 in set(r2opts):
+                M = self._maxReg[n][r2] - 0.00001
+                if R.bounded(M):
+                    r2opts.remove(r2)
+                    reasons.add((n, r2, None, None, None, None, M))
```
roed314: I think the reason should be in terms of `self._maxReg[n][r2]` rather than `M`.

lmfdb/utils/completeness.py
```diff
@@ -1822,25 +2046,50 @@ def describe(tups):
...
+    def clear_regulator(self, n, R, r2opts, reasons):
+        """
+        Remove signatures (n-2*r2, r2) already certified complete using the regulator bounds in self._maxReg[n][r2],
+        and restrict the remaining regulator range accordingly.
+        """
+
+        if 2 <= n < len(self._maxReg):
+            m = infinity
+            for r2 in set(r2opts):
+                M = self._maxReg[n][r2] - 0.00001
```
roed314: Deal with None case.

lmfdb/utils/completeness.py
```diff
@@ -1984,10 +2240,16 @@ def galt(self, n, gal, isgal, cyc, ab, solv):
         return galt
 
     def rd_grd_ratio(self, n, galt):
+        """
+        Return the maximum possible root-discriminant for the given Galois groups.
```
roed314: This description isn't quite right.

lmfdb/utils/completeness.py
rvisser7: sure, done!

lmfdb/utils/completeness.py
rvisser7: Thanks, I agree - done!

lmfdb/utils/completeness.py
rvisser7: Thanks, I'm not sure why I originally wrote that!  I've updated the docstring to the following:
```
Return max(1/r) over the Galois groups in galt, where for each Galois group nTt in galt, r is
a ratio (taken from self._rdgrd[n][t-1]) such that grd <= rd^(1/r) for fields with Galois group nTt.
```

lmfdb/utils/completeness.py
rvisser7: Thanks, I've added the below code which skips over signature $(n - 2r_2, r_2)$ if either self._maxReg[n][r2] doesn't exist or is None.
```
maxReg = self._maxReg[n][r2] if r2 < len(self._maxReg[n]) else None
if maxReg is None:
    continue     
```

lmfdb/utils/completeness.py
rvisser7: Thanks!  This now returns True if `bad_primes == []` and `1 <= self.conductor_bound`.

Although, in practice, I'm not actually sure if it's possible for the user to run a search where `bad_primes = []` ?  (as if the bad_primes search box is empty, I think the search query will just ignore this field).

rvisser7:
For ease of reviewing, in case anyone would like to double check the regulator bounds I've put for each signature, I've attached some links to the papers I've referenced below:

 - Michael Pohst (1977), **Regulatorabschätzungen für total reelle algebraische Zahlkörper**, ([published doi](https://doi.org/10.1016/0022-314X(77)90007-5)), ([MathSciNet](https://mathscinet.ams.org/mathscinet-getitem?mr=460274)).  Used for signature $(2,0)$ - see Satz XIII on pg 485 .

 - T. W. Cusick (1984), **Lower bounds for regulators**. ([published doi](https://doi.org/10.1007/BFb0099441)), ([MathSciNet](https://mathscinet.ams.org/mathscinet-getitem?mr=756083)).  Used for signatures $(3,0)$, $(1,1)$, and $(4,0)$.

 - Sergio Astudillo,  Francisco Diaz y Diaz, Eduardo Friedman (2016) **Sharp lower bounds for regulators of small-degree number fields**. ([published doi](https://doi.org/10.1016/j.jnt.2016.03.002)) ([MathSciNet](https://mathscinet.ams.org/mathscinet-getitem?mr=3504045)).  Used for signatures $(2,1)$, $(0,2)$, $(5,0)$, $(1,2)$, $(6,0)$, $(2,2)$, $(0,3)$, $(7,0)$, $(3,2)$, $(1,3)$, $(8,0)$, $(0,4)$, $(9,0)$.

 - Francesco Battistoni, Giuseppe Molteni,(2025),  **Generalized Pohst inequality and small regulators**.. ([arxiv](https://arxiv.org/pdf/2211.16842)) ([published doi](https://doi.org/10.1090/mcom/3954)), ([MathSciNet](https://mathscinet.ams.org/mathscinet-getitem?mr=4271810)).   Used for signatures $(3,1)$, $(4,1)$, $(5,1)$ and $(6, 1)$.

Please feel free to suggest other papers if anyone finds better regulator bounds!

PR #7016, opened 2026-05-19
fix(layout): push #footer to viewport bottom on content-short pages

edgarcosta:
Previously #footer floated wherever #main's content (with a fixed 600px min-height) ended, leaving large gaps mid-viewport on short pages such as /, /about, and /contact when viewed on tall windows.

Switch to the standard flex sticky-footer pattern: a new #layout wrapper around the page content and footer is a flex column with min-height: 100vh, and an inner #page-wrapper grows to fill any remaining space, so #footer naturally lands at the bottom of the viewport on short pages and continues to flow after content on long pages. No magic numbers, no per-page tuning.

The flex container is a dedicated wrapper rather than body itself because third-party scripts (reCAPTCHA) append elements to body, and those should not become flex items that displace the footer.

Verified with playwright across /, /about, /contact at viewport heights 1080, 1440, 2000: footer's bottom edge is at the viewport bottom on every page.

See it on olive.lmfdb.xyz

### DIFF
--- a/lmfdb/templates/homepage.html
+++ b/lmfdb/templates/homepage.html
@@ -7,6 +7,8 @@
 {% endmacro %}
 
 {% block body -%}
+<div id="layout">
+<div id="page-wrapper">
 <div id="header">
     <div id="logo"><a href="{{ url_for('index') }}">
 {% if ALPHA %}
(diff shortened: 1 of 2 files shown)


jenpaulhus:
I like this update, particularly if we are now going to have the ability to collapse parts of the sidebar.

PR #7044, opened 2026-06-18
Cmdline

roed314:
Add script for searching the LMFDB from the command line.


AndrewVSutherland2:
This looks great and I'm happy to merge it.  My only suggestion for improvement would be to find a way to raise an error when a non-existent column name is specified in the search criteria rather than silently ignoring it (I realize this is not trivial to do, but people are going to make this mistake all the time and be confused by it).

roed314:
Good suggestion.  I also had Claude add a feature where a user can paste a whole URL.

roed314:
@AndrewVSutherland approved in person.

PR #7053, opened 2026-06-30
Speed up construction of torsion growth table

roed314:
A recent change to elliptic curves (showing base change) drastically slowed down page loads for people with a slow connection.  This improves the situation some.

### DIFF
--- a/lmfdb/elliptic_curves/web_ec.py
+++ b/lmfdb/elliptic_curves/web_ec.py
@@ -888,10 +888,14 @@ class WebEC():
         tg['data'] = tgextra = []
 
         # find all base changes of this curve in the database, if any:
-        nf_to_curve = {rec["field_label"]: rec["label"] for rec in db.ec_nfcurves.search({'base_change': {'$contains': [self.lmfdb_label]}}, ["field_label", "label"])}
+        nf_to_curve = {rec["field_label"]: rec["label"]
+                       for rec in db.ec_nfcurves.search({'base_change': {'$contains': [self.lmfdb_label]}},
+                                                        ["field_label", "label"])}
         # look up all the fields in the growth table in a single query
         # (the projection must include the columns needed by field_pretty)
-        pol_to_nf = {tuple(rec["coeffs"]): rec for rec in db.nf_fields.search({"$or": [{"coeffs": tgd['field']} for tgd in tgdata]}, ["label", "coeffs", "disc_abs", "disc_sign", "subfields", "subfield_mults"])}
+        pol_to_nf = {tuple(rec["coeffs"]): rec
+                     for rec in db.nf_fields.search({"$or": [{"coeffs": tgd['field']} for tgd in tgdata]},
+                                                    ["label", "coeffs", "disc_abs", "disc_sign", "subfields", "subfield_mults"])}
         tg['fields_missing'] = False
 
         for tgd in tgdata:
--- a/lmfdb/elliptic_curves/web_ec.py
+++ b/lmfdb/elliptic_curves/web_ec.py
@@ -889,11 +889,9 @@ class WebEC():
 
         # find all base changes of this curve in the database, if any:
         nf_to_curve = {rec["field_label"]: rec["label"] for rec in db.ec_nfcurves.search({'base_change': {'$contains': [self.lmfdb_label]}}, ["field_label", "label"])}
-        if nf_to_curve:
-            # extract the fields from the labels of the base-change curves:
-            pol_to_nf = {tuple(rec["coeffs"]): rec for rec in db.nf_fields.search({"label": {"$in": list(nf_to_curve)}}, ["label", "coeffs", "disc_abs", "disc_sign", "subfields", "subfield_mults"])}
-        else:
-            pol_to_nf = {}
+        # look up all the fields in the growth table in a single query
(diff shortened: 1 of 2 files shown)


lmfdb/elliptic_curves/web_ec.py
```diff
@@ -888,23 +888,29 @@ def make_torsion_growth(self):
         tg['data'] = tgextra = []
 
         # find all base changes of this curve in the database, if any:
-        bcs = list(db.ec_nfcurves.search({'base_change': {'$contains': [self.lmfdb_label]}}, projection='label'))
-        # extract the fields from the labels of the base-change curves:
-        bc_fields = [lab.split("-")[0] for lab in bcs]
-        bc_pols = [db.nf_fields.lookup(lab, projection='coeffs') for lab in bc_fields]
+        nf_to_curve = {rec["field_label"]: rec["label"] for rec in db.ec_nfcurves.search({'base_change': {'$contains': [self.lmfdb_label]}}, ["field_label", "label"])}
```
havarddj: I would probably split this over a line or two, but since the linter doesn't complain, I won't either.

lmfdb/number_fields/web_number_field.py
```diff
@@ -210,12 +210,14 @@ def is_fundamental_discriminant(d):
...
 
-@cached_function
-def field_pretty(label):
+@cached_function(key=lambda label, wnf: label) # ignore field when caching
+def field_pretty(label, wnf=None):
     """
     Given an LMFDB number field label, returns a "pretty" latexed representation of this field (if it exists)
     Otherwise, simply returns back the label itself if unable to find a latex representation.
 
+    field should be a WebNumberField object, which is constructed if needed if not provided
```
havarddj:
```suggestion
    "field" should be a WebNumberField object, which is constructed if needed and not provided.
```

havarddj (commented):
Looks good to me, just added a couple of harmless suggestions. I think there's still potentially work to be done in optimising queries for number fields (which this PR doesn't touch, so can potentially be made into a separate issue) - for example [the biggest multiquadratic field](http://localhost:37777/NumberField/32.0.4026692887688564776141139207792885760000000000000000.1) loads **extremely** slowly on a slow connection.

havarddj:
Actually, judging from the tests, it seems like the change broke some of Robin's functionality: for example the field http://127.0.0.1:37777/NumberField/4.0.2048.2 is not labelled as $$\mathbb{Q}(\sqrt{-2 + \sqrt{2}})$$ anymore.

roed314:
@havarddj I think I've addressed all of your comments.  Let me know if the current version looks good.

lmfdb/number_fields/web_number_field.py
```diff
@@ -210,12 +210,16 @@ def is_fundamental_discriminant(d):
...
-def field_pretty(label):
+@cached_function(key=lambda label, wnf: label) # ignore field when caching
+def field_pretty(label, wnf=None):
     """
     Given an LMFDB number field label, returns a "pretty" latexed representation of this field (if it exists)
     Otherwise, simply returns back the label itself if unable to find a latex representation.
 
+    wnf, if provided, should be a WebNumberField for this label; it may have been
+    built from a search projection, and is only used if it has the needed columns
+    (since the result is cached by label alone, it must not depend on the caller)
```
havarddj: I don't understand what this line means. The result should not depend on the caller? Or the label should not depend on the caller? What exactly do you mean by the caller?

Here's an alternative: 
```
The optional argument 'wnf' should be a WebNumberField, and is used in cases 
where generating the "pretty representation" requires more data than just the label.
The specific data depends on the type of field (multiquadratic, pure cubic, ...), 
and when it is not present in 'wnf', the database will be queried instead.

This argument is cached, and therefore should not "depend on the caller" (perhaps explain this better)
```
I agree it's a little verbose, but I think it makes it easier to understand what is happening and why.

lmfdb/elliptic_curves/web_ec.py
```diff
@@ -888,23 +888,31 @@ def make_torsion_growth(self):
...
-        bcs = list(db.ec_nfcurves.search({'base_change': {'$contains': [self.lmfdb_label]}}, projection='label'))
-        # extract the fields from the labels of the base-change curves:
-        bc_fields = [lab.split("-")[0] for lab in bcs]
-        bc_pols = [db.nf_fields.lookup(lab, projection='coeffs') for lab in bc_fields]
+        nf_to_curve = {rec["field_label"]: rec["label"]
+                       for rec in db.ec_nfcurves.search({'base_change': {'$contains': [self.lmfdb_label]}},
+                                                        ["field_label", "label"])}
+        # look up all the fields in the growth table in a single query
+        # (the projection must include the columns needed by field_pretty)
+        pol_to_nf = {tuple(rec["coeffs"]): rec
```
havarddj, line 896: This looks much better, thanks.

havarddj (approved):
I'm happy with this (modulo a minor suggestion/question for a doc string)! Just to check that it works, I also compared page loading time for [Richard Guy's curve](http://127.0.0.1:37777/EllipticCurve/Q/88/a/1) between this branch and main, and the result is ~15s vs ~30s.

roed314:
I just changed the docstring based on @havarddj's suggestion, and he approved the change.

PR #7058, opened 2026-07-12
Limit displayed Dirichlet coefficients to stored Euler factors

### LINKED ISSUE #7057: L-function page for genus 2 curve displaying bogus a_n for n > 100
The page https://www.lmfdb.org/L/4/5e5/1.1/c1e2/0/0 displays some nonsense a_n for n > 100 (a.g. a_n=101 is wrong) that it has no business displaying (or trying to compute); only a_n for n <= 100 are stored, so the L-function page should not display any a_n for n > 100.

edgarcosta:
# Limit displayed Dirichlet coefficients to stored Euler factors

Fixes #7057.

## Description

L-function pages could display Dirichlet coefficients beyond the range determined by the stored Euler factors. These unknown coefficients were initialized to `1`, making them appear authoritative.

For example, the page for [`4-5e5-1.1-c1e2-0-0`](https://www.lmfdb.org/L/4/5e5/1.1/c1e2/0/0) stores Euler factors through \(p=97\), but displayed incorrect coefficients at \(n=101,103,\ldots\).

This change limits generated display coefficients to indices before the first prime without a stored Euler factor. Explicitly stored coefficient arrays and coefficient downloads are unchanged.

## Testing

- Added a regression test that requests the rendered page for `4-5e5-1.1-c1e2-0-0`.
- Confirmed that the page renders a known coefficient within the stored range.
- Confirmed that the rendered page does not include \(101^{-s}\).
- Scoped `pyflakes` and `git diff --check` pass.

The full L-function test module had 15 passing tests and 6 unrelated failures caused by the Sage 10.8 Galois-group API.

### DIFF
--- a/lmfdb/lfunctions/Lfunction.py
+++ b/lmfdb/lfunctions/Lfunction.py
@@ -262,7 +262,8 @@ def makeLfromdata(L):
         L.dirichlet_coefficients_arithmetic = data['dirichlet_coefficients']
     elif data.get('euler_factors', None) is not None:
         # ask for more, in case many are zero
-        L.dirichlet_coefficients_arithmetic = an_from_data(L.localfactors, 2*L.degree*L.numcoeff)
+        upperbound = min(2*L.degree*L.numcoeff, nth_prime(len(L.localfactors) + 1) - 1)
+        L.dirichlet_coefficients_arithmetic = an_from_data(L.localfactors, upperbound)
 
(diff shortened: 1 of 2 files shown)


AndrewVSutherland:
Looks good, I just restarted the test that failed (due to a database/CI issue, not a regression).  I'll merge when it finishes.

PR #7069, opened 2026-07-19
Make the LMFDB pip installable

roed-math:
This makes the LMFDB pip installable, addressing several long-standing installation warts. From a checkout, `sage -pip install -r requirements.txt` now installs the LMFDB itself in editable mode along with its dependencies, so `import lmfdb` works without adding the checkout to `PYTHONPATH`; `pip install "lmfdb @ git+https://github.com/LMFDB/lmfdb"` also works, and the package is structured so that it could be released on PyPI (see the caveats below).

## What's here

- **Packaging** (`pyproject.toml`, `MANIFEST.in`): metadata, dependencies, and all data files (verified: the wheel ships all 483 non-Python files — templates, static assets, yaml). Console scripts are registered, so the website can be started with `lmfdb` or `sage -python -m lmfdb` from any directory (`start-lmfdb.py` works as before), and the new command-line search is available as `lmfdb_search`. The version number is single-sourced in `lmfdb/version.py`. The dependency list reflects what the code actually imports: `flask-cache`, `flask-markdown`, `cython` and `itsdangerous` were listed but are never imported; `markdown`, `requests`, `werkzeug`, `mpmath` and `pexpect` are imported but were missing.
- **No import-time side effects**: `lmfdb.db` is now a lazy proxy — `import lmfdb` takes a few milliseconds, imports neither sage nor flask nor psycopg2, and creates no files; the postgres connection is established on first use (or explicitly via `db.connect()`). A pleasant consequence: `pip install lmfdb; from lmfdb import db` gives programmatic database access **with no Sage installed**. Website startup connects during imports, as before.
- **Configuration**: the code moved to `lmfdb/config.py`, so that using the database does not pull in `lmfdb.utils` (and hence flask/sage); `lmfdb/utils/config.py` remains as a backwards-compatible shim, still runnable as a script (the CI does this). From a checkout everything is located where it always was (`config.ini` and `secret_key` at the root). When installed as a package they live in `~/.lmfdb`, overridable via `LMFDB_HOME` / `LMFDB_CONFIG` / `--config-file`, with a `config.ini` in the current directory taking precedence.
- **Log files no longer pollute the current directory**: `flasklog`, `slow_queries.log` and verification logs default to `logs/` under the checkout root (or `~/.lmfdb/logs` when installed). The historical default values in existing config files are redirected transparently; explicitly configured paths are honored. **Deployment note**: any logrotate/monitoring pointed at `flasklog` in the repo root will need updating.
- **`CONTRIBUTORS.yaml` moved into the `lmfdb/` package** so the acknowledgments page works when installed. Contributors: add yourself at the new path `lmfdb/CONTRIBUTORS.yaml`.
- The knowl code-reference `git grep`s now run in the checkout root rather than the process working directory, and degrade gracefully (no matches) when lmfdb is not run from a checkout.
- Importing the website without Sage now fails with a clear message explaining that Sage is required and how to install into Sage's python, rather than a bare `ModuleNotFoundError` (pip cannot install Sage, so this is the best we can do; metadata has no way to express a non-PyPI requirement).
- **`.github/workflows/publish.yml`** (not active until configured): pushing a `v*` tag publishes to PyPI, a manual run publishes to TestPyPI as a dry run, via PyPI trusted publishing (no tokens). It builds sdist+wheel, runs `twine check`, checks the tag against the package version, and smoke-tests the wheel.

## Caveats before an actual PyPI release

1. **psycodict is not on PyPI**, so the dependency is currently a git URL — which PyPI rejects at upload time. psycodict needs a PyPI release first; then the dependency becomes a normal versioned requirement.
2. Trusted publishing must be configured on pypi.org / test.pypi.org (see the comments in `publish.yml`). The name `lmfdb` is currently unclaimed on PyPI.
3. Sage ships its own setuptools (73.0.1 in sage 10.7), which can end up performing the build; the packaging deliberately sticks to metadata that setuptools ≥ 64 understands (e.g. the license table rather than a PEP 639 SPDX string).

## Note on versioning and PEP 440

There is a proposal to move LMFDB version numbers to the form `1.8.ba` (major . data version . letter-coded frontend version via `cremona_letter_code`), with per-table versions on Zenodo. A heads-up for that work: **package versions must be PEP 440 compliant**, and release segments are numeric only, so `1.8.ba` cannot be the pip package version. Near-misses are silently dangerous: `1.8b` parses as a *pre-release* of 1.8, which `pip install lmfdb` **skips by default**. The natural fix is a bijection we already have: use the numeric form for the package, `1.8.<class_to_int(code)>`, e.g. `1.8.ba` ↔ `1.8.26` — monotone, sortable, and recoverable in both directions (e.g. on the planned version page). The letter form can remain the display version, the git tag and the Zenodo-facing version; only `lmfdb/version.py` (which `pyproject.toml` reads) needs the numeric mapping, along with the tag-vs-version check in `publish.yml`.

## Testing

- Fresh venv **without Sage**: `pip install .` → `import lmfdb` is clean (no files created, no connection, ~5ms), `db.ec_curvedata.lookup("11.a1")` works, config/logs are created under `~/.lmfdb` (`LMFDB_HOME` respected).
- Venv **with Sage** (editable install): website starts via `python -m lmfdb` from an unrelated directory with no files created there; homepage, `/acknowledgment` (contributors load from the new location), elliptic curve pages, knowls (including code references), and static files all serve; the `start-lmfdb.py` legacy path (no pip install at all) also verified; `lmfdb_search --help` works as a console script.
- `pytest lmfdb/tests/test_acknowlegments.py lmfdb/tests/test_utils.py lmfdb/tests/test_root.py`: 26 passed, 2 skipped.
- `python -m build` + `twine check` pass, and the sdist→wheel path was checked to ship every data file; pyflakes / pylint / ruff pass with the CI's flags.
- Rebased on current main; the TCP keepalives defaults recently added to `lmfdb/utils/config.py` are preserved in the new `lmfdb/config.py`.


GettingStarted.md
```diff
@@ -163,6 +176,17 @@ Running
    on using a different database instance, you can do so by changing
    config.ini in the root of the lmfdb directory.
 
+ * Locations of configuration and log files: when running from a git
+   checkout, the configuration file is `config.ini` at the root of the
+   checkout (as it always was) and log files (`flasklog`,
```
havarddj, line 181: To me, "as it always was" seems like the kind of phrasing that refers to the context of the prompt or the PR. Before merging, I think we should carefully go through the documentation and make sure that it's as clear and un-confusing as possible.

README.md
```diff
@@ -14,7 +14,11 @@ Wiki
 Development
 -----------
 
-* [Getting Started](https://github.com/LMFDB/lmfdb/blob/main/GettingStarted.md) -- cheat sheet for setting up lmfdb
+* [Getting Started](https://github.com/LMFDB/lmfdb/blob/main/GettingStarted.md) -- cheat sheet for setting up lmfdb.
+  The LMFDB is pip installable: `sage -pip install -e .` from a checkout
```
havarddj, line 18: Here we might also want to cherry pick which pieces of information are the most relevant.

lmfdb/__init__.py
```diff
@@ -1,11 +1,30 @@
...
-assert db
+    from lmfdb import db
+    db.ec_curvedata.lookup("11.a1")
+
+or explicitly via ``db.connect()``.  To run the website, use the ``lmfdb``
+command (or ``python -m lmfdb``, or ``start-lmfdb.py`` from a git checkout).
+"""
+
+
+def __getattr__(name):
```
havarddj, line 16: Is it really idiomatic python to override `__getattr__` and `__dir__` in this way?

Also, this gives a double dependency warning, which looks a bit strange: 
```
Missing psycopg dependency; either do "pip install psycopg[binary]" or "pip install psycopg" (requires libpq installed on your system)
Missing dependency; try running "sage -pip install -e ." in the LMFDB home folder.
```
In a similar vein, I get this funny error: 

```
sage: import lmfdb
sage: help(lmfdb)
Missing psycopg dependency; either do "pip install psycopg[binary]" or "pip install psycopg" (requires libpq installed on your system)
Missing dependency; try running "sage -pip install -e ." in the LMFDB home folder.
No module named 'psycopg'
```

Perhaps there's a more natural way to set this up?

PR #7079, opened 2026-07-24
Add Magma support for code snippet tests and snippet log files

rvisser7:
This PR adds support for Magma for generating code snippet log files.  This means we now finally have all snippet tests supported for all languages! :)  Note that this doesn't add Magma to Github's CI (this is still run with the `-i magma` flag, which ignores Magma), but allows users to run `generate_snippet_tests.py` locally and manually generate the snippet log files for Magma.  This also includes support for CHIMP, if it's installed.

This was able to catch some small errors in the Magma code snippets for Dirichlet characters, abstract groups, and number fields (in particular the trivial group and rational number field), which have now all been corrected.

As always, any comments/feedback much appreciated! 🙂  In order to get `pexpect` working properly with Magma in `generate_snippets_tests.py`, I did request some help from Fable 5, but I've checked over all the code myself.


roed314:
Looks good, thanks!

PR #7080, opened 2026-07-25
Remove ANSI colour characters from GAP snippet log files

rvisser7:
This PR removes the ANSI colour escape characters (e.g. `[1m[34mgap> [0m`) from the GAP snippet log files, which originally came from GAP's coloured prompt, and thus addresses [David's comment](https://github.com/LMFDB/lmfdb/pull/6934#discussion_r3363141375) in PR #6934.  The `generate_snippet_tests.py` file now disables all colours from the GAP startup command.  The existing GAP log files have been regenerated, so they are now clean (and far more easily readable!).

As always, any comments/feedback welcome! :)


jenpaulhus:
LGTM!

PR #7081, opened 2026-07-26
Remove some unnecessary code from the code.yaml files

rvisser7:
This PR just removes some old unused code from the `code.yaml` files.

The full names and comment characters for all languages (Sage/Magma/Pari etc.) are now provided globally in `lmfdb/utils/place_code.py`, so there's no need to separately define these in each section anymore.  The logos for each language are also not used anywhere (and anyway should be done globally if they are used in the future).  I haven't touched any sections which still handle code snippets the old way.

As always, comments/feedback welcome! :)

### DIFF
--- a/lmfdb/characters/code.yaml
+++ b/lmfdb/characters/code.yaml
@@ -3,27 +3,6 @@ prompt:
   pari: 'gp'
   magma: 'magma'
 
-logo:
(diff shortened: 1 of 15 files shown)


jenpaulhus:
Looks good!  Thanks, Robin.

PR #7082, opened 2026-07-26
Add Sage and Oscar download links for abstract groups

rvisser7:
The abstract group pages currently offer group data downloads for Gap and Magma (i.e. group constructions as permutation, presentation, matrix groups, boolean invariants, and a character table description).  This PR now also adds "Group to SageMath" and "Group to Oscar" download links for abstract groups.

Examples of the new Sage group data pages:

http://localhost:37777/Groups/Abstract/1.1/download/sage
http://localhost:37777/Groups/Abstract/6.1/download/sage
http://localhost:37777/Groups/Abstract/8.3/download/sage
http://localhost:37777/Groups/Abstract/60.5/download/sage

Examples of the new Oscar group data pages:

http://localhost:37777/Groups/Abstract/1.1/download/oscar
http://localhost:37777/Groups/Abstract/6.1/download/oscar
http://localhost:37777/Groups/Abstract/8.3/download/oscar
http://localhost:37777/Groups/Abstract/60.5/download/oscar

To avoid code duplication, some of the existing code to generate the Gap and Magma data download pages have also been simplified, which include some minor typo fixes.  The Sage download pages have also been added to the snippet testing (this requires getting the snippet tests to work with multi-line inputs, so only Sage data tests are supported at the moment).

Now finally the abstract groups pages have download links for both constructing the groups and their invariants/characters (i.e. "Group to {lang}"), and download links for the code snippets (i.e. "{lang} commands"}, in all four languages Magma/Gap/Sage/Oscar! 🙂 

As always, any comments/feedback very welcome! :)


roed314:
Awesome!  Here's some feedback from GPT: https://chatgpt.com/share/6a6eb459-f9b0-83ea-9740-785bc7a00d9a

rvisser7:
Many thanks for the review!  I've made the following updates:

 - Added two more tests for the code download pages (8.4 and 32.14), which in particular test the matrix group constructions.
 - Added the Magma log files for these tests.
 - Fixed the typo for constructing the GLZN and GLZq groups in Oscar.
 - Fixed typo when renaming the group variable G in the Oscar download pages.
 - Fixed some minor typos in preamble in code download pages.

I agree we should still add GAP and Oscar log files for the group download pages, although this does require some further non-trivial changes to `generate_snippet_tests.py` to make this work.  The current continuation-prompt handling only seems to work for Sage and Magma, whereas Oscar has no continuation prompt at all.  I think this can be a separate PR.

edgarcosta:
This LGTM, but trying out a new thing with ChatGPT with
https://gist.github.com/edgarcosta/ed11a00620a28202dee2a3defa37fb62

edgarcosta:
For Lie representations, this sets `cc_known = False` for Sage/Oscar, which suppresses the entire character table. However, `_char_table_dict` already appears to support the case where class representatives cannot be generated: it omits `UnderlyingGroup` and `ConjugacyClasses` while still exporting the remaining character-table data.

Is suppressing the whole table here intentional? If not, could we leave `cc_known` unchanged and rely on the existing fallback in `_char_table_dict`?

edgarcosta:
Do not pair `c.powers` with `factors_of_order` by position

`WebAbstractConjClass.powers` is indexed by the prime factors of
`|G| * phi(|G|)`, not by `G.factors_of_order`. Thus this code silently
mislabels power maps whenever those prime lists differ.

For example, for `5.1`, `c.powers[0]` is the 2-power map
`[1, 4, 5, 3, 2]`, but the generated Sage/Oscar dictionary stores it
under key `5`. The actual 5-power map is `[1, 1, 1, 1, 1]`.

Please either derive the keys from `factor(|G| * phi(|G|))`, preserving
the indexing of `c.powers`, or use `c.prime_powers()` when pairing maps
with `G.factors_of_order`. A regression test using `5.1` should catch
this.

rvisser7:
Thanks @edgarcosta for the feedback!  I've made the following changes:

 - Now avoids suppressing the character table for Sage/Oscar (even if class representatives cannot be generated).
 - Corrected the indexing for ComputedPowerMaps in the GAP downloads.
 - As suggested, added a test for the cyclic group 5.1.

PR #7088, opened 2026-07-27
Add code snippet tests for the higher genus curves pages

rvisser7:
This PR just adds some code testing for the code downloads for the higher genus pages (as part of my ongoing quest to add tests for all code download pages!).  I went with adding tests for these four pages:

https://beta.lmfdb.org/HigherGenus/C/Aut/2.2-1.0.2-2-2-2-2-2.1  (includes hyperelliptic case)
https://beta.lmfdb.org/HigherGenus/C/Aut/3.168-42.0.2-3-7.1 (no hyperelliptic nor cyclic trigonal)
https://beta.lmfdb.org/HigherGenus/C/Aut/10.9-2.0.3-3-3-3-3-3.1 (includes cyclic trigonal case)
https://beta.lmfdb.org/HigherGenus/C/Aut/13.45-2.0.3-15-15.1 (full auto group bigger than listed group)

The snippet log files for GAP and Magma are included.

As always, feedback/comments welcome!


roed314:
GPT found a few small problems, but it mostly looks good: https://chatgpt.com/share/6a6eb03a-3284-83ea-b07f-4b902307d5f4

rvisser7:
Thanks for the GPT review!

I agree that there is an issue in the code snippet testing logic where we should reset and clear all the data between each snippet log file being generated (which suppresses possible errors that would otherwise be raised).  This will require making some changes to the `generate_snippet_tests.py` file, so I think this is probably an issue for a separate PR ?

Otherwise I'm happy to implement the remaining fixes GPT suggests.  Here's a summary of the main changes:

 - Added some extra tests which now test also the `rep_{lang}`, `topo_{lang}`, and `braid_{lang}` code downloads; this now also includes a test for both the family pages and passport pages.
 - Added guards to safely 404 abort if language in URL is incorrect, instead of raising a server error.
  - Corrected the logic for adding the braid and topological tag.

PR #7091, opened 2026-07-28
Fix TypeError in pAdicSlopeFamily.galois_groups on records with gal=None

paulhusCL:
The second issue Claude identified was an issue with a "None" value for padic Fields.   The fix notes are below. I want to emphasize here this particular warning:

> This fix silently drops fields whose Galois group hasn't been computed from the summary line. It's possible the intended behavior is instead to surface them as "unknown" in the listing. That's a display/math decision for a domain expert, not made here.

I *think* this is the behavior we want but I don't know these pages as well as others, so if we want the unknowns listed, just let me know.

-----

Production traceback on /padicField/family/... pages: sorted() raised "'<' not supported between instances of 'NoneType' and 'int'".

The filter guarding the Counter comprehension checked only for *key presence* ("gal" in rec and "galois_label" in rec), not for a non-None *value*. Records where "gal" is present but null (Galois group not yet computed) passed the filter, so None values reached sorted() alongside ints and comparison failed.

Switched to rec.get(col) is not None checks, matching the idiom already used a few lines above in some_hidden_data_available (rec.get(col) for col in [...]).

### DIFF
--- a/lmfdb/local_fields/family.py
+++ b/lmfdb/local_fields/family.py
@@ -394,7 +394,7 @@ class pAdicSlopeFamily:
     @lazy_attribute
     def galois_groups(self):
         fields, cache = self.fields
-        opts = sorted(Counter((rec["gal"], rec["galois_label"]) for rec in fields if "gal" in rec and "galois_label" in rec).items())
+        opts = sorted(Counter((rec["gal"], rec["galois_label"]) for rec in fields if rec.get("gal") is not None and rec.get("galois_label") is not None).items())
         if not opts:
             return "No Galois groups in this family have been computed"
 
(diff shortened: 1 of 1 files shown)


roed314:
They should be listed as as unknown, since otherwise it's hard to know that there are missing groups in the family.

paulhusCL:
Ok.  The latest commit should now show the unknown counts too.  

Two things to note from Claude about the appearance of these unknown counts:

> The unknown entry is plain text rather than a link: url_for(".family_page",
> gal=label) has no value meaning "missing", so there is nothing to filter on.
> 
> When an unknown bucket is present, a single known group now gets its
> "show N" link (previously suppressed when len(opts) == 1), since filtering
> to that group is meaningful once not all fields are accounted for. Families
> with no unknowns render exactly as before.

If you want a different behavior/look on these pages, just let me know.

paulhusCL:
Should have added some possible test pages (from the original errors)

/padicField/family/5.1.4.3a1.1-1.5.24a 
/padicField/family/2.1.2.3a1.2-1.8.51a?gal=16T1003.

roed314:
It looks good now.

PR #7103, opened 2026-07-31
Set random seed before generating the code snippet log files

rvisser7:
When generating the code snippet log files, some of the code snippets produce output which is non-deterministic (i.e. uses randomised algorithms). E.g. when outputting group generators, the order of the given generators is often randomly chosen.  This causes certain parts of the snippet log files to change every time `generate_snippet_tests.py` is run.  

This PR sets a fixed random seed (currently just set to 1) in each language before every log file is generated.  So now hopefully the code snippet Github CI won't need to update the log files so often. This also addresses David's comments in [PR #6934](https://github.com/LMFDB/lmfdb/pull/6934#discussion_r3363143290).

As always, any comments/feedback much appreciated! 🙂


roed314:
I don't think all of these need to be addressed, but take a look at the GPT review: https://chatgpt.com/share/6a6ed8b5-14b8-83ea-8a16-60bd55421e24

roed314:
@rvisser7, did you have more changes you wanted to make here or are you ready for me to take a look?

PR #7110, opened 2026-08-02
Fix ECNF regulator search box

rvisser7:
This just fixes a typo in the ECNF search when parsing the regulator.  The regulator search box has name set to `regulator` while the column is `reg`; this resulted in the regulator search box having no effect.  Rather surprising this has gone unnoticed for so long!

https://beta.lmfdb.org/EllipticCurve/?regulator=8.4-9.1  (returns all elliptic curves)
http://localhost:37777/EllipticCurve/?regulator=8.4-9.1  (returns the correct search page)


roed314:
Yikes!

PR #7114, opened 2026-08-02
Fix bug in p-adic completeness statement

roed314:
Compare:
* https://beta.lmfdb.org/padicField/?p=2&gal=256.2522
* http://localhost:37777/padicField/?p=2&gal=256.2522


roed314:
@jwj61 There are a few questions here I'd appreciate a second opinion on.

1. If we search for a Galois group using an alias or abstract label, do you think it's reasonable to say that the results are incomplete?  https://beta.lmfdb.org/padicField/?p=2&gal=S4 currently warns about uncomputed Galois groups, but once those are added I think it would currently say complete but with the change this would be incomplete.
2. I wasn't sure what the appropriate behavior for projective images of Artin representations was.  Let me know if you think I should make a different choice.

jwj61:
I am a little unclear as to what is going on.

In the case of p-adic fields with Galois group $$S_4$$, we have all examples with degree $$<24$$, but there are 3 examples in degree 24.  It seems like the right response is to say it is complete (probably coming from completeness.py) if the degree is bounded (either by filling in the degree field or by saying something like 4T5), but not claim completeness for the search above.  Of course none of this has to do with unknown computations.  Also, the time when all p-adic Galois groups are computed may not even be on the horizon.

What is going on with Artin representations?  I looked at the file change for the test and neither says the result is complete, nor does it talk about uncomputed quantities.  On the whole, that seems right.  We know results would be complete for certain characters and conductor bounds (and degree 1 with a conductor bound I don't specifically recall) but I expect these would not come up very often for people.

PR #7140, opened 2026-08-06
Fix invalid values in ECNF search result downloads

### LINKED ISSUE #7004: Error when downloading search results for elliptic curves over number fields

roed-math:
Closes #7004.

Reproducing the three searches from the issue in all six download formats (and running the resulting files) found two remaining bugs after #7005: Oscar downloads serialized non-integral rationals as `-1/2`, which Julia parses as float division, so `make_data()` failed on any curve over a non-monogenic field; and curves with unknown rank downloaded the LaTeX display string `"0 \le r \le 1"` in every language (9974 curves DB-wide).

Fixes: a new `rational_to_lang` hook in the shared downloader (Oscar emits `p//q`; generic for all sections), and the ECNF rank column now downloads the integer rank when known and `[lo, hi]` bounds otherwise, leaving the display unchanged. The per-row base-field polynomial lookup is gone as well: `ECNFDownloader.field_polys` loads the defining polynomials of all 894 base fields occurring in `ec_nfcurves` in a single query the first time a download happens, so generating a download issues no further `nf_fields` queries (~2 min to ~2 s per ~700 rows).

Verified by rebuilding every curve from the downloaded files end to end: sage/magma/gp/oscar rebuild 712/587/1078 curves for the three issue searches with zero bad conductor norms or fields. The full ecnf suite passes (14 tests, including `test_search_download` covering both bugs in all six formats and `test_download_field_polys` checking the bulk fetch and reuse), as do `lmfdb/tests/test_utils.py` and the CMF/local-fields download tests.

---

Ported from https://github.com/roed-math/lmfdb/pull/3, where the full write-up and comment history live.


lmfdb/ecnf/test_ecnf.py
```diff
@@ -77,6 +77,58 @@ def test_download(self):
...
+        rows = [{'field_label': label, 'ainvs': '0,0,0;0,0,0;0,0,0;0,0,0;0,0,0'}
+                for label in ['3.3.1849.1', '2.0.868.1', '3.3.1849.1']]
+        with patch.object(self.db.nf_fields, 'search', wraps=self.db.nf_fields.search) as search, \
+             patch.object(self.db.nf_fields, 'lookup', wraps=self.db.nf_fields.lookup) as lookup:
+            polys = [downloader.postprocess(row, {}, {})['field_coeffs'] for row in rows]
+            # All of the defining polynomials are fetched by the first row, in bulk
+            assert search.call_count == 1
+            assert lookup.call_count == 0
+            # and the dictionary they were saved in is reused afterwards
+            assert downloader.field_polys is downloader.field_polys
```
rvisser7, line 126: I'm just a bit confused about this assert on line 126, isn't this tautologically true?

rvisser7:
Thanks, this looks good to me!

rvisser7 (approved):

PR #7150, opened 2026-08-06
Add curves as a search column for abelian varieties over finite fields

### LINKED ISSUE #6975: Add curves as a search column for abvar/fq
In the abelian varieties section, it would be good to add the list of curves whose Jacobian is in the given isogeny class as a search column (so that they can be downloaded).

From the feedback form:

"I’m trying to get data for a bunch of elliptic curves over various small finite fields and download it to use in gp. i can download various things like the L-polynomial, but i don't see how to download actual equations for the curves. i see that in a given abelian variety A's homepage there is sometimes an equation for a curve with a jacobian for A, which is what i want to get, but i don't see any field to use to download it. is there a good way to get these equations?"

See [here](https://www.lmfdb.org/Variety/Abelian/Fq/?q=2%2C3%2C4%2C5%2C7&g=1) for an example of a search where it would be good to have this option.

roed-math:
Closes #6975.

Adds a default-off "Curves" column to the Variety/Abelian/Fq search results, so the curve equations already shown on isogeny class homepages can be displayed and, in particular, downloaded from search results in all supported formats. The column displays the first equation (TeXified, with a count of the remaining ones) to keep rows compact, while downloads contain the full list of raw equations.

The download description gives the exact convention for the generator of a non-prime base field: for `q = p^r` with `r > 1`, `a` is the image of `x` in `GF(p)[x]/(C(x))`, where `C` is the degree `r` Conway polynomial over `GF(p)` used by the LMFDB. This is the same convention the isogeny class homepage states. The description also distinguishes the two missing-data states: a null value means the curves in that isogeny class have not been computed, while an empty list means the computation was carried out and there are no such curves. (Spot checks back this up: whole `(g, q)` blocks are either uncomputed or fully computed, and inside the computed blocks the empty lists fall exactly on the classes recorded as containing no Jacobian.)

Because the curves arrays are large, they are kept out of the fixed `db_cols` projection:

* `curves` is excluded from the normal fixed projection used by every abvar/Fq search;
* display searches add it to the projection only when the Curves column is shown;
* search-result downloads add it only when the column is included in the output;
* enabling the column can still transfer a few megabytes on unusually large result pages (the heaviest realistic browse window, g=2 and q=167, is about 2.3MB), but ordinary searches and downloads pay no such cost.

The projection hook `Downloader.get_projection` is a behavior-preserving extraction of the previous inline logic, overridden only by `AbvarFq_download`. The TeXification logic is shared with the homepage display.

Tests cover the column dropdown, the compact display, downloads in every language (including gp, which is what the issue asked for), the presence of the Conway polynomial convention in the non-CSV downloads, the projection contract in both directions, and null versus empty curve lists.

---

Ported from https://github.com/roed-math/lmfdb/pull/13, where the full write-up and comment history live.


AndrewVSutherland:
While curves is now a display column, it doesn't appear to be filled in correctly:
image
The isogeny class `1.2.ac` has 'curves': ['y^2+y=x^3+x+1']

PR #7159, opened 2026-08-06
Highlight active search constraints on results pages

### LINKED ISSUE #2770: Show active constraints on search results pages
When displaying the results of a search, it would be nice to change the color of input/select elements if they're actively constraining the results (it's easy to forget that you've constrained something).

roed-math:
Closes #2770.

On search results pages, the inputs that are actually constraining the results being displayed are now tinted (text boxes and selects) or ringed (checkboxes), so it is visible at a glance why the results are filtered.

## How it works

Marking happens server-side in the shared `search_boxes.py` machinery, so every section gets it. `SearchBox._attrs(info)` emits **exactly one** `class` attribute per input, merging the box's own classes with `advanced`, `search_constraint` (this input constrains the results when filled in) and `search_active` (it is filled in right now). Class names are passed with a first-class `classes=` argument rather than hand-written into `extra`, since two `class` attributes on a tag is invalid HTML and browsers silently discard one of them.

`search_active` describes the search that produced the rows on screen, and is deliberately server-owned: nothing rewrites it as the user edits the form, so the highlight never claims a constraint that has not been submitted. The existing fresh/stale "Search again" button remains the signal that the form has been edited.

## What counts as a constraint

`is_constraint` is a bool, or a function of `info` for boxes whose role depends on the rest of the search.

Not constraints: result count, sort and column controllers, dynamic-statistics variable boxes, diagram axis/colour/count boxes, hidden inputs, and the trace and Euler table display controls (`n`, `n_primality`, `view_modp`) in classical modular forms and L-functions. That last group includes the `n_primality=primes` that `set_Trn` inserts on every trace search, which would otherwise look filled in when the user never touched it.

Contextual, via a callable:

- `an_modulo` counts only when there are trace constraints for it to turn into congruences, not when it merely drives `view_modp=reductions`.
- The hypergeometric prime `p` counts only when one of the `Ap`, `Bp`, `Apperp`, `Bpperp` boxes is filled, since it only picks which columns those compare against.
- The abelian variety geometric-decomposition checkbox counts only when one of the inputs it retargets is used.
- A select beside a text box (`TextBoxWithSelect`) counts only once that text box is filled. This is safe because `SearchParser.__call__` returns immediately on an empty field, so a select handed to a `@search_parser` as `mode=` cannot affect the query alone. Six selects are read directly during parsing and do constrain by themselves, so they opt out with `is_constraint=True`.

## Colours

Two colour-scheme keys, so both stay overridable per scheme: `search_active_background` (`#FFF9C4`) for the tint and `search_active_ring` (`#F9A825`) for checkboxes, which are rendered natively and cannot be tinted. The ring uses `box-shadow` rather than `outline` so the browser's keyboard focus indicator survives. The tint uses `background-color` rather than the `background` shorthand so it cannot reset a background image. The exact shades are a maintainer taste call and easy to tune.

## Two bugs found and fixed along the way

Both were pre-existing and unrelated to the highlighting, but each was directly in the path of deciding whether an input constrains the results.

- **Classical modular forms level type.** The dropdown offers "prime squared" and "powerful", but `common_parse` recognised neither, so both fell to the invalid-type branch. That branch flashes an error and returns a redirect, and both callers use the query `common_parse` builds while discarding its return value, so the search then ran with no condition on the level at all. `mf_newforms` and `mf_newspaces` already have `level_is_prime_square` and `level_is_powerful`, so the two values only needed adding to the list. The invalid branch now raises `ValueError` after the flash, the way the rest of that module reports bad input.
- **Abelian variety geometric decomposition.** `CheckBox` renders `value="yes"` but the parser tested `== "on"`, so ticking the box in the form did nothing: the search used the ordinary `dim*`, `number_fields` and `galois_groups` columns instead of the geometric ones. It now accepts both values, keeping hand-written `on` links working.

## Tests

New `SearchBoxTest` in `lmfdb/tests/test_utils.py` covers the rendering rules. Integration assertions in `test_hgm.py`, `test_cmf.py`, `test_cmf2.py`, `test_lfunctions.py`, `test_modular_curves.py` and `test_av.py` check the rendered pages, plus direct `common_parse` query-construction tests for the geometric decomposition checkbox. Tests compare class tokens as sets through a shared helper that asserts there is at most one `class` attribute, so nothing depends on ordering. No new test files, so the CI count is unchanged.

---

Ported from https://github.com/roed-math/lmfdb/pull/23, where the full write-up and comment history live.


lmfdb/elliptic_curves/elliptic_curve.py
```diff
@@ -1291,6 +1291,9 @@ class ECSearchArray(SearchArray):
                                }
 
     def __init__(self):
+        # Unlike most quantifier selects, this one restricts the results on
+        # its own: prime/p-power/sq-free set num_bad_primes and semistable
+        # whether or not a conductor was entered
```
havarddj, line 1296: To me this feels like a comment that only makes sense within the context of this PR. Thoughts?

lmfdb/modular_curves/main.py
```diff
@@ -767,6 +767,10 @@ class ModCurveSearchArray(SearchArray):
     jump_knowl = "modcurve.search_input"
 
     def __init__(self):
+        # Unlike most quantifier selects, this one restricts the results on
+        # its own: prime/p-power/sq-free set num_bad_primes and
+        # level_is_squarefree whether or not a level was entered, and
+        # modcurve_browse.html links straight to ?level_type=... searches
```
havarddj, line 773: Same as above?

lmfdb/utils/color.py
```diff
@@ -196,6 +196,11 @@ class ColorScheme():
         # Buttons
         'button_background': 'col_main_b',
         'select_background': 'col_main_ll',
+        # Background for search inputs actively constraining the results shown
+        'search_active_background': '#FFF9C4', # P2-100
+        # Ring for active checkboxes, which are rendered natively and so
+        # cannot be tinted; darker than the tint so that it stays visible
+        'search_active_ring': '#F9A825', # P2-800
```
havarddj, line 203: I don't think this should be hardcoded here. Instead it should be set as a default (like `c.yellow`, which I dislike), or with a key (like `col_box_constraint`). 

Kiran recently asked about a dark theme for the LMFDB, and I think this is a good excuse to update some of the names/handling here; I'm not a huge fan of names like "col_main_l" because they're not semantic. 

That's worth doing in a separate PR though, happy to take the lead on that.

havarddj:
Okay so even though it's not my favourite, I asked around and people did seem to prefer your shade of yellow to the ones I presented (which were less saturated, but also harder to see). Here are some other options:

### Current PR #FFF9C4 `#FFF9C4`
Screenshot 2026-09-19 at 14 29 01

### Faded Apricot #FBE9E7 `#FBE9E7`
Screenshot 2026-09-19 at 14 28 54

### Muted Mint #E8F5E9 `#E8F5E9`
Screenshot 2026-09-19 at 14 28 58

I prefer the apricot, but feel free to chime in! The mint one is a bit more reminiscent of the old LMFDB colour scheme.

roed314:
@AndrewVSutherland, @jwj61 and I like muted mint, so I'll change it to that.  Further opinions are welcome!

PR #7171, opened 2026-08-06
Add a maximal subgroups mode to the subgroup diagram and profile

### LINKED ISSUE #6891: Maximal subgroups
In cases where the subgroup lattice is too hard to compute, and even in cases where it has been computed but it's really involved, it would be useful to have an option to just display the maximal subgroups, so for example add a button next to "all" and "normal" subgroups to give "maximal".  (Very often, this is what I'd like to see as well, and of course you can always proceed by induction.)

As @roed314 pointed out to me, we already have the data there so it's just a matter of a UI tweak.

Suggested by Kamilla Rekvenyi.

roed-math:
Closes #6891.

Adds a "maximal subgroups" button next to "all subgroups"/"normal subgroups" on abstract group pages, as suggested in the issue. The new mode works for both the diagram and the profile views, crossed with the existing up-to-conjugacy/up-to-automorphism toggle, and gets its own fullpage diagram routes.

The diagram shows the whole group on top with the classes of maximal subgroups below it, built directly from the stored `maximal` column, so it displays even for groups whose full subgroup diagram is disabled for size (e.g. 32.45, 1024.dke). For groups where inclusions were never computed (e.g. 1024.dip) the profile still lists all maximal subgroup classes.

Verified on S4, PGL(2,7), p-groups with huge/partial/missing lattices, and the trivial group; the abstract-groups test suite passes and a new test covers the mode.

---

Ported from https://github.com/roed-math/lmfdb/pull/37, where the full write-up and comment history live.


jenpaulhus:
If I click on "maximal subgroups", it gives them as a vertical line which suggest they are contained in each other.  Here's D20:

Screenshot 2026-08-06 at 17 50 23

PR #7183, opened 2026-08-08
Add "verified" annotations for the 97 genus 2 curves where Sha is rigorously known

AndrewVSutherland:
This PR modifies the display of analytic sha to replace "(rounded)" with "(verified)" when the newly added `sha_proved` column of `g2c_curves` is set to True (which holds for the 97 curves listed in https://github.com/TimoKellerMath/strongBSDgenus2/blob/main/LMFDB-sha-data.txt)

### DIFF
--- a/lmfdb/genus2_curves/templates/g2c_curve.html
+++ b/lmfdb/genus2_curves/templates/g2c_curve.html
@@ -188,7 +188,7 @@ function show_eqns(invstyle) {
     <tr><td>{{ KNOWL('lfunction.leading_coeff', 'Leading coefficient') }}:</td>
         <td>{% if data.leading_coefficient == 'unknown' %} unknown {% else %} \( {{ data.leading_coeff }} \) {% endif %}</td></tr>
     <tr><td>{{ KNOWL('g2c.analytic_sha', 'Analytic order of &#1064;') }}:</td>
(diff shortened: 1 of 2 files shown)


MichaelStollBayreuth:
@AndrewVSutherland Any idea when this will be merged?

AndrewVSutherland:
Given that the data is already there and the code changes are tiny, I am going to take the liberty of merging this now.

AndrewVSutherland:
@MichaelStollBayreuth I will try to get this PR into production today (but there are a bunch of other changes in main that need to run through our CI, so it will likely be a few hours).

MichaelStollBayreuth:
Thanks! A few hours is not a problem. I'm planning to submit the report on Monday (and it most likely will actually be read quite a bit later...).

MichaelStollBayreuth:
I still don't see this on the live database (e.g. on [this curve page](https://www.lmfdb.org/Genus2Curve/Q/529/a/529/1)).

AndrewVSutherland:
The PR to web has been merged, so you should now see "verified" next to the order of Sha on the page you linked to above.

MichaelStollBayreuth:
I do now -- thanks!

MichaelStollBayreuth:
But it could say in addition "Order of Ш: 1 (verified)" -- we compute both the actual order and the analytic order of Ш.

Issue #7099, opened 2026-07-30
Make the belyi_images.txt file into a PostgreSQL table


rvisser7:
Currently, the images for Belyi maps are being stored as a static file [lmfdb/static/images/belyi_images.txt](https://github.com/LMFDB/lmfdb/blob/main/lmfdb/static/images/belyi_images.txt) in the repo, which takes up about 35MB of space.  The whole LMFDB repo is roughly 55MB, so this single file is the majority of it.

Wouldn't it make more sense for this to live in the database as a PostgreSQL table?  The file is already in our standard table-export format; its header is `id|BelyiDB_plabel|DD_images` with types `bigint|text|text[]`, so it looks like it could be uploaded as a PostgreSQL table more or less as-is.

roed314:
Yes, I definitely agree.  Is this file actually being used?  There is already a `belyi_galmap_portraits`.  If it's being used, I'm happy to switch it over to a postgres table and remove it from the repo (and possibly from the git history).

ReymondAkpanya:
Sorry for adding to the bloat to the repo.

 #7189  replaces belyi_images.txt (35MB) with a gzip-compressed belyi_images.txt.gz (~8MB). Still a static file rather than a Postgres table, so this doesn't fully close the issue, but it's roughly a 4x reduction in the meantime.

To answer the question above: yes, this file is actively used. It powers the interactive dessin d'enfant drawings shown via "View Embedding" on galmap pages (get_belyi_images / dessin_svgs in web_belyi.py). It's unrelated to belyi_galmap_portraits, which only supplies the small static thumbnail in the properties box — the two aren't redundant.

One thing worth flagging: merging the PR won't shrink the repo's actual clone size, since the old 35MB blob stays reachable in main's git history either way. Fully removing it needs a history rewrite plus a force-push of main.

PR #7189, opened 2026-08-14
Fix belyi dessin images 7026

### LINKED ISSUE #7026: Incorrect dessin displayed for Belyi map 9T23/6.2.1/6.2.1/6.2.1/b
# Incorrect dessin displayed for Belyi map `9T23/6.2.1/6.2.1/6.2.1/b`

Salim Alloun noticed that the dessin displayed at

https://beta.lmfdb.org/Belyi/9T23/6.2.1/6.2.1/6.2.1/b/

appears to be incorrect. In particular, the white vertices seem to have the correct edge incidences, but the black vertices do not.

The permutation triple shown on the page appears to be correct, so it is not immediately clear where the issue originates. My understanding is that the dessin is generated from the permutation data, which suggests either a problem in the underlying data or in the dessin-generation/display pipeline.

For comparison, the related page

https://beta.lmfdb.org/Belyi/9T23/6.2.1/6.2.1/6.2.1/a/

appears to display the correct dessin.

Could someone check whether the stored dessin data or the rendering process is inconsistent for the `b` entry?

ReymondAkpanya:
This also regenerates all the dessin d'enfant images: 
- grey backgrounds behind edge/boundary labels for readability, 
- switched to unlabelled black/white vertices, 
- fixed a crash on one genus-2 passport,
-  and fixed an SVG glyph-ID collision where multiple embeddings on the same page could silently render each other's digits. 

Closes #7026
Net effect: belyi_images.txt (35MB, tracked) → belyi_images.txt.gz (~8MB, tracked). See #7099  for background.

Note for whoever merges: this shrinks the live file, but the old 35MB blob stays in main's history until someone runs a history rewrite and force-pushes main — a repo-wide disruptive operation (forces everyone to re-clone or clean local history), so probably worth batching with any o rather than doing it just for this.


edgarcosta:
Do you want to change this line also?


https://github.com/ReymondAkpanya/lmfdb/blob/abdaebb5eacb6a83d98b5cd484877556229deb45/lmfdb/belyi/templates/belyi_galmap.html#L147-L152


The new contract here is "one SVG per embedding", but the template still falls
back to dessin_svgs[0] whenever the list is too short. 
```
assert    len(dessin_svgs) == len(embeddings_and_triples)
```
and remove the index-0 fallback, e.g.:
```jinja2
{{ data.dessin_svgs[loop.index0]|safe }}
```

Also, add some tests, given how small the database is, it might more worthwhile to write a check in SQL.

Should we put these images in the database, instead of loading them from a textfile?

ReymondAkpanya:
Fixed — removed the index-0 fallback in `belyi_galmap.html`, indexing `dessin_svgs` directly by embedding position, with an assert in `web_belyi.py` that fails loudly on a count mismatch instead of silently mis-rendering.

Added tests, including the whole-table check you suggested: rather than one example, it validates the image/embedding count invariant across all 1111 galmaps and checks all 82 multi-orbit passports for distinct images — cheap enough to just check everything (~5s total).

PR #7197, opened 2026-09-04
Add comment with download link to n=32 transitive groups

havarddj:
Gap currently does not have transitive groups of degree 32. Therefore the code snippets on the pages for these groups will give errors unless the user has installed the relevant databases. 

This PR adds comments with the relevant links to the transitive groups pages, as well as cleans up newline stripping in utils/place_code.py

### DIFF
--- a/lmfdb/galois_groups/code.yaml
+++ b/lmfdb/galois_groups/code.yaml
@@ -15,6 +15,13 @@ gg:
   oscar: G = transitive_group({n}, {t})
   gap: G := TransitiveGroup({n}, {t});
 
+n32_dbs:
+  comment: Databases for transitive groups of order 32
+  sage: https://www.math.colostate.edu/~hulpke/transgrp/trans32.tgz
+  gap: https://www.math.colostate.edu/~hulpke/transgrp/trans32.tgz
+  magma: https://magma.maths.usyd.edu.au/magma/download/db/TrnGps32.tar.gz
+  oscar: https://www.math.colostate.edu/~hulpke/transgrp/trans32.tgz
+  
 n:
   comment: Degree
   magma: t, n := TransitiveGroupIdentification(G); n;
--- a/lmfdb/galois_groups/transitive_group.py
+++ b/lmfdb/galois_groups/transitive_group.py
@@ -319,7 +319,6 @@ class WebGaloisGroup:
             for lang in n32_dbs:
                 if lang != "comment":
                     self.code['gg'][lang] = f"{comments[lang]} This requires the 32T* database at {n32_dbs[lang]}\n" + self.code['gg'][lang]
-                    print("Name of sage gg key:", self.code['gg']['sage'])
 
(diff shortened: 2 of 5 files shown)


rvisser7:
This looks great, thank you!  I've just put some comments below. :)

lmfdb/galois_groups/code.yaml
```diff
@@ -15,6 +15,13 @@ gg:
   oscar: G = transitive_group({n}, {t})
   gap: G := TransitiveGroup({n}, {t});
 
+n32_dbs:
+  comment: Databases for transitive groups of order 32
```
rvisser7: This should be "degree 32".

lmfdb/galois_groups/transitive_group.py
```diff
@@ -309,6 +309,16 @@ def make_code_snippets(self):
             for lang in self.code[prop]:
                 self.code[prop][lang] = self.code[prop][lang].format(**{'n':self.n(), 't':self.t()})
         self.code['show'] = { lang:'' for lang in self.code['prompt'] }
+        # For transitive groups of order 32, the user needs to load an extra package
```
rvisser7: Also should be "degree 32" :)

lmfdb/utils/place_code.py
```diff
@@ -46,7 +46,9 @@ def place_code(self, is_top_snippet=False):
         if code[item]:
             for L in code[item]:
                 if isinstance(code[item][L],str):
-                    lines = code[item][L].split('\n')[:-1] if '\n' in code[item][L] else [code[item][L]]
+                    lines = code[item][L].split('\n') if '\n' in code[item][L] else [code[item][L]]
+                    # remove trailing empty line if any
```
rvisser7: Technically, I think this would remove *all* blank lines in `lines`, not necessarily just the trailing one.  Though I don't think we have any code snippets with interior blank lines, so I think this is fine :)

rvisser7:
Another side-effect I've noticed is that, in the Galois group code download pages, there is no longer a newline between defining the transitive group G, and the code snippet for the abstract group.  E.g. compare these pages:

https://beta.lmfdb.org/GaloisGroup/32T14/download/gap
http://localhost:37777/GaloisGroup/32T14/download/gap

I think an easy fix is to replace line 131 (second last line) of `lmfdb/utils/place_code.py` with:

```
code += self.code[key][lang].rstrip('\n') + '\n'
```

lmfdb/utils/place_code.py
havarddj: Fair point, I've updated the comment!

havarddj:
Thanks, I've addressed your comments now, let me know if you spot anything else!

lmfdb/utils/place_code.py
roed314: I think this can now just be `code[item][L].split('\n')` since `"".split('\n')` is `[""]`.

PR #7199, opened 2026-09-06
Small cosmetic tweak to the blue left sidebar

rvisser7:
This PR just makes some very small cosmetic tweaks to the blue left sidebar, in particular how the vertical indent guide looks in the collapsible submenus.  I've just shifted the vertical indent guide a small amount to the right, made it a slightly lighter shade of blue, and added 1px of vertical padding in the submenus.

This is certainly a very subjective (and very nitpicky) thing, so please feel free to suggest any changes! :)

E.g. here's a before and after for the "Varieties" section (using Chrome on Windows):

image

### DIFF
--- a/lmfdb/templates/sidebar.html
+++ b/lmfdb/templates/sidebar.html
@@ -24,7 +24,7 @@
     {%- endif -%}
     {%- if entry.collapsible and entry.part2 -%}
       <span class="collapsible-container">
-        {{entry.url | safe}}
+        {{entry.title | safe}}
 	<div class="collapsible-content">
       {%- for pt2 in entry.part2 -%}
         {%- if not BETA and pt2.status == "notonbeta" -%}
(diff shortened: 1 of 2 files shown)


lmfdb/templates/sidebar.html
```diff
@@ -102,7 +102,7 @@
           var contentId = 'lmfdb-sidebar-entry-' + index;
           $content.attr('id', contentId);
 
-          var $chev = $('<button type="button" class="collapsible-chevron open"><svg aria-hidden="true" focusable="false" width="24" height="24" viewBox="0 0 24 24" fill="none" xmlns="http://www.w3.org/2000/svg"><path d="M9 6L15 12L9 18" stroke="currentColor" stroke-width="3"/></svg></button>');
+          var $chev = $('<button type="button" class="collapsible-chevron open"><svg aria-hidden="true" focusable="false" width="24" height="24" viewBox="0 0 24 24" fill="none" xmlns="http://www.w3.org/2000/svg"><path d="M9 6L15 12L9 18" stroke="currentColor" stroke-width="2.5"/></svg></button>');
```
havarddj, line 105: I think this and the next chevron should be refactored so we have a uniform chevron object which is the same in all templates. That way, when we want to restyle it at some point in the future, we don't have to grep around in all the html files.

havarddj (approved):
This PR makes me very happy! It also fixes some spacing issues that I couldn't quite solve when the new sidebar design was first made.

lmfdb/templates/sidebar.html
rvisser7: Thanks, I agree!  I think we'd have to keep it as separate chevron elements, but the html code can definitely be uniformized.  I've now put all the chevron HTML into a `chevronHTML` variable, so that any future changes to this only requires updating one line.

havarddj: That's great! At some point we should update the boxes on right (e.g. props/downloads boxes) to also use this chevron, I think, but doing it now is probably beyond the scope of this PR.

rvisser7: Thanks, I agree!  I've made a new issue #7205 for improving the boxes on the right. :)

Issue #7108, opened 2026-08-01
Document when optional packages are required for code snippets


rvisser7:
Some of the code snippets in the various languages Sage/Gap/Magma/Pari that we serve on the LMFDB depend on optional packages or database files that are often not part of the default install of the relevant CAS.  Whenever this is the case, I think we should at least add a comment in the snippet mentioning the particular package that needs to be installed (or maybe if it's a very obscure package, possibly remove the code snippet?).  For example, the Magma code snippets for Dirichlet characters require CHIMP to be installed, which we already include as a comment in the relevant snippet.

Also, as we're adding more code snippet tests, we should also ensure these optional packages are installed in GitHub's CI for all our tests to pass (or alternatively, we could also in some way flag these code snippets not to be run in the CI?).

Here are the examples I've found so far:

 - **Galois groups of degree > 8**:  In order for Pari to compute Galois groups of number fields in degrees 8 to 11, this requires the `galdata` package.  (Furthermore, I don't think Pari's `polgalois(K.pol)` function works at all for number fields with degrees > 11, so perhaps this Pari snippet should be hidden for large degrees?)

 - **Generators of elliptic curves**:  Also, for Pari to compute generators of elliptic curves (`E.gen`), this requires the `elldata` package to be installed.

 - **Transitive groups of order 32**:  For most CASes, the transitive groups of degree 32 require installing extra packages.  E.g. Magma requires the [TrnGps32](https://magma.maths.usyd.edu.au/magma/download/db/TrnGps32.tar.gz) database file, and Sage/GAP requires the [trans32](https://www.math.colostate.edu/~hulpke/transgrp/trans32.tgz) supplemental download.

 - **Group homology**:  For abstract groups, our code snippet for computing the Schur multiplier (`G.homology(2)`) requires the `hap` package to be installed in Gap.   For this Gap package to be available in Sage, I think we'd also need to make sure `gap_packages` are installed in Sage.  This particular error was already flagged in PR [#6934](https://github.com/LMFDB/lmfdb/pull/6934#discussion_r3363131529).

roed314:
Yes, I agree that we should do this.

I agree about the Pari `polgalois(K.pol)` snippet being hidden for large degrees, though perhaps @AurelPage has some comments on that (is there a better way to do this in Pari)?

PR #7201, opened 2026-09-11
Add degree 32 transitive groups to CI, and upgrade Sage/GAP/Pari packages

rvisser7:
This PR makes some progress towards resolving issue #7108.  In particular, it adds the supplementary degree 32 transitive group GAP database to the CI,  so that the `TransitiveGroup(32, t)` code snippets can actually now be run and tested. This is now available in Sage, GAP, and Oscar when running and testing code snippets.  We've also added GAP (v4.14) and the `pari-galdata` package to the CI.

To check the new packages, a degree 8 number field and degree 32 group was added to the snippet testing.

As part of this PR, I've also upgraded the Sage version in `.environment.yml` from 10.4 to 10.6, in order for Sage to be compatible with GAP 4.14.  To make this work, two small updates to the Siegel modular forms and Bianchi modular forms code were needed, just to ensure all the tests pass, though none of these changes affect the frontend.  (Ideally, it would be great for us to use the latest version of Sage (10.9), though this seemed to cause many more things to break, so I've kept it as 10.6 for this PR.)

I did have some help from Claude Opus 5 in generating the bash script to download and cache the degree 32 transitive group database file to the CI, but all remaining code was written by me.

As always, any comments/feedback very welcome! :)


havarddj:
Since there are some comprehensive changes, I think it would be great to get one of the editors to look at it. In the meantime, perhaps you could squash commits so the commit history looks a bit nicer if/when it's merged?

roed314:
See https://chatgpt.com/share/6aad6e05-47fc-83ea-8638-56874759e751.  Once these two issues are addressed, this change looks good to me.

Issue #5749, opened 2023-11-27
Spacing at the top of the pages


jvoight:
Am I the only one bothered by the vertical spacing between the knowl and the word "Browse" at the top of the browse-search pages?  I feel like it should be the same as the vertical spacing between the Browse and Search sections.  

(I know vertical spacing is always at a premium, but it aids my eye in separating out the info/stats at the top.)

AndrewVSutherland:
I agree with you.  I also think there is more vertical space below the word Browse than is needed.  What do you think of this tweak (which basically moves 5 pixels below above so no net change in vertical real estate)
![image](https://github.com/LMFDB/lmfdb/assets/11425002/5e66acbd-9bd3-410a-bb08-a9880a1efb73)

AndrewVSutherland:
The reason margin-top was set to zero for the first h2 element is so that when there is not a stats line it lines up with the top of the side bar, e.g. as on 
![image](https://github.com/LMFDB/lmfdb/assets/11425002/315c5d42-cedd-4131-bb85-5e4a6f287532)

So the right way to change this is probably not to change the margin-top on the first h2 element but to instead increase the margin-bottom of the stats/description line (which maybe should get its own style class).

jvoight:
Like I said, we tackle the important stuff!  I feel better knowing that this was a result of another change, made for spacing reasons.  Thanks!

davidfarmer:
I think Drew's tweak would look even better with slightly more space above "Browse" and slightly less space below.

PR #7202, opened 2026-09-12
Fix vertical spacing around the Browse heading on browse-search pages

rvisser7:
This PR fixes issue #5749.   On each section homepage, there is not much space above the "Browse" heading and a larger gap just below the heading.  This PR fixes this vertical spacing according to Drew's suggestions [here](https://github.com/LMFDB/lmfdb/issues/5749#issuecomment-1827172476), i.e. by making the top stats/description line its own style class, i.e. adding `class="stats-summary"` to the div wrapper, and adding some custom CSS to increase margin-bottom for this class.

This change will only affect pages containing this `stats-summary` class, so the all the other usual object pages are unaffected.

E.g. Here's a before and after for the number fields homepage (using Chrome on Windows):
https://beta.lmfdb.org/NumberField/
http://localhost:37777/NumberField/

image

Similarly for the elliptic curves homepage:
https://beta.lmfdb.org/EllipticCurve/Q/
http://localhost:37777/EllipticCurve/Q/

As always, any comments/feedback much appreciated! :)


havarddj:
Looks great to me! Something for another day: the fact that you had to add the same bit of code in so many places might hint that we should refactor this into a template. But since the pages you changed all have some amount of special handling, I think this requires a little bit of thinking to get right.

